PR #6, opened 2025-06-10
[python][plan][api] Init Event, Action, Function and WorkflowPlan.

wenjin272:
Linked issue: #12 
This pr import the core concepts Event and Action to Flink-Agents, and import Function which is the dependency of Action. Based on these concepts, we import WorkflowPlan, and add a simplest test case.


python/requirements/build_requirements.txt
```diff
@@ -13,3 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 build
+pydantic==2.11.4
```
HuangXingBo: `build_requirements.txt` don't need this dependency. And we need to put this dependency into `pyproject.toml`

python/flink_agents/plan/workflow_plan.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Attributes:
+    ----------
+    actions : Dict[Type[Event], List[Action]]
+        Mapping of event types to the list of Actions that listen to them.
+    """
+
+    actions: Dict[Type[Event], List[Action]]
+
+    def get_action(self, event_type: Type[Event]) -> List[Action]:
```
codenohup: get_actions?

python/flink_agents/plan/workflow_plan.py
```diff
@@ -0,0 +1,51 @@
...
+        """Get steps that listen to the specified event type.
```
codenohup: steps?

python/flink_agents/api/event.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Attributes:
+    ----------
+    isLegal : bool
+        Is the result legal or not.
+    result : str
+        The final result returned by the workflow.
+    """
+
+    isLegal: bool
```
codenohup: is_legal?

python/flink_agents/plan/action.py
```diff
@@ -0,0 +1,45 @@
...
+    name : str
+        Name/identifier of the workflow Action.
+    exec : Function
+        To be executed when the Action is triggered.
+    listen_event_types : List[Type[Event]]
+        List of event types that will trigger this Action's execution.
+    """
+
+    name: str
+    exec: Function
```
xintongsong, line 44: Shall we check the signature of the function to make sure it's a valid action?

python/flink_agents/plan/function.py
```diff
@@ -0,0 +1,102 @@
...
+
+from pydantic import BaseModel
+
+
+class Function(BaseModel):
+    """Descriptor for a callable function, storing module and qualified name for dynamic
+    retrieval.
+
+    This class allows serialization and lazy loading of functions by storing their
+    module and
```
xintongsong: Odd line break here.

python/flink_agents/plan/function.py
```diff
@@ -0,0 +1,102 @@
...
+    qualified name. The actual callable is loaded on-demand when the instance is called.
+
+    Attributes:
+    ----------
+    module : str
+        Name of the Python module where the function is defined.
+    qualname : str
+        Qualified name of the function (e.g., 'ClassName.method' for class methods).
+    __func: Callable
+        Internal cache for the resolved function
```
xintongsong: Private property should not be documented here.

python/flink_agents/plan/function.py
```diff
@@ -0,0 +1,102 @@
...
+#################################################################################
+
+import importlib
+import inspect
+from typing import Any, Callable, Dict, Tuple
+
+from pydantic import BaseModel
+
+
+class Function(BaseModel):
```
xintongsong: I think this should be PythonFunction, and we need another Function as a common abstraction of Python & Java functions.

python/flink_agents/plan/tests/test_workflow_plan.py
```diff
@@ -0,0 +1,47 @@
...
+from flink_agents.api.event import InputEvent, OutputEvent
+from flink_agents.plan.action import Action
+from flink_agents.plan.function import Function
+from flink_agents.plan.workflow_plan import WorkflowPlan
+
+
+def increment(event: InputEvent) -> OutputEvent: # noqa: D103
+    value = event.input
+    value += 1
+    return OutputEvent(isLegal=True, result=value)
```
xintongsong: This doesn't look right. It assumes an action can only emit one event, as the return value, which is incorrect.

python/flink_agents/api/event.py
xintongsong: We should not need this. If an output is not legal, it should not be outputted at all.

python/flink_agents/api/event.py
```diff
@@ -0,0 +1,59 @@
...
+    result: Any
```
xintongsong: Maybe name it `output`, to align with InputEvent.

python/flink_agents/api/event.py
```diff
@@ -0,0 +1,59 @@
...
+#################################################################################
+
+from abc import ABC
+from typing import Any
+from uuid import UUID, uuid4
+
+from pydantic import BaseModel, Field
+
+
+class Event(BaseModel, ABC, extra="allow"):
```
xintongsong, line 25: If allow extra, how do we make sure it's serializable?

python/flink_agents/plan/workflow_plan.py
wenjin272: yes, thank you xu.

python/flink_agents/plan/action.py
wenjin272: I add a func signature check logic when init one Action, and the check logic will be corrected after we import State and RunnerContext.

python/flink_agents/plan/function.py
wenjin272: make sense

python/flink_agents/plan/tests/test_workflow_plan.py
wenjin272: Fully agree. I add some comments on this function , or I can delete this test case and add it back after we import RunnerContext.

python/flink_agents/api/event.py
wenjin272: I add a validator for extra property, to make sure these are instance of BaseModel, or can be serialized to json.

python/flink_agents/plan/action.py
```diff
@@ -0,0 +1,70 @@
...
+        #TODO: update check logic after import State and RunnerContext.
+        params = inspect.signature(func).parameters
+        if len(params) != 1:
+            err_msg = "Action function must have exactly 1 parameter"
+            raise TypeError(err_msg)
+        for i, param in enumerate(params.values()):
+            if i == 0:
+                if not issubclass(param.annotation, Event):
+                    err_msg = "Action function first parameter must be Event"
+                    raise TypeError(err_msg)
```
xintongsong: What if `exec` is a java function?

python/flink_agents/plan/function.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class JavaFunction(Function):
+    """Descriptor for a java callable function."""
+
+    def __call__(self, *args: Tuple[Any, ...], **kwargs: Dict[str, Any]) -> Any:
+        """Execute the stored function with provided arguments."""
+
+    def check_signature(self, checker: Callable) -> None:
+        """Check function signature is legal or not."""
```
xintongsong: Mark with TODO

python/flink_agents/api/tests/test_event.py
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+#################################################################################
+from typing import Type
+
+import pytest
+
+from flink_agents.api.event import Event
+
+
+def test_event_serializable() -> None: #noqa D103
```
xintongsong: Why do we need this `#noqa D103`?

python/flink_agents/plan/tests/test_workflow_plan.py
xintongsong: I'd suggest to remove it for now. And we probably should also raise a warning when the action has a return value, as it will be ignored.

python/flink_agents/api/event.py
```diff
@@ -0,0 +1,65 @@
...
+    id : UUID
+        Unique identifier for the event, automatically generated using uuid4.
+    """
+    id: UUID = Field(default_factory=uuid4)
+
+    @model_validator(mode='after')
+    def validate_extra(self) -> 'Event':
+        """Make sure all extra properties are serializable."""
+        for value in self.__pydantic_extra__.values():
+            if isinstance(value, BaseModel):
```
xintongsong: What if value as a BaseModel contains non-serializable extras?

python/flink_agents/api/event.py
```diff
@@ -0,0 +1,65 @@
...
+    these properties are required isinstance of BaseModel, or json serializable.
+
+    Attributes:
+    ----------
+    id : UUID
+        Unique identifier for the event, automatically generated using uuid4.
+    """
+    id: UUID = Field(default_factory=uuid4)
+
+    @model_validator(mode='after')
```
xintongsong, line 36: When is this validator called?

python/flink_agents/api/event.py
```diff
@@ -0,0 +1,65 @@
...
+            json.dumps(value)
+        return self
+
+
+class InputEvent(Event):
+    """Event generated by the framework, carrying an input data that
+    arrives at the workflow.
+    """
+
+    input: Any
```
xintongsong, line 53: Is `Any` serializable?

python/flink_agents/plan/tests/test_action.py
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+#################################################################################
+import pytest
+
+from flink_agents.api.event import InputEvent, OutputEvent
+from flink_agents.plan.action import Action
+from flink_agents.plan.function import PythonFunction
+
+
+def increment(event: InputEvent) -> OutputEvent: # noqa: D103
```
xintongsong: increment -> legal_action
decrement -> illegal_action

python/flink_agents/plan/tests/test_action.py
```diff
@@ -0,0 +1,47 @@
...
+    value = event.input
+    value += 1
+    return OutputEvent(output=value)
+
+def decrement(value: int) -> OutputEvent: # noqa: D103
+    value -= 1
+    return OutputEvent(output=value)
+
+def test_action_signature() -> None: # noqa: D103
```
xintongsong: Separate into two text cases

python/flink_agents/plan/tests/test_action.py
```diff
@@ -0,0 +1,47 @@
...
+
+from flink_agents.api.event import InputEvent, OutputEvent
+from flink_agents.plan.action import Action
+from flink_agents.plan.function import PythonFunction
+
+
+def increment(event: InputEvent) -> OutputEvent: # noqa: D103
+    value = event.input
+    value += 1
+    return OutputEvent(output=value)
```
xintongsong: Should not have return value

python/flink_agents/api/event.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Attributes:
+    ----------
+    id : UUID
+        Unique identifier for the event, automatically generated using uuid4.
+    """
+    id: UUID = Field(default_factory=uuid4)
+
+    @model_validator(mode='after')
+    def validate_extra(self) -> 'Event':
```
xintongsong, line 37: This could be a reusable util. There could be more classes that we need to guarantee serializable.

python/flink_agents/plan/action.py
```diff
@@ -0,0 +1,70 @@
...
+        """" Checker for action function signature."""
+        #TODO: update check logic after import State and RunnerContext.
+        params = inspect.signature(func).parameters
+        if len(params) != 1:
+            err_msg = "Action function must have exactly 1 parameter"
+            raise TypeError(err_msg)
+        for i, param in enumerate(params.values()):
+            if i == 0:
+                if not issubclass(param.annotation, Event):
+                    err_msg = "Action function first parameter must be Event"
```
xintongsong: Why having these 2 different error messages? I think it would be more clear to just show the expected and current signatures.

python/flink_agents/api/event.py
wenjin272: 1. Try model_dump_json() to check if any object contains is serializable
2. Override __setattr__  method to check serializable when add field after construct.

python/flink_agents/api/tests/test_event.py
wenjin272: Only for passing code style check that public method needs docs string.

python/flink_agents/plan/action.py
wenjin272: make sense

python/flink_agents/api/tests/test_event.py
```diff
@@ -18,16 +18,25 @@
...
+from pydantic_core import PydanticSerializationError
 
-from flink_agents.api.event import Event
+from flink_agents.api.event import Event, InputEvent, OutputEvent
 
 
-def test_event_serializable() -> None: #noqa D103
-    event = Event(a="1")
-    event.model_dump_json()
+def test_event_serializable_valid() -> None: #noqa D103
```
xintongsong: test_event_init_serializable
test_event_init_non_serializable
test_event_setattr_serializable
test_event_setattr_non_serializable

python/flink_agents/plan/function.py
```diff
@@ -76,9 +76,16 @@ def from_callable(func: Callable) -> Function:
             __func=func,
         )
 
-    def check_signature(self, checker: Callable) -> None:
-        """Apply external check logic to function signature."""
-        checker(self.__get_func())
+    def check_signature(self, *args: Tuple[Any, ...]) -> None:
```
xintongsong, line 81: I think we need more comprehensive tests for this function.

xintongsong (approved):
LGTM.

@lihaosky Please also take a look at this PR.

Issue #13, opened 2025-06-12
[Feature][plan][java] Introduce Workflow, Event, Action & Callable in the Java Workflow Plan


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce Workflow, Event, Action & Callable in the Java Workflow Plan

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #23, opened 2025-06-13
[Feature][plan][java] Introduce Workflow, Event, Action & Callable in…

wenjin272:
Linked issue: #13


api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -0,0 +1,65 @@
...
+    public void setAttr(String name, Object value) {
+        attributes.put(name, value);
+        checkSerializable();
+    }
+
+    public void checkSerializable() {
+        try {
+            SerializableChecker.checkSerializable(this);
+        } catch (JsonProcessingException e) {
+            throw new InvalidParameterException(e.getMessage());
```
xintongsong: Why do we need this wrapping?

api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.core.JsonProcessingException;
+
+import java.io.Serializable;
+import java.security.InvalidParameterException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+/** Base class for all event types in the system. */
+public abstract class Event implements Serializable {
```
xintongsong: Why implementing `Serializable`? Does jackson require it?

api/src/main/java/org/apache/flink/agents/api/InputEvent.java
```diff
@@ -0,0 +1,35 @@
...
+
+/** Event generated by the framework, carrying an input data that arrives at the workflow. */
+public class InputEvent extends Event {
+    private static final long serialVersionUID = 1L;
+    private final Object input;
+
+    public InputEvent(Object input) {
+        super();
+        this.input = input;
+        super.checkSerializable();
```
xintongsong: It's hard to guarantee that all the sub-classes will call this. I think we can only do this when calling `context.sendEvent`.

api/src/main/java/org/apache/flink/agents/api/OutputEvent.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Event representing a result from workflow. By generating an OutputEvent, actions can emit output
+ * data.
+ */
+public class OutputEvent extends Event {
+    private static final long serialVersionUID = 1L;
+    private final Object output;
+
+    public OutputEvent(Object output) throws JsonProcessingException {
```
xintongsong: Unused exception.

api/src/main/java/org/apache/flink/agents/api/utils/SerializableChecker.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.utils;
+
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.core.JsonProcessingException;
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.ObjectMapper;
+
+/** Util class for checking object serializable. */
+public final class SerializableChecker {
```
xintongsong: If we do the check in runtime, we should also move this out of the api module.

api/src/test/java/org/apache/flink/agents/api/TestEvent.java
```diff
@@ -0,0 +1,71 @@
...
+    public static class SerializableClass {
+        private final int a;
+
+        public SerializableClass(int a) {
+            this.a = a;
+        }
+
+        public int getA() {
+            return a;
+        }
```
xintongsong: unused

api/src/test/java/org/apache/flink/agents/api/TestEvent.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public static class UnserializableClass {
```
xintongsong: How is this unserializable?

api/src/test/java/org/apache/flink/agents/api/TestEvent.java
```diff
@@ -0,0 +1,71 @@
...
+        private final int a;
+
+        public UnserializableClass(int a) {
+            this.a = a;
+        }
+    }
+
+    @Test
+    public void testEventInitSerializable() {
+        InputEvent event = new InputEvent(new SerializableClass(1));
```
xintongsong: Unused local variable `event`

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
```diff
@@ -0,0 +1,58 @@
...
+    public JavaFunction(String qualName, String methodName, Class<?>[] parameterTypes) {
+        this.qualName = qualName;
+        this.methodName = methodName;
+        this.parameterTypes = parameterTypes;
+    }
+
+    @Override
+    public Object call(Object... args) throws Exception {
+        Class<?> clazz = Class.forName(qualName);
+        Method method = clazz.getMethod(methodName, parameterTypes);
```
xintongsong: We should only do this once.

plan/src/main/java/org/apache/flink/agents/plan/Action.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Representation of a workflow action with event listening and function execution.
+ *
+ * <p>This class encapsulates a named workflow action that listens for specific event types and
+ * executes an associated function when those events occur.
+ */
+public class Action implements Serializable {
+    private final String name;
+    private final Function exec;
+    private final List<Class<?>> listenEventTypes;
```
xintongsong:
```suggestion
    private final List<Class<? extends Event>> listenEventTypes;
```

plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.apache.flink.agents.plan;
+
+import java.util.List;
+import java.util.Map;
+
+/** Workflow plan compiled from user defined workflow. */
+public class WorkflowPlan {
+    private final Map<Class<?>, List<Action>> actions;
```
xintongsong:
```suggestion
    private final Map<Class<? extends Event>, List<Action>> actions;
```

api/pom.xml
```diff
@@ -28,4 +28,17 @@ under the License.
...
+            <artifactId>flink-shaded-jackson</artifactId>
+            <version>2.14.2-16.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>5.10.1</version>
+            <scope>compile</scope>
+        </dependency>
+    </dependencies>
```
xintongsong: 1. `junit` should be a common dependency. Should not add it for each module.
2. I'm not entirely sure whether jackson should also be a common dependency. But how do you make sure the API and Plan modules depends on the same jackson version?

api/src/main/java/org/apache/flink/agents/api/Event.java
wenjin272: Jackson doesn't require this, I removed it.

api/src/main/java/org/apache/flink/agents/api/InputEvent.java
wenjin272: make sense

plan/src/main/java/org/apache/flink/agents/plan/utils/SerializableChecker.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.plan.utils;
+
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.core.JsonProcessingException;
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.ObjectMapper;
+
+/** Util class for checking object serializable. */
+public final class SerializableChecker {
```
xintongsong: I'd suggest to name this `JsonSerializationUtils` or `JsonUtils`. Later we can include more json serialization related util methods.

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
```diff
@@ -0,0 +1,61 @@
...
+        this.qualName = qualName;
+        this.methodName = methodName;
+        this.parameterTypes = parameterTypes;
+    }
+
+    @Override
+    public Object call(Object... args) throws Exception {
+        if (method == null) {
+            Class<?> clazz = Class.forName(qualName);
+            method = clazz.getMethod(methodName, parameterTypes);
```
lihaosky: Should we validate the method exists or put this in constructor?

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Arrays;
+
+/** Represent a Java function. */
+public class JavaFunction implements Function {
+    private final String qualName;
+    private final String methodName;
+    private final Class<?>[] parameterTypes;
+    private Method method;
+
+    public JavaFunction(String qualName, String methodName, Class<?>[] parameterTypes) {
```
lihaosky: Should there be a constructor taking `Class` directly instead of class name?

plan/src/main/java/org/apache/flink/agents/plan/Action.java
```diff
@@ -0,0 +1,55 @@
...
+    private final String name;
+    private final Function exec;
+    private final List<Class<? extends Event>> listenEventTypes;
+
+    public Action(String name, Function exec, List<Class<? extends Event>> listenEventTypes)
+            throws Exception {
+        this.name = name;
+        this.exec = exec;
+        this.listenEventTypes = listenEventTypes;
+        exec.checkSignature(new Class[] {Event.class});
```
lihaosky, line 41: Looks the function can only take Event as input. Should we just have a `EventFunction` subclass for `Function`?

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
wenjin272: I put this in constructor, so that the validation will be applied when create object rather than call this function in runtime.

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
wenjin272: make sense

plan/src/main/java/org/apache/flink/agents/plan/Action.java
wenjin272: I add an implementation of Function named ActionFunction which only used for Action, and move the signature check from Action to ActionFunction. But I'm not sure whether this wrapper  is appropriate, because signature check is a general feature for function. What do you think about this @xintongsong

xintongsong: @lihaosky
Let me clarify this a bit. The function of an action should take more arguments than just an `Event`. E.g., a runtime context for emitting new events and accessing resources, and maybe also the state (I said "maybe" because alternatively we can get the state from the context). Those classes will be introduced in follow-up PRs, and the target signature for this check will be updated accordingly.

@wenjin272
Checking the signature in `Action`, or introducing another layer of abstraction `ActionFunction` and checking there, I'm okay with either ways.

xintongsong: @wenjin272 
After looking into the current implementation of `ActionFunction`, I tend to think that we should not introduce it.

With the current implementation, we would need to first construct a `JavaFunction` or `PythonFunction`, then wrap it into an `ActionFunction` only for the signature check, and then wrap this `ActionFunction` again into an `Action. Such complexity is unnecessary IMO.

Alternatively, we may introduce `ActionFunction` as an interface rather than a class. That means we would need a `JavaActionFunction` that extends `JavaFunction` and implements the `ActionFunction` interface, and do the same for `PythonActionFunction`. This is also complex.

I think the simplest way is just not to introduce the `ActionFunction` abstraction and do the signature checks directly in `Action`.

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.lang.reflect.Method;
+import java.util.Arrays;
+
+/** Represent a Java function. */
+public class JavaFunction implements Function {
+    private final String qualName;
+    private final String methodName;
+    private final Class<?>[] parameterTypes;
+    private final Class<?> clazz;
```
xintongsong: Seems we don't need the `clazz` as a field of the class.

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
```diff
@@ -0,0 +1,70 @@
...
+    private final Method method;
```
xintongsong: `Method` is non serializable. And we need the Function to be serializable in order to deploy the workflow plan across distributed TMs. This is why the Function only contains qualName, methodName and parameterTypes in the first place.

I'm fine with checking the existence of method early in the constructor. But I think we need a way to tell the serializer that `method` should not be serialized. E.g., in Java Serialization we use the keyword `transient`. Not sure how it's done in Jackson.

plan/src/main/java/org/apache/flink/agents/plan/Action.java
wenjin272: I will remove the ActionFunction wrapper

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
wenjin272: I add JsonProperty and JsonIgnore annotations to mark not serialize `method`

plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
```diff
@@ -0,0 +1,41 @@
...
+
+/** Workflow plan compiled from user defined workflow. */
+public class WorkflowPlan {
+    private final Map<Class<? extends Event>, List<Action>> actions;
+
+    public WorkflowPlan(Map<Class<? extends Event>, List<Action>> actions) {
+        this.actions = actions;
+    }
+
+    public List<Action> getAction(Class<?> type) {
```
xintongsong:
```suggestion
    public List<Action> getAction(Class<? extends Event> type) {
```

Issue #25, opened 2025-06-17
[Feature] Add the basic components required to implement feedback


HuangXingBo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Extract the basic components required to implement feedback from the [flink-statefun](https://github.com/apache/flink-statefun) project

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #30, opened 2025-06-18
[Feature][runtime] Adds the basic components from flink-statefun to implement feedback ability

HuangXingBo:
Linked issue: #25 

### Purpose of change

Adds the basic components from flink-statefun to implement feedback ability

### Tests

The tests from flink-statefun

### API

No

### Documentation

No

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -7,10 +7,5 @@ The Apache Software Foundation (http://www.apache.org/).
 Apache Flink
 Copyright 2014-2025 The Apache Software Foundation
 
-This project contains the following code copied from Apache Flink Stateful Functions:
-runtime/src/main/java/org/apache/flink/agents/runtime/common/*
-runtime/src/main/java/org/apache/flink/agents/runtime/feedback/*
-runtime/src/main/java/org/apache/flink/agents/runtime/logger/*
-runtime/src/main/java/org/apache/flink/agents/runtime/memory/*
-runtime/src/main/java/org/apache/flink/agents/runtime/queue/*
-Original Copyright: 2014-2020 The Apache Software Foundation
+Apache Flink Stateful Functions
+Copyright: 2014-2023 The Apache Software Foundation


HuangXingBo:
cc @xintongsong @GreatEugenius @codenohup

runtime/src/main/java/org/apache/flink/agents/runtime/feedback/Checkpoints.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.io.OutputStream;
+import java.util.Objects;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.function.Supplier;
+
+/**
+ * NOTE: This source code was copied from the <a
+ * href="https://github.com/apache/flink-statefun">flink-statefun</a>
```
lihaosky, line 32: Looks we copied lots of code from other projects directly, what's the main reason for copying/what did we modify after copying? Will there be problem if we need to import their lib later?

lihaosky (commented):
Thanks for the PR. I have a high level question: what's the feedback ability. Is it for subscribe to and run events from FeedbackChannel?

HuangXingBo:
Thanks a lot for your review @lihaosky @codenohup . All components in this PR are extracted from the flink-statefun project. There are two main reasons for not directly depends on the flink-statefun project. The first reason is that the project itself is no longer in a state of development. It would be a burden to maintain an additional branch or release of flink-statefun for us, which goes against our original intention of rapid development and iteration in flink-agents. The second reason is that the core components we need are not so many in the flink-statefun, but they are scattered in various parts of the project. Depending on the entire flink-statefun project will make our project very heavy. Regarding the modification of flink-statefun, you can refer to my poc https://github.com/HuangXingBo/flink-agents/commit/07702dbf4f41edecc1dd01999414e4d0f23b55e7. The main modifications are in operators, and allowing operators to accept Flink DataStream as input and output.

xintongsong (commented):
@HuangXingBo, thanks for working on this. LGTM in general.

Just one comment. I think we should mention in the NOTICE file that the project contains codes copied from Flink and Statefun. See [Paimon's NOTICE file](https://github.com/apache/paimon/blob/master/NOTICE) for an example.

```
This product includes software developed at
The Apache Software Foundation (http://www.apache.org/).

Apache Flink
Copyright 2014-2023 The Apache Software Foundation

...
```

HuangXingBo:
@xintongsong Thanks a lot for the review. I have updated the NOTICE file. Please take another look.

NOTICE
```diff
@@ -2,4 +2,15 @@ Apache Flink Agents
...
+Apache Flink
+Copyright 2014-2025 The Apache Software Foundation
+
+This project contains the following code copied from Apache Flink Stateful Functions:
+runtime/src/main/java/org/apache/flink/agents/runtime/common/*
+runtime/src/main/java/org/apache/flink/agents/runtime/feedback/*
+runtime/src/main/java/org/apache/flink/agents/runtime/logger/*
+runtime/src/main/java/org/apache/flink/agents/runtime/memory/*
+runtime/src/main/java/org/apache/flink/agents/runtime/queue/*
+Original Copyright: 2014-2020 The Apache Software Foundation
```
xintongsong: 1. Let's remove the detailed directories. We need to keep the NOTICE file as concise as possible. And we are likely to make further changes to the copied codes, so the listed 'xx/*' directories will also contain non-copied codes.
2. I think the copyright years of statefun is 2014-2023, as suggested by its [NOTIE file](https://github.com/apache/flink-statefun/blob/master/NOTICE). This is also aligned with its commit history (last commit in Sep 2023).

```suggestion
Apache Flink Stateful Functions
Copyright: 2014-2023 The Apache Software Foundation
```

Issue #33, opened 2025-06-23
[Feature] Adds support of Python Environment Manager


HuangXingBo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Flink Agents use the same python dependency management components of PyFlink

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #37, opened 2025-06-24
[Feature][runtime] Adds support of Python Environment Manager

HuangXingBo:
Linked issue: #33

### Purpose of change

Adds support of python dependency management.

### Tests

`PythonEnvironmentManagerTest`

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -50,7 +50,7 @@ under the License.
         <dependency>
             <groupId>com.alibaba</groupId>
             <artifactId>pemja</artifactId>
-            <version>0.6.0</version>
+            <version>${pemja.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.flink</groupId>
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -47,6 +47,23 @@ under the License.
             <version>${flink.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.alibaba</groupId>
+            <artifactId>pemja</artifactId>
+            <version>0.6.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-python</artifactId>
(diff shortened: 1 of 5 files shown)


runtime/pom.xml
```diff
@@ -47,6 +47,23 @@ under the License.
             <version>${flink.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.alibaba</groupId>
+            <artifactId>pemja</artifactId>
+            <version>0.6.0</version>
```
codenohup: Should the pemja version be specified in the properties section?

runtime/src/main/java/org/apache/flink/agents/runtime/env/PythonEnvironmentManager.java
```diff
@@ -0,0 +1,110 @@
...
+import pemja.core.PythonInterpreterConfig;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.python.util.PythonDependencyUtils.PARAM_DELIMITER;
+
+/**
+ * The base class of python environment manager which is used to create the PythonEnvironment
+ * object. It's used to run python functions in embedded Python environment.
```
codenohup, line 34: Could we add test cases that verify and explain how to execute the Python function with `PythonEnvironmentManager`?

codenohup (commented):
Hi, @HuangXingBo 
Thank you for your PR. This looks good to me overall — I have two small questions, PTAL.

runtime/src/main/java/org/apache/flink/agents/runtime/env/PythonEnvironmentManager.java
HuangXingBo: This will depend on the Python environment and dependencies. I feel this part is more suitable for e2e test.

runtime/pom.xml
HuangXingBo: make sense

Issue #14, opened 2025-06-12
[Feature][plan] Support converting between Java & Python Workflow Plans


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support converting between Java & Python Workflow Plans

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wenjin272:
@PatrickRen and I are willing to submit PRs for this issue.

PR #39, opened 2025-06-25
[plan][python] Support serialize/deserialize workflow plan to/from json.

wenjin272:
Linked issue: #14 

#14 will have 3 pr, include python serialization, java serialization and compatibility check.

This is the first pr which focus on serializing/deserializing python workflow plan to/from json.


python/flink_agents/plan/tests/test_action.py
```diff
@@ -33,15 +35,43 @@ def test_action_signature_legal() -> None: # noqa: D103
...
+    func = PythonFunction.from_callable(legal_signature)
+    return Action(
+        name="legal",
+        exec=func,
+        listen_event_types=[f'{InputEvent.__module__}.{InputEvent.__qualname__}'],
+    )
+
+current_dir = Path(__file__).parent
+
+def test_action_serialize(action: Action) -> None:  # noqa: D103
```
xintongsong, line 61: The test relies on that the serialized string is exactly the same as the json files, which might be too strict.

xintongsong (approved):
LGTM

Issue #27, opened 2025-06-17
[Feature] Adds the support of executing the given workflow


HuangXingBo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Adds the support of executing the given workflow

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #43, opened 2025-06-26
[Feature][runtime] Introducing the FlinkRunnerContext component to support Python workflow execution

GreatEugenius:
Linked issue: #27 

### Purpose of change

Introducing the FlinkRunnerContext component to support Python workflow execution.

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/PythonFunction.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/PythonFunction.java
@@ -37,4 +37,12 @@ public class PythonFunction implements Function {
     public void checkSignature(Class<?>[] parameterTypes) throws Exception {
         throw new UnsupportedOperationException();
     }
+
+    public String getModule() {
+        return module;
+    }
+
+    public String getQualName() {
+        return qualName;
+    }
 }
--- a/python/flink_agents/runtime/flink_runner_context.py
+++ b/python/flink_agents/runtime/flink_runner_context.py
@@ -1,0 +1,59 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#################################################################################
+from typing import Any
+
+import cloudpickle
+from typing_extensions import override
+
+from flink_agents.api.event import Event, OutputEvent
+from flink_agents.api.runner_context import RunnerContext
+
+
+def create_flink_runner_context(j_runner_context: Any):
+    """Used to create a FlinkRunnerContext Python object in Pemja environment."""
+    return FlinkRunnerContext(j_runner_context)
+
+
+def convert_to_python_object(bytesObject: bytes):
+    """Used for deserializing Python objects."""
+    return cloudpickle.loads(bytesObject)
+
+
+class FlinkRunnerContext(RunnerContext):
+    """Providing context for workflow execution in Flink Environment.
+
+    This context allows access to event handling.
+    """
+
+    def __init__(self, j_runner_context: Any):
+        self._j_runner_context = j_runner_context
+
+    @override
+    def send_event(self, event: Event):
+        """Send an event to the workflow for processing.
+        Parameters
+        ----------
+        event : Event
+            The event to be processed by the workflow system.
+        """
+        try:
+            data = event.output if isinstance(event, OutputEvent) else event
+            class_path = f"{event.__class__.__module__}.{event.__class__.__qualname__}"
+            self._j_runner_context.sendEvent(class_path, cloudpickle.dumps(data))
+        except Exception as e:
+            raise RuntimeError(f"Failed to send event {event} to runner context: {e}")
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.runtime;
+
+import org.apache.flink.agents.plan.PythonFunction;
+import org.apache.flink.agents.runtime.context.PythonRunnerContext;
+import org.apache.flink.agents.runtime.env.EmbeddedPythonEnvironment;
+import org.apache.flink.agents.runtime.env.PythonEnvironmentManager;
+import org.apache.flink.agents.runtime.message.PythonEventMessage;
+import pemja.core.PythonInterpreter;
+
+import java.util.List;
+
+/** Execute the corresponding Python action in the workflow. */
+public class PythonRunner {
+
+    private static final String IMPORT_FLINK_RUNNER_CONTEXT =
+            "from flink_agents.runtime import flink_runner_context";
+    private static final String CREATE_FLINK_RUNNER_CONTEXT =
+            "flink_runner_context.create_flink_runner_context";
+    private static final String CONVERT_TO_PYTHON_OBJECT =
+            "flink_runner_context.convert_to_python_object";
+
+    private PythonInterpreter interpreter;
+    private PythonRunnerContext runnerContext;
+
+    public PythonRunner(PythonEnvironmentManager environmentManager) throws Exception {
+        environmentManager.open();
+        EmbeddedPythonEnvironment env =
+                (EmbeddedPythonEnvironment) environmentManager.createEnvironment();
+
+        interpreter = env.getInterpreter();
+        interpreter.exec(IMPORT_FLINK_RUNNER_CONTEXT);
+    }
+
+    public List executePythonFunction(
+            PythonFunction pythonFunction, PythonEventMessage<?> inputMessage) {
+        String modulePath = pythonFunction.getModule();
+        int moduleLastDotIndex = modulePath.lastIndexOf(".");
+        if (moduleLastDotIndex == -1) {
+            throw new IllegalArgumentException(
+                    "Invalid Python module path: " + pythonFunction.getModule());
+        }
+
+        String packagePath = modulePath.substring(0, moduleLastDotIndex);
+        String className = modulePath.substring(moduleLastDotIndex + 1);
+
+        Object eventKey = inputMessage.getKey();
+        runnerContext.setKey(eventKey);
+        runnerContext.clearAllEvents();
+
+        try {
+            interpreter.exec("from " + packagePath + " import " + className);
+
+            Object pythonRunnerContextObject =
+                    interpreter.invoke(CREATE_FLINK_RUNNER_CONTEXT, runnerContext);
+
+            Object pythonEventObject =
+                    interpreter.invoke(CONVERT_TO_PYTHON_OBJECT, inputMessage.getEvent());
+
+            interpreter.invoke(
+                    pythonFunction.getQualName(), pythonEventObject, pythonRunnerContextObject);
+        } catch (Exception e) {
+            throw new RuntimeException(
+                    "Failed to execute Python function: " + pythonFunction.getQualName(), e);
+        }
+
+        return runnerContext.drainEvents();
+    }
+}
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/context/PythonRunnerContext.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/context/PythonRunnerContext.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.runtime.context;
+
+import org.apache.flink.agents.runtime.message.EventMessage;
+import org.apache.flink.agents.runtime.message.PythonEventMessage;
+
+import javax.annotation.concurrent.NotThreadSafe;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class is used to manage the execution context when interacting with Python functions,
+ * including temporarily saving the key of the current Event and collecting new events generated
+ * during the execution of actions.
+ */
+@NotThreadSafe
+public class PythonRunnerContext {
+    private Object key;
+    private List<EventMessage<?, ?>> events;
+
+    public PythonRunnerContext() {
+        this(null);
+    }
+
+    public PythonRunnerContext(Object key) {
+        this.key = key;
+        this.events = new ArrayList<>();
+    }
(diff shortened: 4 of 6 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.flink.agents.runtime.context.PythonRunnerContext;
+import org.apache.flink.agents.runtime.env.EmbeddedPythonEnvironment;
+import org.apache.flink.agents.runtime.env.PythonEnvironmentManager;
+import org.apache.flink.agents.runtime.message.PythonEventMessage;
+import pemja.core.PythonInterpreter;
+
+import java.util.List;
+
+/** Execute the corresponding Python action in the workflow. */
+public class PythonRunner {
```
xintongsong:
```suggestion
public class PythonActionExecutor {
```

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
```diff
@@ -0,0 +1,85 @@
...
+            PythonFunction pythonFunction, PythonEventMessage<?> inputMessage) {
+        String modulePath = pythonFunction.getModule();
+        int moduleLastDotIndex = modulePath.lastIndexOf(".");
+        if (moduleLastDotIndex == -1) {
+            throw new IllegalArgumentException(
+                    "Invalid Python module path: " + pythonFunction.getModule());
+        }
+
+        String packagePath = modulePath.substring(0, moduleLastDotIndex);
+        String className = modulePath.substring(moduleLastDotIndex + 1);
```
xintongsong: This seems not correct. Why would the module path includes the class name?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
```diff
@@ -0,0 +1,85 @@
...
+    public PythonRunner(PythonEnvironmentManager environmentManager) throws Exception {
+        environmentManager.open();
+        EmbeddedPythonEnvironment env =
+                (EmbeddedPythonEnvironment) environmentManager.createEnvironment();
+
+        interpreter = env.getInterpreter();
+        interpreter.exec(IMPORT_FLINK_RUNNER_CONTEXT);
+    }
+
+    public List executePythonFunction(
```
xintongsong: It would be better to implement `PythonFunction.call()`.

1. set the interpreter to the `PythonFunction`
2. call `PythonFunction.call()` and pass in the java arguments
3. convert java arguments into python objects and call the python version `PythonFunction.call()` in the java `PythonFunction.call()`

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
```diff
@@ -0,0 +1,85 @@
...
+public class PythonRunner {
+
+    private static final String IMPORT_FLINK_RUNNER_CONTEXT =
+            "from flink_agents.runtime import flink_runner_context";
+    private static final String CREATE_FLINK_RUNNER_CONTEXT =
+            "flink_runner_context.create_flink_runner_context";
+    private static final String CONVERT_TO_PYTHON_OBJECT =
+            "flink_runner_context.convert_to_python_object";
+
+    private PythonInterpreter interpreter;
```
xintongsong: Should this be `final`?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
```diff
@@ -0,0 +1,85 @@
...
+    private PythonRunnerContext runnerContext;
```
xintongsong: This is never initialized.

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
HuangXingBo: final

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
HuangXingBo: never initialize?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
HuangXingBo: `List<EventMessage<?, ?>>`?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
```diff
@@ -0,0 +1,85 @@
...
+            interpreter.exec("from " + packagePath + " import " + className);
+
+            Object pythonRunnerContextObject =
+                    interpreter.invoke(CREATE_FLINK_RUNNER_CONTEXT, runnerContext);
+
+            Object pythonEventObject =
+                    interpreter.invoke(CONVERT_TO_PYTHON_OBJECT, inputMessage.getEvent());
+
+            interpreter.invoke(
+                    pythonFunction.getQualName(), pythonEventObject, pythonRunnerContextObject);
```
HuangXingBo: Why not expose a method on the Python side and call it once

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
HuangXingBo: Can these logics be extracted into a util method?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonRunner.java
```diff
@@ -0,0 +1,85 @@
...
+    private PythonInterpreter interpreter;
+    private PythonRunnerContext runnerContext;
+
+    public PythonRunner(PythonEnvironmentManager environmentManager) throws Exception {
+        environmentManager.open();
+        EmbeddedPythonEnvironment env =
+                (EmbeddedPythonEnvironment) environmentManager.createEnvironment();
+
+        interpreter = env.getInterpreter();
+        interpreter.exec(IMPORT_FLINK_RUNNER_CONTEXT);
```
HuangXingBo: Can these logics be extracted into init method? Or we can add a new open method. It is not a good idea to put these logics that may throw exceptions into the constructor.

python/flink_agents/runtime/flink_runner_context.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def create_flink_runner_context(j_runner_context: Any) -> "FlinkRunnerContext":
+    """Used to create a FlinkRunnerContext Python object in Pemja environment."""
+    return FlinkRunnerContext(j_runner_context)
+
+
+def convert_to_python_object(bytesObject: bytes) -> Any:
+    """Used for deserializing Python objects."""
+    return cloudpickle.loads(bytesObject)
```
xintongsong: Why is this in the `flink_runner_context` module?

python/flink_agents/runtime/flink_runner_context.py
```diff
@@ -0,0 +1,68 @@
...
+    def send_event(self, event: Event) -> None:
+        """Send an event to the workflow for processing.
+
+        Parameters
+        ----------
+        event : Event
+            The event to be processed by the workflow system.
+        """
+        try:
+            data = event.output if isinstance(event, OutputEvent) else event
```
xintongsong: Why do we have this difference?

runtime/src/main/java/org/apache/flink/agents/runtime/message/PythonEventMessage.java
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.flink.agents.runtime.message;
+
+import java.util.Arrays;
+
+/**
+ * PythonEventMessage is a {@link EventMessage} that carries a Python event in byte array format.
+ *
+ * @param <K> The type of the key.
+ */
+public class PythonEventMessage<K> extends EventMessage<K, byte[]> {
```
xintongsong: What is the purpose of this wrap?

runtime/src/main/java/org/apache/flink/agents/runtime/message/EventMessage.java
```diff
@@ -20,26 +20,37 @@
...
 
-/** Specific message type for data. */
-public class EventMessage<T> implements Message {
+/**
+ * Specific message type for event data.
+ *
+ * @param <K> The type of the key.
+ * @param <T> The type of the event.
+ */
+public class EventMessage<K, T> implements Message {
```
xintongsong:
```suggestion
public class EventMessage<K, E> implements Message {
```

runtime/src/main/java/org/apache/flink/agents/runtime/context/PythonRunnerContext.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public List<EventMessage<?, ?>> drainEvents() {
+        List<EventMessage<?, ?>> list = new ArrayList<>(this.events);
+        clearAllEvents();
+        return list;
+    }
+
+    public void clearAllEvents() {
+        this.events.clear();
+    }
```
xintongsong: Why separating into two methods?

runtime/src/main/java/org/apache/flink/agents/runtime/context/PythonRunnerContext.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public void setKey(Object key) {
+        this.key = key;
+    }
+
+    public void sendEvent(String type, byte[] event) {
+        if (key == null) {
+            throw new IllegalStateException("Key is not set.");
+        }
+        this.events.add(new PythonEventMessage<>(key, event, type));
```
xintongsong: Do we need the key for each message?

xintongsong: Why not wrapping all events into one message?

xintongsong: And why do we need to wrap the events into messages at all?

python/flink_agents/runtime/flink_runner_context.py
```diff
@@ -0,0 +1,68 @@
...
+    return cloudpickle.loads(bytesObject)
+
+
+class FlinkRunnerContext(RunnerContext):
+    """Providing context for workflow execution in Flink Environment.
+
+    This context allows access to event handling.
+    """
+
+    def __init__(self, j_runner_context: Any) -> None:
```
HuangXingBo, line 33: remove `-> None`

runtime/src/main/java/org/apache/flink/agents/runtime/context/PythonRunnerContext.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * This class is used to manage the execution context when interacting with Python functions,
+ * including temporarily saving the key of the current Event and collecting new events generated
+ * during the execution of actions.
+ */
+@NotThreadSafe
+public class PythonRunnerContext {
+    private Object key;
+    private List<EventMessage<?, ?>> events;
```
HuangXingBo: final

runtime/src/main/java/org/apache/flink/agents/runtime/context/PythonRunnerContext.java
```diff
@@ -0,0 +1,67 @@
...
+    public PythonRunnerContext() {
+        this(null);
+    }
+
+    public PythonRunnerContext(Object key) {
+        this.key = key;
+        this.events = new ArrayList<>();
+    }
+
+    public void setKey(Object key) {
```
HuangXingBo: Should we use a method similar to openSession, where the key in the session is the same, and then reset it by closeSession, otherwise there is a possibility that we forget to setKey

python/flink_agents/runtime/flink_runner_context.py
GreatEugenius: This is because of [missing-return-type-special-method (ANN204)], should we remove this format check?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonActionExecutor.java
```diff
@@ -0,0 +1,72 @@
...
+    public List<PythonEvent> executePythonFunction(
+            PythonFunction pythonFunction, PythonEvent inputMessage) {
+        runnerContext.checkNoPendingEvents();
+        pythonFunction.setInterpreter(interpreter);
+
+        try {
+            pythonFunction.call(inputMessage.getEvent());
+        } catch (Exception e) {
+            runnerContext.drainEvents();
+            throw new RuntimeException("Failed to execute the Python function", e);
```
xintongsong: Avoid throwing `RuntimeException` as much as possible. It does not require declaration in the signature, which makes the handling implicit.

runtime/src/main/java/org/apache/flink/agents/runtime/PythonActionExecutor.java
```diff
@@ -0,0 +1,72 @@
...
+        interpreter.exec(IMPORT_FLINK_RUNNER_CONTEXT);
+
+        Object pythonFunctionWrapper =
+                interpreter.invoke(
+                        "flink_runner_context.create_python_function_wrapper", runnerContext);
+        interpreter.set("python_function_wrapper", pythonFunctionWrapper);
+    }
+
+    public List<PythonEvent> executePythonFunction(
+            PythonFunction pythonFunction, PythonEvent inputMessage) {
```
xintongsong:
```suggestion
            PythonFunction function, PythonEvent event) {
```

runtime/src/main/java/org/apache/flink/agents/runtime/context/PythonRunnerContext.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class is used to manage the execution context when interacting with Python functions,
+ * including collecting new events generated during the execution of actions.
+ */
+@NotThreadSafe
+public class PythonRunnerContext {
+    private final List<PythonEvent> events;
```
xintongsong:
```suggestion
    private final List<PythonEvent> pendingEvents;
```

plan/src/main/java/org/apache/flink/agents/plan/PythonFunction.java
```diff
@@ -17,22 +17,40 @@
...
 
+    public void setInterpreter(PythonInterpreter interpreter) {
+        this.interpreter = interpreter;
+    }
+
     @Override
     public Object call(Object... args) throws Exception {
-        throw new UnsupportedOperationException();
+        if (args.length < 1) {
+            throw new IllegalArgumentException("Expected at least 1 arguments: PythonEvent");
```
xintongsong: A `PythonFunction` is not necessarily an action.

python/flink_agents/runtime/flink_runner_context.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        try:
+            class_path = f"{event.__class__.__module__}.{event.__class__.__qualname__}"
+            self._j_runner_context.sendEvent(class_path, cloudpickle.dumps(event))
+        except Exception as e:
+            err_msg = "Failed to send event " + class_path + " to runner context"
+            raise RuntimeError(err_msg) from e
+
+
+class PythonFunctionWrapper:
```
xintongsong: Why is this in the runner context module?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonActionExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public PythonActionExecutor(PythonEnvironmentManager environmentManager) {
+        this.environmentManager = environmentManager;
+        this.runnerContext = new PythonRunnerContext();
+    }
+
+    public void open() throws Exception {
+        environmentManager.open();
+        EmbeddedPythonEnvironment env =
+                (EmbeddedPythonEnvironment) environmentManager.createEnvironment();
```
xintongsong: Shall we change the signature of `PythonEnvironmentManager#createEnvironment` to avoid this casting?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonActionExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+
+        interpreter = env.getInterpreter();
+
+        interpreter.exec(IMPORT_FUNCTION);
+        interpreter.exec(IMPORT_FLINK_RUNNER_CONTEXT);
+        interpreter.exec(IMPORT_PYTHON_JAVA_UTILS);
```
xintongsong: Can these importing statements be combined and executed at once?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonActionExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+                interpreter.invoke(CREATE_FLINK_RUNNER_CONTEXT, runnerContext);
+        interpreter.set(FLINK_RUNNER_CONTEXT_VAR_NAME, pythonRunnerContextObject);
+    }
+
+    public List<PythonEvent> executePythonFunction(PythonFunction function, PythonEvent event)
+            throws Exception {
+        runnerContext.checkNoPendingEvents();
+        function.setInterpreter(interpreter);
+
+        Object pythonRunnerContextObject = interpreter.get(FLINK_RUNNER_CONTEXT_VAR_NAME);
```
xintongsong, line 70: Why do we need to set / get the context to / from the interpreter?

python/flink_agents/plan/function.py
```diff
@@ -141,3 +141,9 @@ def __call__(self, *args: Tuple[Any, ...], **kwargs: Dict[str, Any]) -> Any:
 
     def check_signature(self, *args: Tuple[Any, ...]) -> None:
         """Check function signature is legal or not."""
+
+
+def call_python_function(module: str, qualname: str, func_args: Tuple[Any, ...]) -> Any:
+    """Used to call a Python function in the Pemja environment."""
+    func = PythonFunction(module=module, qualname=qualname)
```
xintongsong, line 148: We probably should avoid creating the function over and over again for each call. But let's keep it as is for now, and leave this as a follow-up issue.

GreatEugenius: Thanks for the feedback! You're absolutely right—repeatedly creating PythonFunction instances could lead to unnecessary overhead. I've created a follow-up issue to track this optimization: [Issue #46](https://github.com/apache/flink-agents/issues/46).

xintongsong (approved):
LGTM

Issue #26, opened 2025-06-17
[Feature] Implements agent execution operator


HuangXingBo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Implements feedback/function operator in java

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #45, opened 2025-06-26
[plan][java] Support serializing / deserializing Java workflow plan to / from JSON

PatrickRen:
Linked issue: #14

### Purpose of change

This pull request implements JSON serializer / deserializer for `Action` and `WorkflowPlan` for Java API. 

There are some known issues considering compatibility with Python, which might be fixed in followup pull requests:
- The type of event that an action listens to can be defined in Java or Python. This PR only considers events defined in Java
- Method `checkSignature` in `PythonFunction` is not implemented, but it is used in the constructor of `Action`. The workaround now when I implement tests is extending the `PythonFunction` and overriding the `checkSignature` method with no-op.

### Tests

Serializers / deserializers are guarded by unit tests.

### API

No.

### Documentation

No.

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/PythonFunction.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/PythonFunction.java
@@ -37,4 +37,12 @@ public class PythonFunction implements Function {
     public void checkSignature(Class<?>[] parameterTypes) throws Exception {
         throw new UnsupportedOperationException();
     }
+
+    public String getModule() {
+        return module;
+    }
+
+    public String getQualName() {
+        return qualName;
+    }
 }
--- a/plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
@@ -19,23 +19,33 @@
 package org.apache.flink.agents.plan;
 
 import org.apache.flink.agents.api.Event;
+import org.apache.flink.agents.plan.serializer.WorkflowPlanJsonDeserializer;
+import org.apache.flink.agents.plan.serializer.WorkflowPlanJsonSerializer;
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
 import java.util.List;
 import java.util.Map;
 
 /** Workflow plan compiled from user defined workflow. */
+@JsonSerialize(using = WorkflowPlanJsonSerializer.class)
+@JsonDeserialize(using = WorkflowPlanJsonDeserializer.class)
 public class WorkflowPlan {
-    private final Map<Class<? extends Event>, List<Action>> actions;
+    private final Map<String, Action> actions;
+    private final Map<Class<? extends Event>, List<Action>> eventTriggerActions;
 
-    public WorkflowPlan(Map<Class<? extends Event>, List<Action>> actions) {
+    public WorkflowPlan(
+            Map<String, Action> actions,
+            Map<Class<? extends Event>, List<Action>> eventTriggerActions) {
         this.actions = actions;
+        this.eventTriggerActions = eventTriggerActions;
     }
 
-    public List<Action> getAction(Class<? extends Event> type) {
-        return actions.get(type);
-    }
-
-    public Map<Class<? extends Event>, List<Action>> getActions() {
+    public Map<String, Action> getActions() {
         return actions;
+    }
+
+    public Map<Class<? extends Event>, List<Action>> getEventTriggerActions() {
+        return eventTriggerActions;
     }
 }
--- a/plan/src/main/java/org/apache/flink/agents/plan/serializer/ActionJsonSerializer.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/serializer/ActionJsonSerializer.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.plan.serializer;
+
+import org.apache.flink.agents.api.Event;
+import org.apache.flink.agents.plan.Action;
+import org.apache.flink.agents.plan.JavaFunction;
+import org.apache.flink.agents.plan.PythonFunction;
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.core.JsonGenerator;
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.SerializerProvider;
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.ser.std.StdSerializer;
+
+import java.io.IOException;
+
+/**
+ * Custom serializer for {@link Action} that handles the serialization of the function and event
+ * types.
+ */
+public class ActionJsonSerializer extends StdSerializer<Action> {
+    public ActionJsonSerializer() {
+        super(Action.class);
+    }
+
+    @Override
+    public void serialize(
+            Action action, JsonGenerator jsonGenerator, SerializerProvider serializerProvider)
+            throws IOException {
+        jsonGenerator.writeStartObject();
+
+        // Write name field
+        jsonGenerator.writeStringField("name", action.getName());
+
+        // Write exec field
+        if (action.getExec() instanceof JavaFunction) {
+            JavaFunction javaFunction = (JavaFunction) action.getExec();
+            serializeJavaFunction(jsonGenerator, javaFunction);
+        } else if (action.getExec() instanceof PythonFunction) {
+            PythonFunction pythonFunction = (PythonFunction) action.getExec();
+            serializePythonFunction(jsonGenerator, pythonFunction);
+        } else {
+            throw new IllegalArgumentException(
+                    "Unsupported function type: " + action.getExec().getClass().getName());
+        }
+
+        // Write listenEventTypes field
+        jsonGenerator.writeFieldName("listenEventTypes");
+        jsonGenerator.writeStartArray();
+        for (Class<? extends Event> eventType : action.getListenEventTypes()) {
+            jsonGenerator.writeString(eventType.getName());
+        }
+        jsonGenerator.writeEndArray();
+
+        jsonGenerator.writeEndObject();
+    }
+
+    private void serializePythonFunction(JsonGenerator gen, PythonFunction func)
+            throws IOException {
+        gen.writeFieldName("exec");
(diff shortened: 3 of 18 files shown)


plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
```diff
@@ -19,23 +19,33 @@
...
 import java.util.List;
 import java.util.Map;
 
 /** Workflow plan compiled from user defined workflow. */
+@JsonSerialize(using = WorkflowPlanJsonSerializer.class)
+@JsonDeserialize(using = WorkflowPlanJsonDeserializer.class)
 public class WorkflowPlan {
-    private final Map<Class<? extends Event>, List<Action>> actions;
+    private final Map<String, Action> actions;
+    private final Map<Class<? extends Event>, List<Action>> eventTriggerActions;
```
wenjin272: As you mentioned in comments, using` Map<Class<? extends Event>, List<Action>>` only works for Java defined Event, because python defined Event may not have correspond Class in java. 

I meet the same issue when implement python part, so I changed the map to `Map<String, String>`,  the key is Event type and the value is action name. This can also decrease the serialize/deserialize complexity.

wenjin272:
As for the second issue, I think we could skip checkSignature of PythonFunction (always return true) currently and add a todo. Because currently the PythonFunction is only used for representing the action defined by Python API, and when construct the python workflow, the signature check will be applied. So there is no need to check signature when construct Action from deserialization. Or, if we can distinguish construct from normal or from deserialization, it may be better to skip check in Action constructor.

codenohup:
Hi, @PatrickRen 
Thank you for your contribution!
I have a small question: From the _checkSignature_ in the _Action_ constructor, it seems that user-defined action methods currently only accept a single event-type parameter. Should we also consider adding a context parameter to support sending events and updating state?

plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
codenohup: Hi, @PatrickRen 
I suggest using _Map<String, List<Action>>_ type to store _eventTriggerActions_, where the key is the unique identifier of the event (e.g., the fully qualified class name).
As @wenjin272  mentioned, there are two main issues:
1. Python events have no corresponding class in Java.
2. Even if we use a _PythonEvent_ wrapper class for Python events, like #43. It would still be difficult to find actions interested in specific events.

codenohup:
emmm, I also encountered an action deserialization failure, which matches the CI error. It seems to be caused by a mismatch in the key names: the _listenEventTypes_ is serialized using the key _listenEventTypes_, but it's deserialized using the key _listen_event_types_.

codenohup:
I'm currently working on implementing a few Flink operators based on this PR to handle workflow processing #26 , but I encountered a non-serializable exception with _WorkflowPlan_. 
image

I resolved this by:
1. Making _WorkflowPlan_ implement the _Serializable_ interface.
2. Overriding the _writeObject_ and _readObject_ methods using _WorkflowPlanJsonSerializer_ and _WorkflowPlanJsonDeserializer_.

You might want to consider incorporating this fix into this PR as well 😊

wenjin272:
> Hi, @PatrickRen Thank you for your contribution! I have a small question: From the _checkSignature_ in the _Action_ constructor, it seems that user-defined action methods currently only accept a single event-type parameter. Should we also consider adding a context parameter to support sending events and updating state?

Currently, there is no implementation of RunnerContext in Java side, I think this should be done in the pr which import Workflow API.

plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
PatrickRen: Thanks for the review! Changed to Map<String, List> as suggested.

wenjin272:
> I also notice another thing: In class `Action` it also a field `listenEventTypes`, which is a list of Java classes currently. Do you think it makes sense to also change it to a list of string? Because type of event that an action listens to could be both Java and Python

I think this makes sense, and actually I have changed it to a list of string in Python Action.

plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
```diff
@@ -18,24 +18,32 @@
...
 import java.util.List;
 import java.util.Map;
 
 /** Workflow plan compiled from user defined workflow. */
+@JsonSerialize(using = WorkflowPlanJsonSerializer.class)
+@JsonDeserialize(using = WorkflowPlanJsonDeserializer.class)
 public class WorkflowPlan {
-    private final Map<Class<? extends Event>, List<Action>> actions;
+    private final Map<String, Action> actions;
+    private final Map<String, List<Action>> eventTriggerActions;
```
xintongsong:
```suggestion
    private final Map<String, Action> actions; // actionName -> action
    private final Map<String, List<Action>> actionsByEvent; // eventClassName -> actions
```

plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
```diff
@@ -18,24 +18,32 @@
...
 
-    public List<Action> getAction(Class<? extends Event> type) {
-        return actions.get(type);
+    public Map<String, Action> getActions() {
+        return actions;
     }
 
-    public Map<Class<? extends Event>, List<Action>> getActions() {
-        return actions;
+    public Map<String, List<Action>> getEventTriggerActions() {
```
xintongsong:
```suggestion
    public Map<String, List<Action>> getActionsByEvent() {
```

plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
xintongsong: I'm aware that the current field name `eventTriggerActions` is aligned with the python workflow plan. But I think `actionsByEvent` is better for readability. Maybe we can have a separate follow-up commit to fix this for both java and python.

plan/src/main/java/org/apache/flink/agents/plan/PythonFunction.java
```diff
@@ -48,7 +48,13 @@ public Object call(Object... args) throws Exception {
 
     // TODO: check Python function signature compatibility with given parameter types
     @Override
-    public void checkSignature(Class<?>[] parameterTypes) throws Exception {
-        throw new UnsupportedOperationException();
+    public void checkSignature(Class<?>[] parameterTypes) throws Exception {}
```
xintongsong, line 51: What is the plan for implementation of this method? Shall we at least create a issue for this? Because the todo comment can easily get overlooked.

codenohup:
> Hi @wenjin272 @codenohup
> 
> Thanks a lot for the review! I addressed your comments and fixed failed tests.
> 
> As suggested by @wenjin272 , I changed checkSignature in PythonFunction to do nothing, and updated related tests.
> 
> @codenohup
> 
> > but I encountered a non-serializable exception with WorkflowPlan.
> 
> Well I think it depends on which serialization style we should use for `WorkflowPlan`: Java object serialization or JSON serialization. If we want to use the Java object serialization way, we need to mark `WorkflowPlan` as `Serializable` as you suggested, but we also need to mark `Action` as `Serializable`, then `Function` also should be `Serializable`, which is a bit difficult to enforce it. So maybe it’s more reasonable to use the JSON serde as replied by @wenjin272 .
> 
> > Should we also consider adding a context parameter to support sending events and updating state?
> 
> This is a good idea, but as @Wenjin said I didn’t find Java implementation of the context, so it’s better to have a followup PR for it.

I have a small question: Would it be possible to implement Java's serializable interface only for _WorkflowPlan_, and then within the Java serialization methods, delegate to your defined JSON-based serialization approach? This would allow us to achieve Java serialization compatibility in a straightforward way, without requiring that _Action_, _Function_, or other components themselves are Java-serializable.

You can take a look at my changes to _WorkflowPlan_ in the #47  — it mainly involves implementing the _Serializable_ interface and overriding the _writeObject_ and _readObject_ methods.

plan/src/main/java/org/apache/flink/agents/plan/serializer/ActionJsonSerializer.java
```diff
@@ -0,0 +1,99 @@
...
+
+        jsonGenerator.writeEndObject();
+    }
+
+    private void serializePythonFunction(JsonGenerator gen, PythonFunction func)
+            throws IOException {
+        gen.writeFieldName("exec");
+        gen.writeStartObject();
+        // Mark it as a Python function
+        gen.writeStringField("func_type", PythonFunction.class.getSimpleName());
```
lihaosky: Define the `func_type` explicitly as static variable `PythonFunction` instead of relying on class name? Serde will break in case `PythonFunction` class gets refactored and change name

plan/src/main/java/org/apache/flink/agents/plan/serializer/ActionJsonSerializer.java
```diff
@@ -0,0 +1,99 @@
...
+        gen.writeStringField("module", func.getModule());
+        gen.writeStringField("qualname", func.getQualName());
+        gen.writeEndObject();
+    }
+
+    private void serializeJavaFunction(JsonGenerator gen, JavaFunction func) throws IOException {
+        gen.writeFieldName("exec");
+        gen.writeStartObject();
+        // Mark it as a Java function
+        gen.writeStringField("func_type", JavaFunction.class.getSimpleName());
```
lihaosky: ditto

plan/src/test/java/org/apache/flink/agents/plan/serializer/ActionJsonSerializerTest.java
```diff
@@ -0,0 +1,193 @@
...
+                        new Class[] {InputEvent.class});
+
+        // Create an Action
+        Action action = new Action("testAction", function, List.of(InputEvent.class));
+
+        // Serialize the action to JSON
+        String json = new ObjectMapper().writeValueAsString(action);
+
+        // Verify the JSON contains the expected fields
+        assertTrue(json.contains("\"name\":\"testAction\""), "JSON should contain the action name");
```
lihaosky, line 55: Curious why not do `assertEquals` with whole serialized json? Will that be more readable?

plan/src/test/java/org/apache/flink/agents/plan/serializer/ActionJsonSerializerTest.java
```diff
@@ -0,0 +1,193 @@
...
+        PythonFunction function = new PythonFunction("test_module", "test_function");
+
+        // Create an Action
+        Action action = new Action("testPythonAction", function, List.of(InputEvent.class));
+
+        // Serialize the action to JSON
+        String json = new ObjectMapper().writeValueAsString(action);
+
+        // Verify the JSON contains the expected fields
+        assertTrue(
```
lihaosky, line 86: ditto

plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
PatrickRen: Thanks for the comment! This makes sense. I also renamed the Python part in my pull request.

plan/src/test/java/org/apache/flink/agents/plan/serializer/ActionJsonSerializerTest.java
PatrickRen: Thanks for the review! I was thinking this could make the assertion failure message more straight-forward, also I'm not sure if the formatting of json could make the test unstable, like with or without new lines, indentations etc.

plan/src/main/java/org/apache/flink/agents/plan/PythonFunction.java
PatrickRen: 👍 I created #53 for tracking this.

PatrickRen:
Thank you all for your review!!

> Would it be possible to implement Java's serializable interface only for WorkflowPlan, and then within the Java serialization methods, delegate to your defined JSON-based serialization approach?

As discussed offline with @codenohup , this part will be completed by a followup PR. 

Also I changed the type of field `listenEventTypes` from `List<Class>` to `List<String>`, in order to be aligned with Python implementation.

PR #47, opened 2025-06-30
[WIP][Feature] Implements flink operators to execute workflow  (#26)

codenohup:
Linked issue: #26

### Purpose of change

Implements flink operators to execute user-defined workflow.

Please note that this PR depends on #43  and #45 , so it should be merged after those two are merged. The first three commits in this PR actually belong to the prerequisite PRs. You can start the review from the fourth commit.

### Tests

Tested manually; I will add automated test cases later.

### API

Yes, add `RunnerContext` api to execute Java action

### Documentation

I will add documentations later.


plan/src/main/java/org/apache/flink/agents/plan/WorkflowPlan.java
```diff
@@ -19,23 +19,56 @@
...
+    public Map<Class<? extends Event>, List<Action>> getEventTriggerActions() {
+        return eventTriggerActions;
+    }
+
+    public List<Action> getEventTriggerActions(Class<? extends Event> event) {
+        return eventTriggerActions.get(event);
+    }
+
+    private void writeObject(ObjectOutputStream out) throws IOException {
+        String serializedStr = new ObjectMapper().writeValueAsString(this);
```
HuangXingBo: new ObjectMapper() -> static final field?

plan/src/main/java/org/apache/flink/agents/plan/serializer/ActionJsonDeserializer.java
```diff
@@ -0,0 +1,118 @@
...
+            func = deserializeJavaFunction(execNode);
+        } else if (PythonFunction.class.getSimpleName().equals(funcType)) {
+            func = deserializePythonFunction(execNode);
+        } else {
+            throw new IOException("Unsupported function type: " + funcType);
+        }
+
+        // Deserialize listenEventTypes
+        List<Class<? extends Event>> listenEventTypes = new ArrayList<>();
+        node.get("listenEventTypes")
```
HuangXingBo: Why is a camel name here, while other places use snake name?

runtime/src/main/java/org/apache/flink/agents/runtime/PythonActionExecutor.java
```diff
@@ -39,13 +40,13 @@ public class PythonActionExecutor {
     private static final String FLINK_RUNNER_CONTEXT_VAR_NAME = "flink_runner_context";
 
     private final PythonEnvironmentManager environmentManager;
-    private final PythonRunnerContext runnerContext;
+    private final PythonRunnerContextImpl runnerContext;
 
     private PythonInterpreter interpreter;
 
     public PythonActionExecutor(PythonEnvironmentManager environmentManager) {
```
HuangXingBo, line 54: Could we put python-related components such as pythonevent and pythonexecutor in specific subdir?

runtime/pom.xml
```diff
@@ -29,29 +29,62 @@ under the License.
     <name>Flink Agents : Runtime</name>
 
     <dependencies>
+        <!-- flink agents -->
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-api</artifactId>
+            <version>0.1-SNAPSHOT</version>
```
HuangXingBo: -> ${project.version}

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+ * href="https://github.com/apache/flink-statefun">flink-statefun</a>.
+ *
+ * @param <K> The type of the key of input event.
+ */
+public class ActionExecutionOperator<K> extends AbstractStreamOperator<EventMessage<K>>
+        implements OneInputStreamOperator<EventMessage<K>, EventMessage<K>> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ActionExecutionOperator.class);
+
+    private static final long serialVersionUID = 1L;
```
HuangXingBo, line 61: put before `private static final Logger LOG = LoggerFactory.getLogger(ActionExecutionOperator.class);`?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/FeedbackOperator.java
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * @param <K> The type of the key of input event.
+ */
+public class FeedbackOperator<K> extends AbstractStreamOperator<EventMessage<K>>
+        implements FeedbackConsumer<Message>,
+                OneInputStreamOperator<EventMessage<K>, EventMessage<K>> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FeedbackOperator.class);
+
+    private static final long serialVersionUID = 1L;
```
HuangXingBo: put before `private static final Logger LOG = LoggerFactory.getLogger(FeedbackOperator.class);`?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+
+    @Override
+    public void open() throws Exception {
+        super.open();
+
+        reusedStreamRecord = new StreamRecord<>(null);
+        reusedSideOutputStreamRecord = new StreamRecord<>(null);
+
+        // init state
+        ValueStateDescriptor<Integer> pendingActionCountDescriptor =
```
HuangXingBo: default value constructor has been deprecated instead use the null to check

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+                // execute action
+                List<Event> actionOutputEvents;
+                if (action.getExec() instanceof JavaFunction) {
+                    action.getExec().call(event, runnerContext);
+                    actionOutputEvents = runnerContext.drainEvents();
+                } else if (action.getExec() instanceof PythonFunction) {
+                    Preconditions.checkState(event instanceof PythonEvent);
+                    actionOutputEvents =
+                            pythonActionExecutor.executePythonFunction(
+                                    (PythonFunction) action.getExec(), (PythonEvent) event);
```
HuangXingBo: When I see this code, I get a sense of fragmentation. Perhaps having a Java executor would feel more natural, but that might fall outside the scope of what this PR needs to address.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+        // If there is some actions in progress with the same key, the subsequent input event should
+        // not be processed immediately.
+        return pendingActionCountState.value() == 0;
+    }
+
+    private void tryProcessPendingInputEvents() throws Exception {
+        if (pendingActionCountState.value() != 0
+                || pendingInputEventsState.get() == null
+                || !pendingInputEventsState.get().iterator().hasNext()) {
+            // If there is some actions in progress with the same key, the pending input event
```
HuangXingBo: is -> are

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+
+        // If there is some actions in progress with the same key, the subsequent input event should
+        // not be processed immediately.
+        return pendingActionCountState.value() == 0;
+    }
+
+    private void tryProcessPendingInputEvents() throws Exception {
+        if (pendingActionCountState.value() != 0
+                || pendingInputEventsState.get() == null
+                || !pendingInputEventsState.get().iterator().hasNext()) {
```
HuangXingBo: -> `if (!canProcessInputEvent())`?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+                    }
+                }
+
+                addPendingActionCount(-1);
+            }
+        }
+    }
+
+    private boolean canProcessInputEvent() throws Exception {
+        if (pendingInputEventsState.get() != null
```
HuangXingBo: Why don't we only use pendingActionCountState to judge it? What's more, ListState has worse performance, should we use valueState to judge it first?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+                && pendingInputEventsState.get().iterator().hasNext()) {
+            // If there is already a pending input event with the same key, the subsequent input
+            // event should not be processed immediately.
+            return false;
+        }
+
+        // If there is some actions in progress with the same key, the subsequent input event should
```
HuangXingBo: is -> are

HuangXingBo (commented):
@codenohup Thanks a lot for the PR. Looks good overall. I only left some minor comments.

runtime/src/main/java/org/apache/flink/agents/runtime/PythonEvent.java
```diff
@@ -35,6 +35,16 @@ public byte[] getEvent() {
         return event;
     }
 
+    @Override
+    public boolean isInputEvent() {
+        return eventType.equals("flink_agents.api.event.InputEvent");
+    }
+
+    public boolean isOutputEvent() {
```
GreatEugenius:
```suggestion
    @Override
    public boolean isOutputEvent() {
```

codenohup:
Hi, @HuangXingBo @GreatEugenius 
Thank you for your reviews. I have addressed the comments and updated the PR — PTAL :)

runtime/src/main/java/org/apache/flink/agents/runtime/python/context/PythonRunnerContextImpl.java
```diff
@@ -15,40 +15,31 @@
...
-import java.util.List;
-
 /**
  * This class is used to manage the execution context when interacting with Python functions,
  * including collecting new events generated during the execution of actions.
  */
 @NotThreadSafe
-public class PythonRunnerContext {
-    private final List<PythonEvent> pendingEvents;
+public class PythonRunnerContextImpl extends RunnerContextImpl {
```
xintongsong, line 30: This is weird. We say `RunnerContextImpl` is "for Java-based actions", and having a `PythonRunnerContextImpl` to extend it.

plan/src/main/java/org/apache/flink/agents/plan/Action.java
```diff
@@ -44,7 +45,7 @@ public Action(String name, Function exec, List<Class<? extends Event>> listenEve
         this.name = name;
         this.exec = exec;
         this.listenEventTypes = listenEventTypes;
-        exec.checkSignature(new Class[] {Event.class});
+        exec.checkSignature(new Class[] {Event.class, RunnerContext.class});
```
xintongsong, line 47: This doesn't belong to this commit.

api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -48,4 +48,20 @@ public Object getAttr(String name) {
...
+        return this.getClass().getCanonicalName();
+    }
+
+    public boolean isInputEvent() {
+        return false;
+    }
+
+    public boolean isOutputEvent() {
+        return false;
+    }
```
xintongsong: Doesn't make sense to have these methods here. It introduce the chance for subclasses to incorrectly override these methods. I'd suggest to make the type of input/output events static constants, and compare in `EventMessage`.

runtime/src/main/java/org/apache/flink/agents/runtime/FlinkAgent.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.flink.agents.runtime.operator.FeedbackOperatorFactory;
+import org.apache.flink.agents.runtime.operator.FeedbackSinkOperator;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.datastream.DataStreamUtils;
+import org.apache.flink.streaming.api.datastream.SingleOutputStreamOperator;
+import org.apache.flink.util.OutputTag;
+
+/** A utility class that bridges Flink DataStream/SQL with the Flink Agents workflow. */
+public class FlinkAgent {
```
xintongsong: 1. Is this class exposed to users?
2. The name doesn't make sense. Maybe `CompileUtils`?

runtime/src/main/java/org/apache/flink/agents/runtime/FlinkAgent.java
```diff
@@ -0,0 +1,112 @@
...
+     * Connects the given DataStream to the Flink Agents workflow.
+     *
+     * @param inputDataStream The input DataStream.
+     * @param keyTypeInfo The type information of the key.
+     * @param workflowPlan The workflow plan.
+     * @param <K> The type of the key.
+     * @return The output DataStream of the workflow.
+     */
+    public static <K> DataStream<EventMessage<K>> connectToWorkflow(
+            DataStream<EventMessage<K>> inputDataStream,
```
xintongsong: We should not require input datastreams to have the `EventMessage` record type.

runtime/src/main/java/org/apache/flink/agents/runtime/FlinkAgent.java
```diff
@@ -0,0 +1,112 @@
...
+        TypeInformation<EventMessage<K>> eventMessageTypeInfo =
+                inputDataStream.getTransformation().getOutputType();
+        FeedbackKey<Message> feedbackKey = new FeedbackKey<>("feedback-pipeline", 1L);
+
+        // 1. Create feedback operator, the feedback operator is responsible for receive event from
+        // {@code inputDataStream} and {@code feedBackSinkOperator} and forward to action execute
+        // operator
+        DataStream<EventMessage<K>> feedbackDatastream =
+                inputDataStream
+                        .keyBy(new EventMessageKeySelector(), keyTypeInfo)
```
xintongsong: Shall we also consider that the input data stream might already be a keyed stream?

HuangXingBo (approved):

xintongsong (commented):
I haven't complete my review on this PR, in particularly the implementation details of the operators. Already have some comments on the other parts. Will work with the remaining parts in a couple of days.

runtime/src/main/java/org/apache/flink/agents/runtime/FlinkAgent.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param <K> The type of the key.
+     * @return The output DataStream of the workflow.
+     */
+    public static <K> DataStream<EventMessage<K>> connectToWorkflow(
+            DataStream<EventMessage<K>> inputDataStream,
+            TypeInformation<K> keyTypeInfo,
+            WorkflowPlan workflowPlan) {
+        TypeInformation<EventMessage<K>> eventMessageTypeInfo =
+                inputDataStream.getTransformation().getOutputType();
+        FeedbackKey<Message> feedbackKey = new FeedbackKey<>("feedback-pipeline", 1L);
```
xintongsong: I think we need to generate different invocation ids for each time calling `connectToWorkflow`. Otherwise, if there are multiple workflows in a flink job, they will be forced in a same colocation group.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/FeedbackOperator.java
```diff
@@ -0,0 +1,203 @@
...
+                                ioManager,
+                                maxParallelism,
+                                totalMemoryUsedForFeedbackCheckpointing,
+                                elementSerializer,
+                                keySelector);
+
+        this.checkpoints = new Checkpoints<>(feedbackLoggerFactory::create);
+
+        //
+        // we first must reply previously check-pointed envelopes before we start
```
xintongsong:
```suggestion
        // we first must replay previously check-pointed envelopes before we start
```

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+                addPendingActionCount(actions.size());
+            }
+            for (Action action : actions) {
+                // execute action
+                List<Event> actionOutputEvents;
+                if (action.getExec() instanceof JavaFunction) {
+                    action.getExec().call(event, runnerContext);
+                    actionOutputEvents = runnerContext.drainEvents();
+                } else if (action.getExec() instanceof PythonFunction) {
+                    Preconditions.checkState(event instanceof PythonEvent);
```
xintongsong: I think eventually we would need to support triggering both java and python actions with one event. Let's mark a todo here.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+            }
+        }
+        if (continuePendingInputEvents.isEmpty()) {
+            pendingInputEventsState.clear();
+        } else {
+            pendingInputEventsState.update(continuePendingInputEvents);
+        }
+    }
+
+    private void addPendingActionCount(int delta) throws IOException {
```
xintongsong: check `delta >= 0`

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+                    if (actionOutputEvent.isOutputEvent()) {
+                        output.collect(
+                                sideOutputTag,
+                                reusedSideOutputStreamRecord.replace(actionOutputEventMessage));
+                    } else {
+                        List<Action> pendingActions =
+                                workflowPlan.getEventTriggerActions(actionOutputEvent.getClass());
+                        addPendingActionCount(pendingActions == null ? 0 : pendingActions.size());
+                        output.collect(reusedStreamRecord.replace(actionOutputEventMessage));
+                    }
```
xintongsong: The assumption here is clear to me. Do we consider triggering an action of another key here?
- If we assume all actions have the same key, then why do we need the event to go through the feedback loop?
- If we assume we might trigger actions with another key, then increasing the pending action count and blocking subsequent input events doesn't make sense, because we will never receive the action in current key.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,271 @@
...
+            // For other Events, which are guaranteed to be generated by an action and enter the
+            // feedback loop, we increment the action count once at generation time.
+            if (eventMessage.isInputEvent()) {
+                addPendingActionCount(actions.size());
+            }
+            for (Action action : actions) {
+                // execute action
+                List<Event> actionOutputEvents;
+                if (action.getExec() instanceof JavaFunction) {
+                    action.getExec().call(event, runnerContext);
```
xintongsong: Execution of action can take long, thus should be asynchronous. Let's mark a todo here.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
xintongsong: They only reason I can think of for the event to go through the feedback loop is that, it allows execution of other keys between actions of the current key. And whether that make sense depends on how we would support asynchronous execution.

runtime/src/main/java/org/apache/flink/agents/runtime/utils/EventUtil.java
wenjin272, line 1: There is no apache license

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -17,38 +17,34 @@
...
  */
-@NotThreadSafe
-public class PythonRunnerContext {
-    private final List<PythonEvent> pendingEvents;
+public class RunnerContextImpl implements RunnerContext {
 
-    public PythonRunnerContext() {
-        this.pendingEvents = new ArrayList<>();
-    }
+    protected final List<Event> events = new ArrayList<>();
```
xintongsong:
```suggestion
    protected final List<Event> pendingEvents = new ArrayList<>();
```

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,210 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(ActionExecutionOperator.class);
+
+    private final WorkflowPlan workflowPlan;
+
+    private final Boolean inputIsJava;
+
+    private transient StreamRecord<OUT> reusedStreamRecord;
+
+    // RunnerContext for Java actions
+    private final RunnerContextImpl runnerContext;
```
xintongsong: These two fields should be either both final or both transient.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -0,0 +1,210 @@
...
+        if (event instanceof OutputEvent) {
+            return (OUT) ((OutputEvent) event).getOutput();
+        } else {
+            Object outputFromOutputEvent =
+                    pythonActionExecutor.getOutputFromOutputEvent(((PythonEvent) event).getEvent());
+            return (OUT) outputFromOutputEvent;
+        }
+    }
+
+    private List<Action> getEventTriggerActions(Event event) {
```
xintongsong:
```suggestion
    private List<Action> getActionsTriggeredBy(Event event) {
```

xintongsong (approved):

Issue #19, opened 2025-06-12
[Feature][api][python] Support DataStream / Table as input / output of Flink Agents in Python


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support DataStream / Table as input / output of Flink Agents in Python

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #48, opened 2025-07-01
[API][Python] Support DataStream / Table as input / output of Flink Agents in Python

wenjin272:
Linked issue: #19 

### Purpose of change
Support DataStream / Table as input / output of Flink Agents in Python

Please note that this PR depends on  #45 and #47 , so it should be merged after these PRs are merged. The first 6 commits in this PR actually belong to the prerequisite PRs.  And the No.7 commit is to solve the workflow plan compatibility between Python and Java, which should be done in #45 . You can start the review from the No.8 commit (currently the last commit). 

### Tests
Tested manually; 
Currently, I add an example to validate the execution, I will add e2e test case later.

### API
Yes, add from_datastream and to_datastream in AgentsExecutionEnvironment to integrate DataStream with agents.

### Documentation
I will add documentations later.


wenjin272:
The python tests with Python 3.12 are failed due to install failure of pyflink. After talking about this with @HuangXingBo offline, we may need modify the gha, and install flink-agents before run tests.

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class MapKeyedProcessFunctionAdapter(KeyedProcessFunction):
+    """Util class for converting element in KeyedStream to Row."""
+
+    def process_element(
+        self, value: Any, ctx: "KeyedProcessFunction.Context"
+    ) -> Generator:
+        """Convert element to Row contains key."""
+        ctx.get_current_key()
```
codenohup: why duplicate `ctx.get_current_key()`

python/flink_agents/api/execution_enviroment.py
```diff
@@ -19,14 +19,31 @@
...
+    @classmethod
+    def from_env(
+        cls, env: StreamExecutionEnvironment
+    ) -> type["AgentsExecutionEnvironment"]:
+        """Set StreamExecutionEnvironment of AgentsExecutionEnvironment.
+
+        Currently, this property is only used for distinguishing execution environment.
+        """
+        cls.__env = env
+        return cls
```
xintongsong: Having a method that modifies the class property doesn't make sense. Why not adding `env` as an argument of `get_execution_environment`?

python/flink_agents/api/execution_enviroment.py
```diff
@@ -54,7 +76,21 @@ def from_list(self, input: List[Dict[str, Any]]) -> 'AgentsExecutionEnvironment'
         """
 
     @abstractmethod
-    def apply(self, workflow: Workflow) -> 'AgentsExecutionEnvironment':
+    def from_datastream(
```
xintongsong: The current api would not work in the following use case:
```
env1 = env.from_datastream(ds1).apply(workflow1)
env2 = env.from_datastream(ds2).apply(workflow2)

output1 = env1.to_datastream()
output2 = env2.to_datastream()
```

I think we would need to return some sort of `AgentBuilder` for `from_datastream/list()`, and make `apply()` and `to_datastream/list()` methods of the builder.

python/flink_agents/api/execution_enviroment.py
```diff
@@ -75,6 +111,16 @@ def to_list(self) -> List[Dict[str, Any]]:
             Outputs of workflow execution.
         """
 
+    @abstractmethod
+    def to_datastream(self) -> DataStream:
```
xintongsong: The commit message mentioned both datastream and table, but it seems only datastream is supported.

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -0,0 +1,165 @@
...
+                output_type=RowTypeInfo(
+                    [PickledBytesTypeInfo(), PickledBytesTypeInfo()]
+                ),
+            )
+        else:
+            self.__input = input.map(
+                lambda x: Row(
+                    key_selector.get_key(x)
+                    if key_selector is not None
+                    else uuid.uuid4(),
```
xintongsong: Not sure if auto generating a uuid is what the user wants. Maybe we can start with always requiring a key selector, and users can explicitly set the key_selector to generate a uuid?

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -0,0 +1,165 @@
...
+                    InputEvent(input=x),
```
xintongsong: I think it's the operator's responsibility to wrap the input into an InputEvent.

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -0,0 +1,165 @@
...
+    def to_list(self) -> List[Dict[str, Any]]:
+        """Get output list of workflow execution.
+
+        This method is not supported for remote execution environments.
+        """
+        msg = "RemoteExecutionEnvironment does not support to_list."
+        raise NotImplementedError(msg)
+
+
+def get_execution_environment(**kwargs: Dict[str, Any]) -> AgentsExecutionEnvironment:
```
xintongsong: Might be better to use a name different from it's supper class. Maybe `create_instance()`?

python/flink_agents/api/execution_enviroment.py
wenjin272: make sense

python/flink_agents/runtime/remote_execution_environment.py
wenjin272: I will update this after @codenohup update PR #47

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -0,0 +1,226 @@
...
+    def build(self) -> AgentInstance:
+        """Build agent instance.
+
+        This method is not supported for RemoteAgentBuilder.
+        """
+        msg = (
+            "RemoteAgentBuilder does not support build Agent Instance and run individually, must run as"
+            "a flink job."
+        )
+        raise NotImplementedError(msg)
```
lihaosky: I don't understand this. I think build is essential for builder? This is inconsistent with local env which you call `to_list` first and then build and execute. We should make it consistent I guess

wenjin272: I think keep consistency makes sense, so I changed this implementation. But when run an agent with datastream/table , it must be executed as a flink job,  so I keep the usage in example.

python/flink_agents/api/event.py
```diff
@@ -19,30 +19,19 @@
...
-    @model_validator(mode='after')
-    def validate_extra(self) -> 'Event':
-        """Ensure init fields is serializable."""
-        self.model_dump_json()
-        return self
 
-    def __setattr__(self, name: str, value: Any) -> None:
-        super().__setattr__(name, value)
-        # Ensure added property can be serialized.
-        self.model_dump_json()
```
xintongsong: Not sure about removing these. If this is because some types currently won't pass the validation, we can bypass those types and leave a todo item.

python/flink_agents/api/execution_environment.py
```diff
@@ -0,0 +1,172 @@
...
+        Table
+            Output table of agent execution.
+        """
+
+
+class AgentsExecutionEnvironment(ABC):
+    """Base class for workflow execution environment."""
+
+    @classmethod
+    def get_execution_environment(
```
xintongsong, line 87: Why is this a class method, rather than a static method?

python/flink_agents/runtime/local_execution_environment.py
```diff
@@ -17,50 +17,92 @@
...
-            err_msg = "LocalExecutionEnvironment doesn't support apply multiple workflows."
+            err_msg = "LocalAgentBuilder doesn't support apply multiple workflows."
             raise RuntimeError(err_msg)
         self.__runner = LocalRunner(workflow)
         return self
 
-    def to_list(self) -> list:
+    def to_list(self) -> List[Dict[str, Any]]:
         """Get output list of execution environment."""
+        self.__env.set_agent(self.__input, self.__output, self.__runner)
```
xintongsong: Why not setting in `apply`? What if the agent does not have any output?

python/flink_agents/runtime/local_execution_environment.py
```diff
@@ -17,50 +17,92 @@
...
 
-    def apply(self, workflow: Workflow) -> 'AgentsExecutionEnvironment':
+    def apply(self, workflow: Workflow) -> AgentBuilder:
         """Create local runner to execute given workflow.
 
         Doesn't support apply multiple workflows.
         """
         if self.__runner is not None:
-            err_msg = "LocalExecutionEnvironment doesn't support apply multiple workflows."
+            err_msg = "LocalAgentBuilder doesn't support apply multiple workflows."
```
xintongsong: Do we support multiple agents in the same job? If yes, `env.set_agent` doesn't make sense. If not, we should check in the env, not here.

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -0,0 +1,215 @@
...
+        agent : Agent
+            The agent user defined to run in execution environment.
+        """
+        if self.__agent_plan is not None:
+            err_msg = "RemoteAgentBuilder doesn't support apply multiple agents yet."
+            raise RuntimeError(err_msg)
+        self.__agent_plan = AgentPlan.from_agent(agent)
+        return self
+
+    def to_datastream(
```
xintongsong, line 71: What happens if this is called multiple times?

xintongsong: How do we make sure this is not called before `apply`?

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def execute(self) -> None:
+        """Execute agent individually.
+
+        RemoteExecutionEnvironment does not support execute agent individually.
+        """
+        err_msg = (
+            "RemoteExecutionEnvironment does not support execute "
+            "agent individually, must run as a flink job."
+        )
```
xintongsong: So how should the flink job be executed?

Issue #55, opened 2025-07-10
[Feature][api][runtime] Support Short-Term Memory


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support Short-Term Memory

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #67, opened 2025-07-16
[Feature][api][runtime] Support Short-Term Memory

KeGu-069:
Linked issue: #55 

### Purpose of change

This pr introduces the short-term memory api and wires it into the operator/runtime layer.


python/flink_agents/api/memoryobject.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Parameters
+        ----------
+        path: str
+          Relative path from the current object to the target field.
+
+        Returns:
+        -------
+        Any
+          The value of the field. If the field is an object, another MemoryObject will be returned.
```
codenohup: not clear, I don't quite understand it

python/flink_agents/api/memoryobject.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        Set the value of a (direct or indirect) field in the object.
+        This will also create the intermediate objects if not exist.
+
+        Parameters
+        ----------
+        path: str
+          Relative path from the current object to the target field.
+        value: Any
+          New value of the field. The type of the value must be either a primary type, or MemoryObject.
```
codenohup: Confirm why we need to set a memory object through the `set` method instead of through `new_object`

python/flink_agents/api/memoryobject.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+
+    @abstractmethod
+    def new_object(self, path: str) -> 'MemoryObject':
+        """
+        Create a new object as the value of a (direct or indirect) field in the object.
```
codenohup: Why can we create a indirect field through `new_object`?

python/flink_agents/examples/agent_example.py
```diff
@@ -40,16 +40,33 @@ class MyAgent(Agent):
     @staticmethod
     def first_action(event: Event, ctx: RunnerContext):  # noqa D102
         input = event.input
-        content = input + " first_action"
+        memory = ctx.get_short_term_memory()
+
+        current = memory.get("counter") or 0
```
codenohup:
```suggestion
current_count = memory.get("counter") or 0
```

codenohup: It seems that this _counter_ is counting actions. You can consider changing the name to _action_counter_.

python/flink_agents/runtime/local_memory_object.py
```diff
@@ -0,0 +1,219 @@
...
+        Parameters
+        ----------
+        path: str
+          Relative path from the current object to the target field.
+
+        Returns:
+        -------
+        Any
+          The value of the field. If the field is an indirect field, another MemoryObject will be returned.
+          If the field doesn't exist, returns None.
```
codenohup, line 69: There are three return values: value, MemoryObject, and None. They should be clearly stated.

runtime/src/main/java/org/apache/flink/agents/runtime/python/context/PythonRunnerContextImpl.java
```diff
@@ -40,4 +47,9 @@ public void sendEvent(String type, byte[] event) {
         // this method will be invoked by PythonActionExecutor's python interpreter.
         sendEvent(new PythonEvent(event, type));
     }
+
+    @Override
```
codenohup, line 51: This method looks the same as the parent class implementation

python/flink_agents/api/memoryobject.py
```diff
@@ -0,0 +1,111 @@
...
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#################################################################################
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List
+
+from pydantic import BaseModel
+
+
+class MemoryObject(BaseModel, ABC):
```
xintongsong: File name: memory_object

python/flink_agents/runtime/local_memory_object.py
```diff
@@ -0,0 +1,229 @@
...
+    Each object keeps a prefix to represent its logical path to a flattened
+    key in the store.
+    """
+
+    ROOT_KEY: ClassVar[str] = ""
+    SEPARATOR: ClassVar[str] = "."
+    NESTED_MARK: ClassVar[str] = "NestedObject"
+
+    _store: dict[str, Any]
+    _prefix: str
```
xintongsong: Should these be private?

python/flink_agents/runtime/local_memory_object.py
```diff
@@ -0,0 +1,229 @@
...
+            The created object.
+        """
+        abs_path = self._full_path(path)
+        parts = abs_path.split(self.SEPARATOR)
+
+        self._fill_parents(parts)
+
+        parent_path = (
+            self.SEPARATOR.join(parts[:-1]) if len(parts) > 1 else self.ROOT_KEY
+        )
```
xintongsong: This has been duplicated a lot. Maybe making it a util method?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/MemoryObjectImpl.java
```diff
@@ -0,0 +1,213 @@
...
+import org.apache.flink.api.common.state.MapState;
+
+import java.util.*;
+
+public class MemoryObjectImpl implements MemoryObject {
+
+    public enum ItemType {
+        OBJECT,
+        VALUE
+    }
```
xintongsong: Should be private

runtime/src/main/java/org/apache/flink/agents/runtime/memory/MemoryObjectImpl.java
```diff
@@ -0,0 +1,213 @@
...
+            parentNode.getSubKeys().add(parts[i]);
+            store.put(parent, parentNode);
+        }
+    }
+
+    /** Represents an item (nested object or primitive value) stored in the short-term memory. */
+    public static final class MemoryItem {
+        private final ItemType type;
+        private Object value;
+        private Set<String> subKeys;
```
xintongsong: Should be final

runtime/src/main/java/org/apache/flink/agents/runtime/python/context/PythonRunnerContextImpl.java
KeGu-069: Event on the Python side lacks a specific type compared to the Java side, so the interface and methods have new implementations. However, MemoryObject has no difference on both sides, so this method can be implemented the same as the parent class.

xintongsong (approved):

Issue #68, opened 2025-07-16
[Feature][plan][runtime][python] Introduce Resource and Resource Provider in python


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce Resource and Resource Provider in python

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #70, opened 2025-07-16
[plan][runtime][python] Introduce Resource and Resource Provider in p…

wenjin272:
Linked issue: #68 

### Purpose of change

Introduce Resource and Resource Provider in python

### Tests

Unit test

### API

Yes, add Resource and ResourceProvider API

### DIFF
--- a/python/flink_agents/api/decorators.py
+++ b/python/flink_agents/api/decorators.py
@@ -38,13 +38,32 @@ def action(*listen_events: Tuple[Type[Event], ...]) -> Callable:
     AssertionError
         If no events are provided to listen to.
     """
-    assert len(listen_events) > 0, 'action must have at least one event type to listen to'
+    assert len(listen_events) > 0, (
+        "action must have at least one event type to listen to"
+    )
 
     for event in listen_events:
-        assert issubclass(event, Event), 'action must only listen to event types.'
+        assert issubclass(event, Event), "action must only listen to event types."
 
     def decorator(func: Callable) -> Callable:
         func._listen_events = listen_events
         return func
 
     return decorator
+
+
+def resource(func: Callable) -> Callable:
+    """Decorator for marking a function as an agent resource.
+
+    Parameters
+    ----------
+    func : Callable
+        Function to be decorated.
+
+    Returns:
+    -------
+    Callable
+        Decorator function that marks the target function as resource.
+    """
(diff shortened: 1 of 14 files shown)


python/flink_agents/api/resource.py
```diff
@@ -0,0 +1,88 @@
...
+
+class ResourceType(Enum):
+    """Type enum of resource."""
+
+    CHAT_MODEL = "chat_model"
+    EMBEDDING_MODEL = "embedding_model"
+    PROMPT = "prompt"
+    TOOL = "tool"
+    VECTOR_STORE = "vector_store"
+    MCP_SERVER = "mcp_server"
```
xintongsong: We may comment out types that we planned to but not yet support.

python/flink_agents/api/resource.py
```diff
@@ -0,0 +1,88 @@
...
+    name : str
+        The name of the resource.
+    type : ResourceType
+        The type of the resource.
+    """
+
+    name: str
+
+    @classmethod
+    def resource_type(cls) -> ResourceType:
```
xintongsong, line 57: This should also be an abstract method

python/flink_agents/api/resource.py
```diff
@@ -0,0 +1,88 @@
...
+    """Resource which is serializable."""
+
+    @model_validator(mode="after")
+    def validate_serializable(self) -> "SerializableResource":
+        """Ensure resource is serializable."""
+        self.model_dump_json()
+        return self
+
+
+class ResourceProvider(BaseModel, ABC):
```
xintongsong: Should this be part of the public API? Does user need to directly implement or call this interface?

python/flink_agents/api/decorators.py
```diff
@@ -38,13 +38,32 @@ def action(*listen_events: Tuple[Type[Event], ...]) -> Callable:
...
+        assert issubclass(event, Event), "action must only listen to event types."
 
     def decorator(func: Callable) -> Callable:
         func._listen_events = listen_events
         return func
 
     return decorator
+
+
+def resource(func: Callable) -> Callable:
```
xintongsong: Why do we have this decorator? We should not allow any custom resource types. So users should always use built-in decorators like `@model`, `@tool`.

PR #72, opened 2025-07-17
[Python][Optimisation]  Introduced caching mechanism to reuse existing PythonFunction instances for identical (module, qualname) pairs.

Kavishankarks:
Python Side (function.py)

Added a global _PYTHON_FUNCTION_CACHE to store reusable function descriptors.

Refactored call_python_function() to retrieve cached instances when available.

Added utility methods for cache management:
clear_python_function_cache()
get_python_function_cache_size()
get_python_function_cache_keys()


Kavishankarks:
> Hi, @Kavishankarks First of all, thank you for your contribution. However, I believe we shouldn't cache all Python functions, as it may cause issues when dealing with Python closures or generators.

yeah, you're right. I'll add function to check if function is closures, generators, coroutine also the bound methods and functions having mutable arguments like dict or list like
  def append_to_list(item, target_list=[]): 
      target_list.append(item)
      return target_list

python/flink_agents/plan/function.py
```diff
@@ -147,6 +231,71 @@ def check_signature(self, *args: Tuple[Any, ...]) -> None:
...
+    if cache_key not in _PYTHON_FUNCTION_CACHE:
+        python_func = PythonFunction(module=module, qualname=qualname)
+        try:
+            actual_func = python_func._PythonFunction__get_func()
+            if _is_function_cacheable(actual_func):
+                _PYTHON_FUNCTION_CACHE[cache_key] = python_func
+            else:
+                return python_func(*func_args)
+        except Exception:
+            return python_func(*func_args)
```
xintongsong: This means for non-cacheable functions, we will go through the `_is_function_cacheable` inspection every time it gets called. Might be better to check it only once on initialization.

xintongsong (commented):
Thanks for working on this, @Kavishankarks.

The PR looks quite good to me. Just one minor comment. Please take a look.

Issue #59, opened 2025-07-10
[Feature][api][runtime] Introduce metric mechanism


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce metric mechanism

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #74, opened 2025-07-23
Feature: [api][runtime] Introduce metric mechanism

GreatEugenius:
Linked issue: #59

### Purpose of change
Introduce metric mechanism

### Tests
Unit test

### API
Yes, Add APIs related to the metrics.


python/flink_agents/runtime/local_runner.py
```diff
@@ -96,6 +97,18 @@ def send_event(self, event: Event) -> None:
     def get_resource(self, name: str, type: ResourceType) -> Resource:
         return self.__agent_plan.get_resource(name, type)
 
+    @override
+    def get_agent_metric_group(self) -> MetricGroup:
+        # TODO: Support metric mechanism for local agent execution.
```
codenohup, line 121: Using metrics with LocalRunner throws error is strange, which implies that an agent needs to have two different code versions — one for use with _LocalRunner_ and another for _FlinkRunner_.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -91,6 +98,12 @@ public void open() throws Exception {
 
         this.runnerContext = new RunnerContextImpl();
 
+        metricGroup = new FlinkAgentsMetricGroupImpl(getMetricGroup());
+
+        builtInMetricGroup = new BuiltInMetricGroup(metricGroup, agentPlan);
+
+        runnerContext.setAgentMetricGroup(metricGroup);
```
codenohup: It looks like RunnerContext and MetricGroup are created at the same time. Why isn't MetricGroup assigned to RunnerContext through its constructor?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -108,6 +121,7 @@ public void processElement(StreamRecord<IN> record) throws Exception {
         events.push(inputEvent);
         while (!events.isEmpty()) {
             Event event = events.pop();
+            builtInMetricGroup.markEventOccurred();
```
codenohup: Should we distinguish between the event generation time and the processing time?

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
```diff
@@ -71,9 +77,11 @@ public void open() throws Exception {
         interpreter.set(FLINK_RUNNER_CONTEXT_VAR_NAME, pythonRunnerContextObject);
     }
 
-    public List<Event> executePythonFunction(PythonFunction function, PythonEvent event)
+    public List<Event> executePythonFunction(
+            PythonFunction function, PythonEvent event, ActionMetricGroup actionMetricGroup)
             throws Exception {
         runnerContext.checkNoPendingEvents();
+        runnerContext.setActionMetricGroup(actionMetricGroup);
```
codenohup: Repeated with line 64

runtime/src/test/java/org/apache/flink/agents/runtime/metrics/FlinkAgentsMetricGroupImplTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.metrics.SimpleCounter;
+import org.apache.flink.runtime.metrics.groups.UnregisteredMetricGroups;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/** Tests for {@link FlinkAgentsMetricGroupImpl}. */
+public class FlinkAgentsMetricGroupImplTest {
```
codenohup, line 34: Typically, we do not write unit tests for an implementation class directly.

runtime/src/test/java/org/apache/flink/agents/runtime/metrics/FlinkAgentsMetricGroupImplTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @BeforeEach
+    public void setUp() {
+        MetricGroup parentMetricGroup =
+                UnregisteredMetricGroups.createUnregisteredOperatorMetricGroup();
+        metricGroup = new FlinkAgentsMetricGroupImpl(parentMetricGroup);
+    }
+
+    @Test
+    public void testGetSubGroup() {
```
codenohup: In JUnit 5, it seems that test classes and test methods do not need to be public.

python/flink_agents/runtime/flink_metric_group.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Parameters
+        ----------
+        j_gauge : Any
+            Java gauge for recording a string value.
+        """
+        self._j_gauge = j_gauge
+
+    @override
+    def update(self, value: str) -> None:
```
codenohup: Can the parameter type be Any, and then converted to a string internally? This would allow users to directly call something like _flink_gauge.update(0.05f)_.

api/src/main/java/org/apache/flink/agents/api/metrics/FlinkAgentsMetricGroup.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    FlinkAgentsMetricGroup getSubGroup(String name);
+
+    /**
+     * Create or retrieve a gauge with the given name.
+     *
+     * @param name The name of the gauge.
+     * @return the gauge instance for string values.
+     */
+    StringGauge getGauge(String name);
```
codenohup: It would be helpful to provide an explanation for why only StringGauge is currently supported.

python/flink_agents/runtime/flink_metric_group.py
GreatEugenius: This is to ensure consistency with the types returned by get_value, preventing confusion for the user.

api/src/main/java/org/apache/flink/agents/api/metrics/StringGauge.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.flink.agents.api.metrics;
+
+import org.apache.flink.metrics.Gauge;
+
+/**
+ * StringGauge interface is designed to measure and update string values. It extends the Gauge
+ * interface with a generic type of String. This interface is used to define standard operations for
+ * components that need to monitor and update string information.
+ */
+public interface StringGauge extends Gauge<String> {
```
xintongsong: What is this and why do we need it?

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -38,8 +40,30 @@ public class RunnerContextImpl implements RunnerContext {
...
+    public RunnerContextImpl(
+            MapState<String, MemoryObjectImpl.MemoryItem> store,
+            FlinkAgentsMetricGroupImpl agentMetricGroup) {
         this.store = store;
+        this.agentMetricGroup = agentMetricGroup;
+    }
+
+    public void setActionMetricGroup(ActionMetricGroup actionMetricGroup) {
+        this.actionMetricGroup = actionMetricGroup;
+    }
```
xintongsong: It might be easier to set the current action, rather than the metric group bound to the action. We may have more and more action related things to be switched.

runtime/src/main/java/org/apache/flink/agents/runtime/metrics/ActionMetricGroup.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.flink.metrics.Meter;
+import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.util.clock.SystemClock;
+
+/**
+ * ActionMetricGroup class extends FlinkAgentsMetricGroupImpl and is used to monitor and measure the
+ * performance metrics of executing actions. It provides metrics for the number of actions currently
+ * executing, the number of actions executed per second, and the execution time of actions.
+ */
+public class ActionMetricGroup extends FlinkAgentsMetricGroupImpl {
```
xintongsong: 1. Should not use inherit here. 
2. Maybe name it `BuiltInActionMetrics`

runtime/src/main/java/org/apache/flink/agents/runtime/metrics/ActionMetricGroup.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private final Counter numOfActionsExecuting;
```
xintongsong: Why do we need this?

runtime/src/main/java/org/apache/flink/agents/runtime/metrics/ActionMetricGroup.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Marks that an action has finished executing. Decrements the executing actions counter, marks
+     * an event on the executed meter, and records the execution time in the histogram.
+     */
+    public void markActionExecuted() {
+        long endTime = SystemClock.getInstance().relativeTimeMillis();
+        numOfActionsExecuting.dec();
+        numOfActionsExecutedPerSec.markEvent();
+        actionExecutionTime.update(endTime - startTime);
```
xintongsong: @codenohup Can these calculations work with async execution?

runtime/src/main/java/org/apache/flink/agents/runtime/metrics/BuiltInMetricGroup.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.HashMap;
+
+/**
+ * Represents a group of built-in metrics for monitoring the performance and behavior of a flink
+ * agent job. This class is responsible for collecting and managing various metrics such as the
+ * number of events processed, the number of actions being executed, and the number of actions
+ * executed per second.
+ */
+public class BuiltInMetricGroup {
```
xintongsong:
```suggestion
public class BuiltInMetrics {
```

runtime/src/main/java/org/apache/flink/agents/runtime/metrics/BuiltInMetricGroup.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private final Meter numOfEventProcessedPerSec;
+
+    private final Counter numOfActionsExecuting;
```
xintongsong: Same here.

runtime/src/main/java/org/apache/flink/agents/runtime/metrics/ActionMetricGroup.java
codenohup: > @codenohup Can these calculations work with async execution?

@xintongsong @GreatEugenius Currently, these calculations won't work correctly under async execution. The reason is that we're using only actionName as the key to track execution times. However, in an asynchronous setup, it's possible for the same action to be triggered by multiple events in parallel, and since they share the same actionName, their timing data will overwrite each other, leading to inaccurate or lost measurements.

PR #80, opened 2025-07-24
[API][Runtime] Support agent and action asynchronous execution

codenohup:
Linked issue: #78 

### Purpose of change
Enable the agent and action to support asynchronous and parallel processing for improved performance. Details can be found in the #78 .

### Tests
Added corresponding unit tests and e2e tests.

### API

Yes, introduce runner_context#execute_async API

### Documentation

No, I'll add the relevant documentation in the future.

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
@@ -38,7 +38,7 @@ public class JavaFunction implements Function {
     @JsonProperty(FIELD_NAME_PARAMETER_TYPES)
     private final Class<?>[] parameterTypes;
 
-    @JsonIgnore private final Method method;
+    @JsonIgnore private transient Method method;
 
     public JavaFunction(
             @JsonProperty(FIELD_NAME_QUAL_NAME) String qualName,
@@ -71,7 +71,10 @@ public class JavaFunction implements Function {
         return parameterTypes;
     }
 
-    public Method getMethod() {
+    public Method getMethod() throws ClassNotFoundException, NoSuchMethodException {
+        if (method == null) {
+            this.method = Class.forName(qualName).getMethod(methodName, parameterTypes);
+        }
         return method;
     }
 
@@ -93,7 +96,7 @@ public class JavaFunction implements Function {
 
     @Override
     public Object call(Object... args) throws Exception {
-        return method.invoke(null, args);
+        return getMethod().invoke(null, args);
     }
 
     @Override
--- a/python/flink_agents/plan/function.py
+++ b/python/flink_agents/plan/function.py
@@ -16,10 +16,11 @@
 # limitations under the License.
 #################################################################################
 
+import logging
 import importlib
 import inspect
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, List, Tuple, get_type_hints
+from typing import Any, Callable, Dict, List, Tuple, get_type_hints, Generator
 
 from pydantic import BaseModel
 
@@ -27,6 +28,11 @@ from flink_agents.plan.utils import check_type_match
 
 # Global cache for PythonFunction instances to avoid repeated creation
 _PYTHON_FUNCTION_CACHE: Dict[Tuple[str, str], "PythonFunction"] = {}
+
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s"
+)
+logger = logging.getLogger(__name__)
 
 
 def _is_function_cacheable(func: Callable) -> bool:
@@ -237,6 +243,28 @@ class JavaFunction(Function):
         """Check function signature is legal or not."""
 
 
+class PythonGeneratorWrapper:
+    """
+    A temporary wrapper class for Python generators to work around a
+    known issue in PEMJA, where the generator type is incorrectly handled.
+
+    TODO: This wrapper is intended to be a temporary solution. Once PEMJA
+    version 0.5.5 (or later) fixes the bug related to generator type conversion,
+    this wrapper should be removed. For more details, please refer to
+    https://github.com/apache/flink-agents/issues/83.
+    """
+
+    def __init__(self, generator: Generator) -> None:
+        """Initialize a PythonGeneratorWrapper. """
+        self.generator = generator
+
+    def __str__(self) -> str:
+        return "PythonGeneratorWrapper, generator=" + str(self.generator)
+
+    def __next__(self) -> Any:
+        return next(self.generator)
+
+
 def call_python_function(module: str, qualname: str, func_args: Tuple[Any, ...]) -> Any:
     """Used to call a Python function in the Pemja environment.
 
@@ -260,19 +288,19 @@ def call_python_function(module: str, qualname: str, func_args: Tuple[Any, ...])
     """
     cache_key = (module, qualname)
 
+    python_func = None
+
     if cache_key not in _PYTHON_FUNCTION_CACHE:
         python_func = PythonFunction(module=module, qualname=qualname)
-        try:
-            if python_func.is_cacheable():
-                _PYTHON_FUNCTION_CACHE[cache_key] = python_func
-            else:
-                return python_func(*func_args)
-        except Exception:
-            return python_func(*func_args)
+        if python_func.is_cacheable():
+            _PYTHON_FUNCTION_CACHE[cache_key] = python_func
+    else:
+        python_func = _PYTHON_FUNCTION_CACHE[cache_key]
 
-    # Use cached instance
-    func = _PYTHON_FUNCTION_CACHE[cache_key]
-    return func(*func_args)
+    func_result = python_func(*func_args)
+    if isinstance(func_result, Generator):
+        return PythonGeneratorWrapper(func_result)
+    return func_result
 
 
 def clear_python_function_cache() -> None:
@@ -305,3 +333,34 @@ def get_python_function_cache_keys() -> List[Tuple[str, str]]:
         List of (module, qualname) tuples representing cached functions.
     """
     return list(_PYTHON_FUNCTION_CACHE.keys())
+
+
+def call_python_generator(generator_wrapper: PythonGeneratorWrapper) -> (bool, Any):
+    """
+    Invokes the next step of a wrapped Python generator and returns whether
+    it is done, along with the yielded or returned value.
+
+    Args:
+        generator_wrapper (PythonGeneratorWrapper): A wrapper object that
+        contains a `generator` attribute. This attribute should be an instance
+        of a Python generator.
+
+    Returns:
+        Tuple[bool, Any]:
+            - The first element is a boolean flag indicating whether the generator
+            has finished:
+                * False: The generator has more values to yield.
+                * True: The generator has completed.
+            - The second element is either:
+                * The value yielded by the generator (when not exhausted), or
+                * The return value of the generator (when it has finished).
+    """
+    try:
+        result = next(generator_wrapper.generator)
+    except StopIteration as e:
+        return True, e.value if hasattr(e, 'value') else None
+    except Exception:
+        logger.exception("Error in generator execution")
+        raise
+    else:
+        return False, result
--- a/python/flink_agents/runtime/flink_runner_context.py
+++ b/python/flink_agents/runtime/flink_runner_context.py
@@ -15,7 +15,7 @@
 #  See the License for the specific language governing permissions and
 # limitations under the License.
 #################################################################################
-from typing import Any
+from typing import Any, Callable, Tuple, Dict
 
 import cloudpickle
 from typing_extensions import override
@@ -105,6 +105,18 @@ class FlinkRunnerContext(RunnerContext):
         """
         return FlinkMetricGroup(self._j_runner_context.getActionMetricGroup())
 
+    @override
+    def execute_async(
+            self,
+            func: Callable[[Any], Any],
+            *args: Tuple[Any, ...],
+            **kwargs: Dict[str, Any],
+    ) -> Any:
+        """Asynchronously execute the provided function. Access to memory
+         is prohibited within the function.
+        """
+        # TODO: Implement in a future commit.
+        pass
 
 def create_flink_runner_context(j_runner_context: Any, agent_plan_json: str) -> FlinkRunnerContext:
     """Used to create a FlinkRunnerContext Python object in Pemja environment."""
--- a/python/flink_agents/runtime/flink_runner_context.py
+++ b/python/flink_agents/runtime/flink_runner_context.py
@@ -26,6 +26,8 @@ from flink_agents.api.runner_context import RunnerContext
 from flink_agents.plan.agent_plan import AgentPlan
 from flink_agents.runtime.flink_memory_object import FlinkMemoryObject
 from flink_agents.runtime.flink_metric_group import FlinkMetricGroup
+from concurrent.futures import ThreadPoolExecutor
+import os
 
 
 class FlinkRunnerContext(RunnerContext):
@@ -36,7 +38,9 @@ class FlinkRunnerContext(RunnerContext):
 
     __agent_plan: AgentPlan
 
-    def __init__(self, j_runner_context: Any, agent_plan_json: str) -> None:
+    def __init__(
+        self, j_runner_context: Any, agent_plan_json: str, executor: ThreadPoolExecutor
+    ) -> None:
         """Initialize a flink runner context with the given java runner context.
 
         Parameters
@@ -46,6 +50,7 @@ class FlinkRunnerContext(RunnerContext):
         """
         self._j_runner_context = j_runner_context
         self.__agent_plan = AgentPlan.model_validate_json(agent_plan_json)
+        self.executor = executor
 
     @override
     def send_event(self, event: Event) -> None:
@@ -115,9 +120,29 @@ class FlinkRunnerContext(RunnerContext):
         """Asynchronously execute the provided function. Access to memory
          is prohibited within the function.
         """
-        # TODO: Implement in a future commit.
-        pass
+        future = self.executor.submit(func, *args, **kwargs)
+        while not future.done():
+            # TODO: Currently, we are using a polling mechanism to check whether
+            #  the future has completed. This approach should be optimized in the
+            #  future by switching to a notification-based model, where the Flink
+            #  operator is notified directly once the future is completed.
+            yield
+        return future.result()
 
-def create_flink_runner_context(j_runner_context: Any, agent_plan_json: str) -> FlinkRunnerContext:
+
+def create_flink_runner_context(
+    j_runner_context: Any, agent_plan_json: str, executor: ThreadPoolExecutor
+) -> FlinkRunnerContext:
     """Used to create a FlinkRunnerContext Python object in Pemja environment."""
-    return FlinkRunnerContext(j_runner_context, agent_plan_json)
+    return FlinkRunnerContext(j_runner_context, agent_plan_json, executor)
+
+
+def create_async_thread_pool() -> ThreadPoolExecutor:
+    """Used to create a thread pool to execute asynchronous
+    code block in action."""
+    return ThreadPoolExecutor(max_workers=os.cpu_count() * 2)
+
+
+def close_async_thread_pool(executor: ThreadPoolExecutor) -> None:
+    """Used to close the thread pool."""
+    executor.shutdown()
--- a/python/flink_agents/runtime/local_runner.py
+++ b/python/flink_agents/runtime/local_runner.py
@@ -18,7 +18,7 @@
 import logging
 import uuid
 from collections import deque
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Callable, Tuple, Generator
 
 from typing_extensions import override
 
@@ -128,6 +128,23 @@ class LocalRunnerContext(RunnerContext):
         err_msg = "Metric mechanism is not supported for local agent execution yet."
         raise NotImplementedError(err_msg)
 
+    def execute_async(
+        self,
+        func: Callable[[Any], Any],
+        *args: Tuple[Any, ...],
+        **kwargs: Dict[str, Any],
+    ) -> Any:
+        """Asynchronously execute the provided function. Access to memory
+         is prohibited within the function.
+        """
+        logger.warning(
+            "Local runner does not support asynchronous execution; falling back to synchronous execution."
+        )
+        func_result = func(*args, **kwargs)
+        yield
+        return func_result
+
+
 class LocalRunner(AgentRunner):
     """Agent runner implementation for local execution, which is
     convenient for debugging.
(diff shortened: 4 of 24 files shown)


python/flink_agents/plan/function.py
```diff
@@ -146,7 +146,59 @@ def check_signature(self, *args: Tuple[Any, ...]) -> None:
...
 
+class PythonGeneratorWrapper:
+    """
+    A temporary wrapper class for Python generators to work around a
+    known issue in PEMJA, where the generator type is incorrectly handled.
+
+    Note: This wrapper is intended to be a temporary solution. Once PEMJA
+    version 0.6.0 (or later) fixes the bug related to generator type conversion,
+    this wrapper should be removed.
+    """
```
wenjin272, line 255: Maybe need add TODO and create correspond issue to track this.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -223,4 +239,16 @@ private List<Action> getActionsTriggeredBy(Event event) {
             return agentPlan.getActionsTriggeredBy(event.getClass().getName());
         }
     }
+
+    private MailboxProcessor getMailboxProcessor() throws Exception {
+        Field field = MailboxExecutorImpl.class.getDeclaredField("mailboxProcessor");
+        field.setAccessible(true);
+        return (MailboxProcessor) field.get(mailboxExecutor);
+    }
```
xintongsong: I guess the elegant way is to add an interface on MailboxExecutor for getting the mailbox main thread, or verifying whether we are in the main thread.

I'm okay with the workaround here, but let's add a TODO to fix this latter.

python/flink_agents/runtime/local_runner.py
```diff
@@ -115,6 +115,23 @@ def get_short_term_memory(self) -> MemoryObject:
...
+    def execute_async(
+        self,
+        func: Callable[[Any], Any],
+        *args: Tuple[Any, ...],
+        **kwargs: Dict[str, Any],
+    ) -> Any:
+        """Asynchronously execute the provided function. Access to memory
+         is prohibited within the function.
+        """
+        logger.info(
```
xintongsong: warn

python/flink_agents/runtime/local_runner.py
```diff
@@ -115,6 +115,23 @@ def get_short_term_memory(self) -> MemoryObject:
...
+            "Local runner does not support asynchronous execution; falling back to synchronous execution."
+        )
+        func_result = func(*args, **kwargs)
+        yield func_result
```
xintongsong: Why do we need this yeild here?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -116,7 +137,23 @@ public void open() throws Exception {
...
+        pendingInputEventsState =
+                getRuntimeContext()
+                        .getListState(
+                                new ListStateDescriptor<>(
+                                        "pendingInputEvents", TypeInformation.of(Event.class)));
+        currentProcessingKeysOperatorState =
+                getOperatorStateBackend()
+                        .getListState(
+                                new ListStateDescriptor<>(
+                                        "currentProcessingKeys", TypeInformation.of(Object.class)));
```
xintongsong: 1. The fact, that `actionTasksState` and `pendingInputEventsState` are keyed states and `currentProcessingKeysOperatorState` is operator states, can be more explicit.
- `actionTasksKState`
- `pendingInputEventsKState`
- `currentProcessingKeysOpState`

xintongsong: 2. How do we handle the redistribution of the operator state?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -240,6 +308,90 @@ private List<Action> getActionsTriggeredBy(Event event) {
...
+        } else {
+            throw new IllegalStateException(
+                    "Unsupported action type: " + action.getExec().getClass());
+        }
+    }
+
+    private void addActionTask(ActionTask actionTask) throws Exception {
+        actionTasksState.add(actionTask);
+        mailboxExecutor.submit(
+                () -> processActionTaskWrapper(actionTask.getKey()), "process action task");
```
xintongsong: processActionTaskWrapper -> tryProcessActionFromKey

xintongsong: It seems we are assuming the actions in `actionTasksState` are consistent with `processActionTaskWrapper` in the mailbox queue. This seems fragile. We should control the order of actions at one place.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -126,60 +163,91 @@ public void open() throws Exception {
...
+
+    private void processActionTask(Object key) throws Exception {
+        setCurrentKey(key);
+        ActionTask actionTask = pollFromListState(actionTasksState);
+        if (actionTask == null) {
+            throw new RuntimeException(
+                    "A null value was encountered while trying to process the ActionTask. This likely indicates a bug in the internal logic.");
+        }
+        boolean finished = actionTask.invoke();
+        List<Event> outputEvents = actionTask.getOutputEvents();
```
xintongsong: It seems `getOutputEvents()` and `getGeneratedActionTask()` are only expected to be called after `invoke()`. Why not combine these three into a single method?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -240,6 +308,90 @@ private List<Action> getActionsTriggeredBy(Event event) {
         }
     }
 
+    private <T> T pollFromListState(ListState<T> listState) throws Exception {
```
xintongsong: 1. These list state operations seems heavy. They should be avoid as much as possible.
2. We might move them into a util.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -251,4 +403,17 @@ private void checkMailboxThread() {
                 mailboxProcessor.isMailboxThread(),
                 "Expected to be running on the task mailbox thread, but was not.");
     }
+
+    private void checkError() {
+        if (errorCause != null) {
```
xintongsong: 1. Is `errorCause` always accessed in the mailbox thread? Do we need to consider the thread safety?
2. Is it guaranteed there's only one error?
3. In which case will the error not be handled immediately?

runtime/src/main/java/org/apache/flink/agents/runtime/python/operator/PythonActionTask.java
```diff
@@ -0,0 +1,84 @@
...
+                event,
+                key);
+        runnerContext.checkNoPendingEvents();
+
+        String pythonGeneratorVarName =
+                pythonActionExecutor.executePythonFunction(
+                        (PythonFunction) action.getExec(), (PythonEvent) event, runnerContext);
+        if (pythonGeneratorVarName != null) {
+            // The Python action generates a generator. We need to execute it once, which will
+            // submit an asynchronous task and return whether the action has been completed.
```
xintongsong: I think the key information here, which should be explained, is that we don't want to convert actions which don't call `async_execute()` into generators.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
```diff
@@ -39,62 +37,103 @@ public class PythonActionExecutor {
...
-        // TODO: remove the set and get runner context after updating pemja to version 0.5.3
-        Object pythonRunnerContextObject =
-                interpreter.invoke(CREATE_FLINK_RUNNER_CONTEXT, runnerContext, agentPlanJson);
-        interpreter.set(FLINK_RUNNER_CONTEXT_VAR_NAME, pythonRunnerContextObject);
+        // TODO: remove the set and get thread pool after updating pemja to version 0.5.3
+        Object pythonAsyncThreadPool = interpreter.invoke(CREATE_ASYNC_THREAD_POOL);
+        this.pythonAsyncThreadPoolObjectName =
+                PYTHON_ASYNC_THREAD_POOL_VAR_NAME_PREFIX
+                        + PYTHON_ASYNC_THREAD_POOL_VAR_ID.incrementAndGet();
+        interpreter.set(pythonAsyncThreadPoolObjectName, pythonAsyncThreadPool);
```
xintongsong: 1. What does this todo mean?
2. Is it necessary to have separate thread pools for each operator?

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
```diff
@@ -39,62 +37,103 @@ public class PythonActionExecutor {
...
-        Object pythonRunnerContextObject = interpreter.get(FLINK_RUNNER_CONTEXT_VAR_NAME);
+        Object pythonRunnerContextObject =
+                interpreter.invoke(
+                        CREATE_FLINK_RUNNER_CONTEXT,
+                        runnerContext,
+                        agentPlanJson,
+                        interpreter.get(pythonAsyncThreadPoolObjectName));
+        String pythonRunnerContextObjectName =
+                FLINK_RUNNER_CONTEXT_VAR_NAME_PREFIX
+                        + FLINK_RUNNER_CONTEXT_VAR_ID.incrementAndGet();
```
xintongsong: Why do we need to create a new object for each time calling the function?

runtime/src/main/java/org/apache/flink/agents/runtime/python/operator/PythonActionTask.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public boolean invoke() throws Exception {
+        LOG.debug(
+                "Try execute python action {} for event {} with key {}.",
+                action.getName(),
+                event,
+                key);
+        runnerContext.checkNoPendingEvents();
+
+        String pythonGeneratorVarName =
```
xintongsong: The name is confusing. Maybe `pythonGeneratorRef`?

xintongsong: Or create a Java class for python generator?

runtime/src/main/java/org/apache/flink/agents/runtime/python/operator/PythonGeneratorActionTask.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public boolean invoke() throws Exception {
+        LOG.debug(
+                "Try execute python generator action {} for event {} with key {}.",
+                action.getName(),
+                event,
+                key);
+        boolean finished = pythonActionExecutor.callPythonGenerator(pythonGeneratorVarName).f0;
```
xintongsong: `f1` is never used?

python/flink_agents/runtime/local_runner.py
codenohup: > Why do we need this yeild here?

Since the user is using _yield from_ to await the result of an asynchronous function in their _action_, we should also use yield to return the result of the function.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
codenohup: > Why do we need to create a new object for each time calling the function?

Since actions may be executed in parallel, to avoid interference between different action instances, we create a new instance for each run of an action.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
codenohup: > 1. What does this todo mean?
Due to a bug in Pemja that may prevent proper cleanup of Python objects, we are using get and set methods as a temporary workaround. For detailed context, please refer to: #83 .

> 2. Is it necessary to have separate thread pools for each operator?
Since each operator has its own Python interpreter, and the thread pool is created and managed within the interpreter, it is necessary to allocate a separate thread pool for each operator.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
codenohup: > 2. How do we handle the redistribution of the operator state?

I suggest replacing _currentProcessingKeysOpState_ to _UnionList_. This change would ensure that all keys are retained even after a parallelism modification or restart. After that, we can simply iterate through the list and clean up any keys that fall outside the current key range.

python/flink_agents/runtime/local_runner.py
```diff
@@ -128,6 +128,23 @@ def get_action_metric_group(self) -> MetricGroup:
...
+    ) -> Any:
+        """Asynchronously execute the provided function. Access to memory
+         is prohibited within the function.
+        """
+        logger.warning(
+            "Local runner does not support asynchronous execution; falling back to synchronous execution."
+        )
+        func_result = func(*args, **kwargs)
+        yield func_result
+        return func_result
```
xintongsong, line 145: Should not need this `return` statement.

python/flink_agents/runtime/local_runner.py
```diff
@@ -203,7 +220,13 @@ def run(self, **data: Dict[str, Any]) -> Any:
...
                     "key: %s, performing action: %s", key, action.name
                 )
-                action.exec(event, context)
+                func_result = action.exec(event, context)
+                if isinstance(func_result, Generator):
+                    try:
+                        while True:
+                            next(func_result)
+                    except StopIteration:
+                        pass
```
xintongsong:
```suggestion
func_result = action.exec(event, context)
if isinstance(func_result, Generator):
    try:
        for _ in func_result:
            pass
    except Exception as e:
        logger.error(f"Error in async execution: {e}")
        raise
```

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
```diff
@@ -71,7 +71,10 @@ public Class<?>[] getParameterTypes() {
         return parameterTypes;
     }
 
-    public Method getMethod() {
+    public Method getMethod() throws ClassNotFoundException, NoSuchMethodException {
+        if (method == null) {
+            this.method = Class.forName(qualName).getMethod(methodName, parameterTypes);
+        }
```
xintongsong, line 77: Why do we need this lazy initialization?

python/flink_agents/plan/function.py
```diff
@@ -237,6 +237,27 @@ def check_signature(self, *args: Tuple[Any, ...]) -> None:
...
 
+class PythonGeneratorWrapper:
+    """
+    A temporary wrapper class for Python generators to work around a
+    known issue in PEMJA, where the generator type is incorrectly handled.
+
+    TODO: This wrapper is intended to be a temporary solution. Once PEMJA
+    version 0.6.0 (or later) fixes the bug related to generator type conversion,
+    this wrapper should be removed.
+    """
```
xintongsong, line 255: Better link to the tracking issue.

python/flink_agents/plan/function.py
```diff
@@ -305,3 +326,31 @@ def get_python_function_cache_keys() -> List[Tuple[str, str]]:
...
+            - The second element is either:
+                * The value yielded by the generator (when not exhausted), or
+                * The return value of the generator (when it has finished).
+    """
+    try:
+        result = next(generator_wrapper.generator)
+    except StopIteration as e:
+        return True, e.value
+    else:
+        return False, result
```
xintongsong:
```suggestion
def call_python_generator(generator_wrapper: PythonGeneratorWrapper) -> (bool, Any):
    try:
        result = next(generator_wrapper.generator)
        return False, result
    except StopIteration as e:
        return True, e.value if hasattr(e, 'value') else None
    except Exception as e:
        # log and throw error
        logger.error(f"Error in generator execution: {e}")
        raise
```

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTask.java
```diff
@@ -0,0 +1,111 @@
...
+
+        public boolean isFinished() {
+            return finished;
+        }
+
+        public List<Event> getOutputEvents() {
+            return outputEvents;
+        }
+
+        public ActionTask getGeneratedActionTask() {
```
xintongsong: Maybe return an `Optional<ActionTask>` and force the checking and null-handling at the caller side?

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
```diff
@@ -40,63 +36,103 @@ public class PythonActionExecutor {
...
         interpreter = env.getInterpreter();
         interpreter.exec(PYTHON_IMPORTS);
 
-        runnerContext = new PythonRunnerContextImpl(shortTermMemState, metricGroup);
-
-        // TODO: remove the set and get runner context after updating pemja to version 0.5.3
-        Object pythonRunnerContextObject =
-                interpreter.invoke(CREATE_FLINK_RUNNER_CONTEXT, runnerContext, agentPlanJson);
-        interpreter.set(FLINK_RUNNER_CONTEXT_VAR_NAME, pythonRunnerContextObject);
+        // TODO: remove the set and get thread pool after updating pemja to version 0.5.3
```
xintongsong: same here for the issue link

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -112,78 +149,151 @@ public void open() throws Exception {
...
+            // event should be queued and processed later. Therefore, we add it to
+            // pendingInputEventsState.
+            pendingInputEventsKState.add(inputEvent);
+        } else {
+            // Otherwise, the new event is processed immediately.
+            processEvent(getCurrentKey(), inputEvent, true);
+        }
+    }
+
+    private void processEvent(Object key, Event event, boolean shouldSubmitMail) throws Exception {
```
xintongsong: In terms of readability, I find this `shouldSubmitMail` a bit hard to understand. It's not straightforward that `processEvent` sometimes need to submit the mail to the mailbox and sometimes not.

I think the rules for notifying the mailbox executor about pending action tasks in a key is quite simple. We should notify in two cases:
1. When `actionTasksKState` changes from empty to non-empty. That is when the InputEvent is processed, because it is guaranteed that `actionTasksKState` is empty before processing the next InputEvent.
2. When we processed an action task from `actionTasksKState` and there are more remaining. 

A straightforward approach would be `processEvent` always submit to the mailbox when an input event is processed, and `processActionTaskForKey` always submit to the mailbox when there are remaining action tasks. In this way, we won't need the `shouldSubmitMail` argument here. 

IIUC, the problem is when we finish all action tasks of the previous input event, and pick up a new pending input event. Currently, we check the remaining action tasks at the end of `processActionTaskForKey`. If we move that check to before processing the a new pending input event, the problem should be solved.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -240,4 +362,70 @@ private List<Action> getActionsTriggeredBy(Event event) {
...
+
+    private void tryResumeProcessActionTasks() throws Exception {
+        Iterable<Object> keys = currentProcessingKeysOpState.get();
+        if (keys != null) {
+            for (Object key : keys) {
+                mailboxExecutor.submit(
+                        () -> tryProcessActionTaskForKey(key), "process action task");
+            }
+        }
+    }
```
xintongsong, line 442: This is smart!

python/flink_agents/runtime/flink_runner_context.py
```diff
@@ -105,7 +110,34 @@ def get_action_metric_group(self) -> FlinkMetricGroup:
...
+) -> FlinkRunnerContext:
     """Used to create a FlinkRunnerContext Python object in Pemja environment."""
-    return FlinkRunnerContext(j_runner_context, agent_plan_json)
+    return FlinkRunnerContext(j_runner_context, agent_plan_json, executor)
+
+
+def create_async_thread_pool() -> ThreadPoolExecutor:
+    """Used to create a thread pool to execute asynchronous
+    code block in action."""
+    return ThreadPoolExecutor(max_workers=os.cpu_count() * 2)
```
xintongsong, line 143: How do we clean-up the thread pool?

python/flink_agents/runtime/local_runner.py
codenohup: > Should not need this `return` statement.

Since the action needs to obtain the result of the asynchronous function via `yield from runner_context.execute_async(...)`, it is necessary to return the value here. It should be modified as follows:

```python
func_result = func(*args, **kwargs)
yield
return func_result
```

plan/src/main/java/org/apache/flink/agents/plan/JavaFunction.java
codenohup: > Why do we need this lazy initialization?

Since we need to store the `ActionTask` in the state, and it indirectly contains the `method` from the `JavaFunction`, which is not serializable in Java.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTask.java
```diff
@@ -0,0 +1,112 @@
...
+        private final Optional<ActionTask> generatedActionTaskOpt;
+
+        public ActionTaskResult(
+                boolean finished,
+                List<Event> outputEvents,
+                Optional<ActionTask> generatedActionTaskOpt) {
+            this.finished = finished;
+            this.outputEvents = outputEvents;
+            this.generatedActionTaskOpt = generatedActionTaskOpt;
+        }
```
xintongsong:
```suggestion
        public ActionTaskResult(
                boolean finished,
                List<Event> outputEvents,
                @Nullable ActionTask generatedActionTaskOpt) {
            this.finished = finished;
            this.outputEvents = outputEvents;
            this.generatedActionTaskOpt = Optional.ofNullable(generatedActionTaskOpt);
        }
```

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -112,75 +150,167 @@ public void open() throws Exception {
...
+            List<Action> triggerActions = getActionsTriggeredBy(event);
+            if (triggerActions != null && !triggerActions.isEmpty()) {
+                shouldSubmitMail = !listStateNotEmpty(actionTasksKState);
+                for (Action triggerAction : triggerActions) {
+                    actionTasksKState.add(createActionTask(key, triggerAction, event));
+                }
+            }
+        }
+
+        return shouldSubmitMail;
```
xintongsong: Why not submit the mail here?

Issue #76, opened 2025-07-23
[Feature][api][python] Introduce Prompt, Tool and ChatModel in Python


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce Prompt, Tool and ChatModel in Python

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #81, opened 2025-07-24
[api][python] Introduce Prompt, Tool and ChatModel in python

wenjin272:
Linked issue: #76 

### Purpose of change

Introduce Prompt, Tool and ChatModel in python

### Tests

unit tests

### API

Yes, add Prompt, Tool and ChatModel api.

### DIFF
--- a/python/flink_agents/plan/agent_plan.py
+++ b/python/flink_agents/plan/agent_plan.py
@@ -21,7 +21,9 @@ from pydantic import BaseModel, field_serializer, model_validator
 
 from flink_agents.api.agent import Agent
 from flink_agents.api.resource import Resource, ResourceType
-from flink_agents.plan.action import Action
+from flink_agents.plan.actions.action import Action
+from flink_agents.plan.actions.chat_model_action import CHAT_MODEL_ACTION
+from flink_agents.plan.actions.tool_call_action import TOOL_CALL_ACTION
 from flink_agents.plan.function import PythonFunction
 from flink_agents.plan.resource_provider import (
     JavaResourceProvider,
@@ -30,7 +32,9 @@ from flink_agents.plan.resource_provider import (
     PythonSerializableResourceProvider,
     ResourceProvider,
 )
-from flink_agents.plan.tools.function_tool import FunctionTool
+from flink_agents.plan.tools.function_tool import from_callable
+
+BUILT_IN_ACTIONS = [CHAT_MODEL_ACTION, TOOL_CALL_ACTION]
 
 
 class AgentPlan(BaseModel):
@@ -116,7 +120,7 @@ class AgentPlan(BaseModel):
         """Build a AgentPlan from user defined agent."""
         actions = {}
         actions_by_event = {}
-        for action in _get_actions(agent):
+        for action in _get_actions(agent) + BUILT_IN_ACTIONS:
             assert action.name not in actions, f"Duplicate action name: {action.name}"
             actions[action.name] = action
             for event_type in action.listen_event_types:
@@ -169,7 +173,8 @@ class AgentPlan(BaseModel):
             self.__resources[type] = {}
         if name not in self.__resources[type]:
             resource_provider = self.resource_providers[type][name]
-            self.__resources[type][name] = resource_provider.provide()
+            resource = resource_provider.provide(get_resource=self.get_resource)
+            self.__resources[type][name] = resource
         return self.__resources[type][name]
 
 
@@ -222,30 +227,33 @@ def _get_resource_providers(agent: Agent) -> List[ResourceProvider]:
 
             if callable(value):
                 clazz, kwargs = value()
-                module = clazz.__module__
                 provider = PythonResourceProvider(
                     name=name,
                     type=clazz.resource_type(),
-                    module=module,
+                    module=clazz.__module__,
                     clazz=clazz.__name__,
                     kwargs=kwargs,
                 )
                 resource_providers.append(provider)
-        if hasattr(value, "_is_tool"):
+        elif hasattr(value, "_is_tool"):
             if isinstance(value, staticmethod):
                 value = value.__func__
 
             if callable(value):
                 # TODO: support other tool type.
-                func = PythonFunction.from_callable(value)
-                tool = FunctionTool(name=name, func=func)
-                provider = PythonSerializableResourceProvider(
-                    name=tool.name,
-                    type=tool.resource_type(),
-                    serialized=tool.model_dump(),
-                    module=tool.__module__,
-                    clazz=tool.__class__.__name__,
-                    resource=tool,
+                tool = from_callable(name=name, func=value)
+                resource_providers.append(
+                    PythonSerializableResourceProvider.from_resource(
+                        name=name, resource=tool
+                    )
                 )
-                resource_providers.append(provider)
+        elif hasattr(value, "_is_prompt"):
+            if isinstance(value, staticmethod):
+                value = value.__func__
+            prompt = value()
+            resource_providers.append(
+                PythonSerializableResourceProvider.from_resource(
+                    name=name, resource=prompt
+                )
+            )
     return resource_providers
--- a/python/flink_agents/plan/resource_provider.py
+++ b/python/flink_agents/plan/resource_provider.py
@@ -17,6 +17,7 @@
 #################################################################################
 import importlib
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from typing import Any, Dict, Optional
 
 from pydantic import BaseModel
@@ -44,8 +45,14 @@ class ResourceProvider(BaseModel, ABC):
     type: ResourceType
 
     @abstractmethod
-    def provide(self) -> Resource:
-        """Create resource in runtime."""
+    def provide(self, get_resource: Callable) -> Resource:
+        """Create resource in runtime.
+
+        Parameters
+        ----------
+        get_resource : Callable
+            The helper function to get other resource declared in the same Agent.
+        """
 
 
 class SerializableResourceProvider(ResourceProvider, ABC):
@@ -81,11 +88,12 @@ class PythonResourceProvider(ResourceProvider):
     clazz: str
     kwargs: Dict[str, Any]
 
-    def provide(self) -> Resource:
+    def provide(self, get_resource: Callable) -> Resource:
         """Create resource in runtime."""
         module = importlib.import_module(self.module)
         cls = getattr(module, self.clazz)
-        return cls(**self.kwargs)
+        resource = cls(**self.kwargs, get_resource=get_resource)
+        return resource
 
 
 class PythonSerializableResourceProvider(SerializableResourceProvider):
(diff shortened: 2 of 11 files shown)


python/flink_agents/api/prompt.py
```diff
@@ -0,0 +1,71 @@
...
+    template: Sequence[ChatMessage]
+
+    @staticmethod
+    def from_messages(name: str, messages: Sequence[ChatMessage]) -> "Prompt":
+        """Create prompt from sequence of ChatMessage."""
+        return Prompt(name=name, template=messages)
+
+    @staticmethod
+    def from_text(
+        name: str, text: str, role: MessageRole = MessageRole.USER
```
xintongsong: Why is the default role `USER`?

python/flink_agents/api/prompt.py
```diff
@@ -0,0 +1,71 @@
...
+    @classmethod
+    def resource_type(cls) -> ResourceType:
+        """Get the resource type."""
+        return ResourceType.PROMPT
+
+    def format_string(self, **kwargs: str) -> str:
+        """Generate text string from template with input arguments."""
+        msgs = []
+        for m in self.template:
+            msg = f"{m.role.value}: {Formatter().format(m.content, **kwargs)}"
```
xintongsong: This means if I create a prompt from a string, it will always add a "USER: " before my prompt.

xintongsong: Should not always create a new formatter.

xintongsong: What happens if an argument is not provided?

python/flink_agents/api/tools/utils.py
```diff
@@ -0,0 +1,171 @@
...
+    "exclusiveMinimum": "gt",
+    "exclusiveMaximum": "lt",
+    "inclusiveMinimum": "ge",
+    "inclusiveMaximum": "le",
+    "minItems": "min_length",
+    "maxItems": "max_length",
+}
+
+
+def get_field_params_from_field_schema(field_schema: dict) -> dict:
```
xintongsong: Should this be private?

python/flink_agents/plan/tools/function_tool.py
```diff
@@ -15,19 +15,77 @@
...
     func: Function
-    def call(self, *args: Tuple[Any, ...], **kwargs: Dict[str, Any]) -> Any:
-        """Call function."""
+
+    @field_serializer("func")
+    def __serialize_func(self, func: Function) -> dict:
+        # append meta info to help deserialize exec
+        data = func.model_dump()
+        data["func_type"] = func.__class__.__qualname__
+        return data
```
xintongsong: Why do we need this here? `Function` should be self contained.

python/flink_agents/plan/action.py
```diff
@@ -41,40 +49,109 @@ class Action(BaseModel):
...
+    exec=PythonFunction.from_callable(process_tool_request),
+    listen_event_types=[f"{ToolRequestEvent.__module__}.{ToolRequestEvent.__name__}"],
+)
+PROCESS_TOOL_RESPONSE = Action(
+    name="process_tool_response",
+    exec=PythonFunction.from_callable(process_tool_response),
+    listen_event_types=[f"{ToolResponseEvent.__module__}.{ToolResponseEvent.__name__}"],
+)
+
+BUILT_IN_ACTIONS = [PROCESS_CHAT_REQUEST, PROCESS_TOOL_REQUEST, PROCESS_TOOL_RESPONSE]
```
xintongsong: I'd suggest to organize builtin actions and events by functionality. E.g., put ChatRequestEvent, ChatResponseEvent and ChatAction in one file.

python/flink_agents/plan/action.py
```diff
@@ -41,40 +49,109 @@ class Action(BaseModel):
...
+        if tool_call_id in tool_context:
+            # get the tool call context from short term memory
+            tool_context = tool_context[tool_call_id]
+            tool_context.messages.append(
+                ChatMessage(role=MessageRole.TOOL, content=str(event.response))
+            )
+            ctx.send_event(tool_context)
+
+
+PROCESS_CHAT_REQUEST = Action(
```
xintongsong:
```suggestion
CHAT_MODEL_ACTION = Action(
```

python/flink_agents/plan/action.py
```diff
@@ -41,40 +49,109 @@ class Action(BaseModel):
...
+    name="process_chat_request",
+    exec=PythonFunction.from_callable(process_chat_request),
+    listen_event_types=[f"{ChatRequestEvent.__module__}.{ChatRequestEvent.__name__}"],
+)
+PROCESS_TOOL_REQUEST = Action(
```
xintongsong:
```suggestion
TOOL_USE_ACTION = Action(
```

python/flink_agents/plan/action.py
```diff
@@ -41,40 +49,109 @@ class Action(BaseModel):
...
         super().__init__(name=name, exec=exec, listen_event_types=listen_event_types)
-        #TODO: Update expected signature after import State and Context.
+        # TODO: Update expected signature after import State and Context.
         self.exec.check_signature(Event, RunnerContext)
 
+
+def process_chat_request(event: ChatRequestEvent, ctx: RunnerContext) -> None:
+    """Built-in action for processing a chat request."""
+    chat_model = ctx.get_resource(event.model, ResourceType.CHAT_MODEL)
+    response = chat_model.chat(event.messages)
```
xintongsong: Model and tool should be called asynchronously. Let's add a todo here and complete it once the async execution is completed.

python/flink_agents/plan/action.py
```diff
@@ -41,40 +49,109 @@ class Action(BaseModel):
...
+
+
+def process_tool_request(event: ToolRequestEvent, ctx: RunnerContext) -> None:
+    """Built-in action for processing a tool call request."""
+    tool = ctx.get_resource(event.tool, ResourceType.TOOL)
+    response = tool.call(**event.kwargs)
+    ctx.send_event(ToolResponseEvent(request=event, response=response))
+
+
+def process_tool_response(event: ToolResponseEvent, ctx: RunnerContext) -> None:
```
xintongsong: We can process tool response in chat model action

python/flink_agents/plan/action.py
```diff
@@ -41,40 +49,109 @@ class Action(BaseModel):
...
+    """Built-in action for processing a tool call response."""
+    state = ctx.get_short_term_memory()
+
+    if state.is_exist("__tool_context"):
+        tool_context = state.get("__tool_context")
+        tool_call_id = event.request.id
+        if tool_call_id in tool_context:
+            # get the tool call context from short term memory
+            tool_context = tool_context[tool_call_id]
```
xintongsong: The name here is confusing.

xintongsong: Shall we also remove the context?

python/flink_agents/plan/agent_plan.py
```diff
@@ -124,6 +124,15 @@ def from_agent(agent: Agent) -> "AgentPlan":
                     actions_by_event[event_type] = []
                 actions_by_event[event_type].append(action.name)
 
+        # append built-in actions
+        for action in BUILT_IN_ACTIONS:
```
xintongsong: Is the handling of built-in actions different from user defined actions? Shall we just put them in the same list before handling?

python/flink_agents/plan/agent_plan.py
```diff
@@ -169,7 +178,23 @@ def get_resource(self, name: str, type: ResourceType) -> Resource:
...
+                            for name in resource.tools
+                        ]
+                    )
+                # bind prompt
+                if resource.prompt is not None and isinstance(resource.prompt, str):
+                    resource.prompt = self.get_resource(
+                        resource.prompt, ResourceType.PROMPT
+                    )
+
+            self.__resources[type][name] = resource
```
xintongsong: This should be done inside the provider.

python/flink_agents/plan/action.py
wenjin272: Make sense. But ChatRequestEvent/ChatResponseEvent should be in api module, for user can sent or listen them in user defined action, while ChatAction should in plan module. So I organize builtin actions in plan module by functionality and organize events in api module by functionality.

python/flink_agents/api/prompts/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        if isinstance(key, int):
+            return args[key]
+        else:
+            if key in kwargs:
+                return kwargs[key]
+            else:
+                return str(key)
+
+
+formatter = SafeFormatter()
```
xintongsong: Uppercase

python/flink_agents/plan/resource_provider.py
```diff
@@ -44,7 +44,7 @@ class ResourceProvider(BaseModel, ABC):
     type: ResourceType
 
     @abstractmethod
-    def provide(self) -> Resource:
+    def provide(self, **kwargs: Any) -> Resource:
```
xintongsong: We should use explicit argument here.

python/flink_agents/plan/resource_provider.py
```diff
@@ -81,11 +81,27 @@ class PythonResourceProvider(ResourceProvider):
...
+                    [
+                        plan.get_resource(name, ResourceType.TOOL)
+                        for name in resource.tools
+                    ]
+                )
+            # bind prompt
+            if resource.prompt is not None and isinstance(resource.prompt, str):
+                resource.prompt = plan.get_resource(
+                    resource.prompt, ResourceType.PROMPT
+                )
```
xintongsong: This should be done in the chat model provider.

xintongsong: Where is `resource.tools` set?

It is unclear that who is responsible for binding the model with tools.
- Looking at this piece of codes, it seems the provider is responsible for binding the model with tools, by calling `bind_tools`. But the problems is how does the provider know which tools to bind?
- Looking at `test_build_in_actions.py`, we provide tools as an argument of the model resource provider, which will be passed into the constructor of the model. That means we leave it to the model implementation to decide how to use the arguments.

wenjin272: Yes, in the previous Implementation, the model and the framework should both be aware of the tools that model need bind, which is not elegant and clean.
Currently, it is the responsibility of model implementation to decide which tool to bind and how to bind, and resource provider only provide fetcher to resource to get other resource declared in the same Agent.

python/flink_agents/plan/resource_provider.py
```diff
@@ -81,11 +88,12 @@ class PythonResourceProvider(ResourceProvider):
     clazz: str
     kwargs: Dict[str, Any]
 
-    def provide(self) -> Resource:
+    def provide(self, fetcher: Callable) -> Resource:
```
xintongsong:
```suggestion
    def provide(self, get_resource: Callable) -> Resource:
```

python/flink_agents/runtime/tests/test_built_in_actions.py
```diff
@@ -0,0 +1,171 @@
...
+    """Mock ChatModel for testing integrating prompt and tool."""
+
+    __tools: List[ToolMetadata]
+
+    def __init__(self, /, **kwargs: Any) -> None:
+        """Init method of MockChatModel."""
+        super().__init__(**kwargs)
+        # bind tools
+        if self.tools is not None:
+            self.bind_tools(
```
xintongsong: Why do we still need `bind_tools` in `BaseChatModel`?

Issue #20, opened 2025-06-12
[Feature][api][java] Support DataStream / Table as input / output of Flink Agents in Java


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support DataStream / Table as input / output of Flink Agents in Java

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

xli1996:
Hi, I'm willing to submit a PR!

PR #85, opened 2025-08-03
[api][java] Support DataStream / Table as input / output of Flink Agents in Java

letaoj:
Linked issue: #20

### Purpose of change

Support DataStream and Table as input/output of Agent in java


api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
```diff
@@ -0,0 +1,161 @@
...
+     * @return AgentsExecutionEnvironment appropriate for the execution context.
+     */
+    public static AgentsExecutionEnvironment getExecutionEnvironment(
+            StreamExecutionEnvironment env) {
+        if (env == null) {
+            // Return local execution environment for testing/development
+            try {
+                Class<?> localEnvClass =
+                        Class.forName(
+                                "org.apache.flink.agents.runtime.env.LocalExecutionEnvironment");
```
letaoj, line 58: use class reference to break the cyclic dependency problem.

examples/pom.xml
```diff
@@ -28,4 +28,27 @@ under the License.
...
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-runtime</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
xintongsong: Which part of this module needs to depend on flink-agents-runtime? I'm asking because by design users should not need to depend on non-api modules when building their agents. If there're something they need, we probably should consider moving it to the api module.

runtime/src/main/java/org/apache/flink/agents/runtime/env/LocalExecutionEnvironment.java
```diff
@@ -0,0 +1,242 @@
...
+        throw new UnsupportedOperationException(
+                "LocalExecutionEnvironment does not support fromTable. Use RemoteExecutionEnvironment instead.");
+    }
+
+    @Override
+    public void execute() throws Exception {
+        // For local execution, the execution happens immediately when toList() is
+        // called
+        // This method is kept for API compatibility
+    }
```
xintongsong: I think this is against intuition. In python, `to_list()` provides an empty output list, which will be filled in after `execute()`. We may want to keep the same behavior here.

runtime/src/main/java/org/apache/flink/agents/runtime/env/LocalExecutionEnvironment.java
```diff
@@ -0,0 +1,242 @@
...
+                    "LocalAgentBuilder does not support toDataStream. Use toList instead.");
+        }
+
+        @Override
+        public Table toTable(Schema schema) {
+            throw new UnsupportedOperationException(
+                    "LocalAgentBuilder does not support toTable. Use toList instead.");
+        }
+
+        private List<Map<String, Object>> executeAgent() {
```
xintongsong: I'm not sure about implementing this in the LocalExecutionEnvironment. As the projects evolves, the agent execution logics can become more and more complex.

In python, we introduced the local mode so that users can easily execute the agent in their python environment, without relying on a JDK/JVM or a Flink cluster.

For java users, they already have the java environment, and Flink also supports local execution (MiniCluster). So I think it might make sense to rely on Flink's MiniCluster for executing the agents locally. In this way, we can reuse the same agent execution implementation as in the remote mode, rather than re-implement it here.

xintongsong: I would also suggest to keep this PR concise and leave supporting local execution in java as a separate follow-up issue.

runtime/src/main/java/org/apache/flink/agents/runtime/env/RemoteExecutionEnvironment.java
```diff
@@ -0,0 +1,154 @@
...
+
+            return outputDataStream;
+        }
+
+        @Override
+        public Table toTable(Schema schema) {
+            if (tableEnv == null) {
+                throw new IllegalStateException(
+                        "Table environment not available. Use fromTable() method to enable table output.");
+            }
```
xintongsong, line 148: Nice catch~! I checked the python implementation, and it is indeed a problem that if you create the agent builder with `from_datastream`, you won't be able output an table with `to_table`. But this is not the desired behavior

@wenjin272, I think we should fix this. Let's first create an issue to track it.

wenjin272: Yes, there is bug in python implementation. I create issue #93 to track this and will fix it these days. Thanks @letaoj for the carefully work.

examples/pom.xml
letaoj: I think I was testing out something in this folder, which requires to use the `LocalExecutionEnvironment` (which is now removed). I will remove this dependency for good.

runtime/src/main/java/org/apache/flink/agents/runtime/env/LocalExecutionEnvironment.java
letaoj: Thanks for the suggestion. I'm still new to Flink so I do not have much experience with Local testing. I will remove this file for now and look into the MiniCluster that you referred in a follow-up PR for the local testing environment

runtime/src/main/java/org/apache/flink/agents/runtime/env/RemoteExecutionEnvironment.java
letaoj: Should the correct logic be, if the table env is null, then we will create one using `StreamTableEnvironment.create(env)` and then use it?

wenjin272: @letaoj Yes, I think the correct logic is as your description.

xintongsong: I would suggest to pass `tableEnv` into `AgentsExecutionEnvironment` in `getExecutionEnvironment()`, rather than in `fromTable()`.

That means when the user creates an `AgentsExecutionEnvironment`, they can choose to provide both `StreamExecutionEnvironment` and `StreamTableEnvironment` with custom settings, or one / none of them. When not provided, a default `StreamExecutionEnvironment` / `StreamTableEnvironment` will be created.

Imagine a user wants to apply an agent from a data stream and to a table, and at the same time needs a custom table environment. In that case, you won't have chance to set the custom table environment via `from_table` because it's never called.

runtime/src/main/java/org/apache/flink/agents/runtime/env/LocalExecutionEnvironment.java
xintongsong: Thank you. I also created an issue (#95) to track this.

xintongsong (approved):
Thanks for addressing my comments. LGTM. Merging.

PR #88, opened 2025-08-05
[infra] fix: improve lint.sh and add license header check tool

codenohup:
### Purpose of change
1. The current code formatting tool _lint.sh_ modifies source files by default when running style checks, which allows non-compliant code to pass the validation unintentionally.
2. There is currently no tool in place to check for the presence of Apache license headers

### Tests
Manually tested locally and confirmed to be working.

### API
No.

### Documentation
Updated the relevant documentation.

### DIFF
--- a/python/flink_agents/api/agent.py
+++ b/python/flink_agents/api/agent.py
@@ -18,7 +18,7 @@
 from abc import ABC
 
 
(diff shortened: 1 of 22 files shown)


xintongsong (approved):
LGTM

Issue #69, opened 2025-07-16
[Feature][plan][runtime][java] Introduce Resource and Resource Provider in java


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce Resource and Resource Provider in java

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

kitalkuyo-gita:
# Resource and Resource Provider Feature Design Proposal

Based on my analysis of the Apache Flink Agents codebase, I've designed a comprehensive proposal for introducing Resource and Resource Provider functionality in Java:

## 1. Overall Architecture Design

### 1.1 Core Concept Definitions

**Resource**: An abstracted manageable resource including compute resources, storage resources, network resources, external service connections, etc. Each resource has lifecycle management (create, use, release).

**Resource Provider**: A component responsible for creating, configuring, and managing specific types of resources, similar to the existing [1](#1-0) .

### 1.2 Module Structure

I recommend extending the existing module structure:

- **API Module**: Add Resource-related core interfaces
- **Runtime Module**: Implement resource management runtime components  
- **Plan Module**: Support resource declaration and binding

## 2. API Module Design

### 2.1 Core Interface Design

Add the following interfaces in the `api/src/main/java/org/apache/flink/agents/api/` directory, maintaining consistency with the existing Event system [2](#1-1) :

**Resource Interface**:
```java
// api/src/main/java/org/apache/flink/agents/api/Resource.java
public interface Resource extends AutoCloseable {
    String getId();
    ResourceType getType();
    ResourceStatus getStatus();
    Map<String, Object> getProperties();
    void initialize() throws ResourceException;
    boolean isReady();
}
```

**ResourceProvider Interface**:
```java
// api/src/main/java/org/apache/flink/agents/api/ResourceProvider.java
public interface ResourceProvider<T extends Resource> {
    ResourceType getSupportedType();
    T createResource(ResourceConfig config) throws ResourceException;
    void releaseResource(T resource) throws ResourceException;
    boolean isHealthy();
}
```

### 2.2 Supporting Types and Enums

```java
// api/src/main/java/org/apache/flink/agents/api/ResourceType.java
public enum ResourceType {
    PYTHON_ENVIRONMENT,
    DATABASE_CONNECTION,
    HTTP_CLIENT,
    FILE_SYSTEM,
    MEMORY_CACHE,
    CUSTOM
}

[...]

PR #91, opened 2025-08-05
[plan][runtime][java] Introduce Resource and Resource Provider in java

letaoj:
Linked issue: #69 

### Purpose of change

Introduce resource and resource provider concept in java to mimic the same behavior in python. This is the first step to introduce Tool, chat_model and prompt.

### Tests

Unit test added, however some of the code are not covered as we only have base class that is abstract which can't be instantiate for use.

### API

Added resource and resource provider in java

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/resource/ResourceType.java
+++ b/api/src/main/java/org/apache/flink/agents/api/resource/ResourceType.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.resource;
+
+/**
+ * Type enum of resource.
+ *
+ * <p>Currently, only support chat models and tools.
+ */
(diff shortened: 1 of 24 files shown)


api/src/main/java/org/apache/flink/agents/api/resource/ResourceType.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Type enum of resource.
+ *
+ * <p>Currently, only support chat models and tools.
+ */
+public enum ResourceType {
+    CHAT_MODEL("chat_model"),
+    TOOL("tool");
+    // EMBEDDING_MODEL("embedding_model"),
```
istavrak: NIT: We may want to remove the commented out enums, as at the moment of update in the future the maintainer will need to check the Python implementation anyhow.

letaoj: removed.

api/src/main/java/org/apache/flink/agents/api/Resource.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.io.Serializable;
+import java.util.function.BiFunction;
+
+/**
+ * Base interface for all kinds of resources, including chat models, tools, prompts and so on.
+ *
+ * <p>Resources are components that can be used by agents during action execution.
+ */
+public interface Resource extends Serializable {
```
wenjin272: In python implementation, there are two type of Resource:
1. Resource which is not json serializable, link ChatModel, for it may contain some fields which are not json serializable.
2. SerializableResource which is json serializable, link function tool.

For Resource we construct it in runtime, and for SerializableResource we can get or deserialize it in runtime. 

In currently implementation, for there is no resource implementation which is not json serializable, only provide resource is enough, but for clarity of semantics, I think may be we can provide distinction at the beginning

letaoj: Hi @wenjin272 Thank you for the review. I will update the PR to reflect the concept described in this comment.

wenjin272:
LGTM, please take a look at your convenience @xintongsong.

xintongsong:
Hi @letaoj,
The PR also looks good to me. However, it cannot be merged due to conflicts. I think it's because you have used `git merge` to adapt latest changes from the main branch. Please use `git rebase` instead.

xintongsong (approved):
Thanks, @letaoj. LGTM. Merging.

Issue #77, opened 2025-07-23
[Feature][api][java] Introduce Prompt, Tool and ChatModel in java


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce Prompt, Tool and ChatModel in java

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

letaoj:
@xintongsong https://github.com/apache/flink-agents/issues/56#issuecomment-3106914894 This is what I was referring to during the community sync. Should I wait for Amen to implement this issue?

PR #94, opened 2025-08-06
WIP: [Feature][API] Introduce ChatModel,Tool and Prompt in Java

AmenAdem:
Linked issue: #77

### Purpose of change

This PR implements the core tool system infrastructure for Flink Agents, including:

### Tests
No unit tests implemented yet

### API
Yes, this change introduces new public APIs

### Documentation
es, this should be documented as it provides the foundational tool execution system. Documentation will be added once the implementation is finalized and unit tests are complete.


api/src/main/java/org/apache/flink/agents/api/chat/messages/AbstractMessage.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * The AbstractMessage class is an abstract implementation of the Message interface. It provides a
+ * base implementation for message content, media attachments, metadata, and message type.
+ *
+ * @see Message
+ */
+public abstract class AbstractMessage implements Message {
```
wenjin272: In your implementation, there are AbstractMessage and different AbstractMessage implementation, like UserMessage, SystemMessage and ToolMessage, which is similar to langChain.
In python implementation, there is only ChatMessage with different message role, can represent all user、system、tool message, which is similar to llama_index.

I think the former is more flexible but more complex, and the latter is more simple but enough for current project. 

For api compatibility between python and java, I suggest use the latter abstraction.

api/src/main/java/org/apache/flink/agents/api/prompt/Prompt.java
```diff
@@ -0,0 +1,221 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * A simplified Prompt class that represents a conversation prompt for AI models. Contains a list of
+ * messages and optional chat options.
+ */
+public class Prompt implements ChatRequest<List<Message>> {
```
wenjin272: Could provide create Prompt form text string?

api/src/main/java/org/apache/flink/agents/api/tool/ToolMetadata.java
```diff
@@ -0,0 +1,241 @@
...
+ * Metadata for a tool including its definition, invocation method, and parameters. This class holds
+ * all the information needed to register and invoke a tool.
+ */
+public class ToolMetadata {
+
+    private final ToolDefinition definition;
+    private final ToolType type;
+    private final Method method;
+    private final Object instance;
+    private final List<ToolParameterInfo> parameters;
```
wenjin272: What's the usage of parameters? Dose the inputSchema in ToolDefinition is enough?

api/src/main/java/org/apache/flink/agents/api/tool/ToolRegistry.java
```diff
@@ -0,0 +1,139 @@
...
+
+import java.lang.reflect.Method;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Registry for managing tools available to agents. Works with the merged design where BaseTool
+ * contains ToolMetadata.
+ */
+public class ToolRegistry {
```
wenjin272: Does the ToolRegistry will be used when convert a user declared Agent to AgentPlan? Maybe we need add an integration test to explain the usage of ToolRegistry like test_get_resource_in_action.py in python.

api/src/main/java/org/apache/flink/agents/api/tool/FunctionTool.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Function-based tool implementation that wraps annotated methods. Now uses the merged design with
+ * ToolMetadata.
+ */
+public class FunctionTool extends BaseTool {
+
+    private final Method method;
```
wenjin272: Maybe we can reuse the JavaFunction to represent the function in FunctionTool.

api/src/main/java/org/apache/flink/agents/api/tool/ToolRegistry.java
```diff
@@ -0,0 +1,139 @@
...
+        String name = tool.getName();
+        if (tools.containsKey(name)) {
+            throw new IllegalArgumentException(
+                    "Tool with name '" + name + "' is already registered");
+        }
+        tools.put(name, tool);
+    }
+
+    /** Register all @Tool annotated methods from an object instance. */
+    public void registerToolsFromObject(Object toolProvider) {
```
wenjin272: Maybe we can only support static method as FunctionTool at first to simplify this pr, and add other support later if necessary.

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModel.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.flink.agents.api.prompt.Prompt;
+
+public interface BaseChatModel {
+    /**
+     * Process a chat request and return a chat response.
+     *
+     * @param request the chat request containing messages, options, etc.
+     * @return the chat response containing model outputs
+     */
+    ChatResponse call(Prompt request);
```
wenjin272: Maybe we can rename this method to 'chat'.

I suggest the input and output to be consistent with python implementation, or the api compatibility between python and java will be hard to maintain.

wenjin272 (commented):
Hi, @AmenAdem, thanks for working on this PR. I believe this must have taken a lot of your effort.

I left some comments on this pr, and apart from these, I have several overall comments

- The python implementation #76 has been merged, which you can refer to. For the api compatibility between Python and Java, I suggest to keep the consistency of interface signature between Python and Java. And if you think there are unreasonable design in Python part, also can point it.
- Add some integration tests or examples to show how can user declare prompt, tool and chat model in agent, and how the framework use it when run an agent.

wenjin272:
The pr this pr depends on #91 is nearly merged. You could rebase on it.

api/src/main/java/org/apache/flink/agents/api/chat/messages/AbstractMessage.java
AmenAdem: To be honest, I'm not fully convinced that the single ChatMessage class approach is the most optimal design choice - I can see the benefits of the more flexible AbstractMessage hierarchy (like LangChain's approach) for extensibility and type safety.
However, for the sake of API alignment between Python and Java implementations, I agree we should go with the single ChatMessage + MessageRole pattern that matches the existing Python codebase.
done !

api/src/main/java/org/apache/flink/agents/api/tool/ToolRegistry.java
AmenAdem: Ok, but Do we want to throw an exception if someone tries to register non-static @Tool methods? or should we only look for @Tool on static methods in the initial implementation?

api/src/main/java/org/apache/flink/agents/api/tool/ToolMetadata.java
AmenAdem: You re absolutely right the inputSchema in ToolDefinition should be sufficient.
I added the parameters field because I was thinking about runtime performance optimization. my reasoning was:
Avoiding repeated JSON schema parsing - the List<ToolParameterInfo> would be pre-parsed parameter metadata for faster runtime access

api/src/main/java/org/apache/flink/agents/api/tool/ToolRegistry.java
AmenAdem: You're right that the integration between ToolRegistry and the Agent → AgentPlan conversion process needs clarification.
Currently, the ToolRegistry operates more as a standalone tool management system (as shown in ToolExample.java with registry.invokeTool()), but its role in the AgentPlan compilation pipeline isn't clear.
I'm working on this integration now and thinking through two possible approaches:
* AgentPlan auto-discovery: When creating an AgentPlan from an agent, automatically scan for @Tool annotated methods and register them in a plan-specific registry.
* Explicit registration: Developers register tools in a ToolRegistry first, then reference them in agents

api/src/main/java/org/apache/flink/agents/api/tool/example/ToolExample.java
```diff
@@ -0,0 +1,131 @@
...
+    public static void main(String[] args) {
+        // Create tool registry
+        ToolRegistry registry = new ToolRegistry();
+
+        // Register tools from an instance
+        ToolExample example = new ToolExample();
+        registry.registerToolsFromObject(example);
+
+        // Register static tools from class
+        registry.registerToolsFromClass(ToolExample.class);
```
wenjin272: This will cause already registered exception

api/src/main/java/org/apache/flink/agents/api/tool/example/ToolExample.java
```diff
@@ -0,0 +1,131 @@
...
+
+        // Example 1: Use the calculate tool
+        ToolParameters calcParams = new ToolParameters();
+        calcParams.addParameter("operation", "add");
+        calcParams.addParameter("a", 10.5);
+        calcParams.addParameter("b", 5.2);
+
+        ToolResponse calcResult = registry.invokeTool("calculate", calcParams);
```
wenjin272: This will cause Invalid parameters error, for parameters name is not match. The parameter name in tool is arg0, arg1 and arg2 for `method.getParameters()` in `createSchemaFromMethod` not returns the declared name in method.

api/src/main/java/org/apache/flink/agents/api/tool/ToolRegistry.java
wenjin272: I think the both approaches are look good to me, and we can provide the first approach now, and provide the second approach later.  
For the first approach, we can use a plan-specific registry to help tool register, and maybe we need store the registered tools in registry directly in AgentPlan rather than the registry for keep compatibility with python AgentPlan.

api/src/main/java/org/apache/flink/agents/api/tool/ToolRegistry.java
wenjin272: The reason I propose only support static method currently is as a part of  Flink job, the AgentPlan will be compiled in JobManager, and will be distributed to TaskManagers, so the tools in AgentPlan need to support serialization. For tools from non-static methods, the object should also be serializable, which may cause some complexity. 

I think maybe we can let the ToolRegistry support register both static and non-static methods, but only support static method as Tool when declare an Agent.

xintongsong:
@AmenAdem Thanks for working on this. The PR currently contains commits that do not belong to it, which makes it hard to review and merge. Could you please re-organize the commits?

FYI, we use `git rebase` rather than `git merge` in this project, to keep a clear linear git history. That means PRs containing commits that are already merged in the codebase can meet conflicts during rebasing. Such conflicts are better resolved before reviewing.

AmenAdem:
@xintongsong thanks for pointing that out. I ve dropped the unrelated commit, and the branch now contains only the relevant changes commits are clean and ready for review.

api/src/main/java/org/apache/flink/agents/api/chat/messages/ToolCall.java
```diff
@@ -0,0 +1,85 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.chat.messages;
+
+import java.util.Objects;
+
+public class ToolCall {
```
wenjin272: Class 'ToolCall' is never used.

api/src/main/java/org/apache/flink/agents/api/tools/BaseTool.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /** Get the metadata of this tool. */
+    public final ToolMetadata getMetadata() {
+        return metadata;
+    }
+
+    /** Return resource type of class. */
+    @Override
+    public final ResourceType getResourceType() {
```
wenjin272: Maybe this could be static method

api/src/main/java/org/apache/flink/agents/api/util/CollectionUtils.java
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.util;
+
+import java.util.Collection;
+import java.util.Map;
+
+/** Utility methods for Collection operations in the Flink Agents framework. */
+public final class CollectionUtils {
```
wenjin272: Class 'CollectionUtils' is never used

api/src/main/java/org/apache/flink/agents/api/util/StringUtils.java
```diff
@@ -0,0 +1,79 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.util;
+
+/** Utility methods for String operations in the Flink Agents framework. */
+public final class StringUtils {
```
wenjin272: Class 'StringUtils' is never used.

api/src/main/java/org/apache/flink/agents/api/chat/messages/ChatMessage.java
```diff
@@ -0,0 +1,173 @@
...
+    public static final String MESSAGE_TYPE = "messageType";
+
+    private MessageRole role;
+    private String content;
+    private List<Map<String, Object>> toolCalls;
+    private Map<String, Object> extraArgs;
+
+    /** Default constructor with USER role */
+    public ChatMessage() {
+        this.role = MessageRole.USER;
```
wenjin272: The default role could be System to be aligned with python implementation.

plan/src/main/java/org/apache/flink/agents/plan/resourceprovider/ToolResourceProvider.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.concurrent.Callable;
+
+/**
+ * Resource provider for tools created from static Java methods annotated with @Tool. Stores the
+ * reflective identity to reconstruct a FunctionTool at runtime.
+ */
+public class ToolResourceProvider extends ResourceProvider {
```
wenjin272: Can we use JavaSerializableResourceProvider rather than introduce a new type of ResourceProvider, which add the complexity of the serialization/deserialization of AgentPlan and also break the AgentPlan compatibility between python and java.

You may modify the currently implementation of JavaSerializableResourceProvider and you can refer to the PythonSerializableResourceProvider in python. The key assumption is the resource in SerializableResourceProvider should be json serializable. 

In our design, Tool should be regarded as SerializableResource, and be json serializable.

plan/src/main/java/org/apache/flink/agents/plan/resourceprovider/JavaSerializableResourceProvider.java
```diff
@@ -39,14 +41,18 @@ public JavaSerializableResourceProvider(
 
     @Override
     public Resource provide(Callable<Resource> getResource) throws Exception {
```
wenjin272: For SerializableResourceProvider, we should get the SerializableResource by deserialization rather than construct it.  You can refer to the PythonSerializableResourceProvider in python implementation.

api/src/main/java/org/apache/flink/agents/api/prompt/Prompt.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Prompt for a language model
+ *
+ * <p>The template can be either a string or a sequence of ChatMessage objects.
+ */
+public class Prompt extends SerializableResource {
```
xintongsong, line 44: I'd suggest to make `PromptTemplate`, `MessagesTemplate`, `StringTemplate` private inner classes of `Prompt`. Because while these classes are in the api module, they are for internal usages and we do not want users to use them.

api/src/main/java/org/apache/flink/agents/api/prompt/PromptUtils.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /** Extract variable names from template */
+    public static java.util.Set<String> extractVariables(String template) {
+        java.util.Set<String> variables = new java.util.HashSet<>();
+        Matcher matcher = BRACE_PATTERN.matcher(template);
+        while (matcher.find()) {
+            variables.add(matcher.group(1));
+        }
+        return variables;
+    }
```
xintongsong: I'd suggest to make these methods private static methods of `Prompt`.

api/src/main/java/org/apache/flink/agents/api/prompt/PromptUtils.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /** Validate that all required variables are provided */
+    public static void validateTemplate(String template, Map<String, String> kwargs) {
```
xintongsong: This method is never used. Shall we remove it?

api/src/main/java/org/apache/flink/agents/api/annotation/Tool.java
```diff
@@ -24,18 +24,34 @@
...
+     */
+    String description() default "";
+
+    /**
+     * Whether the tool result should be returned directly or passed back to the model.
+     *
+     * @return true if the tool result should be returned directly, false if it should be passed
+     *     back to the model
+     */
+    boolean returnDirect() default false;
```
xintongsong: Unused

api/src/main/java/org/apache/flink/agents/api/annotation/Tool.java
```diff
@@ -24,18 +24,34 @@
...
+@Target({ElementType.FIELD, ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Tool {
     /**
-     * The name of the tool resource. If not specified, the field name will be used.
+     * The name of the tool resource. If not specified, the field/method name will be used.
      *
      * @return the resource name
      */
     String name() default "";
```
xintongsong: We should not accept empty string as default name.

api/src/main/java/org/apache/flink/agents/api/annotation/ToolParam.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    boolean required() default true;
+
+    /**
+     * The name of the tool argument.
+     *
+     * @return a string representing the name of the argument, which can be used to identify it in
+     *     the tool's context.
+     */
+    String name() default "";
```
xintongsong, line 47: I'd suggest to use the parameter name, rather than introducing an extra name in the annotation.

api/src/main/java/org/apache/flink/agents/api/tools/BaseTool.java
```diff
@@ -0,0 +1,62 @@
...
+package org.apache.flink.agents.api.tools;
+
+import org.apache.flink.agents.api.resource.ResourceType;
+import org.apache.flink.agents.api.resource.SerializableResource;
+
+/**
+ * Simplified base class for all tools. Focus on essential tool functionality without
+ * overcomplicated features.
+ */
+public abstract class BaseTool extends SerializableResource {
```
xintongsong: I think the following classes should belong to either the plan or runtime module, because we do not support users to extend `BaseTool` with custom implementations. They can only either annotate a method as a tool, or leverage tools from a mcp server.
- BaseTool
- SchemaUtils
- ToolMetadata
- ToolParameters
- ToolResponse
- ToolType

api/src/main/java/org/apache/flink/agents/api/tools/SchemaUtils.java
```diff
@@ -0,0 +1,259 @@
...
+        TYPE_MAPPING.put(Boolean.class, "boolean");
+        TYPE_MAPPING.put(boolean.class, "boolean");
+        TYPE_MAPPING.put(Object.class, "object");
+        TYPE_MAPPING.put(Map.class, "object");
+        TYPE_MAPPING.put(java.util.List.class, "array");
+        TYPE_MAPPING.put(java.util.Collection.class, "array");
+    }
+
+    /** Create a JSON schema from a method annotated with @Tool. */
+    public static String createSchemaFromMethod(Method method) {
```
xintongsong: Never used

api/src/main/java/org/apache/flink/agents/api/tools/SchemaUtils.java
```diff
@@ -0,0 +1,259 @@
...
+        } else if (targetType == Float.class || targetType == float.class) {
+            return Float.parseFloat(defaultValue);
+        } else if (targetType == Boolean.class || targetType == boolean.class) {
+            return Boolean.parseBoolean(defaultValue);
+        }
+        return defaultValue;
+    }
+
+    /** Validate parameters against a JSON schema. */
+    public static boolean validateParameters(ToolParameters parameters, String schema) {
```
xintongsong: never used

api/src/main/java/org/apache/flink/agents/api/tools/ToolException.java
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.tools;
+
+/**
+ * Exception thrown when tool operations fail. Provides specific error handling for tool-related
+ * failures.
+ */
+public class ToolException extends RuntimeException {
```
xintongsong: Never used

api/src/main/java/org/apache/flink/agents/api/chat/messages/Message.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.chat.messages;
+
+import java.util.Map;
+
+public interface Message {
```
xintongsong: Why do we need this abstraction? There's no implementation other than `ChatMessage`, and the methods `getMessageType()` and `getText()` are never used.

api/src/main/java/org/apache/flink/agents/api/chat/messages/ToolCall.java
xintongsong: +1

plan/src/main/java/org/apache/flink/agents/plan/resourceprovider/ToolResourceProvider.java
xintongsong: +1

plan/src/main/java/org/apache/flink/agents/plan/serializer/ResourceProviderJsonDeserializer.java
```diff
@@ -115,6 +119,24 @@ private JavaSerializableResourceProvider deserializeJavaSerializableResourceProv
                 name, ResourceType.fromValue(type), module, clazz);
     }
 
+    private ToolResourceProvider deserializeToolResourceProvider(JsonNode node) {
```
xintongsong: I think the serializer and deserializer should not understand anything about the tools. They should only see SerializableResource as a blackbox.

AmenAdem:
Hi @wenjin272 Sorry for the delayed response I ve been quite busy lately. I ll definitely have time this weekend to address the remaining comments and get this PR updated.

xintongsong:
Hi @AmenAdem,

@wenjin272 and I have updated the PR, based on the work you have already completed, and addressed the remaining comments.

Sorry we had to take over like this. We are approaching the target date of the first release of this project, and there are quite some subsequent tasks blocked by this PR. I hope you can understand.

xintongsong (approved):

Issue #102, opened 2025-08-10
[Feature] [api] Support Event Logger and Event Listener in Java


alnzng:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce event logger and event listener mechanism 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #103, opened 2025-08-11
[Feature] [API] Support Event Logger and Event Listener in Java

alnzng:
Linked issue: #102 

### Purpose of change

This PR introduces a new feature which allows the users to enable event logger and event listeners to monitor the action events.

### Tests

Unit tests and e2e example

### API

Add new `EventLogger` and `EventListener` APIs

### Documentation

I will add a doc for this separately

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
+++ b/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
@@ -129,7 +129,23 @@ public abstract class AgentsExecutionEnvironment {
      * @return The current AgentsExecutionEnvironment instance for method chaining.
      */
     public AgentsExecutionEnvironment registerEventListener(EventListener listener) {
-        eventListeners.add(listener);
+        return registerEventListener(listener, EventFilter.ACCEPT_ALL);
+    }
+
+    /**
+     * Register an event listener with a filter for the execution environment.
+     *
+     * <p>This method allows users to register custom EventListeners that will only be notified of
+     * events matching the specified EventFilter. This is useful for filtering events before
+     * processing them in the listener.
+     *
+     * @param listener Custom EventListener instance to be registered.
+     * @param filter EventFilter to apply to the events processed by the listener.
+     * @return The current AgentsExecutionEnvironment instance for method chaining.
+     */
+    public AgentsExecutionEnvironment registerEventListener(
+            EventListener listener, EventFilter filter) {
+        eventListeners.add(EventListener.withFilter(listener, filter));
         return this;
     }
 
--- a/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
+++ b/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
@@ -92,16 +92,30 @@ public abstract class AgentsExecutionEnvironment {
     }
 
     /**
-     * Set an event logger for the execution environment.
+     * Enable an event logger for the execution environment.
      *
      * <p>This method allows users to set a custom EventLogger that will be used to log events
-     * during agent execution. This is useful for monitoring and debugging agent behavior.
+     * during agent execution.
      *
      * @param eventLogger Custom EventLogger instance.
      * @return The current AgentsExecutionEnvironment instance for method chaining.
      */
-    public AgentsExecutionEnvironment setEventLogger(EventLogger eventLogger) {
-        this.eventLogger = eventLogger;
+    public AgentsExecutionEnvironment enableEventLogger(EventLogger eventLogger) {
+        return enableEventLogger(eventLogger, EventFilter.ACCEPT_ALL);
+    }
+
+    /**
+     * Enable an event logger with a filter for the execution environment.
+     *
+     * <p>This method allows users to set a custom EventLogger along with an EventFilter to control
+     * which events are logged. This is useful for reducing noise and focusing on relevant events.
+     *
+     * @param logger Custom EventLogger instance.
+     * @param filter EventFilter to apply to logged events.
+     * @return The current AgentsExecutionEnvironment instance for method chaining.
+     */
+    public AgentsExecutionEnvironment enableEventLogger(EventLogger logger, EventFilter filter) {
+        this.eventLogger = EventLogger.withFilter(logger, filter);
         return this;
     }
 
--- a/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
+++ b/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
@@ -108,8 +108,8 @@ public abstract class AgentsExecutionEnvironment {
     /**
      * Register an event listener for the execution environment.
      *
-     * <p>This method allows users to register custom EventListeners that will be notified of
-     * events during agent execution. This is useful for implementing custom event handling logic.
+     * <p>This method allows users to register custom EventListeners that will be notified of events
+     * during agent execution. This is useful for implementing custom event handling logic.
      *
      * @param listener Custom EventListener instance to be registered.
      * @return The current AgentsExecutionEnvironment instance for method chaining.
--- a/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
+++ b/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
@@ -18,12 +18,15 @@
 
 package org.apache.flink.agents.api;
 
+import org.apache.flink.agents.api.listener.EventListener;
+import org.apache.flink.agents.api.logger.EventLogger;
 import org.apache.flink.api.java.functions.KeySelector;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.table.api.Table;
 import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
 
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -34,6 +37,8 @@ import java.util.List;
  */
 public abstract class AgentsExecutionEnvironment {
 
+    protected EventLogger eventLogger;
+    protected List<EventListener> eventListeners = new ArrayList<>();
     /**
      * Get agents execution environment.
      *
@@ -84,6 +89,34 @@ public abstract class AgentsExecutionEnvironment {
      */
     public static AgentsExecutionEnvironment getExecutionEnvironment() {
         return getExecutionEnvironment(null);
+    }
+
+    /**
+     * Set an event logger for the execution environment.
+     *
+     * <p>This method allows users to set a custom EventLogger that will be used to log events
+     * during agent execution. This is useful for monitoring and debugging agent behavior.
+     *
+     * @param eventLogger Custom EventLogger instance.
+     * @return The current AgentsExecutionEnvironment instance for method chaining.
+     */
+    public AgentsExecutionEnvironment setEventLogger(EventLogger eventLogger) {
+        this.eventLogger = eventLogger;
+        return this;
+    }
+
+    /**
+     * Register an event listener for the execution environment.
+     *
+     * <p>This method allows users to register custom EventListeners that will be notified of
+     * events during agent execution. This is useful for implementing custom event handling logic.
+     *
+     * @param listener Custom EventListener instance to be registered.
+     * @return The current AgentsExecutionEnvironment instance for method chaining.
+     */
+    public AgentsExecutionEnvironment registerEventListener(EventListener listener) {
+        eventListeners.add(listener);
+        return this;
     }
 
     /**
--- a/api/src/main/java/org/apache/flink/agents/api/EventContext.java
+++ b/api/src/main/java/org/apache/flink/agents/api/EventContext.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api;
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Context for events in the Flink Agents API.
+ *
+ * <p>EventContext provides metadata about the event, including event key, timestamp, and additional
+ * attributes that can be used to filter or categorize events.
+ */
+public class EventContext {
+    private final Object key;
+    private final Instant timestamp;
+    private final Map<String, Object> attributes;
+
+    public EventContext(Object key) {
+        this.key = key;
+        this.timestamp = Instant.now();
+        this.attributes = new HashMap<>();
+    }
+
+    public Object getKey() {
+        return key;
+    }
+
+    public Instant getTimestamp() {
+        return timestamp;
+    }
+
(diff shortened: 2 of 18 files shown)


api/src/main/java/org/apache/flink/agents/api/logger/EventLogger.java
```diff
@@ -0,0 +1,116 @@
...
+     * Unwraps the EventLogger if it is wrapped by another logger or filter.
+     *
+     * <p>This method can be overridden by implementations that wrap other loggers or filters to
+     * return the underlying logger.
+     *
+     * @return the underlying EventLogger instance
+     */
+    default EventLogger unwrap() {
+        return this;
+    }
```
xintongsong: Why do we need this?

api/src/main/java/org/apache/flink/agents/api/logger/EventLogger.java
```diff
@@ -0,0 +1,116 @@
...
+import org.apache.flink.agents.api.Event;
+import org.apache.flink.agents.api.EventContext;
+import org.apache.flink.agents.api.EventFilter;
+
+/**
+ * Interface for logging action events in the Flink Agents framework.
+ *
+ * <p>EventLogger provides a unified interface for capturing, filtering, and persisting events as
+ * they flow through the agent execution pipeline. Implementations can target different storage
+ * backends such as Flink state, external databases, or file systems.
```
xintongsong, line 29: I'm not sure about logging the events into the Flink state.
1. The purpose of adding event log is for users to understand what happened in the agent. If the logs are in state, how does the user read them?
2. The size of the list state will keep increasing, since we never removed anything from it. This could be problematic.
3. During failover, the state will be restored from a checkpoint. This means events after the checkpoint and before the failure will be lost. (Not entirely sure whether we need them or not.)

As the first event logger implementation, I would suggest to start with logging the events into a file. (Just another log file of the Flink cluster). Then in the next step we can apply rolling strategies on the files.

WDYT?

api/src/main/java/org/apache/flink/agents/api/EventContext.java
```diff
@@ -0,0 +1,53 @@
...
+        return key;
+    }
+
+    public Instant getTimestamp() {
+        return timestamp;
+    }
+
+    public Map<String, Object> getAttributes() {
+        return attributes;
+    }
```
xintongsong: Why do we need this?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -176,13 +184,26 @@ public void open() throws Exception {
...
         tryResumeProcessActionTasks();
     }
 
+    private void initEventLogger() throws Exception {
+        if (eventLogger == null) {
+            return;
+        }
+        if (eventLogger.unwrap() instanceof FlinkStateEventLogger) {
+            ((FlinkStateEventLogger) eventLogger).setStreamingRuntimeContext(getRuntimeContext());
+        }
```
xintongsong: It seems this is the only place we need `unwrap()`. In this case, it might be easier to pass the runtime context into `open()`. To make this extendable, we can introduce an `EventLoggerOpenParams`, so that we can add more things in it as needed by new event logger implementations, without modifying existing loggers.

xintongsong: IIUC, the reason we need `EventFilter.ACCEPT_ALL` is also that we need to make sure here the `eventLogger` is a wrapped one? Moving the arguments to `open()`, we can remove `unwrap()`, as well as `ACCEPT_ALL`.

api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
```diff
@@ -86,6 +91,64 @@ public static AgentsExecutionEnvironment getExecutionEnvironment() {
...
+     * which events are logged. This is useful for reducing noise and focusing on relevant events.
+     *
+     * @param logger Custom EventLogger instance.
+     * @param filter EventFilter to apply to logged events.
+     * @return The current AgentsExecutionEnvironment instance for method chaining.
+     */
+    public AgentsExecutionEnvironment enableEventLogger(EventLogger logger, EventFilter filter) {
+        this.eventLogger = EventLogger.withFilter(logger, filter);
+        return this;
+    }
```
xintongsong: I think we need someway to support users setting the event logger and filter with configs. @GreatEugenius is working on the configuration mechanism of Flink Agents. Once the mechanism is ready, I think it should be straightforward to enable the event logger. But for the filters, we might need some further design on how to express the filter rules with configurations.

xintongsong (commented):
Thanks for working on this, @alnzng. I've left some comments. PTAL.

api/src/main/java/org/apache/flink/agents/api/logger/EventLogger.java
alnzng: Great questions.

I was thinking that we could leverage on Flink internal state mechanism to help event log replay or time travel kind of requierments. The end users should be able to store these (event log) states into (centrialized) remote storage(e.g. hdfs, s3, etc) and build query on top of it.

Yes, it doesn't have the retention policy setup in this PR, so you are right on that the size will keep increasing. I thought it is possible that we apply some retention policy(e.g. Flink State TTL `StateTtlConfig` or implement a new one either time or size based) on it, and I was thinking we can enable this in the next iteration.

Good point. Agreed that event logs could be lost in these failure scenarios, and I can understand your concern here. From event processing perspective, this doesn't look like a concern. But from event logging perspective, seems we'd better to record all the occurred events and this is important in general in the scenarios that event processing failed.

This PR provides an implementation `FlinkStateEventLogger`, and it is not enabled by default. But with above reasonable concern on data loss, Let me remove it. Technically, the end users are still be able to implement a similar solution by themselves, but it's their responsibilities to maintain it.

Sounds good on offering a default File based event logger, and agree that we should also have rolling / retention strategies for the log files supportted, similar like log4j offerring.

api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
alnzng: Exactly! I was considering to uses configs as the user-facing APIs to enable event logger or listener, but I didn't do this just because we don't have config supporrted. I'm happy to refactor this part once the config mechanism is available.

@GreatEugenius Let me know once you have some things about new config mechanism can share, I can help take a look.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
alnzng: Yes, this `unwrap` is used only for `FlinkStateEventLogger` at this moment. Since we are going to remove this `FlinkStateEventLogger`, this `unwrap` is not that useful so that I can remove it.

>IIUC, the reason we need EventFilter.ACCEPT_ALL is also that we need to make sure here the eventLogger is a wrapped one?

No. Having `EventFilter.ACCEPT_ALL` is just for allowing EventFilter as an optional while enablign the EventLogger. 

```java
public AgentsExecutionEnvironment enableEventLogger(EventLogger eventLogger) {
        return enableEventLogger(eventLogger, EventFilter.ACCEPT_ALL);
    }
```

The wappered EventLogger is for enabling the filter function by default, it is a decorator pattern for EventLogger.

``` java
public void append(EventContext context, Event event) throws Exception {
                if (filter.accept(event, context)) {
                    delegate.append(context, event);
                }
            }
```

api/src/main/java/org/apache/flink/agents/api/EventContext.java
alnzng: The intention was just give the ability to setup more information as the context of event log, but it is not used actually in this PR. I can remove this if you think this is confusing.

api/src/main/java/org/apache/flink/agents/api/logger/EventLogger.java
alnzng: Explained this in below comment. TLDR; this is just for FlinkStateEventLogger, we can remove it since we are going to remove FlinkStateEventLogger.

api/src/main/java/org/apache/flink/agents/api/EventContext.java
xintongsong: I'm leaning towards remove it for the moment. I see another possible alternative that stores such addition information just in the events. Without concrete use cases, I can not tell which approach would be better, and we don't have to decide before the use cases show up.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
xintongsong: But why not just use the original event logger when we need to accept all event? 

It's probably not a big deal, just different ways of implementation. Just it's a bit against intuition for me, that we turn one event logger into another event logger with a filter that does nothing.

api/src/main/java/org/apache/flink/agents/api/logger/EventLogger.java
alnzng: @xintongsong As we discussed, I added a File-based Event Logger implementation `FileEventLogger`. Please take a look and let me know if that makes sense.

examples/src/main/java/org/apache/flink/agents/examples/EventLoggingAndListenerExample.java
```diff
@@ -0,0 +1,293 @@
...
+        // Configure file-based event logging with filtering
+        FileEventLoggerConfig loggerConfig =
+                FileEventLoggerConfig.builder()
+                        .baseEventLogDir("/tmp/flink-agents-logs")
+                        .eventFilter(
+                                EventFilter.byEventType(
+                                        InputEvent.class,
+                                        OrderProcessedEvent.class,
+                                        OutputEvent.class))
+                        .build();
```
xintongsong: It should be a lot easier for users to configure the event logger through a configuration file, rather than the programming api. I even think we should not expose these in the programming api, for simplicity. As the configuration mechanism is not yet ready, maybe we can just hardcode it to always print all events into a default file path, and make it configurable latter.

api/src/main/java/org/apache/flink/agents/api/logger/EventLogRecord.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * <p>This class is used to encapsulate the details of an event as it is logged, allowing for
+ * structured logging and retrieval of event information.
+ *
+ * <p>The class uses custom JSON serialization/deserialization to handle polymorphic Event types by
+ * leveraging the eventType information stored in the EventContext.
+ */
+@JsonSerialize(using = EventLogRecordJsonSerializer.class)
+@JsonDeserialize(using = EventLogRecordJsonDeserializer.class)
+public class EventLogRecord {
```
xintongsong: I think this record class and its (de)serializer should belong to the `runtime.logger` module, rather than `api.logger`.

api/src/main/java/org/apache/flink/agents/api/logger/EventLoggerFactory.java
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * @see EventLogger
+ * @see EventLoggerConfig
+ */
+public final class EventLoggerFactory {
+
+    /** Thread-safe registry of factory functions keyed by config class type. */
+    private static final Map<
+                    Class<? extends EventLoggerConfig>, Function<EventLoggerConfig, EventLogger>>
+            FACTORIES = new ConcurrentHashMap<>();
```
xintongsong: Using the class of config as the key is a bit hacky. This requires each logger implementation to have a dedicated config, which is unnecessary.

I think we can have one unified EventLoggerConfig, with the common params for all logger implementations (e.g., the filters), and a map of arbitrary key-value pairs for implementation-specific params (e.g., log file path).

And we can have a string identifier of logger implementation as the key here. The identifier is also helpful for users to specify which logger should be used with a short string.

examples/pom.xml
```diff
@@ -34,6 +34,11 @@ under the License.
             <artifactId>flink-agents-api</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-runtime</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
xintongsong: Having the example module depending on the runtime module can be an dangerous indicator that the user program needs to depend on some internal, non-api stuff, which is undesired. 

In this case, we are using `FileEventLoggerConfig` from the runtime module, which shouldn't be a problem if we make the event logger hardcoded for the moment.

examples/src/main/java/org/apache/flink/agents/examples/EventLoggingAndListenerExample.java
xintongsong: That also means we don't need to pass the loggerConfig all the way through the long path into the ActionExecutionOperator.

runtime/src/main/java/org/apache/flink/agents/runtime/logger/FileEventLogger.java
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * <pre>
+ * {baseLogDir}/
+ *   └── {jobId}/
+ *       ├── 0/
+ *       │   └── events.log    (subtask 0)
+ *       ├── 1/
+ *       │   └── events.log    (subtask 1)
+ *       └── 2/
+ *           └── events.log    (subtask 2)
```
xintongsong: 1. We may have multiple agents in one flink job. So I think we should also include the task name/id.
2. I'd suggest write all files in one flat directory, with meaningful structured names (e.g., events-<jobid>-<taskid>-<subtaskid>.log). This is more aligned with how flink currently organize log files.

examples/src/main/java/org/apache/flink/agents/examples/EventLoggingAndListenerExample.java
alnzng: I'm ok with not exposing Event Logger/Listener as programming APIs, since I can't see any reason we have to keep it as programming APIs. And one potential benefit is this makes flink-agent programming APIs as (agent features) focused as possible. 

And yes, let's enable this by default, and expose it to end users once we have the new configuration mechanism.

api/src/main/java/org/apache/flink/agents/api/logger/EventLogRecord.java
alnzng: Sounds good. Moving this to `runtime` module, but I would prefer adding it into a dedicated folder `eventlog` to group all the classes related to `event log` together.

runtime/src/main/java/org/apache/flink/agents/runtime/logger/FileEventLogger.java
alnzng: IIUC, the existing implementation doesn't support multiple agents in a same Flink job: https://github.com/apache/flink-agents/blob/d282e6e334ad785ec42ddead0e9ce492e942ed13/runtime/src/main/java/org/apache/flink/agents/runtime/env/RemoteExecutionEnvironment.java#L108C9-L115C10

```java
public AgentBuilder apply(Agent agent) {
            try {
                this.agentPlan = new AgentPlan(agent);
                return this;
            } catch (Exception e) {
                throw new RuntimeException("Failed to create agent plan from agent", e);
            }
        }
```

but I think you are right on that we should support multiple agents in one Flink job, this seems to be a gap in current implementation.

Let me add task name to differeniate the logs from different tasks. And thanks for the suggestion that using flatten way to organize the log files.

api/src/main/java/org/apache/flink/agents/api/logger/EventLoggerFactory.java
alnzng: Great suggestion. I like this idea "offerring an unitified EventLoggerConfig".

runtime/src/main/java/org/apache/flink/agents/runtime/logger/FileEventLogger.java
xintongsong: > IIUC, the existing implementation doesn't support multiple agents in a same Flink job:

Actually, I think multiple agents are supported. You cannot call `apply()` multiple times on one AgentBuilder, but you can create multiple AgentBuilder-s by calling `fromDataStream/Table()` multiple times.

> with meaningful structured names (e.g., events---.log)

I just noticed that my example file name here doesn't display as expected. I meant `events-<jobid>-<taskid>-<subtaskid>.log`

xintongsong (approved):
Thanks for addressing my comments, @alnzng. This PR looks great to me. Merging.

PR #107, opened 2025-08-12
Mcp def

lihaosky:
Linked issue: #xxx

### DIFF
--- a/python/flink_agents/api/chat_models/chat_model.py
+++ b/python/flink_agents/api/chat_models/chat_model.py
@@ -23,9 +23,7 @@ from typing_extensions import override
 from flink_agents.api.chat_message import ChatMessage
 from flink_agents.api.prompts.prompt import Prompt
 from flink_agents.api.resource import Resource, ResourceType
-
-if TYPE_CHECKING:
-    from flink_agents.api.tools.mcp import MCPServer
+from flink_agents.api.tools.mcp import MCPServer
 
 
 class BaseChatModel(Resource, ABC):
@@ -38,13 +36,13 @@ class BaseChatModel(Resource, ABC):
         or str indicate the Prompt resource in Agent.
     tools : Optional[List[str]] = None
         Tools name can be call by the ChatModel.
-    mcp_servers : Optional[List["MCPServer"]] = None
+    mcp_servers : Optional[List[MCPServer]] = None
         MCP servers whose tools should be available to the ChatModel.
     """
 
     prompt: Optional[Union[Prompt, str]] = None
     tools: Optional[List[str]] = None
-    mcp_servers: Optional[List["MCPServer"]] = None
+    mcp_servers: Optional[List[MCPServer]] = None
 
     @classmethod
     @override
--- a/python/flink_agents/api/chat_models/chat_model.py
+++ b/python/flink_agents/api/chat_models/chat_model.py
@@ -16,13 +16,16 @@
 # limitations under the License.
 #################################################################################
 from abc import ABC, abstractmethod
-from typing import List, Optional, Sequence, Union
+from typing import List, Optional, Sequence, Union, TYPE_CHECKING
 
 from typing_extensions import override
 
 from flink_agents.api.chat_message import ChatMessage
 from flink_agents.api.prompts.prompt import Prompt
 from flink_agents.api.resource import Resource, ResourceType
+
+if TYPE_CHECKING:
+    from flink_agents.api.tools.mcp import MCPServer
 
 
 class BaseChatModel(Resource, ABC):
@@ -35,10 +38,13 @@ class BaseChatModel(Resource, ABC):
         or str indicate the Prompt resource in Agent.
     tools : Optional[List[str]] = None
         Tools name can be call by the ChatModel.
+    mcp_servers : Optional[List["MCPServer"]] = None
+        MCP servers whose tools should be available to the ChatModel.
     """
 
     prompt: Optional[Union[Prompt, str]] = None
     tools: Optional[List[str]] = None
+    mcp_servers: Optional[List["MCPServer"]] = None
 
     @classmethod
     @override
--- a/python/flink_agents/api/tools/mcp.py
+++ b/python/flink_agents/api/tools/mcp.py
@@ -1,0 +1,132 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#################################################################################
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List, Optional, Tuple
+
+from pydantic import BaseModel, Field
+from typing_extensions import override
+
+from flink_agents.api.resource import ResourceType
+from flink_agents.api.chat_models.chat_model import BaseChatModel
+from flink_agents.api.tools.tool import BaseTool, ToolMetadata, ToolType
+
+
+class MCPToolDefinition(BaseModel, ABC):
+    """MCP tool definition returned by list_tools or get_tool_definition."""
+
+    name: str
+    description: str
+    parameters: Dict[str, Any] = Field(default_factory=dict)
+
+    def to_model_tool(self, chat_model: BaseChatModel) -> Dict[str, Any]:
+        """Convert this MCP tool definition to a model-specific tool dict.
+
+        Default implementation is not provided; subclasses must override.
+        """
+        raise NotImplementedError(
+            "to_model_tool must be implemented by MCPToolDefinition subclasses"
+        )
+
+
+class MCPPrompt(BaseModel):
+    """MCP prompt definition returned by list_prompts/get_prompt."""
+
+    name: str
+    description: Optional[str] = None
+    arguments: Dict[str, Any] = Field(default_factory=dict)
+
+
+class MCPServer(BaseTool, ABC):
+    """Resource representing an MCP server and exposing its tools/prompts.
+
+    This is a logical tool container; it is not directly invokable with call().
+    """
+
+    endpoint: str
+    metadata: ToolMetadata = Field(  # default metadata for the server wrapper
+        default_factory=lambda: ToolMetadata(
+            name="mcp_server",
+            description="A container for tools provided by an MCP server.",
+            args_schema=BaseModel,  # not used
+        )
+    )
+
+    @classmethod
+    @override
+    def resource_type(cls) -> ResourceType:  # type: ignore[override]
+        return ResourceType.MCP_SERVER
+
+    @classmethod
+    @override
+    def tool_type(cls) -> ToolType:
+        return ToolType.MCP
+
+    def call_tool(self, name: str,*args: Any, **kwargs: Any) -> Any:  # noqa: D401
+        """MCPServer isn't directly invokable; raise to signal misuse."""
+        raise NotImplementedError("MCPServer does not support direct call().")
+
+    # Listing and retrieval APIs
+    def list_tools(self, cursor: Optional[str] = None) -> Tuple[List[MCPToolDefinition], Optional[str]]:
+        """List available tool definitions from the MCP server and an optional cursor.
+
+        Parameters
+        ----------
+        cursor : Optional[str]
+            Optional pagination cursor/state.
+
+        Returns
+        -------
+        Tuple[List[MCPToolDefinition], Optional[str]]
+            A tuple of (tools, next_cursor). next_cursor may be None when there are no more pages.
+        """
+        raise NotImplementedError
+
+    def get_tool_definition(self, name: str) -> MCPToolDefinition:
+        """Get a single tool definition by name."""
+        raise NotImplementedError
+
+    def list_prompts(self, cursor: Optional[str] = None) -> Tuple[List[MCPPrompt], Optional[str]]:
+        """List available prompts from the MCP server and an optional cursor.
+
+        Parameters
+        ----------
+        cursor : Optional[str]
+            Optional pagination cursor/state.
+
+        Returns
+        -------
+        Tuple[List[MCPPrompt], Optional[str]]
+            A tuple of (prompts, next_cursor). next_cursor may be None when there are no more pages.
+        """
+        raise NotImplementedError
+
+    def get_prompt(self, name: str) -> MCPPrompt:
+        """Get a single prompt definition by name."""
+        raise NotImplementedError
+
+    # Conversion helpers
+    def to_model_tool(self, name: str, chat_model: BaseChatModel) -> Dict[str, Any]:
+        """Convert a named MCP tool to a model-specific tool dict.
+
+        Default behavior is not implemented; subclasses should override.
+        """
+        raise NotImplementedError(
+            "MCPServer.to_model_tool must be implemented by subclasses"
+        )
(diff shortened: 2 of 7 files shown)


python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,132 @@
...
+from flink_agents.api.chat_models.chat_model import BaseChatModel
+from flink_agents.api.tools.tool import BaseTool, ToolMetadata, ToolType
+
+
+class MCPToolDefinition(BaseModel, ABC):
+    """MCP tool definition returned by list_tools or get_tool_definition."""
+
+    name: str
+    description: str
+    parameters: Dict[str, Any] = Field(default_factory=dict)
```
lihaosky: Maybe we should use Function call or Function tool directly here which supports callable

lihaosky:
I put mcp server related stuff directly in this PR so we can discuss here. @xintongsong

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,132 @@
...
+            A tuple of (prompts, next_cursor). next_cursor may be None when there are no more pages.
+        """
+        raise NotImplementedError
+
+    def get_prompt(self, name: str) -> MCPPrompt:
+        """Get a single prompt definition by name."""
+        raise NotImplementedError
+
+    # Conversion helpers
+    def to_model_tool(self, name: str, chat_model: BaseChatModel) -> Dict[str, Any]:
```
lihaosky: Passing mcp directly as tool to model instead of passing its individual tools to model

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,132 @@
...
+
+from pydantic import BaseModel, Field
+from typing_extensions import override
+
+from flink_agents.api.resource import ResourceType
+from flink_agents.api.chat_models.chat_model import BaseChatModel
+from flink_agents.api.tools.tool import BaseTool, ToolMetadata, ToolType
+
+
+class MCPToolDefinition(BaseModel, ABC):
```
xintongsong: I think this should extends either `BaseTool` or `ToolMetadata`?

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,132 @@
...
+    """MCP tool definition returned by list_tools or get_tool_definition."""
+
+    name: str
+    description: str
+    parameters: Dict[str, Any] = Field(default_factory=dict)
+
+    def to_model_tool(self, chat_model: BaseChatModel) -> Dict[str, Any]:
+        """Convert this MCP tool definition to a model-specific tool dict.
```
xintongsong: Does this mean a tool implementation needs to understand how to convert itself into a model-specific form for all supported chat models? I though it should be responsibility of the chat model implementations to convert the unified representation of mcp tools to what ever forms they need.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Default implementation is not provided; subclasses must override.
+        """
+        raise NotImplementedError(
+            "to_model_tool must be implemented by MCPToolDefinition subclasses"
+        )
+
+
+class MCPPrompt(BaseModel):
```
xintongsong: Should this extends `Prompt`?

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,132 @@
...
+    """MCP prompt definition returned by list_prompts/get_prompt."""
+
+    name: str
+    description: Optional[str] = None
+    arguments: Dict[str, Any] = Field(default_factory=dict)
+
+
+class MCPServer(BaseTool, ABC):
```
xintongsong: I think this should extend `Resource`, rather than `BaseTool`.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,132 @@
...
+    @override
+    def resource_type(cls) -> ResourceType:  # type: ignore[override]
+        return ResourceType.MCP_SERVER
+
+    @classmethod
+    @override
+    def tool_type(cls) -> ToolType:
+        return ToolType.MCP
+
+    def call_tool(self, name: str,*args: Any, **kwargs: Any) -> Any:  # noqa: D401
```
xintongsong: Would it be nicer to call the tools directly on the `MCPTool` returned from `MCPServer.list_tools()` or `MCPServer.get_tool()`? In this way, the caller side won't need to understand whether the tool being called is form a mcp server or a local java/python function.

xintongsong: Like:
```
tool = mcp_server.get_tool(name)
tool.call()
```

python/flink_agents/api/tools/mcp.py
lihaosky: I was thinking about it as well. It should be better to extend tool

python/flink_agents/api/tools/mcp.py
lihaosky: Yes. ChatModel understands how to convert it. That's why the chatModel is passed here. Maybe ChatModel should have a convert to tool method. And this `to_model_tool` should be in `ToolMetaData`. https://github.com/apache/flink-agents/blob/main/python/flink_agents/api/tools/tool.py#L101 only convert to open ai tool which isn't generic

python/flink_agents/api/tools/mcp.py
lihaosky: Yeah. If we make `MCPToolDefinition` extends tool, we can do this

python/flink_agents/api/tools/mcp.py
wenjin272: Yes, I think this is the responsibility of chat model implementations to covert the Tool of flink-agents to the form they need.
Because almost all chat models support open ai tool form, so I add the to_openai_tool in ToolMetaData, but we also need add a custom function which support convert it to model-specific form.

python/flink_agents/api/tools/mcp.py
wenjin272: I think so, because MCPServer can't be call directly, and more likely a resource provide tools and prompts.

python/flink_agents/api/tools/mcp.py
wenjin272: Yes, I think maybe we can directly use FunctionTool here.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,132 @@
...
+        """MCPServer isn't directly invokable; raise to signal misuse."""
+        raise NotImplementedError("MCPServer does not support direct call().")
+
+    # Listing and retrieval APIs
+    def list_tools(self, cursor: Optional[str] = None) -> Tuple[List[MCPToolDefinition], Optional[str]]:
```
wenjin272: The methods in MCPServer may generate network IO, should we mark these as async function?

python/flink_agents/api/tools/mcp.py
xintongsong: I think it would be more straightforward to pass the tool into the model, and let the model read the tool meta in a common form and convert into the form it needs. Passing the model into the tool seems unnecessary, since we still need to call the model for the conversion inside the tool.

In other words, I think we should not have a `to_model_tool()` function on `ToolMetadata`. Instead, we should have  a `convert_tool()` on `ChatModel`. Or do we need this as an interface at all? This can be completely internal to the `ChatModel`.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,131 @@
...
+    @classmethod
+    @override
+    def tool_type(cls) -> ToolType:
+        return ToolType.MCP
+
+    @classmethod
+    @override
+    def call(self, *args: Any, **kwargs: Any) -> Any:
+        """Call the MCP tool with the given arguments."""
+        raise NotImplementedError
```
xintongsong: This is not a class method.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class MCPPrompt(Prompt):
+    """MCP prompt definition that extends the base Prompt class.
+
+    This represents a prompt template from an MCP server.
+    """
+
+    name: str
```
xintongsong: `name` is already define in the super class `Resource`.

xintongsong (commented):
API-wise, this looks good to me. I think we can move forward with the implementation.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,131 @@
...
+    description: Optional[str] = None
+    arguments: Dict[str, Any] = Field(default_factory=dict)
```
istavrak, line 69: Is this about the schema of the template? It could be beneficial to elaborate more in the docstring of the class.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+
+class MCPServer(Resource, ABC):
+    """Resource representing an MCP server and exposing its tools/prompts.
+
+    This is a logical container for MCP tools and prompts; it is not directly invokable.
+    """
+
+    endpoint: str
```
istavrak, line 93: Do we need any runtime parameters about the MCP Server?

python/flink_agents/api/tools/mcp.py
yanand0909: I don't know what do you mean by template but yes this is for all the arguments and their type, description that the function needs. I can elaborate in the docstring

python/flink_agents/api/tools/mcp.py
yanand0909: We might need auth headers and connection config but we can leave that to user?

python/flink_agents/api/tools/mcp.py
istavrak: Thanks! The template here refers to the prompt template.

yanand0909: Oh yes this one is for prompt template

python/flink_agents/integrations/chat_models/ollama_chat_model.py
```diff
@@ -25,6 +25,7 @@
 from flink_agents.api.chat_message import ChatMessage, MessageRole
 from flink_agents.api.chat_models.chat_model import BaseChatModel
 from flink_agents.api.resource import ResourceType
+from flink_agents.api.tools.mcp import MCPServer
```
wenjin272: Unused import statement

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,281 @@
...
+
+        return tools
+
+    def get_tool(self, name: str) -> MCPToolDefinition:
+        """Get a single callable tool by name."""
+        return asyncio.run(self._get_tool_async(name))
+
+    async def _get_tool_async(self, name: str) -> MCPToolDefinition:
+        """Async implementation of get_tool."""
+        tools, _ = await self._list_tools_async()
```
wenjin272: The return value type of `_list_tools_async` is `List[MCPToolDefinition]`, why here use `tools, _` to receive return value?

wenjin272 (commented):
The mcp implementation alone looks good to me. But I think there may lack some implementation about how does chat model use mcp server internally.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,281 @@
...
+from typing_extensions import override
+
+from flink_agents.api.resource import Resource, ResourceType
+from flink_agents.api.tools.tool import BaseTool, ToolMetadata, ToolType
+from flink_agents.api.tools.utils import extract_mcp_content_item
+from flink_agents.api.prompts.prompt import Prompt
+
+
+
+class MCPToolDefinition(BaseTool):
```
xintongsong: Maybe just call it `MCPTool`, to align with `FunctionTool`.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,281 @@
...
+    @override
+    def call(self, *args: Any, **kwargs: Any) -> Any:
+        """Call the MCP tool with the given arguments."""
+        if self.mcp_server is None:
+            raise ValueError("MCP tool call requires a reference to the MCP server")
+
+        return asyncio.run(self.mcp_server._call_tool_async(self.metadata.name, *args, **kwargs))
+
+
+class MCPPrompt(Prompt):
```
xintongsong, line 62: How does user use this prompt? I'd expect to use a `MCPPrompt` just like using any other `Prompt`. That means calling `from_text()` or `format_messages()` on it. That may not work because `MCPPrompt` neither override these methods, nor set the properties that these methods use in its super class.

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,281 @@
...
+
+    async def _get_tool_async(self, name: str) -> MCPToolDefinition:
+        """Async implementation of get_tool."""
+        tools, _ = await self._list_tools_async()
+        for tool in tools:
+            if tool.metadata.name == name:
+                return tool
+        raise ValueError(f"Tool '{name}' not found on MCP server at {self.endpoint}")
+
+    def get_tool_definition(self, name: str) -> ToolMetadata:
```
xintongsong:
```suggestion
    def get_tool_metadata(self, name: str) -> ToolMetadata:
```

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,281 @@
...
+            if prompt_data.arguments:
+                for arg in prompt_data.arguments:
+                    arguments_dict[arg.name] = {
+                        "description": arg.description,
+                        "required": getattr(arg, 'required', False)
+                    }
+
+            prompt = MCPPrompt(
+                name=prompt_data.name,
+                template="",  # MCP prompts don't have templates in metadata - content comes from get_prompt call
```
xintongsong, line 237: I think you mean `get_content`, not `get_prompt`?

Still, users should not be aware of any differences in using a `Prompt` and a `MCPPrompt`. We might consider overriding `format_string()` and `from_text()` rather than introducing a new `get_content`.

And if `template` doesn't make sense for `MCPPrompt`, that probably indicates that we should refactor the hierarchy of prompts. Maybe introducing a new `BasePrompt` abstraction, and change the current `Prompt` into something like `LocalPrompt`?

python/flink_agents/api/tools/mcp.py
```diff
@@ -0,0 +1,281 @@
...
+        return asyncio.run(self.mcp_server._get_prompt_content_async(self.name, arguments))
+
+
+class MCPServer(Resource, ABC):
+    """Resource representing an MCP server and exposing its tools/prompts.
+
+    This is a logical container for MCP tools and prompts; it is not directly invokable.
+    """
+
+    model_config = {"arbitrary_types_allowed": True}
```
xintongsong, line 90: What is this and how is it used?

python/flink_agents/api/chat_models/chat_model.py
```diff
@@ -60,3 +65,20 @@ def chat(self, messages: Sequence[ChatMessage]) -> ChatMessage:
         ChatMessage
             Response from the ChatModel.
         """
+
+    @abstractmethod
+    def convert_tool_to_model_format(self, tool_metadata: ToolMetadata) -> Dict[str, Any]:
```
xintongsong: Why do we need this method? It's never implemented or called.

python/flink_agents/api/tools/mcp.py
yanand0909: Yes I think we should have introduce new base prompt that this MCPPrompt and LocalPrompt should extend

python/flink_agents/api/tools/mcp.py
yanand0909: Yes this is what I wanted to discuss as MCPPrompt is conceptually different from Prompt we have we might need to make some changes to our prompt design

python/flink_agents/api/tools/mcp.py
yanand0909: Initially this function was returning cursor as well but we don't need it as StreamableHttp client does not support it. I will remove it thanks for pointing out

python/flink_agents/api/chat_models/chat_model.py
yanand0909: User might need this function to convert McpTools to model's expected format? I am still not very sure of how these tools will be used

xintongsong: We just pass the available tools into `chat()`, and the model implementation will do the conversion. E.g., `OllamaChatModel` does the conversion in its `chat()` method, and `AnthropicChatModel` (PR #139) does it with a method `to_anthropic_tool()` which is also called in `chat()`.

At the moment, I don't see any need for calling `convert_tool_to_model_format` from outside of the chat model. So let's leave it to the model implementation for now, rather than introducing a common abstract method.

Issue #109, opened 2025-08-12
[Feature] [api][runtime]Introduce Memory Reference


KeGu-069:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce Memory Reference 

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #110, opened 2025-08-12
[API][Runtime]Introduce Memory Reference

KeGu-069:
Linked issue: #109 

### Purpose of change

This pr introduces MemoryRef as an optimization of the short-term memory.

### DIFF
--- a/python/flink_agents/api/memory_object.py
+++ b/python/flink_agents/api/memory_object.py
@@ -32,7 +32,7 @@ class MemoryObject(BaseModel, ABC):
     """
 
     @abstractmethod
-    def get(self, path: Union[str,"MemoryRef"] ) -> Any:
+    def get(self, path_or_ref: Union[str,"MemoryRef"] ) -> Any:
         """Get the value of a (direct or indirect) field or a MemoryRef in the object.
 
         Parameters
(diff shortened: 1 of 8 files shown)


python/flink_agents/api/memory_object.py
```diff
@@ -30,25 +32,27 @@ class MemoryObject(BaseModel, ABC):
     """
 
     @abstractmethod
-    def get(self, path: str) -> Any:
-        """Get the value of a (direct or indirect) field in the object.
+    def get(self, path: Union[str,"MemoryRef"] ) -> Any:
```
wenjin272: parameter name is not consistent with doc string.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/MemoryObjectImpl.java
```diff
@@ -197,23 +207,26 @@ private void fillParents(String[] parts) throws Exception {
     }
 
     /** Represents an item (nested object or primitive value) stored in the short-term memory. */
-    public static final class MemoryItem {
+    public static final class MemoryItem implements Serializable {
         private final ItemType type;
         private final Object value;
         private final Set<String> subKeys;
+        private final String typeName;
```
wenjin272: Why we need `typeName`, I can't find the usage.

wenjin272:
LGTM. Please take a look at your convenience @xintongsong

Issue #118, opened 2025-08-19
[Feature][api][runtime] Introduce configuration mechanism


GreatEugenius:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce configuration mechanism

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #122, opened 2025-08-20
Feature: [api][runtime] Introduce configuration mechanism

GreatEugenius:
Linked issue: #118 

### Purpose of change
Introduce a configuration mechanism.

### Tests
Unit test

### API
Yes, add APIs related to configurations.


python/flink_agents/runtime/local_configuration.py
```diff
@@ -0,0 +1,225 @@
...
+        else:
+            items[new_key] = v
+    return items
+
+class LocalConfiguration(WritableConfiguration, ReadableConfiguration):
+    """Base class for config objects in the system.
+    Provides a flat dict interface to access nested config values.
+    """
+
+    _conf_data: ClassVar[dict[str, Any]] = {}
```
Sxnan: Why is this a class variable instead of a instance variable?

python/flink_agents/runtime/local_configuration.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def get_int(self, key: str, default: int=0) -> int:
```
Sxnan: The default value of an int configuration is not always 0. Similar to other get methods.

It is more intuitive to throw an exception if `get_xxx` method is called when the key is missing and the default value is not given.

python/flink_agents/api/configuration.py
```diff
@@ -0,0 +1,148 @@
...
+    """A configuration option defines a configuration key with its type and default
+    value.
+
+    Args:
+        key: The configuration key name
+        type: The expected type of the configuration value (int, float, str, or bool)
+        default: The default value for this configuration option
+    """
+
+    def __init__(self, key: str, type: Type[Any], default: Any) -> None:
```
Sxnan: We should support a config option without a default value, which means that the ConfigOption is not optional and we should throw an exception if it is missing.

python/flink_agents/examples/chat_ollama_exmaple.py
```diff
@@ -58,7 +57,6 @@ def math_chat_model() -> Tuple[Type[BaseChatModelSetup], Dict[str, Any]]:
         """ChatModel which focus on math, and reuse ChatModelServer."""
         return OllamaChatModelSetup, {
             "name": "math_chat_model",
-            "connection": "ollama_server",
```
Sxnan: Why do we remove this?

python/flink_agents/examples/chat_ollama_exmaple.py
```diff
@@ -39,7 +39,6 @@
 
 model = os.environ.get("OLLAMA_CHAT_MODEL", "qwen3:8b")
 
-
```
Sxnan: nit: unnecessary change

api/src/main/java/org/apache/flink/agents/api/FlinkAgentsCoreOptions.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api;
+
+import org.apache.flink.configuration.ConfigOption;
+import org.apache.flink.configuration.ConfigOptions;
+
+public final class FlinkAgentsCoreOptions {
+    public static final ConfigOption<String> EXAMPLE_OPTION =
```
Sxnan: Can we think of a better config option to showcase? Adding a temporary field to a public API class may not be a good idea.

python/requirements/test_requirements.txt
```diff
@@ -17,3 +17,4 @@ setuptools>=75.3
 pytest==8.4.0
 pydantic==2.11.4
 apache-flink==1.20.1
+pyyaml==6.0.2
```
Sxnan: Why do we maintain two sets of requirements? One in the `pyproject.toml` and one in the requirements folder? And they are inconsistent already.

python/flink_agents/plan/agent_plan.py
```diff
@@ -229,35 +229,51 @@ def _get_actions(agent: Agent) -> List[Action]:
...
             if isinstance(value, staticmethod):
                 value = value.__func__
 
             if callable(value):
+                final_kwargs = {}
                 clazz, kwargs = value()
+                resource_kwargs = _get_module_conf(config_data, name)
+                clazz_kwargs = _get_module_conf(config_data, clazz.__name__)
+
+                final_kwargs.update(clazz_kwargs)
```
Sxnan: Instead of injecting the configuration in the kwargs, it is better to pass the configuration to the resource provider. 
Then, the resource provider will pass the configuration to the Resource when creating the Resource. 
The Resource implementation will decide how to use the configuration properly.

Sxnan (changes requested):
Thanks for the PR! I left some comments below.

api/src/main/java/org/apache/flink/agents/api/FlinkAgentsCoreOptions.java
alnzng: +1 on this. For educating purpose, I would suggest we add javadoc to explain the usage of this class, instead of using this placeholder field.

A clarification quesiton:
- is this class intended for defining all FlinkAgents options? or this is only for `core` options and we expected there could be some `non-core` options

python/flink_agents/api/core_options.py
```diff
@@ -0,0 +1,52 @@
...
+# limitations under the License.
+#################################################################################
+from typing import Any
+
+from pyflink.java_gateway import get_gateway
+
+from flink_agents.api.configuration import ConfigOption
+
+
+def covert_j_option_to_python_option(j_option: Any) -> ConfigOption:
```
alnzng, line 25: Nit: I would suggest we keep the naming consistent. If we prefer short names(e.g. j -> java), let's do it for python as well(e.g. p -> python). Or versa vice.

python/requirements/test_requirements.txt
GreatEugenius: The requirements folder is a legacy artifact from earlier development stages and is no longer actively maintained. I have created an issue to delete it in the future: https://github.com/apache/flink-agents/issues/125

python/flink_agents/api/configuration.py
```diff
@@ -0,0 +1,148 @@
...
+            value: The boolean value to set for the key
+        """
+
+    @abstractmethod
+    def set(self, option: ConfigOption, value: Any) -> None:
+        """Set the configuration value using the ConfigOption.
+
+        Args:
+            option: The config option to set
+            value: The boolean value to set for the key
```
alnzng: remove `boolean `

python/flink_agents/api/configuration.py
```diff
@@ -0,0 +1,148 @@
...
+    def set_bool(self, key: str, value: bool) -> None:  # noqa: FBT001
+        """Set the boolean configuration value using the key.
+
+        Args:
+            key: The configuration key to set
+            value: The boolean value to set for the key
+        """
+
+    @abstractmethod
+    def set(self, option: ConfigOption, value: Any) -> None:
```
alnzng, line 93: Should we support `get(option: ConfigOption) -> Any` method in ReadableConfiguration?

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -133,10 +138,26 @@ class RemoteExecutionEnvironment(AgentsExecutionEnvironment):
     """Implementation of AgentsExecutionEnvironment for execution with DataStream."""
 
     __env: StreamExecutionEnvironment
+    __config: LocalConfiguration
 
     def __init__(self, env: StreamExecutionEnvironment) -> None:
         """Init method of RemoteExecutionEnvironment."""
         self.__env = env
+        self.__config = LocalConfiguration()
```
alnzng: Seems both RemoteExecutionEnvironment and LocalExecutionEnvironment loads `LocalConfiguration`, is it expected?  just from `LocalConfiguration` literally, it should be used for  LocalExecutionEnvironment only.

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -133,10 +138,26 @@ class RemoteExecutionEnvironment(AgentsExecutionEnvironment):
...
+        flink_conf_dir = os.environ.get("FLINK_CONF_DIR")
+        if flink_conf_dir is not None:
+            config_dir = Path(flink_conf_dir) / "config.yaml"
```
alnzng, line 149: Does this `config.yaml` mean that flink-agents requires Flink 2.0+?

GreatEugenius: No, currently flink-agents depends on Flink 1.20 and uses config.yaml as the configuration file.

python/flink_agents/api/core_options.py
GreatEugenius: This is a good suggestion. We are indeed using j_xx to represent Java variables in many places.

python/flink_agents/runtime/remote_execution_environment.py
alnzng: Does Flink 1.20 support the old config file `flink-config.yaml`? If yes, do we want to support it in flink-agents for consistency?

python/flink_agents/api/execution_environment.py
```diff
@@ -108,6 +109,16 @@ def get_execution_environment(
                 "flink_agents.runtime.remote_execution_environment"
             ).create_instance(env=env, **kwargs)
 
+    @abstractmethod
+    def get_config(self, path: Optional[str] = None) -> WritableConfiguration:
```
Sxnan: Should we return a Configuration that supports both read and write?

python/flink_agents/plan/configuration.py
```diff
@@ -0,0 +1,224 @@
...
+    def __init__(self, conf_data: Optional[Dict[str, Any]] = None) -> None:
+        """Initialize with optional configuration data."""
+        if conf_data is None:
+            super().__init__(conf_data = {})
+        else:
+            super().__init__(conf_data = conf_data)
+
+    @override
+    def get_int(self, key: str, default: Optional[int]=None) -> int:
+        value = self.conf_data.get(key)
```
Sxnan, line 84: I think we can extract a helper method for all the get_xxx to avoid duplicates code

python/flink_agents/plan/configuration.py
```diff
@@ -0,0 +1,224 @@
...
+        if value is None:
+            if default is None:
```
Sxnan: After rethinking the behavior of throwing an exception in case of None default value, I think such behavior is too strict that we cannot have a default value of None. 

I think we can just return the deafult value anyway, and let the caller decide whether to throw an exception in case of a None value. 

WDYT?

python/flink_agents/plan/configuration.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def get_conf_data(self) -> dict:
+        """Get the configuration data dictionary.
+
+        Returns:
+            dict: A dictionary containing all configuration items
+        """
+        return self.conf_data
+
+    def revert_to_flink_config(self) -> Configuration:
```
Sxnan: The naming is confusing. Maybe `convert_to_flink_config` or just `to_flink_config`

python/flink_agents/plan/configuration.py
```diff
@@ -0,0 +1,224 @@
...
+            elif isinstance(value, float):
+                flink_config.set_float(key, value)
+            elif isinstance(value, str):
+                flink_config.set_string(key, value)
+            else:
+                flink_config.set_string(key, str(value))
+        return flink_config
+
+    def get_config_data_by_prefix(self, prefix: str) -> dict:
+        """Extract configuration items for a specific module from the configuration
```
Sxnan: The comment is inaccurate. There should not be a notion of 'module'

python/flink_agents/plan/resource_provider.py
```diff
@@ -88,11 +92,21 @@ class PythonResourceProvider(ResourceProvider):
...
-    def provide(self, get_resource: Callable) -> Resource:
+    def provide(self, get_resource: Callable, config: AgentConfiguration) -> Resource:
         """Create resource in runtime."""
         module = importlib.import_module(self.module)
         cls = getattr(module, self.clazz)
-        resource = cls(**self.kwargs, get_resource=get_resource)
+
+        final_kwargs = {}
+
+        clazz_kwargs = config.get_config_data_by_prefix(self.clazz)
```
Sxnan: Maybe 'resource_class_config' and `resource_config` make more sense

python/flink_agents/plan/configuration.py
```diff
@@ -0,0 +1,224 @@
...
+        """Load configuration from a YAML file and update current configuration data.
+
+        Args:
+            config_path (str, optional): Path to the configuration file.
+        """
+        if config_path:
+            path = Path(config_path)
+            with path.open() as f:
+                raw_config = yaml.safe_load(f)
+                self.conf_data.update(flatten_dict(raw_config.get('Agent', {})))
```
Sxnan: Maybe use the lower case `agent`

python/flink_agents/runtime/local_execution_environment.py
```diff
@@ -86,6 +88,13 @@ class LocalExecutionEnvironment(AgentsExecutionEnvironment):
     __output: List[Any] = None
     __runner: LocalRunner = None
     __executed: bool = False
+    __config: AgentConfiguration = AgentConfiguration()
+
+    def get_config(self, path: Optional[str] = None) -> AgentConfiguration:
+        """Get configuration of execution environment."""
+        if path is not None:
+            self.__config.load_from_file(path)
```
Sxnan, line 96: The load_from_file only looks for configurations under key `agent`. For local configuration, do we need to put all the configuration under `agent`?

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -202,6 +223,8 @@ def from_list(self, input: List[Dict[str, Any]]) -> "AgentsExecutionEnvironment"
 
     def execute(self) -> None:
         """Execute agent."""
+        flink_config = self.__config.revert_to_flink_config()
+        self.__env.configure(flink_config)
```
Sxnan: Why do we need to configure Flink configuration?

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -109,4 +110,10 @@ public Resource getResource(String name, ResourceType type) throws Exception {
         }
         return agentPlan.getResource(name, type);
     }
+
+    @Override
+    public ReadableConfiguration getConfig() {
+        System.out.println(agentPlan.getConfigData());
```
Sxnan: Should this be removed?

api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
```diff
@@ -86,6 +87,13 @@ public static AgentsExecutionEnvironment getExecutionEnvironment() {
         return getExecutionEnvironment(null);
     }
 
+    /**
+     * Returns a writable configuration object for setting configuration values.
+     *
+     * @return the WritableConfiguration instance used to modify configuration settings
+     */
+    public abstract WritableConfiguration getConfig();
```
Sxnan: Maybe return a readable and writable?

plan/src/main/java/org/apache/flink/agents/plan/AgentConfiguration.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public <T> void set(ConfigOption<T> option, T value) {
+        if (value == null && option.getDefaultValue() != null) {
+            return;
+        }
+        confData.put(option.getKey(), value);
+    }
+
+    @SuppressWarnings("unchecked")
```
Sxnan: Why do we need this?

plan/src/main/java/org/apache/flink/agents/plan/AgentConfiguration.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private static Map<String, Object> flatten(
+            Map<String, Object> config, String keyPrefix, String keySeparator) {
+        final Map<String, Object> flattenedMap = new HashMap<>();
+
+        config.forEach(
+                (key, value) -> {
+                    String flattenedKey = keyPrefix + key;
+                    if (value instanceof Map) {
+                        Map<String, Object> e = (Map<String, Object>) value;
```
Sxnan, line 167: Nit: We have an unchecked cast warning here

plan/src/main/java/org/apache/flink/agents/plan/serializer/AgentPlanJsonDeserializer.java
```diff
@@ -119,6 +120,23 @@ public AgentPlan deserialize(JsonParser parser, DeserializationContext ctx)
             }
         }
 
-        return new AgentPlan(actions, actionsByEvent, resourceProviders);
+        // Deserialize config data
```
Sxnan, line 124: Do we test for the serde of the configuration?

runtime/src/main/java/org/apache/flink/agents/runtime/env/RemoteExecutionEnvironment.java
```diff
@@ -55,27 +74,39 @@ public AgentBuilder fromList(List<Object> input) {
...
     }
 
     @Override
     public void execute() throws Exception {
         env.execute();
     }
 
+    private static Map<String, Object> loadAgentConfiguration(String configDir) {
+        try {
+            return (Map<String, Object>)
```
Sxnan: Nit: unchecked cast warning

runtime/src/main/java/org/apache/flink/agents/runtime/env/RemoteExecutionEnvironment.java
```diff
@@ -43,8 +48,22 @@ public class RemoteExecutionEnvironment extends AgentsExecutionEnvironment {
 
     private final StreamExecutionEnvironment env;
 
+    private final AgentConfiguration config;
+
     public RemoteExecutionEnvironment(StreamExecutionEnvironment env) {
         this.env = env;
+        final String configDir = System.getenv(ConfigConstants.ENV_FLINK_CONF_DIR);
```
Sxnan, line 57: Add tests for configuration loading

Sxnan (changes requested):
Thanks for the update! It is in pretty shape. I left some comments below.

python/flink_agents/plan/configuration.py
GreatEugenius: You are right. I also discovered while writing tests that if the default value is set to None, an error occurs. This can be quite harsh for the user.

python/flink_agents/runtime/local_execution_environment.py
GreatEugenius: To ensure consistency in configuration and develop user habits, I believe it is better to place them under the agent in both local and remote configurations.

api/src/main/java/org/apache/flink/agents/api/configuration/FullConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.api.configuration;
+
+/**
+ * A configuration interface that provides both read and write access. Combines the capabilities of
+ * {@link ReadableConfiguration} and {@link WritableConfiguration}.
+ */
+public interface FullConfiguration extends ReadableConfiguration, WritableConfiguration {}
```
Sxnan: We can just name it `Configuration`

api/src/main/java/org/apache/flink/agents/api/configuration/ReadableConfiguration.java
```diff
@@ -26,7 +26,16 @@ public interface ReadableConfiguration {
      * @param defaultValue The default value to return if key is not found
      * @return The integer value associated with the key or the default value
      */
-    int getInt(String key, Integer defaultValue);
+    int getInt(String key, int defaultValue);
```
Sxnan: Should we use  `Integer` instead of `int` so that we can have a null default value. Same for the other getter. For setter, it is fine to use the `int`, etc.

python/flink_agents/plan/resource_provider.py
```diff
@@ -99,11 +99,11 @@ def provide(self, get_resource: Callable, config: AgentConfiguration) -> Resourc
...
         final_kwargs = {}
 
-        clazz_kwargs = config.get_config_data_by_prefix(self.clazz)
+        resource_class_config = config.get_config_data_by_prefix(self.clazz)
         resource_kwargs = config.get_config_data_by_prefix(self.name)
 
         final_kwargs.update(self.kwargs)
-        final_kwargs.update(clazz_kwargs)
+        final_kwargs.update(resource_class_config)
         final_kwargs.update(resource_kwargs)
```
Sxnan: resource_kwargs -> resource_config

Sxnan (commented):
Thanks for the update! Just left 3 minor comments.

Sxnan (approved):
Thanks for the update! LGTM.

@xintongsong Please, take a final look at your convenience.

python/pyproject.toml
```diff
@@ -50,6 +50,7 @@ dependencies = [
     #TODO: Seperate integration dependencies from project
     "ollama==0.4.8",
     "dashscope~=1.24.2",
+    "pyyaml==6.0.2",
```
Sxnan: This may be put above the TODO comments

Sxnan (commented):
Thanks for the update! I left one nit comment

xintongsong (approved):
LGTM

Issue #57, opened 2025-07-10
[Feature][api][python] Support ReAct Agent in Python


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support ReAct Agent in Python

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

kitalkuyo-gita:
# Technical Design for Supporting ReAct Agent Feature

## 1. Current Architecture Analysis

Apache Flink Agents currently uses an event-driven architecture , with actions defined through the `@action` decorator that respond to specific event types , and uses WorkflowPlan to compile user-defined workflows into executable plans .

## 2. ReAct Agent Core Design

### 2.1 New Event Type Design

Building on the existing Event base class architecture, we need to add the following ReAct-specific event types:

**Core Event Types:**
- `ReasoningEvent`: Contains reasoning requests and context information
- `ActionEvent`: Contains specific action instructions and parameters  
- `ObservationEvent`: Contains action execution results and feedback
- `ToolCallEvent`: Contains tool invocation requests
- `ToolResultEvent`: Contains tool execution results
- `PlanningEvent`: Contains task decomposition and planning information

### 2.2 ReAct Agent State Management

Extend the existing RunnerContext mechanism to add ReAct-specific state management:

**State Components:**
- `ReActMemory`: Stores reasoning history, action records, and observation results
- `TaskState`: Tracks current task state and objectives
- `ToolRegistry`: Manages available tools and their invocation states
- `ReasoningChain`: Maintains reasoning chains and decision processes

### 2.3 ReAct Workflow Core Classes

**Main Class Design:**
- `ReActAgent`: Inherits from Workflow, implements ReAct loop logic
- `ReActPlanner`: Responsible for task decomposition and high-level planning
- `ToolExecutor`: Unified tool execution interface
- `LLMInterface`: Abstract interface for LLM calls

## 3. Detailed Implementation Design

### 3.1 ReAct Loop Control Mechanism

Leverage the existing event processing loop to implement the ReAct reasoning-action-observation cycle:

**Control Flow:**
1. Receive InputEvent to trigger initial reasoning
2. ReasoningAction generates ActionEvent
3. ActionEvent triggers specific tool calls
4. Tool execution results produce ObservationEvent
5. ObservationEvent triggers next round of reasoning or termination

### 3.2 Tool Integration Framework

**Tool Abstraction Layer:**
[...]

PR #126, opened 2025-08-21
WIP: [api][python] Introduce built-in ReAct Agent.

wenjin272:
Linked issue: #57 

### Purpose of change

Introduce built-in ReAct Agent.

### Tests

No

### API

Yes, provide ReActAgent

### Documentation

No


wenjin272:
One point that has been on my mind is whether we should implicitly inject built-in actions into all types of Agents, including user-defined ones. My concern is that this might reduce the flexibility for users. Alternatively, we could provide built-in actions only for the ReAct Agent. Based on the current implementation of the ReAct Agent, it appears that most of its core logic is already implemented through built-in actions

python/flink_agents/api/agents/react_agent.py
```diff
@@ -0,0 +1,75 @@
...
+    implementation we use the tool/function call ability of current llm, and get
+    the tool calls from response directly.
+
+    User can extend this class and overwrite the start_action and stop_action
+    to define their own logic to handle input and output.
+    """
+
+    @action(InputEvent)
+    @staticmethod
+    def start_action(event: InputEvent, ctx: RunnerContext) -> None:
```
wenjin272, line 209: Should we provide start_action and stop_action in ReActAgent?

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -25,59 +25,65 @@
...
+            #  After memory supports remove, we can use "TOOL_CALL_CONTEXT/request_id"
+            #  to store and remove the specific tool context directly.
+            # save tool call context
+            tool_call_context = short_term_memory.get(TOOL_CALL_CONTEXT)
+            if not tool_call_context:
+                tool_call_context = {}
+            if request_id not in tool_call_context:
+                tool_call_context[request_id] = event
+            # append response to request event messages
+            tool_call_context[request_id].messages.append(response)
```
xintongsong: Why using the event as the context? It's against intuition that we are modifying a received event.

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -25,59 +25,65 @@
...
+            ctx.send_event(
+                ToolRequestEvent(
+                    id=event.id,
+                    tool_calls=response.tool_calls,
                 )
-
+            )
         # send response
         else:
+            # clear tool call context related to specific request id
```
xintongsong: Is it possible that the chat response contains both text and tool calls?

xintongsong: We should make it explicit that a response without tool call means end of the chat.

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -25,59 +25,65 @@
...
+        tool_call_context = short_term_memory.get(TOOL_CALL_CONTEXT)
+        for response in responses.values():
+            tool_call_context[request_id].messages.append(
+                ChatMessage(role=MessageRole.TOOL, content=str(response))
+            )
+        # update tool call context
+        short_term_memory.set(TOOL_CALL_CONTEXT, tool_call_context)
+        process_chat_request_or_tool_response(
+            event=tool_call_context[request_id], ctx=ctx
+        )
```
xintongsong: I'd suggest to refactor the codes so that only necessary codes are called recursively.

python/flink_agents/api/execution_environment.py
```diff
@@ -170,3 +197,122 @@ def from_table(
     @abstractmethod
     def execute(self) -> None:
         """Execute agent individually."""
+
+    def add_action(
```
xintongsong: Do we need this on the execution environment?

python/flink_agents/runtime/local_execution_environment.py
```diff
@@ -54,6 +54,14 @@ def apply(self, agent: Agent) -> AgentBuilder:
         if self.__runner is not None:
             err_msg = "LocalAgentBuilder doesn't support apply multiple agents."
             raise RuntimeError(err_msg)
+        # inspect refer actions and resources from env to agent.
+        for action_name in agent._action_names:
+            agent.actions[action_name] = self.__env.actions[action_name]
+        for type, names in agent._resource_names.items():
```
xintongsong: How do we know `agent._action_names` and `agent._resource_names`?

python/flink_agents/examples/react_agent_example.py
```diff
@@ -0,0 +1,138 @@
...
+    ----------
+    output : str
+        The llm output.
+
+    Returns:
+    -------
+    OutputData:
+        The parsed output.
+    """
+    return OutputData.model_validate_json(output)
```
xintongsong: This is just throwing the work to users.

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -15,69 +15,125 @@
...
+    # if there is no tool call generated, return chat response directly
+    else:
+        # clear tool call context related to specific request id
+        tool_call_context = short_term_memory.get(TOOL_CALL_CONTEXT)
+        if tool_call_context and request_id in tool_call_context:
+            tool_call_context.pop(request_id)
+            short_term_memory.set(TOOL_CALL_CONTEXT, tool_call_context)
+        return ChatResponseEvent(
+            request=ChatRequestEvent(id=request_id, model=model, messages=messages),
```
xintongsong: We should try to avoid re-construct the `ChatRequestEvent`.

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -15,69 +15,125 @@
...
+
+        event = chat(
+            request_id=event.id,
+            model=event.model,
+            chat_model=chat_model,
+            messages=event.messages,
+            short_term_memory=short_term_memory,
+        )
+
+        ctx.send_event(event)
```
xintongsong: Why sending the event here outside `chat()`?

python/flink_agents/runtime/local_execution_environment.py
```diff
@@ -54,6 +54,10 @@ def apply(self, agent: Agent) -> AgentBuilder:
         if self.__runner is not None:
             err_msg = "LocalAgentBuilder doesn't support apply multiple agents."
             raise RuntimeError(err_msg)
+        # inspect resources from environment to agent instance.
+        registered_resources = self.__env.resources
+        for type, name_to_resource in registered_resources.items():
+            agent.resources[type] = name_to_resource | agent.resources[type]
```
xintongsong, line 62: Should not do this in `apply()`. What if the resources are registered afterwards?

python/flink_agents/runtime/remote_execution_environment.py
```diff
@@ -65,6 +71,14 @@ def apply(self, agent: Agent) -> "AgentBuilder":
         if self.__agent_plan is not None:
             err_msg = "RemoteAgentBuilder doesn't support apply multiple agents yet."
             raise RuntimeError(err_msg)
+
+        # inspect refer actions and resources from env to agent.
+        for type, names in agent._resource_names.items():
```
xintongsong: What is this `_resource_names`?

python/flink_agents/examples/react_agent_example.py
```diff
@@ -0,0 +1,149 @@
...
+        ],
+    )
+
+    # create ReAct agent.
+    agent = ReActAgent(
+        chat_model_setup=OllamaChatModelSetup,
+        connection="ollama",
+        prompt=prompt,
+        tools=["add", "multiply"],
+        output_schema_provider=get_output_row_type
```
xintongsong: Why do we need a provider here, rather than just provide the schema?

python/flink_agents/api/agents/react_agent.py
```diff
@@ -0,0 +1,251 @@
...
+            schema_prompt = f"The final response should be json format, and match the schema {json_schema}."
+            self._resources[ResourceType.PROMPT][DEFAULT_SCHEMA_PROMPT] = (
+                Prompt.from_text(name="output_schema", text=schema_prompt)
+            )
+
+            self._resources[ResourceType.TOOL][OUTPUT_SCHEMA] = output_schema_provider
+        if prompt:
+            self._resources[ResourceType.PROMPT][DEFAULT_USER_PROMPT] = prompt
+        if output_parser:
+            self._resources[ResourceType.TOOL][OUTPUT_PARSER] = output_parser
```
xintongsong: Why are `output_schema_provider` and `output_parser` tools?

python/flink_agents/runtime/local_execution_environment.py
wenjin272: Currently, the compile from Agent to AgentPlan must apply before to_datastream/to_table, because construct the output datastream/table depend on transferring AgentPlan from Python to Java. Thus, we can't do this in env.execute, for AgentPlan is already generate and we can't inspect registered resource any more. Should we do this in to_datastream？

xintongsong: Can we add more resources to the AgentPlan after compiling it from the Agent? Not saying we should do it in this PR. Just trying to understand.

python/flink_agents/api/events/tool_event.py
```diff
@@ -42,9 +43,9 @@ class ToolResponseEvent(Event):
     ----------
     request : ToolRequestEvent
         The correspond request of the response.
-    response : Any
-        The response from the tool.
+    responses : Dict[UUID, Any]
```
xintongsong, line 46: Why do we need `ToolRequestEvent` in `ToolResponseEvent`? Can we replace it with the request id, like what we do for `ChatResponseEvent`?

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -15,69 +15,129 @@
...
 from flink_agents.api.resource import ResourceType
 from flink_agents.api.runner_context import RunnerContext
 from flink_agents.plan.actions.action import Action
 from flink_agents.plan.function import PythonFunction
 
+TOOL_CALL_CONTEXT = "_TOOL_CALL_CONTEXT"
+
+
+def chat(
+    request_id: UUID,
```
xintongsong: What is this `request_id`? IIUC, this is the id of the initial `ChatRequestEvent`? We should make that explicit, perhaps with the name?

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -15,69 +15,129 @@
...
+        # append response to tool call context
+        tool_call_context[request_id].append(response)
+        # update tool call context
+        short_term_memory.set(TOOL_CALL_CONTEXT, tool_call_context)
+        ctx.send_event(
+            ToolRequestEvent(
+                id=request_id,
+                model=model,
+                tool_calls=response.tool_calls,
+            )
```
xintongsong: It seems we are creating a `ToolRequestEvent` with the id of the `ChatRequestEvent`? We should never have multiple events with the same id.

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -15,69 +15,129 @@
...
-                specific_tool_ctx = tool_context.pop(tool_call_id)
-                specific_tool_ctx.messages.append(
-                    ChatMessage(role=MessageRole.TOOL, content=str(event.response))
-                )
-                ctx.send_event(specific_tool_ctx)
-                # update short term memory to remove the specific tool call context
-                state.set("__tool_context", tool_context)
+        request_id = event.request.id
+        model = event.request.model
+        responses = event.responses
```
xintongsong: I assume this is why we need to keep `ToolRequestEvent` in `ToolResponseEvent`? Alternatively, we can save the tool request id and model in state, and fetch it when receiving the tool response.

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -15,69 +15,129 @@
...
 from flink_agents.plan.actions.action import Action
 from flink_agents.plan.function import PythonFunction
 
+TOOL_CALL_CONTEXT = "_TOOL_CALL_CONTEXT"
+
+
+def chat(
+    request_id: UUID,
+    model: str,
+    chat_model: BaseChatModelSetup,
```
xintongsong: Why do we need both `model` and `chat_model`?

xintongsong: Seems `chat_model` is converted from `model`. Why not do it inside `chat()`?

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -15,69 +15,129 @@
...
+    messages: List[ChatMessage],
+    short_term_memory: MemoryObject,
+    ctx: RunnerContext,
```
xintongsong: Why do we need `short_term_memory` as an argument, while we already have `ctx`?

plan/src/main/java/org/apache/flink/agents/plan/Action.java
```diff
@@ -58,4 +68,8 @@ public Function getExec() {
     public List<String> getListenEventTypes() {
         return listenEventTypes;
     }
+
+    public Map<String, Object> getParams() {
```
xintongsong: I'd suggest the name `getActionConfig`.

python/flink_agents/api/runner_context.py
```diff
@@ -53,6 +54,16 @@ def get_resource(self, name: str, type: ResourceType) -> Resource:
...
 
+    @abstractmethod
+    def get_action_params(self, name: str) -> Dict[str, Any]:
+        """Get additional parameters of action.
+
+        Parameters
+        ----------
+        name : str
+            The name of the action.
+        """
```
xintongsong: We should not need the user to pass in the action name. The framework should know this information.

xintongsong: And maybe we can have an additional method for getting one specific config / param, rather than returning all of them.

python/flink_agents/plan/tests/test_action.py
```diff
@@ -58,14 +60,22 @@ def action() -> Action:  # noqa: D103
         name="legal",
         exec=func,
         listen_event_types=[f"{InputEvent.__module__}.{InputEvent.__qualname__}"],
+        params={
```
xintongsong, line 63: `Action` is not a public api. The question is, do we want to expose the ability of adding parameters to users. If true, how can user use this feature via public apis? If not, we should not expose this via runner context, which is also public api.

python/flink_agents/api/agents/react_agent.py
```diff
@@ -0,0 +1,251 @@
...
+from flink_agents.api.events.event import InputEvent, OutputEvent
+from flink_agents.api.prompts.prompt import Prompt
+from flink_agents.api.resource import ResourceType
+from flink_agents.api.runner_context import RunnerContext
+
+DEFAULT_CHAT_MODEL = "_default_chat_model"
+DEFAULT_SCHEMA_PROMPT = "_default_schema_prompt"
+DEFAULT_USER_PROMPT = "_default_user_prompt"
+OUTPUT_PARSER = "_output_parser"
+OUTPUT_SCHEMA = "_output_schema"
```
xintongsong: Should these be private?

xintongsong (approved):
LGTM

Issue #120, opened 2025-08-20
[Feature] [integration][python] Integrate built-in openai chat model.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Provide built-in openai chat model.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #128, opened 2025-08-26
[integration][python] Integrate OpenAI chat model.

wenjin272:
Linked issue: #120 

### Purpose of change

Integrate OpenAI chat model.

### Tests

ut

### API

No

### Documentation

No


python/flink_agents/integrations/chat_models/utils.py
```diff
@@ -0,0 +1,40 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#################################################################################
+from typing import Any, Dict
+
+from flink_agents.api.tools.tool import ToolMetadata
+
+
+def to_openai_tool(
```
alnzng: is this method specific to OpenAI APIs? should we put this method inside OpenAI specific [utils.py](python/flink_agents/integrations/chat_models/openai/utils.py)?

python/flink_agents/integrations/chat_models/openai/openai_chat_model.py
```diff
@@ -0,0 +1,280 @@
...
+        tool_specs = None
+        if tools is not None:
+            tool_specs = [to_openai_tool(tool.metadata) for tool in tools]
+            strict = kwargs.get("strict", False)
+            for tool_spec in tool_specs:
+                if tool_spec["type"] == "function":
+                    tool_spec["function"]["strict"] = strict
+                    tool_spec["function"]["parameters"]["additionalProperties"] = False
+
+        response = self.client.chat.completions.create(
```
alnzng, line 168: Should we use OpenAI latest `Responses API` for the implementation? Looks like OpenAI recommends this `Responses API` for new projects: https://platform.openai.com/docs/guides/migrate-to-responses

>While Chat Completions remains supported, Responses is recommended for all new projects.

python/flink_agents/integrations/chat_models/openai/openai_chat_model.py
```diff
@@ -0,0 +1,280 @@
...
+        How long to wait, in seconds, for an API call before failing.
+    default_headers : Optional[Dict[str, str]]
+        The default headers for API requests.
+    reuse_client : bool
+        Whether to reuse the OpenAI client between requests.
+    """
+
+    api_key: str = Field(default=None, description="The OpenAI API key.")
+    api_base_url: str = Field(description="The base URL for OpenAI API.")
+    api_version: str = Field(description="The API version for OpenAI API.")
```
alnzng: Why these two `api_base_url ` and `api_version` fields need to be set explictly? Or in what scenarios the users should customize these two information?

Looks like current implementation leverages on OpenAI python SDK `openai>=1.66.3` to build the connection and interact, the official python SDK should set these two information properly by default, right?

python/flink_agents/integrations/chat_models/openai/openai_chat_model.py
```diff
@@ -0,0 +1,280 @@
...
+        config = self.__get_credential_kwargs()
+
+        if not self.reuse_client:
+            return OpenAI(**config)
+
+        if self._client is None:
+            self._client = OpenAI(**config)
+        return self._client
+
+    def __get_credential_kwargs(self) -> Dict[str, Any]:
```
alnzng: Maybe `_get_client_kwargs` a better name? This method return more than credential related configs.

python/flink_agents/integrations/chat_models/openai/utils.py
```diff
@@ -0,0 +1,131 @@
...
+import openai
+from openai.types.chat import ChatCompletionMessage, ChatCompletionMessageParam
+
+from flink_agents.api.chat_message import ChatMessage, MessageRole
+
+DEFAULT_OPENAI_API_BASE_URL = "https://api.openai.com/v1"
+DEFAULT_OPENAI_API_VERSION = ""
+
+
+def resolve_openai_credentials(
```
xintongsong: We have too many `utils.py` files. Just in this PR, there are two of them. Let's give them some meaningful names.

python/flink_agents/integrations/chat_models/openai/utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+    return final_api_key, str(final_api_base_url), final_api_version
+
+
+def get_from_param_or_env(
+    key: str,
+    param: Optional[str] = None,
+    env_key: Optional[str] = None,
+    default: Optional[str] = None,
+) -> str:
```
xintongsong:
```suggestion
def resolve_parameter(
    param_name: str,
    value_from_args: Optional[str] = None,
    env_var_name: Optional[str] = None,
    default_value: Optional[str] = None,
) -> str:
```

xintongsong: And we need more explain about the arguments and behavior of this function.

python/flink_agents/integrations/chat_models/openai/openai_chat_model.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+class OpenAIChatModelSetup(BaseChatModelSetup):
+    """The settings for the OpenAI LLM.
+
+    Attributes:
+    ----------
+    model : str
+        The OpenAI model to use.
+    temperature : str
```
alnzng: nit: `str` -> `float`

python/flink_agents/integrations/chat_models/utils.py
wenjin272: This method is also used by Ollama chat model. I rename this file from utils to chat_model_utils  to represent it is common utils can be used by all chat model implementations.

python/flink_agents/integrations/chat_models/openai/openai_chat_model.py
wenjin272: I think we should provide implementation base on the `Responses API` for it support more built-in tools, native remote MCP and reasoning summary.
But I think we also need provide implementation base on the `Chat Completions API`, for many platforms have built-in compatibility with it but have not support `Responses API`.
In summary, this implementation will keep using `Chat Completions API`, and I will create a new issue and add `Responses API` support later.

python/flink_agents/integrations/chat_models/openai/openai_chat_model.py
wenjin272: The api_version is not need any more, and I removed it.
The api_base_url is still need, for user can configure it to 
* https://api.deepseek.com for DeepSeek 
* https://dashscope.aliyuncs.com/compatible-mode/v1 for DashScope 
* https://api.chatanywhere.org/v1 for free OpenAI request.

and default is https://api.openai.com/v1 for official OpenAI

alnzng: I understand that different LLM providers(e.g. DeepSeek, DashScope) offer APIs tha are OpenAI compatible, so technically you are right on we can set the different `api_base_url ` to this OpenAI chat mode. 

But I thought our implementation follows the idea `one ChatModel implementation per LLM provider`, for example, openai_chat_model.py for OpenAI, tongyi_chat_model.py for alibaba, anthropic_chat_model.py for Anthropic, etc.  I see the industry doing similar strategy like this, for example:
- LlamaIndex: https://github.com/run-llama/llama_index/tree/main/llama-index-integrations/llms
- LangChain: https://github.com/pcliupc/langchain/tree/master/libs/community/langchain_community/chat_models

does our current plan use a different strategy? cc: @xintongsong

alnzng: But keep the `api_base_url ` be configurable makes sense to me, since the users might want to connect to a different proxy URL which point to OpenAI under the hood.

python/flink_agents/integrations/chat_models/openai/openai_chat_model.py
alnzng: >But I think we also need provide implementation base on the Chat Completions API, for many platforms have built-in compatibility with it but have not support Responses API.

This basically my question in this comment: https://github.com/apache/flink-agents/pull/128#discussion_r2308056685

If this chat model implementation is inended for OpenAI service dedicately, then we don't need to worry about the API compatibility with other LLM providers.

python/flink_agents/integrations/chat_models/openai/openai_chat_model.py
wenjin272: This implementation is more likely https://github.com/run-llama/llama_index/tree/main/llama-index-integrations/llms/llama-index-llms-openai-like. 

I think provide an implementation based on  `Chat Completions API` makes sense.  There are two main reasons: 
* Some LLM providers doesn't provide their own API, they only support `Chat Completions API`, like DeepSeek.
* It is convenient for user to transfer from different LLM providers. For example, user has used DeepSeek, and he want to transfer to DashScope or OpenAI, he just need modify the `api_base_url`.

I agree the strategy that provide one ChatModel implementation per LLM provider. But I think this has no conflict with providing an implementation based on `Chat Completions API`. Perhaps we could modify this implementation name from `OpenAIChatModel` to `OpenAIChatCompletionModel`. And I create the issue #132 which plan to add `OpenAIResponseChatModel`. For other LLM providers which have their own API, we also will provide correspond implementations.

xintongsong: > But I thought our implementation follows the idea `one ChatModel implementation per LLM provider`, for example, openai_chat_model.py for OpenAI, tongyi_chat_model.py for alibaba, anthropic_chat_model.py for Anthropic, etc. 

Not necessarily.

In long term, I think for most model providers having dedicated implementations would be helpful, even if the they are openai compatible. This allows us to fully leverage the unique features and avoid issues caused by inconsistent behaviors.

However, at this moment, I'd like to see as many model providers supported as possible, with minimum efforts. To that end, a general implementation for all open-ai compatible model providers makes a good first step.

alnzng: Thanks for answering my commetns @wenjin272 and @xintongsong .

@wenjin272 +1 on this more like a openai-like intention, which is similar with the LlamaIndex link you pointed out above. maybe we can follow a similar strategy like LlamaIndex to name the implementaiton as `openai_like_chat_model.py`. But I guess we can start the name `openai_chat_model.py` and iterate/refactor once we work on #132 

@xintongsong I'm not against we provide a general API for all openai-compabitle LLM provider. I was just confusing on the naming pattern of our implementation. My thought was that we should use name `openai_like_chat_model` or `openai_compatible_chat_model` if the implementation was for a general purpose. That being said, I thought `openai_chat_model` or `anthropic_chat_model` or `tongyi_chat_model` naming means a dedicated implementation for each LLM provider.

As a conclusion, I think we aligned that long term we should use `per implementation per LLM provider` strategy. Let's keep current implementation to help us move faster, and let's reiterate / refactor it in the future.

alnzng (approved):
LGTM.

python/flink_agents/integrations/chat_models/openai/openai_chat_model.py
xintongsong: Thanks, @alnzng. I'll merge this PR for now. And let's change the name once we work on the openai dedicated implementation, as you suggested.

PR #138, opened 2025-08-31
[Feature] [api] Introduce per action state consistency API

letaoj:
Linked issue: #152 

### Purpose of change

This change added a new set of APIs to make sure the action level consistency after a crash and recovery. The design was discussed in #108 

### Tests

Unit test added for the integration of ActionStateStore

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
+++ b/api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
@@ -23,4 +23,8 @@ public class AgentConfigOptions {
     /** The config parameter specifies the directory for the FileEvent file. */
     public static final ConfigOption<String> BASE_LOG_DIR =
             new ConfigOption<>("baseLogDir", String.class, null);
+
+    /** The config parameter specifies the backend for action state store. */
+    public static final ConfigOption<String> ACTION_STATE_STORE_BACKEND =
+            new ConfigOption<>("actionStateStoreBackend", String.class, "inmemory");
 }
--- a/api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
+++ b/api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
@@ -26,5 +26,5 @@ public class AgentConfigOptions {
 
     /** The config parameter specifies the backend for action state store. */
     public static final ConfigOption<String> ACTION_STATE_STORE_BACKEND =
-            new ConfigOption<>("actionStateStoreBackend", String.class, "inmemory");
+            new ConfigOption<>("actionStateStoreBackend", String.class, null);
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
@@ -30,7 +30,7 @@ public interface ActionStateStore {
      * @param action the action the agent is taking
      * @param state the current state of the whole task
      */
-    public void put(Object key, Action action, ActionState state);
+    void put(Object key, Action action, ActionState state);
 
     /**
      * Retrieve the state of a specific action associated with a given key from the backend storage.
@@ -39,7 +39,7 @@ public interface ActionStateStore {
      * @param action the action the agent is taking
      * @return the state of the action, or null if not found
      */
-    public ActionState get(Object key, Action action);
+    ActionState get(Object key, Action action);
 
     /**
      * Retrieve all states associated with a given key from the backend storage.
@@ -47,5 +47,5 @@ public interface ActionStateStore {
      * @param key the key associated with the message
      * @return a map of key of action to action states associated with the key
      */
-    public Map<String, ActionState> getAll(Object key);
+    Map<String, ActionState> rebuildState(Object key);
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.runtime.actionstate;
+
+import org.apache.flink.agents.plan.Action;
+
+import java.util.Map;
+
+/** Interface for storing and retrieving the state of actions performed by agents. */
+public interface ActionStateStore {
+    /**
+     * Store the state of a specific action associated with a given key to the backend storage.
+     *
+     * @param key the key associate with the message
+     * @param action the action the agent is taking
+     * @param state the current state of the whole task
+     */
+    public void put(Object key, Action action, ActionState state);
+
+    /**
+     * Retrieve the state of a specific action associated with a given key from the backend storage.
+     *
+     * @param key the key associated with the message
+     * @param action the action the agent is taking
+     * @return the state of the action, or null if not found
+     */
+    public ActionState get(Object key, Action action);
+
+    /**
+     * Retrieve all states associated with a given key from the backend storage.
+     *
+     * @param key the key associated with the message
+     * @return a map of key of action to action states associated with the key
+     */
+    public Map<String, ActionState> getAll(Object key);
+}
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
@@ -20,6 +20,8 @@ package org.apache.flink.agents.runtime.actionstate;
 import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.plan.Action;
 
+import java.io.IOException;
+
 /** Interface for storing and retrieving the state of actions performed by agents. */
 public interface ActionStateStore {
     enum BackendType {
@@ -43,8 +45,9 @@ public interface ActionStateStore {
      * @param action the action the agent is taking
      * @param event the event that triggered the action
      * @param state the current state of the whole task
+     * @throws IOException when key generation failed
      */
-    void put(Object key, Action action, Event event, ActionState state);
+    void put(Object key, Action action, Event event, ActionState state) throws IOException;
 
     /**
      * Retrieve the state of a specific action associated with a given key from the backend storage.
@@ -53,8 +56,9 @@ public interface ActionStateStore {
      * @param action the action the agent is taking
      * @param event the event that triggered the action
      * @return the state of the action, or null if not found
+     * @throws IOException when key generation failed
      */
-    ActionState get(Object key, Action action, Event event);
+    ActionState get(Object key, Action action, Event event) throws IOException;
 
     /**
      * Retrieve all states associated with a given key from the backend storage and recover the
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
@@ -17,6 +17,7 @@
  */
 package org.apache.flink.agents.runtime.actionstate;
 
+import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.plan.Action;
 
 import java.util.Map;
@@ -28,18 +29,20 @@ public interface ActionStateStore {
      *
      * @param key the key associate with the message
      * @param action the action the agent is taking
+     * @param event the event that triggered the action
      * @param state the current state of the whole task
      */
-    void put(Object key, Action action, ActionState state);
+    void put(Object key, Action action, Event event, ActionState state);
 
     /**
      * Retrieve the state of a specific action associated with a given key from the backend storage.
      *
      * @param key the key associated with the message
      * @param action the action the agent is taking
+     * @param event the event that triggered the action
      * @return the state of the action, or null if not found
      */
-    ActionState get(Object key, Action action);
+    ActionState get(Object key, Action action, Event event);
 
     /**
      * Retrieve all states associated with a given key from the backend storage.
@@ -48,4 +51,7 @@ public interface ActionStateStore {
      * @return a map of key of action to action states associated with the key
      */
     Map<String, ActionState> rebuildState(Object key);
+
+    /** Clean up state store to avoid evergrowing storage usage. */
+    void cleanUpState();
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
@@ -20,10 +20,22 @@ package org.apache.flink.agents.runtime.actionstate;
 import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.plan.Action;
 
-import java.util.Map;
-
 /** Interface for storing and retrieving the state of actions performed by agents. */
 public interface ActionStateStore {
+    enum BackendType {
+        INMEMORY("inmemory"),
+        KAFKA("kafka");
+
+        private final String type;
+
+        BackendType(String type) {
+            this.type = type;
+        }
+
+        public String getType() {
+            return type;
+        }
+    }
     /**
      * Store the state of a specific action associated with a given key to the backend storage.
      *
@@ -45,13 +57,20 @@ public interface ActionStateStore {
     ActionState get(Object key, Action action, Event event);
 
     /**
-     * Retrieve all states associated with a given key from the backend storage.
-     *
-     * @param key the key associated with the message
-     * @return a map of key of action to action states associated with the key
+     * Retrieve all states associated with a given key from the backend storage and recover the
+     * local action states.
      */
-    Map<String, ActionState> rebuildState(Object key);
+    void rebuildState(Object recoveryMarker);
 
     /** Clean up state store to avoid evergrowing storage usage. */
     void cleanUpState();
+
+    /**
+     * Get a marker object representing the current recovery point in the state store.
+     *
+     * @return a marker object, or null if not supported
+     */
+    default Object getRecoveryMarker() {
+        return null;
+    }
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
@@ -25,7 +25,6 @@ import java.io.IOException;
 /** Interface for storing and retrieving the state of actions performed by agents. */
 public interface ActionStateStore {
     enum BackendType {
-        INMEMORY("inmemory"),
         KAFKA("kafka");
 
         private final String type;
@@ -42,23 +41,28 @@ public interface ActionStateStore {
      * Store the state of a specific action associated with a given key to the backend storage.
      *
      * @param key the key associate with the message
+     * @param seqNum the sequence number of the key
      * @param action the action the agent is taking
      * @param event the event that triggered the action
      * @param state the current state of the whole task
      * @throws IOException when key generation failed
      */
-    void put(Object key, Action action, Event event, ActionState state) throws IOException;
+    void put(Object key, long seqNum, Action action, Event event, ActionState state)
+            throws IOException;
 
     /**
      * Retrieve the state of a specific action associated with a given key from the backend storage.
+     * It any of the sequence number for a key can't be found, all the states associated with the
+     * key after the sequence number should be ignored and null will be returned.
      *
      * @param key the key associated with the message
+     * @param seqNum the sequence number of the key
      * @param action the action the agent is taking
      * @param event the event that triggered the action
      * @return the state of the action, or null if not found
      * @throws IOException when key generation failed
      */
-    ActionState get(Object key, Action action, Event event) throws IOException;
+    ActionState get(Object key, long seqNum, Action action, Event event) throws IOException;
 
     /**
      * Retrieve all states associated with a given key from the backend storage and recover the
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
@@ -21,6 +21,7 @@ import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.plan.Action;
 
 import java.io.IOException;
+import java.util.List;
 
 /** Interface for storing and retrieving the state of actions performed by agents. */
 public interface ActionStateStore {
@@ -37,6 +38,7 @@ public interface ActionStateStore {
             return type;
         }
     }
+
     /**
      * Store the state of a specific action associated with a given key to the backend storage.
      *
@@ -65,13 +67,18 @@ public interface ActionStateStore {
     ActionState get(Object key, long seqNum, Action action, Event event) throws IOException;
 
     /**
-     * Retrieve all states associated with a given key from the backend storage and recover the
-     * local action states.
+     * Rebuild the in-memory state from the backend storage using the provided recovery markers.
+     *
+     * @param recoveryMarkers a list of markers representing the recovery points
      */
-    void rebuildState(Object recoveryMarker);
+    void rebuildState(List<Object> recoveryMarkers);
 
-    /** Clean up state store to avoid evergrowing storage usage. */
-    void cleanUpState();
+    /**
+     * Prune the state for a given key.
+     *
+     * @param key the key whose state should be pruned
+     */
+    void pruneState(Object key);
 
     /**
      * Get a marker object representing the current recovery point in the state store.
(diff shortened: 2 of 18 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -301,20 +309,47 @@ private void processActionTaskForKey(Object key) throws Exception {
...
             builtInMetrics.markActionExecuted(actionTask.action.getName());
             currentInputEventFinished = !currentKeyHasMoreActionTask();
         } else {
-            // If the action task not finished, we should get a new action task to execute continue.
-            Optional<ActionTask> generatedActionTaskOpt = actionTaskResult.getGeneratedActionTask();
+            // If the action task is not finished, we should get a new action task to continue the
+            // execution.
             checkNotNull(
-                    generatedActionTaskOpt.isPresent(),
+                    generatedActionTaskOpt.get(),
```
letaoj, line 393: this is a potential bug that was fixed in this PR, isPresent() returns true/false and will never by null. By doing a get on optional, it will return the actual value in the optional so that we can properly check the nullity

runtime/src/main/java/org/apache/flink/agents/runtime/memory/MemoryObjectImpl.java
```diff
@@ -35,6 +36,7 @@ private enum ItemType {
     private static final String SEPARATOR = ".";
 
     private final MapState<String, MemoryItem> store;
+    private final List<MemoryUpdate> memoryUpdates;
```
xintongsong, line 45: I'd suggest to store the memory updates in the runner context. We can pass it in when creating the memory object.

1. `MemoryObject` is an user-facing API. We should at least keep `getAllUpdates()` as part of the internal implementation, so that user don't have the chance to mess-up with it.
2. `MemoryObject` can be hierachical. User may call `set()` on child memory objects. So calling `getAllUpdates()` for persistence only on the root memory object is not enough.
3. Conceptually, it's more straightforward that memory object only keeps the content that should be memorized, and all execution contextual information is maintained in the runner context.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -138,6 +145,7 @@ public ActionExecutionOperator(
         this.mailboxExecutor = mailboxExecutor;
         this.eventLogger = EventLoggerFactory.createLogger(EventLoggerConfig.builder().build());
         this.eventListeners = new ArrayList<>();
+        this.actionStateStore = new KafkaActionStateStore();
```
xintongsong: It might be better making this configurable. Although Kafka-based store is the only option we currently provide, users should have another option to not providing an external kafka service and disable this feature.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -301,20 +309,47 @@ private void processActionTaskForKey(Object key) throws Exception {
 
         // 2. Invoke the action task.
         createAndSetRunnerContext(actionTask);
-        ActionTask.ActionTaskResult actionTaskResult = actionTask.invoke();
-        for (Event actionOutputEvent : actionTaskResult.getOutputEvents()) {
+
+        boolean isFinished = false;
+        List<Event> outputEvents;
+        Optional<ActionTask> generatedActionTaskOpt;
+        ActionState actionState = actionStateStore.get(key, actionTask.action);
```
xintongsong: The key and action do not uniquely identify an action task. An action can be triggered multiple times, by different events. I think we should add the event id as part of the identifier.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -301,20 +309,47 @@ private void processActionTaskForKey(Object key) throws Exception {
...
+        if (actionState != null) {
+            isFinished = actionState.getGeneratedActionTask().isPresent();
```
xintongsong: `isFinished` should be `false` when generated action task is present.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -301,20 +309,47 @@ private void processActionTaskForKey(Object key) throws Exception {
...
+            outputEvents = actionState.getOutputEvents();
+            generatedActionTaskOpt = actionState.getGeneratedActionTask();
+            for (MemoryUpdate memoryUpdate : actionState.getMemoryUpdates()) {
+                actionTask
+                        .getRunnerContext()
+                        .getShortTermMemory()
+                        .set(memoryUpdate.getPath(), memoryUpdate.getValue());
+            }
+        } else {
```
xintongsong, line 366: In python async execution, if an action task is invoked but not finished, it will be put back to `actionTasksKState` and re-invoked later. In that case, it will falls into the `if (actionState != null)` clause, and treated as replay rather than resume.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/MemoryObjectImpl.java
letaoj: sounds good. The reason why I keep it inside of the memory object is because it will ease how we keep track of the memory updates. We can tracks the update when the put is called but if the update is inside of the context, i might need to figure out a way to find the updates.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: Originally, I was thinking that for the same action, we should always produce the same result but think about it twice, even the same action with different memory, the result could be different. I will update the key to include event id

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: make sense, i will make this configurable.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: Thanks for catching this! Fixed.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: Updated the condition to also check the generated actiontask is null

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: @xintongsong any suggestion on how to make it configurable? I do not see any Configuration passed in to this operator. I might miss a bit context here.

xintongsong: There's an `AgentConfiguration` in `AgentPlan`. It was introduced in PR #122

letaoj: Ah. nice. Thanks for the pointer

letaoj:
@xintongsong ack. Rebased the PR to get rid of the merge commit.

api/src/main/java/org/apache/flink/agents/api/context/RunnerContext.java
```diff
@@ -73,4 +75,12 @@ public interface RunnerContext {
...
      */
     ReadableConfiguration getConfig();
+
+    /**
+     * Gets all the updates made to this MemoryObject since it was created or the last time this
+     * method was called.
+     *
+     * @return list of memory updates
+     */
+    List<MemoryUpdate> getAllMemoryUpdates();
```
Sxnan: The `RunnerContext` is a user-facing API, and this method is intended to be invoked by the framework only. To avoid the user accidentally calling this method, we may introduce an internal interface for the framework.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
```diff
@@ -0,0 +1,37 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.UUID;
+
+/** Utility class for action state related operations. */
+public class ActionStateUtil {
+
+    public static String generateKey(Object key, Action action, Event event) {
+        return key
+                + "-"
+                + event.getId().toString()
```
Sxnan: Correct me if I am wrong, the ID (UUID) of the event is randomly generated while instantiating the event, which means that the ID will be different when the same event is reprocessed during recovery. 

I think we need to generate the ID based on the content of the event. cc @xintongsong

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -119,4 +124,12 @@ public ReadableConfiguration getConfig() {
     public String getActionName() {
         return actionName;
     }
+
+    @Override
+    public List<MemoryUpdate> getAllMemoryUpdates() {
+        mailboxThreadChecker.run();
+        List<MemoryUpdate> memoryUpdatesCopy = List.copyOf(memoryUpdates);
+        memoryUpdates.clear();
```
Sxnan: Why do we need to clear updates when getting? It feels unintuitive and unnecessary at the same time, as the runner context is instantiated on every org.apache.flink.agents.runtime.operator.ActionTask#invoke.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -301,20 +311,48 @@ private void processActionTaskForKey(Object key) throws Exception {
...
         // 2. Invoke the action task.
         createAndSetRunnerContext(actionTask);
-        ActionTask.ActionTaskResult actionTaskResult = actionTask.invoke();
-        for (Event actionOutputEvent : actionTaskResult.getOutputEvents()) {
+
+        boolean isFinished;
+        List<Event> outputEvents;
+        Optional<ActionTask> generatedActionTaskOpt;
+        ActionState actionState = maybeGetActionState(key, actionTask.action, actionTask.event);
+        if (actionState != null && actionState.getGeneratedActionTask() == null) {
```
Sxnan: Should be `actionState.getGeneratedActionTask().isEmpty()`

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -301,20 +311,48 @@ private void processActionTaskForKey(Object key) throws Exception {
...
+            isFinished = !actionState.getGeneratedActionTask().isPresent();
```
Sxnan: Is it always true?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
Sxnan: Good catch!

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -392,6 +430,22 @@ public void close() throws Exception {
         super.close();
     }
 
+    @Override
+    public void initializeState(StateInitializationContext context) throws Exception {
```
Sxnan, line 473: I am wondering whether we need this and the `snapshotState` methods here, as it just call the super's.

api/src/main/java/org/apache/flink/agents/api/context/RunnerContext.java
letaoj: I will move it to the RunnerContextImpl since I do see some method that lives in the RunnerContextImpl and seems to be internal. ActionTask's `getRunnerContext` method is also returning the `RunnerContextImpl` because I think it has some internal only method that needs to be accessed by the operator. If you think we should make an InternalRunnerContext, let me know.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
letaoj: Thanks for catching this. Updated the logic to generate the id from the attributes map in it.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
letaoj: I was thinking that the runner context will be reused over time so that it will be better to clear the memory updates each time after the task is persisted, but you are right, the runner context is recreated for every single action task so there is no need to clear the updates.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: :D yeah, same issue as the bug that i fixed lol, thanks!

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: Yes, it is. Updated to true directly

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: This is for my next PR since we are going to persist some state to the flink state store.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
```diff
@@ -23,4 +23,8 @@ public class AgentConfigOptions {
     /** The config parameter specifies the directory for the FileEvent file. */
     public static final ConfigOption<String> BASE_LOG_DIR =
             new ConfigOption<>("baseLogDir", String.class, null);
+
+    /** The config parameter specifies the backend for action state store. */
+    public static final ConfigOption<String> ACTION_STATE_STORE_BACKEND =
+            new ConfigOption<>("actionStateStoreBackend", String.class, "inmemory");
```
Sxnan: The default value may be null, which disables the action state altogether.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
```diff
@@ -0,0 +1,38 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.UUID;
+
+/** Utility class for action state related operations. */
+public class ActionStateUtil {
+
+    public static String generateKey(Object key, Action action, Event event) {
+        return key
+                + "-"
+                + UUID.nameUUIDFromBytes(
```
Sxnan: I have an offline discussion with @xintongsong about the key of the event. Here is what we have in consensus:

We can only ensure per‑action state consistency on a best‑effort basis. Here are the limitations to ensure per-action state consistency:
1. The order and the content of input events for each key should be consistent before and after recovery.
2. One action cannot depend on the state of another action if they can be triggered by the same event. 

Here is what the implementation should look like:
1. Only the InputEvent and PythonEvent with `flink_agents.api.events.event.InputEvent` as type name need a consistent ID.
2. The ID of those events should be computed by both the content of the event and a per-key sequence number. 
3. Whenever we cannot find an action state for a given action state key, which means that the input event arrived in a different order after recovery, we stop looking up the action state for the input event with the same key afterward. 

@xintongsong let me know if I missed something

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/InMemoryActionStateStore.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.flink.agents.api.Event;
+import org.apache.flink.agents.plan.Action;
+import org.apache.flink.annotation.VisibleForTesting;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.agents.runtime.actionstate.ActionStateUtil.generateKey;
+
+public class InMemoryActionStateStore implements ActionStateStore {
```
Sxnan: If this cannot be used in any way for production, how about putting it to the test code.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -392,6 +440,29 @@ public void close() throws Exception {
         super.close();
     }
 
+    @Override
+    public void snapshotState(StateSnapshotContext context) throws Exception {
+        super.snapshotState(context);
+        if (actionStateStore != null) {
+            Object recoveryMarker = actionStateStore.getRecoveryMarker();
```
Sxnan, line 501: IIUC, there should be a place where we pass the recovery marker to the action state store so that it knows where to replay from.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
letaoj: I thought the last field is the default value, let me take a look at the configOption implementation again.

letaoj:
```
    public ConfigOption(String key, Class<T> type, T defaultValue) {
        this.key = Objects.requireNonNull(key);
        this.type = Objects.requireNonNull(type);
        this.defaultValue = defaultValue;
    }
```

yeah, the last value is default  which is `inmemory`. Do you think I should make the default null so that when not enabled explicitly, it will disable all of them?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/InMemoryActionStateStore.java
letaoj: Sure. I will move it to test

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: Yeah, it is in the initializeState which I accidentally removed. Updated it.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
letaoj:
```suggestion
            new ConfigOption<>("actionStateStoreBackend", String.class, null);
```

Sxnan: Yes, I think ActionStateStore should be optional to run a Flink agent job, so it makes more sense to make the default null.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
Sxnan: I think key generation should not include the uuid from the event, the uuid is randomly generated. And it is currently not generated with a per-key sequence number. For that sequence number, I think we can use the Flink state to store a monotonically increasing number, assign and increase the number whenever an input event arrives.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.apache.flink.agents.api.Event;
+import org.apache.flink.agents.plan.Action;
+
+import java.io.IOException;
+
+/** Interface for storing and retrieving the state of actions performed by agents. */
+public interface ActionStateStore {
+    enum BackendType {
+        INMEMORY("inmemory"),
```
Sxnan: We can remove this

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -392,6 +447,71 @@ public void close() throws Exception {
...
+            if (recoveryMarkers != null) {
+                for (Object marker : recoveryMarkers) {
+                    // there should be only one recovery marker
+                    recoveryMarker = marker;
+                    break;
+                }
+            }
+            actionStateStore.rebuildState(recoveryMarker);
+        }
+        ListState<Long> msgSequenceNumberOpState =
```
Sxnan: I think we should use the keyed value state for the sequence number for each key

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
Sxnan: Since one action can depend on the state produced by a previous action, if the input event arrived in a different order after recovery, we have to drop the action state for that key. I think this is missing in the PR.

One possible implementation is that if we cannot find an action state for a key `k` and sequence number `n`, we clear the action state for key `k` whose sequence number is greater than `n`.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: Now I got your point. This is not a globally increased sequence number, instead, a per-key based. This comes to the part that I'm not familiar with. Should I retrieve the keyed state as part of operator run logic instead of just recover it during init state?

letaoj: can you check if my implementation of using keyed state is correct or not?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
letaoj: removed

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
Sxnan: Could we use the value state here? It can reduce the complexity of the code.

```java
sequenceNumberKState = getRuntimeContext().getState(new ValueStateDescriptor<>(RECOVERY_MARKER_STATE_NAME, Long.class))
```

letaoj: Yeah, I think I was having trouble to get the keyed state work when I used ValueState and then I switched to ListState. However, the issue isn't because the ValueState vs ListState, it is because I was calling the value() on the ValueState out of a keyed context. Updated the logic.

letaoj:
@Sxnan For the pruning logic, I implemented in a different way where I remember the last seqNum that we did not find in the state map and for any sequence that comes in for the same key, it will only return the state when the sequence number is smaller than the last not found sequence.

Sxnan:
I spotted a few bugs that need to be addressed:

1. There is a fundamental problem that Python async execution cannot work with per-action state and checkpoint at the moment. I created an issue https://github.com/apache/flink-agents/issues/185. But it should not block this PR, as it is not introduced by this PR.
2. The operator state used to store the `recoveryMarks` has to be handled carefully. We should take the `currentProcessingKeysOpState` as an example. During recovery, we should pass the unioned list of markers to the ActionStateStore, and let the action state store decide where to recover from given all the markers.
3. We cannot simply clean up the state in notifyCheckpointComplete, since we don't know if there are states that are yet to be used to recover. Therefore, we can only clean the state up to the sequence number for each key. 

For 2 and 3, it requires some advanced usage of the Flink state, so I prepare some code for your reference. Feel free to use the code. https://github.com/Sxnan/flink-agents/commits/per-stask-state-poc/

letaoj:
Had a offline discussion with @Sxnan and here's the conclusion

1. We will use UnionState to guarantee the correctness during recovery even though for Kafka based state store, there could be read amplification happens. This should be okay for v0.1 and we will further optimized it once we launched the agent.
2. As @Sxnan mentioned, the state recovery finishes when the state is processed. Then during replay, the checkpoint could happen at any time so we should not simply clean up the entire map instead remember the last proceed sequence and then clean up upto the latest proceed sequence number

letaoj:
@Sxnan I took a different approach for state cleanup. Since the in-memory state is only useful during recover so we will not put new state into the map when we persist the state during `put` for Kafka-based state store. We will clean up the keyed state when an output event triggered (indicates the end of a key), let me know what do you think!

letaoj:
@Sxnan I guess I will move it to when there is no more pending event then clear the key. I thought the output event is the one indicate the end of a key but it looks like it is only a special event that will produce output to the output stream.

letaoj:
@Sxnan updated.

letaoj:
@Sxnan UPDATED the code to leverage the POC code that you provided! Thanks for the code pointer. I've also added a new test case to make sure the replay will be used when the state store has the key and sequence number.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Rebuild the in-memory state from the backend storage using the provided recovery markers.
+     *
+     * @param recoveryMarkers a list of markers representing the recovery points
+     */
+    void rebuildState(List<Object> recoveryMarkers);
+
+    /**
+     * Prune the state for a given key.
```
Sxnan, line 77: Just to make sure that, pruning here means removing from the in-memory(cache), instead of the actual state from the remote persistent storage.

If so, it should be worth mentioning in the docs.

letaoj:
@xintongsong 

> When calling put() with an ActionState whose generatedActionTask is not null (meaning the action is not finished), the action state will not be flushed to the external Kafka and will be kept in memory, until another put() that changes the generatedActionTask to null (meaning the action is finished). Am I understanding correctly?

We will flush the data to the backend store in every put request. Kafka's compaction logic should ideally compact the same key with two different value when later the action is completed and present the latest value when read it. During recovery, if the action is not finished, then we will go ahead and execute the action directly instead of looking it up from the store.

> When we say get() will ignore states after a not-found sequence number, does it include new states written after the not-founding get()? E.g., we put s1, s2 into the store, and then try to get s1 with a mismatched action or seq-num, then its expected that both s1, s2 will not be used. But what if we then put a s3 into the store?

Yes, it will ignore it. Since the newly written s3 will not and also should not be used for recovery.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
letaoj: Ideally, prune should not only prune the in-memory state but also clean up state in the backend storage because we do not want to have overgrowing state. However, i think this requires more discussion when we use a real database as the backend storage.

xintongsong (approved):
Thanks both @letaoj and @Sxnan for working on this PR and for the indepth discussion. I think this is good enough as the first version.

Issue #129, opened 2025-08-27
[Feature] [integration][python] Integrate built-in Anthropic chat model.


alnzng:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Provide built-in Anthropic chat model

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #139, opened 2025-09-02
[integration][python] Anthropic Chat Model APIs Integration

alnzng:
Linked issue: #129

### Purpose of change

Integrate Anthropic Chat Model APIs.

### Tests

Unit tests, also use `chat_model_example` as template to verify the Anthropic APIs.

```
2025-09-02 11:37:31,357 - INFO - HTTP Request: POST https://api.anthropic.com/v1/messages "HTTP/1.1 200 OK"
2025-09-02 11:37:31,359 - INFO - key: 0002, send_event: id=UUID('5445c578-d209-4520-b0d1-ec9b44667df1') request=ChatRequestEvent(id=UUID('3f708c50-7d06-4326-ab54-b58091be8b3b'), model='creative_chat_model', messages=[ChatMessage(role=<MessageRole.USER: 'user'>, content='Tell me a joke about cats.', tool_calls=[], extra_args={})]) response=ChatMessage(role=<MessageRole.ASSISTANT: 'assistant'>, content="Why don't cats play poker in the jungle?\n\nToo many cheetahs! 🐱", tool_calls=[], extra_args={})
2025-09-02 11:37:31,360 - INFO - key: 0002, performing action: process_chat_response
2025-09-02 11:37:31,360 - INFO - key: 0002, send_event: id=UUID('6c29e9dc-f5e7-4602-84ce-44e2d0f6686c') output="Why don't cats play poker in the jungle?\n\nToo many cheetahs! 🐱"
0001: The sum of 1 and 2 is 3.
0002: Why don't cats play poker in the jungle?

Too many cheetahs! 🐱
```
### API

  - ToolRequestEvent: Added optional external_id field for storing original tool call IDs from external systems
  (e.g., Anthropic tool_use_id)

### Documentation

n/a

### DIFF
--- a/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
+++ b/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
@@ -59,6 +59,14 @@ def convert_to_anthropic_message(message: ChatMessage) -> MessageParam:
                 }
             ],
         }
+    elif message.role == MessageRole.ASSISTANT:
+        # Use original Anthropic content blocks if available for context
+        anthropic_content_blocks = message.extra_args.get("anthropic_content_blocks")
+        content = anthropic_content_blocks if anthropic_content_blocks is not None else message.content
+        return {
+            "role": message.role.value,
+            "content": content,  # type: ignore
+        }
     else:
         return {
             "role": message.role.value,
@@ -176,8 +184,12 @@ class AnthropicChatModelConnection(BaseChatModelConnection):
 
             return ChatMessage(
                 role=MessageRole(message.role),
-                content=message.content,
-                tool_calls=tool_calls
+                content=message.content[0].text,
+                tool_calls=tool_calls,
+                extra_args={
+                    "anthropic_content_blocks": message.content
+                }
+
             )
         else:
             # TODO: handle other stop_reason values according to Anthropic API:
--- a/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
+++ b/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
@@ -15,18 +15,17 @@
 #  See the License for the specific language governing permissions and
 # limitations under the License.
 #################################################################################
-import typing
 import uuid
+from typing import Sequence, Optional, List, Any, Dict
 
+from anthropic import Anthropic
+from anthropic._types import NOT_GIVEN
+from anthropic.types import MessageParam, ToolParam, TextBlockParam
 from pydantic import Field, PrivateAttr
-from typing import Sequence, Optional, List, Any, Dict, Iterable
 
 from flink_agents.api.chat_message import ChatMessage, MessageRole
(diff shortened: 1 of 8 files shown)


wenjin272:
LGTM

wenjin272:
LGTM~

python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
```diff
@@ -0,0 +1,245 @@
...
+                    },
+                    "original_id": content_block.id,
+                }
+                for content_block in message.content
+                if content_block.type == 'tool_use'
+            ]
+
+            return ChatMessage(
+                role=MessageRole(message.role),
+                content=message.content,
```
xintongsong: IIUC, this is the reason why we need to change the type of `ChatMessage.content` into `Union[str, List[Any]]`?

But do we need all the content blocks in the chat message? I think here we only care about the text block and tool use block (which is already converted into `tool_calls`.

alnzng: Yes, we need the previous content blocks as context. This implementation follows Anthropic official SDK example: https://github.com/anthropics/anthropic-sdk-python/blob/main/examples/tools.py#L39

IIUC, This is one of few major differences between Anthropic SDK and other SDKs(like OpenAI SDK). It requires the client pass the previous tool call context, otherwise it throws runtime errors like below example:

```
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.2.content.0: unexpected `tool_use_id` found in `tool_result` blocks: toolu_01RPcDHy7vKQpAYgzL8Cd5Dr. Each `tool_result` block must have a corresponding `tool_use` block in the previous message.'}, 'request_id': 'req_011CSomfwQM8agx2hSbivysG'}
```

xintongsong: I see. In that case, I'd suggest to put the context in `ChatMessage.extra_args`, rather than the `content`. `role`, `content` and `tool_calls` are common fields of chat message that we should not assume only the anthropic chat model will access them. On the other hand, `extra_args` is a dict that accept arbitrary key-values, which is probably more suitable for storing implementation-specific information. WDYT?

alnzng: Good question. The current proposal should work well with the single model / LLM provider execution(which was my assumption), however as you pointed that the framework might need to support multi-models execution, which requires the `ChatMessage.content` to store model agnostic information only.

That being said, it makes sense to me that moving these anthropic specific content / context into `extra_args`. Let me refactor the codes accordingly. Thanks for the suggestion.

alnzng: Refactored the codes with your suggestion, please take a look. Thanks.

xintongsong (approved):
Thanks for addressing my comments. LGTM. Merging.

Issue #149, opened 2025-09-10
[Feature] [Integration] Add Embedding Model abstraction and provide a built-in Ollama embedding model implementation


alnzng:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Add Embedding Model abstraction and provide a built-in Ollama embedding model implementation

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #150, opened 2025-09-10
[api][python] Add Embedding Model abstraction and provide a built-in Ollama embedding model implementation

alnzng:
Linked issue: #149 

### Purpose of change

This PR adds embedding model support to Flink Agents by defining base abstractions for embedding model connections and setups, along with a complete Ollama embedding model implementation for local development and testing.

### Tests

Unit tests

### API

 New Base Classes:
  - `BaseEmbeddingModelConnection` - Abstract base for embedding model connections
  - `BaseEmbeddingModelSetup` - Abstract base for embedding model setups

  New Resource Types:
  - `ResourceType.EMBEDDING_MODEL_CONNECTION`
  - `ResourceType.EMBEDDING_MODEL`

  New Decorators:
  - `@embedding_model_connection` - For declaring embedding model connections in agents
  - `@embedding_model_setup` - For declaring embedding model setups in agents

  New Concrete Implementations:
  - `OllamaEmbeddingModelConnection` - Ollama-specific connection implementation
  - `OllamaEmbeddingModelSetup` - Ollama-specific setup implementation

### Documentation

[Vector Store and Embedding Model Integration for Flink Agents](https://github.com/apache/flink-agents/discussions/143)

### DIFF
--- a/python/flink_agents/integrations/embedding_models/local/tests/start_ollama_server.sh
+++ b/python/flink_agents/integrations/embedding_models/local/tests/start_ollama_server.sh
@@ -26,7 +26,15 @@ if [[ $os == "Linux" ]]; then
   then
     exit $ret
   fi
-  ollama serve
+
+  if curl -f http://localhost:11434/api/tags >/dev/null 2>&1; then
+      echo "Ollama is already running"
+  else
+      echo "Starting Ollama server"
+      ollama serve &
+      sleep 10 # wait for ollama to start
+  fi
+
   ollama pull all-minilm:22m
   ollama run all-minilm:22m
 fi
--- a/python/flink_agents/integrations/chat_models/tests/start_ollama_server.sh
+++ b/python/flink_agents/integrations/chat_models/tests/start_ollama_server.sh
@@ -26,7 +26,15 @@ if [[ $os == "Linux" ]]; then
   then
     exit $ret
   fi
-  ollama serve
+
+  if curl -f http://localhost:11434/api/tags >/dev/null 2>&1; then
+      echo "Ollama is already running"
+  else
+      echo "Starting Ollama server"
+      ollama serve &
+      sleep 10 # wait for ollama to start
+  fi
+
   ollama pull qwen3:0.6b
   ollama run qwen3:0.6b
 fi


python/flink_agents/api/embedding_models/embedding_model.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    @classmethod
+    @override
+    def resource_type(cls) -> ResourceType:
+        """Return resource type of class."""
+        return ResourceType.EMBEDDING_MODEL_CONNECTION
+
+    @abstractmethod
+    def embed_query(self, text: str, **kwargs: Any) -> list[float]:
```
xintongsong: Minor: for simplicity
```suggestion
    def embed(self, text: str, **kwargs: Any) -> list[float]:
```

python/flink_agents/api/embedding_models/embedding_model.py
```diff
@@ -0,0 +1,107 @@
...
+    """Base abstract class for text embedding model setup.
+
+    Responsible for managing embedding model configurations, such as:
+    - Connection to embedding model service (connection_name)
+    - Model name (model_name)
+
+    Provides the basic embedding interface for generating embeddings from text inputs.
+    """
+
+    connection_name: str = Field(description="Name of the referenced connection.")
```
xintongsong: Minor:
```suggestion
    connection: str = Field(description="Name of the referenced connection.")
```

python/flink_agents/api/embedding_models/embedding_model.py
```diff
@@ -0,0 +1,107 @@
...
+    model_name: str = Field(description="Name of the embedding model to use.")
```
xintongsong: Minor:
```suggestion
    model: str = Field(description="Name of the embedding model to use.")
```

xintongsong: It seems we have multiple ways to specify which model to use.
- For ollama and tongyi chat model, it's specified in the connection.
- For anthropic and openai chat model, it's specified in the setup.
- For both chat model connection and setup, `model` is a implementation specific parameter, rather than commonly defined in the abstraction.
- For this embedding model abstraction, it's defined as part of the setup abstraction.

I think we'd better have a unified way, so that users don't get confused. IMO, having it in the setup abstraction (like in this PR) might make more sense. @wenjin272, WDYT?

python/flink_agents/api/decorators.py
```diff
@@ -87,6 +87,41 @@ def chat_model_setup(func: Callable) -> Callable:
...
+
+    Returns:
+    -------
+    Callable
+        Decorator function that marks the target function declare an embedding model.
+    """
+    func._is_embedding_model_setup = True
+    return func
+
+
```
xintongsong, line 124: In order for the decorated functions to be accessed as embed model connections and setups, we also need to convert the functions into resource providers and put them into the agent plan. See `AgentPlan._get_resource_providers()`.

python/flink_agents/integrations/embedding_models/local/ollama_embedding_model.py
```diff
@@ -0,0 +1,165 @@
...
+    ----------
+    base_url : str
+        Base url the Ollama server is hosted under.
+    model : str
+        Embedding model name to use.
+    request_timeout : float
+        The timeout for making http request to Ollama API server.
+    """
+
+    model: str = Field(description="Embedding model name to use.")
```
xintongsong: This also confuses me. We have a `model` in the `OllamaEmbeddingModelConnection`, and also a `model_name` in `BaseEmbeddingModelSetup`. What is the relationship between these two?

python/flink_agents/integrations/embedding_models/local/tests/start_ollama_server.sh
```diff
@@ -0,0 +1,32 @@
...
+  curl -fsSL https://ollama.com/install.sh | sh
+  ret=$?
+  if [ "$ret" != "0" ]
+  then
+    exit $ret
+  fi
+  ollama serve
+  ollama pull all-minilm:22m
+  ollama run all-minilm:22m
+fi
```
xintongsong, line 40: There's a ci failure related to this.

We have 2 test cases, for chat and embed models respectively, that rely on the ollama server. We start the ollama server with this script, but never shut it down. So whichever test case is executed later, the script will fail at `ollama serve` because the port is already taken. And we did not check whether the script fail in the test case. So the test case continues, and connect to the server started by the previous test case, where the required model is not running.

xintongsong: I see several improvements needed for these test cases

1. We might want to reuse the script, passing in the required models as arguments.
2. In the script, we may check whether the server is already started. If true, we can just start the required model on the server, if not already running.
3. We should check whether the script fails in the test cases.
4. We'd better shutdown the server at some point. We may rely on GHA to clear the environment, but when executed locally this will leave a running progress.

cc @wenjin272

python/flink_agents/api/embedding_models/embedding_model.py
alnzng: This `embed_query` naming was influenced by the LangChain's implementation which separate `query` and `documents` embeddings: https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/embeddings/embeddings.py
```
    @abstractmethod
    def embed_documents(self, texts: list[str]) -> list[list[float]]:
        """Embed search docs.

        Args:
            texts: List of text to embed.

        Returns:
            List of embeddings.
        """

    @abstractmethod
    def embed_query(self, text: str) -> list[float]:
        """Embed query text.

        Args:
            text: Text to embed.

        Returns:
            Embedding.
        """
```

But I don't see any issue that if we want to use a generic method for handing both `query` and `documents` embeddings, especially I confirmed that `ollama` and `openai` APIs did follow this generic method implementation.

So let me change it to `embed`.

python/flink_agents/api/embedding_models/embedding_model.py
alnzng: +1 on this, and below was my thoughts on the pros that providing model in `setup` for either Chat or Embedding Models:

 Logical Separation:
  ```
  # Connection = "HOW to connect" 
  ChatModelConnection(api_key="...", base_url="...")

  # Setup = "WHAT to use" + "HOW to configure"
  ChatModelSetup(connection="conn", model="gpt-4", temperature=0.7)
```
Resource Reuse:
```
  # One connection, multiple model configs
  openai_conn = OpenAIConnection(api_key="...")

  gpt4_setup = OpenAISetup(connection="openai_conn", model="gpt-4")
  gpt35_setup = OpenAISetup(connection="openai_conn", model="gpt-3.5-turbo")
```

File: python/flink_agents/integrations/embedding_models/local/ollama_embedding_model.py (hunk 5 above)
Comment (alnzng) [reply python/flink_agents/integrations/embedding_models/local/ollama_embedding_model.py]: Good catch! I should remove this `model` and we should only reply on the `model` give in `setup`. (I copied the code from ollama chat model, but didn't do clean up properly). 

I shared my thought above why I think we'd better define `model` in `setup`:  from either Logical Separation and  Resource Reuse perspectives.

File: python/flink_agents/integrations/embedding_models/local/tests/start_ollama_server.sh (hunk 6 above)
Comment (alnzng) [reply python/flink_agents/integrations/embedding_models/local/tests/start_ollama_server.sh:40]: Hmm, I haven't took a look this failure. I thought this might be caused by some env issues, because I saw the tests were executed successfully on different Python versions.

Screenshot 2025-09-10 at 11 03 57 PM

I feel your theory is reasonable and let me verify it. Agreed with your suggestions on the improvements, let me think more about it.

File: python/flink_agents/api/embedding_models/embedding_model.py (hunk 3 above)
Comment (wenjin272) [reply python/flink_agents/api/embedding_models/embedding_model.py]: +1 on having model name in the setup abstraction. I will refactor the ChatModel api and implementation to align the behavior. #159

File: python/flink_agents/api/decorators.py [hunk 7]
```diff
@@ -87,6 +87,41 @@ def chat_model_setup(func: Callable) -> Callable:
     return func
 
 
+def embedding_model_connection(func: Callable) -> Callable:
```
wenjin272, line 90: Should also add `add_embedding_model_setup` and `add_embedding_model_connection` in `AgentsExecutionEnvironment` and `Agent`.

alnzng: Yep! I will add this missing part, thanks!

python/flink_agents/integrations/embedding_models/local/tests/start_ollama_server.sh
alnzng: @xintongsong to unblock this PR, I added the logic to detect if the ollama service is up to avoid ` bind: address already in use` conflict. I created a issue for improving it in long term: https://github.com/apache/flink-agents/issues/160

Can you take a look again and see if the change makes sense to you?

xintongsong: Sure. Improving the tests in a separate follow-up issue makes sense.

And the failure only happens in 3.9 because `test_ollama_chat_model` is only executed in 3.9. We did this because having 6 python testing jobs trying to download the model at the same time makes the CI unstable. But admittedly this is confusing and misleading.

@wenjin272, I think we should separate all the heavy integration tests into a dedicated job.

PR #158, opened 2025-09-11
[python][examples] Add quickstart workflow examples

Sxnan:
### Purpose of change

This PR introduces a comprehensive quickstart workflow for Flink Agents with practical examples demonstrating how to integrate LLM-powered agents with Flink streaming pipelines. The changes include:

1. **Two complete example workflows:**
   - `product_review_analysis.py`: A simple streaming pipeline that analyzes product reviews using an LLM agent to extract review scores and unsatisfied reasons
   - `product_improve_suggestion.py`: A more complex multi-stage pipeline that aggregates review analysis results and generates product improvement suggestions

2. **Agent implementations:**
   - `ReviewAnalysisAgent`: Analyzes individual product reviews
   - `ProductSuggestionAgent`: Generates improvement suggestions based on aggregated review data

3. **Sample data:** Product review data in JSON format for testing the examples

These examples serve as a practical entry point for users to understand how to build streaming applications that leverage LLM agents for real-time data processing and analysis.

### API

This change does not modify any public APIs. It only adds example code that demonstrates usage of existing Flink Agents APIs

### DIFF
--- a/python/flink_agents/examples/quickstart/agents/product_suggestion_agent.py
+++ b/python/flink_agents/examples/quickstart/agents/product_suggestion_agent.py
@@ -24,12 +24,10 @@ from pydantic import BaseModel
 from flink_agents.api.agent import Agent
 from flink_agents.api.chat_message import ChatMessage, MessageRole
 from flink_agents.api.chat_models.chat_model import (
-    BaseChatModelConnection,
     BaseChatModelSetup,
 )
 from flink_agents.api.decorators import (
     action,
-    chat_model_connection,
     chat_model_setup,
     prompt,
 )
@@ -38,7 +36,6 @@ from flink_agents.api.events.event import InputEvent, OutputEvent
 from flink_agents.api.prompts.prompt import Prompt
 from flink_agents.api.runner_context import RunnerContext
 from flink_agents.integrations.chat_models.ollama_chat_model import (
-    OllamaChatModelConnection,
     OllamaChatModelSetup,
 )
 
@@ -112,16 +109,6 @@ class ProductSuggestionAgent(Agent):
         {input}
         """
         return Prompt.from_text("generate_suggestion_prompt", prompt_str)
-
-    @chat_model_connection
-    @staticmethod
-    def ollama_server() -> Tuple[Type[BaseChatModelConnection], Dict[str, Any]]:
-        """ChatModelServer responsible for model service connection."""
-        return OllamaChatModelConnection, {
-            "name": "ollama_server",
-            "model": "qwen3:8b",
-            "request_timeout": 120,
-        }
 
     @chat_model_setup
     @staticmethod
--- a/python/flink_agents/examples/quickstart/agents/review_analysis_agent.py
+++ b/python/flink_agents/examples/quickstart/agents/review_analysis_agent.py
@@ -24,12 +24,10 @@ from pydantic import BaseModel
 from flink_agents.api.agent import Agent
 from flink_agents.api.chat_message import ChatMessage, MessageRole
 from flink_agents.api.chat_models.chat_model import (
-    BaseChatModelConnection,
     BaseChatModelSetup,
 )
 from flink_agents.api.decorators import (
     action,
-    chat_model_connection,
     chat_model_setup,
     prompt,
 )
@@ -38,7 +36,6 @@ from flink_agents.api.events.event import InputEvent, OutputEvent
 from flink_agents.api.prompts.prompt import Prompt
 from flink_agents.api.runner_context import RunnerContext
 from flink_agents.integrations.chat_models.ollama_chat_model import (
-    OllamaChatModelConnection,
     OllamaChatModelSetup,
 )
 
@@ -107,20 +104,10 @@ class ReviewAnalysisAgent(Agent):
        ]
     }}
 
-    input: 
+    input:
     {input}
     """
         return Prompt.from_text("review_analysis_prompt", prompt_str)
-
-    @chat_model_connection
-    @staticmethod
-    def ollama_server() -> Tuple[Type[BaseChatModelConnection], Dict[str, Any]]:
-        """ChatModelServer responsible for model service connection."""
-        return OllamaChatModelConnection, {
-            "name": "ollama_server",
-            "model": "qwen3:8b",
-            "request_timeout": 120,
-        }
 
     @chat_model_setup
     @staticmethod
@@ -162,7 +149,7 @@ class ReviewAnalysisAgent(Agent):
                     )
                 )
             )
-        except Exception as e:
+        except Exception:
             logging.exception(
                 f"Error processing chat response {event.response.content}"
             )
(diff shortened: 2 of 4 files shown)


python/flink_agents/examples/quickstart/resources/product_review.txt
```diff
@@ -0,0 +1,34 @@
+{"id":1,"review":"Great product! Works perfectly and lasts a long time."}
```
xintongsong: IIUC, the data here is a simplified and selected version from the [Amazon Review Data](https://nijianmo.github.io/amazon/index.html). While the data set does not come with any license, it does require citing the original paper. We may either add a header in the file and skip it when reading from the agent, or add a notice file in the directory about this.

python/flink_agents/examples/quickstart/product_review_analysis.py
```diff
@@ -0,0 +1,96 @@
...
+    # Add required flink-agents jars to the environment.
+    env.add_jars(
+        f"file:///{current_dir}/../../../../runtime/target/flink-agents-runtime-0.1-SNAPSHOT.jar"
+    )
+    env.add_jars(
+        f"file:///{current_dir}/../../../../plan/target/flink-agents-plan-0.1-SNAPSHOT.jar"
+    )
+    env.add_jars(
+        f"file:///{current_dir}/../../../../api/target/flink-agents-api-0.1-SNAPSHOT.jar"
+    )
```
xintongsong, line 67: We should not require manually adding the framework jars in user code. Ideally, we should have everything needed by flink-agents as a uber jar, and let user place it into the flink `lib/` directory.

python/flink_agents/examples/quickstart/product_review_analysis.py
```diff
@@ -0,0 +1,96 @@
...
+        watermark_strategy=WatermarkStrategy.no_watermarks(),
+        source_name="streaming_agent_example",
+    ).map(
+        lambda x: ProductReview.model_validate_json(x)  # Deserialize JSON to ProductReview.
+    )
+
+    # Use the ReviewAnalysisAgent to analyze each product review.
+    review_analysis_res_stream = (
+        agents_env.from_datastream(
+            input=product_review_stream, key_selector=MyKeySelector()
```
xintongsong: Is there any way we can set a key selector without defining a custom class? Like a java lambda function?

python/flink_agents/examples/quickstart/agents/review_analysis_agent.py
```diff
@@ -0,0 +1,167 @@
...
+                        id=ctx.get_short_term_memory().get("id"),
+                        score=json_content["score"],
+                        reasons=json_content["reasons"],
+                    )
+                )
+            )
+        except Exception as e:
+            logging.exception(
+                f"Error processing chat response {event.response.content}", exc_info=e
+            )
```
xintongsong: How can a user fail the agent?

python/flink_agents/examples/quickstart/product_improve_suggestion.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+    # Aggregate the analysis results in 1-minute tumbling windows.
+    # This produces a score distribution and collects all unsatisfied reasons for each
+    # product.
+    aggregated_analysis_res_stream = (
+        review_analysis_res_stream.key_by(lambda x: x.id)
+        .window(TumblingProcessingTimeWindows.of(Time.minutes(1)))
+        .process(AggregateScoreDistributionAndDislikeReasons())
+    )
```
xintongsong, line 146: I wonder if we can convert the output of ReviewAnalysisAgent into a table, and apply some event-time windowing. The purpose is to show how to integrate the agent with not only datastream but also table api, and event-time is more frequently used by flink users.

python/flink_agents/examples/quickstart/product_review_analysis.py
```diff
@@ -0,0 +1,96 @@
...
+
+    This example demonstrates how to use Flink Agents to analyze product reviews in a
+    streaming pipeline. The pipeline reads product reviews from a file, deserializes
+    each review, and uses an LLM agent to extract review scores and unsatisfied reasons.
+    The results are printed to stdout. This serves as a minimal, end-to-end example of
+    integrating LLM-powered agents with Flink streaming jobs.
+    """
+    # Set up the Flink streaming environment and the Agents execution environment.
+    env = StreamExecutionEnvironment.get_execution_environment()
+    agents_env = AgentsExecutionEnvironment.get_execution_environment(env=env)
```
xintongsong: Minor: does this work?
```suggestion
    agents_env = AgentsExecutionEnvironment.get_execution_environment(env)
```

python/flink_agents/examples/quickstart/agents/product_suggestion_agent.py
```diff
@@ -0,0 +1,177 @@
...
+            "prompt": "generate_suggestion_prompt",
+            "extract_reasoning": True,
+        }
+
+    @action(InputEvent)
+    @staticmethod
+    def process_input(event: InputEvent, ctx: RunnerContext) -> None:
+        """Process input event."""
+        input: ProductReviewSummary = event.input
+        ctx.get_short_term_memory().set("id", input.id)
```
xintongsong, line 129: Minor: Would this be better?
```suggestion
        ctx.short_term_memory().set("id", input.id)
```

python/flink_agents/examples/quickstart/agents/product_suggestion_agent.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @chat_model_connection
+    @staticmethod
+    def ollama_server() -> Tuple[Type[BaseChatModelConnection], Dict[str, Any]]:
+        """ChatModelServer responsible for model service connection."""
+        return OllamaChatModelConnection, {
+            "name": "ollama_server",
+            "model": "qwen2.5:7b",
+            "request_timeout": 120,
+        }
```
xintongsong: Is this identical to the model connection used in the other agent? We might show how to reuse resources with this example.

python/flink_agents/examples/quickstart/agents/product_suggestion_agent.py
Sxnan: This requires a change to the RunnerContext. 

Issue is created https://github.com/apache/flink-agents/issues/175

python/flink_agents/examples/quickstart/product_improve_suggestion.py
Sxnan: I found that the ActionExecutionOperator cannot properly work with event time at the moment. I created an issue for it. https://github.com/apache/flink-agents/issues/178

The aggregation of the review analysis results is not trivial. We may do that in a follow-up PR. I created a follow-up issue. https://github.com/apache/flink-agents/issues/179

WDYT?

Sxnan:
@xintongsong Thanks for the review! All the comments are addressed. Please take another look.

Issue #157, opened 2025-09-11
[Feature] [integrations][java] Add support for OllamaChatModel


twosom:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Summary
Add support for Ollama as a chat model provider in the `integrations/chat-models` module.

## Related Issue
#156 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Issue #156, opened 2025-09-11
[Feature] Introduce new module `integrations`


twosom:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Summary
Add a new `integrations` module to Apache Flink Agents for external service integrations.

## Motivation
- Centralize integration logic for external services
- Keep core agent logic separate from integration-specific code
- Enable easy addition of new integrations (chat models, vector stores, etc.)

## Proposed Structure
```
flink-agents/
├── integrations/
│   ├── chat-models/          # LLM providers (OpenAI, Anthropic, etc.)
│   ├── vector-stores/        # Vector databases (maybe future)
│   └── data-connectors/      # Specialized connectors (maybe future)
```

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #162, opened 2025-09-11
[Feature][Integrations][java] Integration with OllamaChatModel

twosom:
Linked issue: #156, #157 

### Purpose of change

This PR introduces support for Ollama as a chat model provider in the `integrations/chat-models` module, enabling local LLM inference capabilities for Flink Agents.

### DIFF
--- a/integrations/chat-models/ollama/src/main/java/org/apache/flink/agents/integrations/chatmodels/ollama/OllamaChatModel.java
+++ b/integrations/chat-models/ollama/src/main/java/org/apache/flink/agents/integrations/chatmodels/ollama/OllamaChatModel.java
@@ -44,32 +44,13 @@ import java.util.stream.Collectors;
  * A chat model integration for Ollama powered by the ollama4j client.
  *
  * <p>This implementation adapts the generic Flink Agents chat model interface to Ollama's
- * conversation API. It supports: - Converting framework ChatMessage instances into Ollama message
- * format. - Optional tool/function registration using JSON schema metadata provided by BaseTool. -
- * Synchronous chat completion returning an assistant message as the result.
+ * conversation API.
  *
- * <p>See also BaseChatModel for the common resource abstractions and lifecycle.
+ * <p>See also {@link BaseChatModel} for the common resource abstractions and lifecycle.
  *
- * <p>Example usage (pattern taken from AgentWithOllamaExample):
+ * <p>Example usage:
  *
  * <pre>{@code
- * import org.apache.flink.agents.api.Agent;
- * import org.apache.flink.agents.api.InputEvent;
- * import org.apache.flink.agents.api.OutputEvent;
- * import org.apache.flink.agents.api.annotation.Action;
- * import org.apache.flink.agents.api.annotation.ChatModel;
- * import org.apache.flink.agents.api.chat.messages.ChatMessage;
- * import org.apache.flink.agents.api.chat.messages.MessageRole;
- * import org.apache.flink.agents.api.chat.model.BaseChatModel;
- * import org.apache.flink.agents.api.context.RunnerContext;
- * import org.apache.flink.agents.api.resource.ResourceType;
- * import org.apache.flink.agents.integrations.chatmodels.ollama.OllamaChatModel;
- *
- * import java.util.Collections;
- * import java.util.HashMap;
- * import java.util.List;
- * import java.util.Map;
- *
  * public class MyAgent extends Agent {
  *   // Register the chat model via @ChatModel metadata.
  *   @ChatModel
@@ -86,17 +67,6 @@ import java.util.stream.Collectors;
  *         List.of(String.class.getName(), String.class.getName(),
  *                String.class.getName(), List.class.getName()));
  *     return meta;
- *   }
- *
- *   // Use the chat model inside an action.
- *   @Action(listenEvents = {InputEvent.class})
- *   public static void process(InputEvent event, RunnerContext ctx) throws Exception {
- *     // IMPORTANT: use the name of the @ChatModel-annotated method ("ollama") as the resource name
- *     BaseChatModel chatModel =
- *         (BaseChatModel) ctx.getResource("ollama", ResourceType.CHAT_MODEL);
- *     ChatMessage response = chatModel.chat(
- *         Collections.singletonList(new ChatMessage(MessageRole.USER, (String) event.getInput())));
- *     ctx.sendEvent(new OutputEvent(response.getContent()));
  *   }
  * }
  * }</pre>


integrations/chat-models/ollama/src/main/java/org/apache/flink/agents/integrations/chatmodels/ollama/OllamaChatModel.java
```diff
@@ -0,0 +1,246 @@
...
+        try {
+            final List<OllamaChatMessage> ollamaChatMessages =
+                    messages.stream()
+                            .map(this::convertToOllamaChatMessages)
+                            .collect(Collectors.toList());
+
+            final OllamaChatResult ollamaChatResult =
+                    this.client.chat(this.modelName, ollamaChatMessages);
+
+            return ChatMessage.assistant(ollamaChatResult.getResponse());
```
wenjin272, line 211: Here we only extract the response content, but when response contains tool calls, we should also extract them to our ChatMessage.

twosom: > Here we only extract the response content, but when response contains tool calls, we should also extract them to our ChatMessage.

Thank you @wenjin272 

I'll quickly fix the part that returns ChatMessage and upload it.
Thank you.

integrations/chat-models/ollama/src/main/java/org/apache/flink/agents/integrations/chatmodels/ollama/OllamaChatModel.java
```diff
@@ -0,0 +1,246 @@
...
+ *     // IMPORTANT: use the name of the @ChatModel-annotated method ("ollama") as the resource name
+ *     BaseChatModel chatModel =
+ *         (BaseChatModel) ctx.getResource("ollama", ResourceType.CHAT_MODEL);
+ *     ChatMessage response = chatModel.chat(
+ *         Collections.singletonList(new ChatMessage(MessageRole.USER, (String) event.getInput())));
+ *     ctx.sendEvent(new OutputEvent(response.getContent()));
+ *   }
+ * }
+ * }</pre>
+ */
```
xintongsong, line 73: The example might be a bit too long. I think we don't need the imports, as well as the action which is no different from using any other chat model. Just showcase how to define the model should be good enough.

twosom: @xintongsong 
Hello. As you requested, I've simplified the document.
Thank you.

twosom:
> > > Hi, @twosom , thanks for your contribution. I left a minor comment about converting ollama chat response to our chat message.
> > > Currently, because there is no built-in actions in java like `chat_model_action` and `tool_call_action` in python, the example actually doesn't call the function tool. I will add these built-in actions to keep align with python side later.
> > 
> > 
> > Hi @wenjin272. Thanks to Ollama4j's `registerTool` function, It's actually calling the function.
> > Therefore, the example works correctly.
> > image
> 
> Ok, I see the tool calls are actually executed internally in Ollama4j after we register the tools to the client.
> 
> ```
> public OllamaChatResult chatStreaming(OllamaChatRequest request, OllamaTokenHandler tokenHandler) throws OllamaBaseException, IOException, InterruptedException, ToolInvocationException {
>        OllamaChatEndpointCaller requestCaller = new OllamaChatEndpointCaller(this.host, this.auth, this.requestTimeoutSeconds, this.verbose);
>        request.setTools((List)this.toolRegistry.getRegisteredSpecs().stream().map(Tools.ToolSpecification::getToolPrompt).collect(Collectors.toList()));
>        OllamaChatResult result;
>        if (tokenHandler != null) {
>            request.setStream(true);
>            result = requestCaller.call(request, tokenHandler);
>        } else {
>            result = requestCaller.callSync(request);
>        }
> 
>        List<OllamaChatToolCalls> toolCalls = result.getResponseModel().getMessage().getToolCalls();
> 
>        for(int toolCallTries = 0; toolCalls != null && !toolCalls.isEmpty() && toolCallTries < this.maxChatToolCallRetries; ++toolCallTries) {
>            for(OllamaChatToolCalls toolCall : toolCalls) {
>                String toolName = toolCall.getFunction().getName();
>                ToolFunction toolFunction = this.toolRegistry.getToolFunction(toolName);
>                if (toolFunction == null) {
>                    throw new ToolInvocationException("Tool function not found: " + toolName);
>                }
> 
>                Map<String, Object> arguments = toolCall.getFunction().getArguments();
>                Object res = toolFunction.apply(arguments);
>                String argumentKeys = (String)arguments.keySet().stream().map(Object::toString).collect(Collectors.joining(", "));
>                request.getMessages().add(new OllamaChatMessage(OllamaChatMessageRole.TOOL, "[TOOL_RESULTS] " + toolName + "(" + argumentKeys + "): " + String.valueOf(res) + " [/TOOL_RESULTS]"));
>            }
> 
>            if (tokenHandler != null) {
>                result = requestCaller.call(request, tokenHandler);
>            } else {
>                result = requestCaller.callSync(request);
>            }
> 
>            toolCalls = result.getResponseModel().getMessage().getToolCalls();
>        }
> 
>        return result;
>    }
> ```
> 
> This indeed reduce the complexity of API user, but maybe this behavior is not what we want.
> 
> In our design, flink-agents framework will manage the tools and related context(store in short-memory). Chat request & Tool call & Tool response will all generate `Event`. Based on this, flink-agents will provide Event log for improving observability. And the per-action consistency may also depend on this.
> 
> You can take a look at `process_chat_request_or_tool_response` and `process_tool_request` in Python. And I will add correspond implementation later in Java side.
> 
> If the tool calls are hidden within API calls, I'm worried this might break some of our assumptions.

@wenjin272 

Hello. Thank you for your review.
As you mentioned, the ollama client returns a list of functions to execute from the chat message, and these must be executed directly by the flink-agents. (At least, that's how it appears in the Python code I analyzed.)

To achieve this, Java implementations of `ToolRequestEvent` and `ToolResponseEvent` are required.

Additionally, to proceed identically to Python, Java implementations of `ChatRequestEvent` and `ChatResponseEvent` are also needed. (At least to my knowledge.)
However, this seems to fall outside the scope of this PR.

Therefore, my suggestion is to first have the client make the call and then proceed with the `ToolRequestEvent` and `ToolResponseEvent` work as the next task. (Including this work in this PR would make the scope too broad.)
Of course, there is also the option not to proceed with this PR. 
I would appreciate it if you could choose whichever is more convenient.

Best regards

wenjin272:
Hi, @twosom, thanks for your reply.

I totally agree your analysis that flink-agents need Java implementation of  `ToolRequestEvent `, `ToolResponseEvent `, `ChatRequestEvent ` and `ChatResponseEvent ` and two built-in Actions  `process_chat_request_or_tool_response` and  `process_tool_request`  to proceed identically to Python. And these are outside the scope of this PR.

In this pr, I think what we should change is only directly return the response contains tool calls in `chat` of `OllamaChatModel` rather than process tool calls in Ollama4j API internally.

After we apply this change, the `AgentWithOllamaExample` actually will not work. I will submit another PR to implement the built-in events and actions mentioned above these days, and then this example should work as it does now. And I believe this will not block the merging of this PR.

Issue #163, opened 2025-09-12
[Feature][api][java] Separate the abstraction of ChatModel to connection and setup in Java.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Separate the abstraction of ChatModel to connection and setup in Java to keep align with Python API.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #164, opened 2025-09-12
[api][java] Separate the abstraction of ChatModel to connection and setup in Java.

wenjin272:
Linked issue: #163 

### Purpose of change

Separate the abstraction of ChatModel to connection and setup in Java.

### Tests

ut

### API

Yes, remove `BaseChatModel`, and add `BaseChatModelConnection`, `BaseChatModelSetup`

### Documentation

No

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelConnection.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelConnection.java
@@ -20,49 +20,40 @@ package org.apache.flink.agents.api.chat.model;
 
 import org.apache.flink.agents.api.chat.messages.ChatMessage;
 import org.apache.flink.agents.api.resource.Resource;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/apache/flink/agents/api/resource/Resource.java
```diff
@@ -18,12 +18,23 @@
...
  *
  * <p>Resources are components that can be used by agents during action execution.
  */
 public abstract class Resource {
+    protected BiFunction<String, ResourceType, Resource> getResource;
+
+    protected Resource(
+            BiFunction<String, ResourceType, Resource> getResource, ResourceDescriptor descriptor) {
+        this.getResource = getResource;
+    }
```
xintongsong: Minor:
```suggestion
    protected Resource(
            ResourceDescriptor descriptor, BiFunction<String, ResourceType, Resource> getResource) {
        this.getResource = getResource;
    }
```

xintongsong: And same for the model connection and setup.

Issue #130, opened 2025-08-27
[Feature] support VectorStore


alnzng:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

support VectorStore

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #165, opened 2025-09-15
[api][python] support VectorStore and provide a built-in Chroma implementation

alnzng:
Linked issue: #130

### Purpose of change

This PR adds vector store support to Flink Agents by defining base abstractions for vector store connections and setups, along with a complete Chroma vector store implementation which provides both local(in-memory) mode and server mode.

### Tests

Unit tests

### API

 Core Vector Store Abstractions:
  - BaseVectorStoreConnection - Abstract base for vector store connections with raw embedding search
  - BaseVectorStoreSetup - Abstract base coordinating embedding models with vector stores
  - VectorStoreQuery - Structured query object for text-based semantic search
  - VectorStoreQueryResult - Standardized result containing retrieved documents
  - Document - Document model with content, metadata, and optional ID
  - VectorStoreQueryMode - Enum for query types (currently SEMANTIC)

  ChromaDB Implementation:
  - ChromaVectorStoreConnection - Chroma-specific connection implementation
  - ChromaVectorStoreSetup - Chroma-specific setup implementation

### Documentation

https://github.com/apache/flink-agents/discussions/143


alnzng:
@xintongsong @wenjin272 I would like to know your thoughts if we can disable Python 3.9. 

I'm thinking that we might have to disable Python 3.9 to unblock some new features(include this PR) based on my observation: 

- For this PR, it brings a new dependency `chroma` which requires Python `>= 3.10`. 
- I observed this PR https://github.com/apache/flink-agents/pull/107 tries to bring a new `mcp` dependency which requires Python `>= 3.10` as well: https://github.com/apache/flink-agents/pull/107/files#diff-1a45bd938160f7761769e7b6fbc819d1cc7f84ce37f1a6a23ef83cc4e214abc8R32

It might be a common requirements on Python `>= 3.10` once we add more dependencies to offer built-in implementation.

python/flink_agents/api/resource.py
```diff
@@ -25,16 +25,18 @@
...
+    vector_store and prompt.
     """
 
     CHAT_MODEL = "chat_model"
     CHAT_MODEL_CONNECTION = "chat_model_connection"
     TOOL = "tool"
     EMBEDDING_MODEL = "embedding_model"
     EMBEDDING_MODEL_CONNECTION = "embedding_model_connection"
+    VECTOR_STORE = "vector_store"
+    VECTOR_STORE_CONNECTION = "vector_store_connection"
```
wenjin272, line 38: After adding resource type, there are some things to do to enable use the resource in Agent.

1. Add correspond decorator @vector_store_setup and @vector_store_connection
2. Insert `add_vector_store_setup` and `add_vector_store_connection` in `AgentsExecutionEnvironment` and `Agent`
3. Support retrieve the resource in `AgentPlan`

Additionally, to enable using it in a Agent in flink job, should also add the resource type in java `ResourceType`, for `ResourceProviderJsonDeserializer` depends it.

wenjin272 (commented):
Hi, @alnzng, it's really a great work. The vector store api and implementation look good to me. And I left some comments about how to use vector store in Agent.

About the python version, I will disable python 3.9 and reformat code in #168.

python/flink_agents/api/resource.py
alnzng: Thank you @wenjin272 ! You are totally correct that I should add these kind of changes in this PR. Let me add these missing pieces. 
(you reminded me a similar thing while I was implementing embedding model support, and I forgot it 😞 )

>Additionally, to enable using it in a Agent in flink job, should also add the resource type in java ResourceType, for ResourceProviderJsonDeserializer depends it.

Let me take a look this part, didn't know this gap. Thanks for your suggestion!

alnzng: Fixed these gaps, thanks!

Issue #111, opened 2025-08-13
[Feature] Support MCP integration


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support leveraging tools and prompts from mcp servers.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

yanand0909:
Hi @lihaosky @xintongsong, I can work on this issue

PR #168, opened 2025-09-15
[api][python] Add built-in mcp support

wenjin272:
Linked issue: #111 

### Purpose of change

Add built-in mcp support in Python

### Tests

No, will add ut test later.

### API

Yes, add `MCPPrompt`, `MCPTool` and `MCPServer`

### Documentation

No

### DIFF
--- a/python/flink_agents/api/resource.py
+++ b/python/flink_agents/api/resource.py
@@ -35,7 +35,7 @@ class ResourceType(Enum):
     EMBEDDING_MODEL_CONNECTION = "embedding_model_connection"
     PROMPT = "prompt"
     # VECTOR_STORE = "vector_store"
-    # MCP_SERVER = "mcp_server"
+    MCP_SERVER = "mcp_server"
 
 
 class Resource(BaseModel, ABC):
--- a/python/flink_agents/api/agent.py
+++ b/python/flink_agents/api/agent.py
@@ -29,6 +29,7 @@ from flink_agents.api.embedding_models.embedding_model import (
 from flink_agents.api.events.event import Event
 from flink_agents.api.prompts.prompt import Prompt
 from flink_agents.api.resource import ResourceType
+from flink_agents.api.tools.mcp import MCPServer
 
 
 class Agent(ABC):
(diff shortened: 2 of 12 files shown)


python/flink_agents/api/prompts/prompt.py
```diff
@@ -15,39 +15,57 @@
...
         """Get the resource type."""
         return ResourceType.PROMPT
 
+
+class LocalPrompt(Prompt):
+    """Prompt for a language model.
+
+    Attributes:
+    ----------
+    template : Union[Sequence[ChatMessage], str]
```
alnzng, line 63: nit: change to new syntax `Sequence[ChatMessage] | str`

.github/workflows/ci.yml
```diff
@@ -98,7 +98,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ 'macos-latest', 'ubuntu-latest' ]
-        python-version: [ '3.9', '3.10', '3.11' ]
+        python-version: [ '3.10', '3.11' ]
```
alnzng, line 101: Thank you for disabling Python `3.9`, this will help unblock my PR!

python/flink_agents/api/resource.py
```diff
@@ -35,7 +35,7 @@ class ResourceType(Enum):
     EMBEDDING_MODEL_CONNECTION = "embedding_model_connection"
     PROMPT = "prompt"
     # VECTOR_STORE = "vector_store"
-    # MCP_SERVER = "mcp_server"
+    MCP_SERVER = "mcp_server"
```
alnzng, line 38: Should we add this in `ResourceType.java`?

xintongsong: Not necessarily. Because we do not support mcp server integration in java yet, and we also don't have cross-language resource usage. So at this moment, users won't have change to use mcp server in java anyway.

xintongsong (approved):
LGTM. I'll merge this after the latest comments from Alan are addressed.

python/flink_agents/api/resource.py
wenjin272: Yes, we should add this in ResourceType.java. Thanks for the reminder.

@xintongsong , the reason we should add is when deserialize the AgentPlan json created from python in java side. It will reconstruct the resource type from string value
https://github.com/apache/flink-agents/blob/1e2098c88d7085e0d74072c684b141d3c0230330/plan/src/main/java/org/apache/flink/agents/plan/serializer/ResourceProviderJsonDeserializer.java#L104

But may be we must add comments in `ResourceType.java`, the MCPServer is only supported in python.

Issue #171, opened 2025-09-16
[Feature][api][java] Support interactive interface of Agent and AgentExecutionEnvironment.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

1. Support register resources in environment and use in agent.
2. Support add resource and add action of agent.
3. Support pass additional config of action to agent.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #172, opened 2025-09-16
[api][java] Support interactive interface of Agent and AgentExecutionEnvironment.

wenjin272:
Linked issue: #171 

### Purpose of change

1. Support register resources in environment and use in agent.
2. Support add resource and add action of agent.
3. Support pass additional config of action to agent.

### Tests

unit test

### API

Yes, add interactive interface of agent and environment.

### Documentation

No


api/src/main/java/org/apache/flink/agents/api/Agent.java
```diff
@@ -18,11 +18,133 @@
...
+     *
+     * @param events The event types this action listened.
+     * @param method The method of this action, should be static method.
+     * @param config The optional config can be used by this action.
+     */
+    public Agent addAction(
+            Class<? extends Event>[] events,
+            Class<?> clazz,
+            Method method,
+            @Nullable Map<String, Object> config) {
```
xintongsong: 1. `clazz` is not in JavaDoc
2. We should not need `clazz`, as it can be fetched from `method.getDeclaringClass()`

api/src/main/java/org/apache/flink/agents/api/Agent.java
```diff
@@ -18,11 +18,133 @@
...
+        if (actions.containsKey(name)) {
+            throw new IllegalArgumentException(String.format("Action %s already defined.", name));
+        }
+        actions.put(name, new Tuple4<>(events, clazz, method, config));
+        return this;
+    }
+
+    public void mergeResource(Map<ResourceType, Map<String, Object>> resources) {
+        resources.forEach(this.resources::putIfAbsent);
+    }
```
xintongsong:
```suggestion
    public void addResourcesIfAbsent(Map<ResourceType, Map<String, Object>> resources) {
        resources.forEach(this.resources::putIfAbsent);
    }
```

api/src/main/java/org/apache/flink/agents/api/Agent.java
xintongsong: Maybe adding another `addAction` without config?

Issue #153, opened 2025-09-10
[Feature] Provide Kafka-based implementation of ActionState


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Provide Kafka-based implementation of ActionState

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #181, opened 2025-09-17
[Feature] Implemenet Kafka-based state store

letaoj:
Linked issue: #153 

### Purpose of change

The main changes in this commits are

* Introduce the Kafka-based state store backend for Flink Agent to use
* Introduce Seder for the ActionState
* Introduce Custom Partitioner used to partition the Kafka record

### Tests

All of them are covered by unit tests.


letaoj:
I will need to rebase after the API change is merged..

tools/lint.sh
```diff
@@ -67,7 +67,7 @@ install_python_deps() {
   if command -v uv >/dev/null 2>&1; then
     echo "Using uv for dependency management"
     pushd python
-    uv sync --extra lint
+    uv sync --extra lint --index-strategy unsafe-best-match
```
Sxnan: Why do we need this change?

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
```diff
@@ -27,4 +27,20 @@ public class AgentConfigOptions {
...
+    /** The config parameter specifies the Kafka bootstrap server. */
+    public static final ConfigOption<String> KAFKA_BOOTSTRAP_SERVERS =
+            new ConfigOption<>("kafkaBootstrapServers", String.class, "localhost:9092");
+
+    /** The config parameter specifies the Kafka topic for action state. */
+    public static final ConfigOption<String> KAFKA_ACTION_STATE_TOPIC =
+            new ConfigOption<>("kafkaActionStateTopic", String.class, "flink-agents-action-state");
+
+    /** The config parameter sepcifies the number of partitions for the Kafka action state topic. */
+    public static final ConfigOption<Integer> KAFKA_ACTION_STATE_TOPIC_NUM_PARTITIONS =
```
Sxnan, line 40: Why do we need to specify the number of partitions? Should we allow creating Kafka topic during job run?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateKafkaSeder.java
```diff
@@ -0,0 +1,110 @@
...
+
+        return mapper;
+    }
+
+    /** Mixin to add type information for Event hierarchy. */
+    @JsonTypeInfo(
+            use = JsonTypeInfo.Id.CLASS,
+            include = JsonTypeInfo.As.PROPERTY,
+            property = "@class")
+    public abstract static class EventTypeInfoMixin {}
```
Sxnan, line 113: Same as the one in `ActionStateKafkaSerializer`, can we reuse?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
```diff
@@ -33,38 +57,303 @@
...
-    public KafkaActionStateStore() {
-        this(new HashMap<>());
+    /** Constructs a new KafkaActionStateStore with custom Kafka configuration. */
+    public KafkaActionStateStore(AgentConfiguration agentConfiguration) {
+        this.actionStates = new HashMap<>();
+        this.latestKeySeqNum = new HashMap<>();
+        this.agentConfiguration = agentConfiguration;
+        this.topic = agentConfiguration.get(KAFKA_ACTION_STATE_TOPIC);
+        // create the topic if not exists
+        maybeCreateTopic();
```
Sxnan, line 119: Should we support creating the topic if it is not exist, or we require that the topic is created beforehand.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
```diff
@@ -27,4 +27,20 @@ public class AgentConfigOptions {
...
     public static final ConfigOption<String> ACTION_STATE_STORE_BACKEND =
             new ConfigOption<>("actionStateStoreBackend", String.class, null);
+
+    /** The config parameter specifies the Kafka bootstrap server. */
+    public static final ConfigOption<String> KAFKA_BOOTSTRAP_SERVERS =
+            new ConfigOption<>("kafkaBootstrapServers", String.class, "localhost:9092");
+
+    /** The config parameter specifies the Kafka topic for action state. */
+    public static final ConfigOption<String> KAFKA_ACTION_STATE_TOPIC =
+            new ConfigOption<>("kafkaActionStateTopic", String.class, "flink-agents-action-state");
```
Sxnan: I think it is better to use `null` as the default value and throw an exception if it is null. Otherwise, it is problematic when multiple jobs use the same default topic.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
```diff
@@ -33,38 +57,303 @@
...
+            LOG.error("Producer is null, cannot put action state to Kafka");
+            return;
+        }
+
+        String stateKey = generateKey(key, seqNum, action, event);
+        try {
+            ProducerRecord<String, ActionState> kafkaRecord =
+                    new ProducerRecord<>(topic, stateKey, state);
+            Future<RecordMetadata> recordFuture = producer.send(kafkaRecord);
+            RecordMetadata recordMetadata =
```
Sxnan: The `recordMetadata` is unused.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
```diff
@@ -33,38 +57,303 @@
...
     }
 
     @Override
-    public ActionState get(Object key, long seqNum, Action action, Event event) throws IOException {
-        // TODO: Implement me
-        return null;
+    public ActionState get(Object key, long seqNum, Action action, Event event) throws Exception {
+        String stateKey = generateKey(key, seqNum, action, event);
+
+        boolean hasDivergence = checkDivergence(key.toString(), seqNum);
```
Sxnan, line 151: Can you elaborate on what case we have divergence? Why do we need to check?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
```diff
@@ -24,7 +24,7 @@
 import java.util.List;
 
 /** Interface for storing and retrieving the state of actions performed by agents. */
-public interface ActionStateStore {
+public interface ActionStateStore extends AutoCloseable {
```
Sxnan, line 27: `close` method is not called. Should it be call in the `ActionExecutionOperator`?

Sxnan (commented):
Thanks for the PR. Please see my comments below.

tools/lint.sh
letaoj: :( This is to mitigate the issue where our internal pip repo does not have the necessary dependency thus i have to add this to the lint for the lint to work. I will revert it to keep the main branch clean.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
letaoj: Make sense. Updated.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
letaoj: We are auto-creating the state topic if it is not created thus I think it would be beneficial if we can let the user to specify the number of partition.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateKafkaSeder.java
letaoj: Sure, I combined the two class together. This class is a abstract class, i do not want to cross reference the class in two different classes...

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateStore.java
letaoj: Yes, thank you for catching this!

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
letaoj: maybeCreateTopic method will create the topic if the topic does not exist with default config.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
letaoj: Yep. Removed. I was using it to get the partition -> offset map but we are now getting the map in the `getRecoveryMarker` method

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
letaoj: This is what @xintongsong and I was discussing, think about a case like this

1 -> 2 (crash)
   -> 2' -> 3 -> 4 (crash)

When the second crash happened and we are trying to recover from the state and the input event comes with an order of `1, 2, 3, 4` (action 2 also generate action 3,4 in this case), we should not reuse the state of 3,4 because state of action 3, 4 was based on 2'. So to make 100% correct, we will ignore states after the diverge happens

Sxnan: Thanks for the explanation! LGTM

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -350,13 +358,12 @@ private void processActionTaskForKey(Object key) throws Exception {
         long sequenceNumber = sequenceNumberKState.value();
         boolean isFinished;
         List<Event> outputEvents;
-        Optional<ActionTask> generatedActionTaskOpt;
+        Optional<ActionTask> generatedActionTaskOpt = Optional.empty();
         ActionState actionState =
                 maybeGetActionState(key, sequenceNumber, actionTask.action, actionTask.event);
-        if (actionState != null && actionState.getGeneratedActionTask().isEmpty()) {
+        if (actionState != null) {
```
xintongsong, line 365: I assume this means we expect an incompleted action should not have a action state in the store? How is this guaranteed? Looks like we always call `maybePersistTaskResult` after `actionTask.invoke()`.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -190,9 +190,17 @@ public void open() throws Exception {
...
+            actionStateStore = new KafkaActionStateStore(agentPlan.getConfig());
         }
 
+        // init recovery marker state for recovery marker persistence
+        recoveryMarkerOpState =
+                getOperatorStateBackend()
+                        .getUnionListState(
+                                new ListStateDescriptor<>(
+                                        RECOVERY_MARKER_STATE_NAME,
+                                        TypeInformation.of(Object.class)));
```
xintongsong: Minor: We might not need this if `actionStateStore == null`

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
xintongsong: I think we need to either check that action task is finished before calling `maybePersistTaskResult()`, or skip the persisting for unfinished actions inside `maybePersistTaskResult()`.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: Updated

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
letaoj: > I assume this means we expect an incompleted action should not have a action state in the store?

Yes, this is because I realize it will not be easy to serializer the ActionTask when the async action is only completed halfway. I will add a check in `maybePersistTaskResult()` so that unfinished job will not be persisted.

xintongsong (approved):
LGTM

Issue #175, opened 2025-09-16
[Feature] Use Python property to simplify the getter in RunnerContext


Sxnan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Use Python property to simplify the getter in RunnerContext

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

GreatEugenius:
I'm willing to submit a PR!

PR #183, opened 2025-09-17
[Feature][API][python] Use Python property to simplify the getter in RunnerContext

GreatEugenius:
Linked issue: #175

### Purpose of change
Use Python property to simplify the getter in RunnerContext.


GreatEugenius:
@xuannan @wenjin272 please help take a look this PR, thanks.

wenjin272:
LGTM

python/flink_agents/api/runner_context.py
```diff
@@ -65,6 +65,11 @@ def get_action_config(self) -> Dict[str, Any]:
           The configuration of the action executed.
         """
 
+    @property
+    def action_config(self) -> Dict[str, Any]:
```
Sxnan: Instead of introducing new property methods, can we rename the current get method and make it both abstract and property method?

Sxnan (commented):
Can you also update the quickstart example to use the property method?

python/flink_agents/api/runner_context.py
GreatEugenius: Resolved.

Issue #178, opened 2025-09-16
[Feature] ActionExecutionOperator handle watermark properly


Sxnan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, the ActionExecutionOperator process records asynchronously. This is problematic in event time processing, since the operator doesn't wait for the async processing of the records before the watermark to finish before emitting the watermark downstream. In other words, a watermark can surpass the events.

We should have a similar watermark handling in the Flink async operator, see org.apache.flink.streaming.api.operators.async.AsyncWaitOperator#processWatermark

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #186, opened 2025-09-19
[Feature][Runtime] ActionExecutionOperator handle watermark properly

GreatEugenius:
Linked issue: #178 

### Purpose of change
Let the ActionExecutionOperator handle the watermark properly.

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/KeySegment.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/KeySegment.java
@@ -56,4 +56,9 @@ public class KeySegment {
     public boolean isEmpty() {
         return keyReferenceCounts.isEmpty();
     }
+
+    // Package-private getter for test access
+    Map<Object, Integer> getKeyReferenceCounts() {
+        return keyReferenceCounts;
+    }
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/SegmentedQueue.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/SegmentedQueue.java
@@ -95,4 +95,14 @@ public class SegmentedQueue {
     private boolean isFirstSegmentEmpty() {
         return !this.segments.isEmpty() && segments.getFirst().isEmpty();
     }
+
+    // Package-private getter for test access
+    Deque<KeySegment> getSegments() {
+        return segments;
+    }
+
+    // Package-private getter for test access
+    Deque<Watermark> getWatermarks() {
+        return watermarks;
+    }
 }
--- a/runtime/src/test/java/org/apache/flink/agents/runtime/operator/queue/KeySegmentTest.java
+++ b/runtime/src/test/java/org/apache/flink/agents/runtime/operator/queue/KeySegmentTest.java
@@ -19,8 +19,6 @@ package org.apache.flink.agents.runtime.operator.queue;
 
 import org.junit.jupiter.api.Test;
 
-import java.util.Map;
-
 import static org.junit.jupiter.api.Assertions.*;
 
 class KeySegmentTest {
@@ -45,7 +43,7 @@ class KeySegmentTest {
         keySegment.incrementKeyReference("key1");
 
         assertTrue(keySegment.hasActiveKey("key1"));
-        assertEquals(3, getKeyReferenceCount(keySegment, "key1"));
+        assertEquals(3, keySegment.getKeyReferenceCounts().get("key1"));
     }
 
     @Test
@@ -74,7 +72,7 @@ class KeySegmentTest {
         keySegment.decrementKeyReference("key1");
         keySegment.decrementKeyReference("key1");
         assertTrue(keySegment.hasActiveKey("key1"));
-        assertEquals(1, getKeyReferenceCount(keySegment, "key1"));
+        assertEquals(1, keySegment.getKeyReferenceCounts().get("key1"));
 
         // Decrement once more - should be removed
         keySegment.decrementKeyReference("key1");
@@ -135,20 +133,5 @@ class KeySegmentTest {
         keySegment.decrementKeyReference("key1");
         keySegment.decrementKeyReference("key3");
         assertTrue(keySegment.isEmpty());
-    }
-
-    // Helper method to access the private map for testing purposes
-    private int getKeyReferenceCount(KeySegment keySegment, Object key) {
-        // Using reflection to access the private field for testing
-        try {
-            java.lang.reflect.Field field = KeySegment.class.getDeclaredField("keyReferenceCounts");
-            field.setAccessible(true);
-            @SuppressWarnings("unchecked")
-            Map<Object, Integer> map = (Map<Object, Integer>) field.get(keySegment);
-            return map.getOrDefault(key, 0);
-        } catch (Exception e) {
-            fail("Failed to access keyReferenceCounts field");
-            return -1;
-        }
     }
 }
--- a/runtime/src/test/java/org/apache/flink/agents/runtime/operator/queue/SegmentedQueueTest.java
+++ b/runtime/src/test/java/org/apache/flink/agents/runtime/operator/queue/SegmentedQueueTest.java
@@ -92,15 +92,15 @@ class SegmentedQueueTest {
 
         // Add a key first
         queue.addKeyToLastSegment("key1");
-        int initialSegments = getSegmentCount(queue);
+        int initialSegments = queue.getSegments().size();
 
         // Add a watermark
         Watermark watermark = new Watermark(1000L);
         queue.addWatermark(watermark);
 
         // Verify a new segment was created
-        assertEquals(initialSegments + 1, getSegmentCount(queue));
-        assertEquals(1, getWatermarkCount(queue));
+        assertEquals(initialSegments + 1, queue.getSegments().size());
+        assertEquals(1, queue.getWatermarks().size());
     }
 
     @Test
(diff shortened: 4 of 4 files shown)


api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -28,6 +28,10 @@
 public abstract class Event {
     private final UUID id;
     private final Map<String, Object> attributes;
+    /** The timestamp of the record. */
+    private long timestamp;
```
Sxnan: Can we use Long type and use the null value to indicate the absence of the timestamp? Would that be simpler?

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -97,6 +97,14 @@ public List<Event> drainEvents() {
         return list;
     }
 
+    public List<Event> drainEvents(long timestamp) {
```
Sxnan: I think we can take a Long type timestamp here. We skip setting the timestamp if it is null. 

This way, the caller does not need to figure which drainEvents method to call

api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -54,4 +58,21 @@ public Object getAttr(String name) {
     public void setAttr(String name, Object value) {
         attributes.put(name, value);
     }
+
+    public boolean hasTimestamp() {
+        return hasTimestamp;
+    }
+
+    public long getTimestamp() {
```
Sxnan: We can just return Long here and remove the `hasTimestamp` method

runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/KeySegment.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+package org.apache.flink.agents.runtime.operator.queue;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** A group of keys with reference counting to track active keys. */
+public class KeySegment {
+    /** Maps keys to their reference counts (number of active occurrences). */
```
Sxnan: I think it is more accurate to say `(number of unfinished events of the specific key)`

runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/KeySegment.java
```diff
@@ -0,0 +1,59 @@
...
+                        return null; // Remove the key if count is 1 or less
+                    } else {
+                        return count - 1;
+                    }
+                });
+    }
+
+    /** Checks if a key is active (i.e., its reference count is greater than zero). */
+    public boolean hasActiveKey(Object key) {
+        return keyReferenceCounts.containsKey(key) && keyReferenceCounts.get(key) > 0;
```
Sxnan: As we remove the key if its count goes to 0, if the key is in the map, then it implies that the count is greater than 0.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/SegmentedQueue.java
```diff
@@ -0,0 +1,95 @@
...
+        if (segments.isEmpty()) {
+            lastSegment = appendNewSegment();
+        } else {
+            lastSegment = segments.getLast();
+        }
+        lastSegment.incrementKeyReference(key);
+    }
+
+    /**
+     * Removes a key from all segments, decrementing its reference count. Returns true if the key
```
Sxnan: The doc is incorrect; we actually removed the key from the first segment that contains the key

runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/SegmentedQueue.java
```diff
@@ -0,0 +1,95 @@
...
+        KeySegment newSegment = new KeySegment();
+        segments.addLast(newSegment);
+        return newSegment;
+    }
+
+    /**
+     * Pops the oldest watermark from the watermark deque and removes the corresponding key segment
+     * from the segments queue.
+     */
+    public Watermark popOldestWatermark() {
```
Sxnan, line 81: How about we make this method pop the completed watermark? And return null if no more completed watermark. This way we can ensure that the caller will never mistakenly pop an uncompleted watermark

runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/SegmentedQueue.java
```diff
@@ -0,0 +1,95 @@
...
+        segments.pop();
+        return watermarks.pop();
+    }
+
+    /** Checks if a watermark is ready to be processed (i.e., oldest segment is empty). */
+    public boolean canProcessWatermark() {
```
Sxnan: This and `isFirstSegmentEmpty` can be private

runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/SegmentedQueue.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /** Adds a watermark and creates a new segment to associate with it. */
+    public void addWatermark(Watermark watermark) {
+        watermarks.addLast(watermark);
+        appendNewSegment();
+    }
+
+    /** Creates a new key segment and appends it to the end of the queue. */
+    public KeySegment appendNewSegment() {
```
Sxnan: This can be private

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
Sxnan, line 1: How do we handle recovery from a checkpoint? I think we should add the key of the `actionTasksKState` to the segment during recovery.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -97,6 +97,16 @@ public List<Event> drainEvents() {
         return list;
     }
 
+    public List<Event> drainEvents(Long timestamp) {
```
Sxnan, line 93: Can we remove the `drainEvents()` method? 

Only PythonActionExecutor.java:137 is calling it, and it has access to the event sourceTimestamp

runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/KeySegment.java
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.runtime.operator.queue;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** A group of keys with the number of unfinished input records for each specific key. */
+public class KeySegment {
```
Sxnan, line 24: We should have unit tests for this class.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/queue/SegmentedQueue.java
```diff
@@ -0,0 +1,98 @@
...
+ * limitations under the License.
+ */
+package org.apache.flink.agents.runtime.operator.queue;
+
+import org.apache.flink.streaming.api.watermark.Watermark;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+public class SegmentedQueue {
```
Sxnan, line 25: We should have unit tests for this class.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -282,7 +301,11 @@ private void processEvent(Object key, Event event) throws Exception {
         if (EventUtil.isOutputEvent(event)) {
             // If the event is an OutputEvent, we send it downstream.
             OUT outputData = getOutputFromOutputEvent(event);
-            output.collect(reusedStreamRecord.replace(outputData));
+            if (event.hasSourceTimestamp()) {
+                output.collect(reusedStreamRecord.replace(outputData, event.getSourceTimestamp()));
+            } else {
+                output.collect(reusedStreamRecord.replace(outputData));
```
Sxnan, line 313: We should also clear the timestamp by `StreamRecord#eraseTimestamp`

Sxnan (commented):
Thanks for the update! Please see my comments below.

GreatEugenius:
Hi @Sxnan, thank you for the review! All the comments have been addressed. Please take another look.

runtime/src/test/java/org/apache/flink/agents/runtime/operator/queue/KeySegmentTest.java
```diff
@@ -0,0 +1,154 @@
...
+        assertFalse(keySegment.isEmpty());
+
+        // Remove remaining keys
+        keySegment.decrementKeyReference("key1");
+        keySegment.decrementKeyReference("key3");
+        assertTrue(keySegment.isEmpty());
+    }
+
+    // Helper method to access the private map for testing purposes
+    private int getKeyReferenceCount(KeySegment keySegment, Object key) {
```
Sxnan: We should avoid using reflection as much as possible, as it is very hard to maintain. 

For testing, we can use a package-private getter to access the field we want to inspect, as the unit test should typically be placed in the same package as the class being tested.

Sxnan:
@xintongsong Could you have a look at this PR?

Issue #134, opened 2025-08-28
[Feature] Improve the docstring of public facing API


Sxnan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The docstring of the public API should include all the available arguments from the class itself or from the parent. For example, the docstring of `OllamaChatModelSetup` does not show that it can take `prompt`, `tools`,  etc. as arguments.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #195, opened 2025-09-24
[Improvement] Add inherited parameters to ChatModelSetup/EmbeddingModelSetup docstrings

GreatEugenius:
Linked issue: #134 

### Purpose of change
Updated the docstrings for all `ChatModelSetup` and `EmbeddingModelSetup` classes to explicitly document parameters inherited from parent classes (e.g., `prompt`, `tools`, etc.). This ensures public API documentation reflects the full set of available arguments, improving clarity and consistency.

### DIFF
--- a/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
+++ b/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
@@ -210,6 +210,12 @@ class AnthropicChatModelSetup(BaseChatModelSetup):
 
     Attributes:
     ----------
+    connection : str
+        Name of the referenced connection. (Inherited from BaseChatModelSetup)
+    prompt : Optional[Union[Prompt, str]
+        Prompt template or string for the model. (Inherited from BaseChatModelSetup)
+    tools : Optional[List[str]]
+        List of available tools to use in the chat. (Inherited from BaseChatModelSetup)
     model : str
         Specifies the Anthropic model to use. Defaults to claude-sonnet-4-20250514.
     max_tokens: int
--- a/python/flink_agents/integrations/chat_models/ollama_chat_model.py
+++ b/python/flink_agents/integrations/chat_models/ollama_chat_model.py
@@ -167,6 +167,12 @@ class OllamaChatModelSetup(BaseChatModelSetup):
 
     Attributes:
     ----------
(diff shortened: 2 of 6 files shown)


python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
```diff
@@ -210,6 +210,12 @@ class AnthropicChatModelSetup(BaseChatModelSetup):
 
     Attributes:
     ----------
+    connection : str
```
alnzng, line 213: I’m not a Python expert, so this is more of a curiosity question: is it considered a best practice or common pattern to repeat inherited attributes in subclass docstrings?

One potential drawback I see is maintainability: if the parent attribute docs change, we’d need to update every subclass that repeats them, otherwise the subclass docs could become stale or misleading.

I understand the intent here is to make the subclass docstring self-contained so users can see all fields without checking the parent. That makes sense. But (if I understand correctly) most Python doc generators (e.g., Sphinx) can automatically include inherited attributes from the parent class, so we might get the same clarity without duplicating the docs.

@wenjin272 @Sxnan What are your thoughts?

GreatEugenius: Hi @alnzng, thank you very much for your suggestion! You're absolutely right — using tools like Sphinx can automatically inherit and document attributes from parent classes, which would avoid duplication and reduce maintenance overhead. We do plan to adopt such documentation generators in the future.

However, we are currently approaching a release deadline and wanted to ensure that subclass docstrings remain as self-contained as possible at this stage. The goal is to make it easier for users to understand all relevant fields without needing to refer back to the parent class documentation, even if it means some temporary duplication.

We appreciate your feedback and will definitely keep this in mind for future improvements!

alnzng: Thanks for the reply @GreatEugenius!

I appreciate the explanation, though I'm still not fully convinced of the necessity. Since we're building comprehensive documentation that will list all available attributes(including inherited ones), users should have visibility into all options without needing to check the docstrings. For (advanced) developers who do read the source codes directly, checking parent classes for inherited attributes is standard practice.

However, I'm ok with it if that's the team's decision.

Issue #185, opened 2025-09-18
[Feature] Async action properly handle checkpoint and per-action state


Sxnan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, with Python async execution supported, a Python action can be divided into multiple invocations before it is finished. And each invocation can potentially update the state. 

It is problematic when a checkpoint is taken while an async execution is not finished, but has already updated the state. Say we have an async action divided into two steps, 1 and 2. After step 1, it updates the state from S0 to S1. Now, a checkpoint is made. Now the checkpoint contains S1. When we recover from the checkpoint, we have to run the async action from step 1, but now it sees the state S1 instead of S0.

To fix the problem, we have to keep the update to the state in memory, and only persist it to the state when the async action is finished. 

Note that the per-action state consistency has the same problem, so we can only put the state update to the action state store when the async action is all finished.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #219, opened 2025-09-25
[runtime] Only persist memory changes to flink state when action is finished

Sxnan:
Linked issue: #185

### Purpose of change

This PR updates the `ActionExecutionOperator` to only persist memory changes to flink state when the action is finished.

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/memory/CachedMemoryStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/memory/CachedMemoryStore.java
@@ -55,5 +55,6 @@ public class CachedMemoryStore implements MemoryStore {
         for (Map.Entry<String, MemoryObjectImpl.MemoryItem> entry : cache.entrySet()) {
             store.put(entry.getKey(), entry.getValue());
         }
+        cache.clear();
     }
 }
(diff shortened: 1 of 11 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/memory/CachedMemoryStore.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public boolean contains(String key) throws Exception {
+        return cache.containsKey(key) || store.contains(key);
+    }
+
+    public void persistCache() throws Exception {
+        for (Map.Entry<String, MemoryObjectImpl.MemoryItem> entry : cache.entrySet()) {
+            store.put(entry.getKey(), entry.getValue());
+        }
+    }
```
xintongsong, line 59: Shall we clear the cache after persisting?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -374,6 +380,7 @@ private void processActionTaskForKey(Object key) throws Exception {
         } else {
             maybeInitActionState(key, sequenceNumber, actionTask.action, actionTask.event);
             ActionTask.ActionTaskResult actionTaskResult = actionTask.invoke();
+            actionTaskRunnerContexts.remove(actionTask);
```
xintongsong, line 386: I think this is correct but a bit hard to understand. We might need to explain that context is only cached for one round of execution. IIUC, this is because we don't want to assume that they key of `actionTaskRunnerContexts` remains the same action task after invoking?

xintongsong (approved):
LGTM

Issue #204, opened 2025-09-24
[Feature][docs] How to build an agent - Embedding and Vector Store docs


Sxnan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Documentation of How to build an agent - Embedding and Vector Store

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #221, opened 2025-09-26
[Feature][docs] add docs for introducing embedding models and vector stores

alnzng:
Linked issue: #204

### Purpose of change

add docs for introducing embedding models and vector stores

### Tests

Screenshot 2025-09-26 at 12 00 39 AM

### API

n/a

### Documentation

n/a


docs/content/docs/development/vector_stores.md
```diff
@@ -0,0 +1,250 @@
...
+        return ChromaVectorStoreConnection, {
+            "persist_directory": "/path/to/chroma/data",  # For persistent storage
+            # Or use other modes:
+            # "host": "localhost", "port": 8000  # For server mode
+            # "api_key": "your-chroma-cloud-key"  # For cloud mode
+        }
+
+    @vector_store_setup
+    @staticmethod
+    def chroma_store() -> Tuple[Type[ChromaVectorStoreSetup], Dict[str, Any]]:
```
alnzng: @xintongsong I'd like to discuss improving the vector store interface design for better usability. I've outlined a proposal below and would appreciate your thoughts before proceeding with implementation. If you're aligned with this direction, I can create follow-up PRs to refactor both the code and documentation.

------
**Problem**
The current two-component architecture (Connection + Setup) adds unnecessary complexity for vector stores without clear practical benefits.

Why it works for embedding / chat models:
  - One embedding connection → multiple embedding models (e.g. 3-small, 3-large, ada-002)
  - Clear separation: auth/limits vs model selection

Why it doesn't fit vector stores:
  - Connection and collection are tightly coupled to the same data store
  - No clear use case for sharing connections across different setups
  - Adds cognitive overhead without meaningful resource sharing

**Proposed Change**

  Current (Complex):
```
  @vector_store_connection + @vector_store_setup  # Two decorators
```

  Proposed (Simple):
```
  @vector_store  # One decorator
  def chroma_vector_store():
      return ChromaVectorStore, {
          "persist_directory": "/data",
          "collection": "docs",
          "embedding_model": "openai_embedding"
      }
```
(This single-component design idea is similar like how we implement `mcp_server` currently)

alnzng:
@xintongsong I added docs for embedding models and vector stores based on the existing implementations, please help take a look.

docs/content/docs/development/vector_stores.md
xintongsong: Thanks for pointing this out. It makes perfect sense to me.

xintongsong:
@alnzng, thanks for the really nice work on these documentations. A few suggestions from my side.

### Embedding Models

In general, I'd suggest the following order to introduce things to our users.
1. How to use embedding models in general
2. What built-in providers are available, and their specifics
3. How to provide custom provider support, in the built-in ones do not meet requirements

In that sense, I'd suggest the following changes.
1. Move the introduction of the connection and setup concepts, and why we have this separation, to the `Overview` section.
2. Change the `Embedding Interfaces` section name into something like `Use Embedding Models in Agents`, move `BaseEmbeddingModelConnection` and `BaseEmbeddingModelSetup` out, and add a general example that not only shows how to define the model connection and setup with the decorators, but also how to obtain the embedding model with `RunnerContext.get_resource()` and call the `embed()` method on it. 
3. Add another section `Support Custom Providers` after `Built-in Providers`, and introduce `BaseEmbeddingModelConnection` and `BaseEmbeddingModelSetup`, which are the interfaces need to be implemented for custom supports.

In addition, I'd suggest to add a notice for the `Support Custom Providers` section, reminding users that the APIs are experimental and unstable, subject to in-compatible changes in future releases.

### Vector Stores

My suggestions are similar to those on Embedding Models, to separate the sections from the user's perspective: how to use it, what built-in supports are available, and how to customize in built-in supports don't meet requirements.

docs/content/docs/development/vector_stores.md
alnzng: Great, created a issue for this: https://github.com/apache/flink-agents/issues/230

alnzng:
Great suggestions, @xintongsong ! I updated the docs, please take another look, thanks.

xintongsong:
Thanks for addressing my comment, @alnzng. 

Now I think the documentation reads really well. However, it seems some latest changes are not captured by the sample codes. Have you tried running the sample codes on the latest main branch?
- We refactored the signature of decorated functions. They used to return a class, followed by a list of arguments used for initializing the class. Now we expect the function to return a ResourceDescriptor. See #211 for more details.
- `ctx.get_resource()` should also have an argument of the resource type.

@wenjin272, could you also help checking on the sample codes?

alnzng:
@xintongsong Oh, I should use the latest API in the doc. I know we have changed to `ResourceDescriptor` recently, but forgot to update the example codes here. (I copied this from one of my very old example). Thanks for pointing this out! I updated the doc accordingly.

@wenjin272 Please help take a second eye on this, thanks!

wenjin272:
Hi, @alnzng and @xintongsong, I build the docs locally, read the document and check the code.  All things look good to me.

PR #226, opened 2025-09-28
[docs] Update installation section.

wenjin272:
Linked issue: #198

### Purpose of change

Update the installation section.

### Tests

No

### API

No

### Documentation

Yes

### DIFF
--- a/docs/content/docs/get-started/installation.md
+++ b/docs/content/docs/get-started/installation.md
@@ -53,11 +53,10 @@ To run on a Flink cluster, ensure the Flink Agents Java JARs are placed in the F
 
 <!-- TODO: fill in the command after Flink Agents is released -->
 ```shell
-# Download the Flink Agents released tar
+# Download the Flink Agents released flink-agents-dist jar.
 
-# After downloading and extracting the Flink Agents release bundle,
-# copy the Flink Agents JARs to Flink's lib directory
-
+# After downloading the bundle jar, copy it to Flink's lib directory.
+cp flink-agents-dist-$VERSION.jar $FLINK_HOME/lib/
 ```
 
 <!-- TODO: link to flink quickstart example docs -->
@@ -80,9 +79,18 @@ To clone from Git, run:
 git clone https://github.com/apache/flink-agents.git
 ```
 
+### Build
+To run on a Flink cluster, we need build the whole project.
+
+We provide a script to run:
+
+```shell
+bash tools/build.sh
+```
+
 ### Java Build
 
-The Python wheel depends on the Flink Agents Java modules, so we need to build Flink Agents Java modules first, run:
+To try Flink Agents in Java, user can build java independently.
 
 ```shell
 cd flink-agents
@@ -91,7 +99,8 @@ mvn clean install -DskipTests
 
 ### Python Build and Install
 
-Then we can build and install the Flink Agents wheel.
+To try Flink Agents in Python on a local executor,
+user can build Python independently.
 
 {{< tabs>}}
 {{< tab "uv (Recommended)" >}}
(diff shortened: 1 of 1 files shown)


docs/content/docs/get-started/installation.md
```diff
@@ -22,47 +22,46 @@ specific language governing permissions and limitations
...
 
-# Installation
-
+## Overview
 Flink Agents provides both Python and Java APIs to define a Flink Agents job.
 
-To try Flink Agents in Python on a local executor, you only need to install the Flink Agents Python package.
-
-To define a Flink Agents job using the Java API, or to run the job in a Flink cluster, you need to install both the Flink Agents Python and Java dependencies. 
+To try Flink-Agents in Python or Java, user can only install the correspond package.
```
xintongsong: This is incorrect. To run a Python job in a Flink cluster, you'll also need the java packages.

I think it's too complicated to explain when to install python / java only, and when to install both. The benefit for users to install one less package is probably weaker than the confusion for users to understand what packages they need. So let's just tell our users to always install both.

docs/content/docs/get-started/installation.md
```diff
@@ -80,68 +79,24 @@ To clone from Git, run:
...
-
 ### Install Flink Agents to Flink
 
 
 To install the Java dependencies to Flink, run:
 
-<!-- TODO: fill in the command after Flink Agents produce uber jar -->
 ```shell
 
+cd flink-agents
```
xintongsong, line 93: Remote the empty line.

docs/content/docs/get-started/installation.md
```diff
@@ -22,47 +22,46 @@ specific language governing permissions and limitations
...
 ```shell
 python -m pip install flink-agents
 ```
 
-<!-- TODO: link to local quickstart example docs -->
-Now you can run a Flink Agents job on the local executor.
-See [local quickstart example]() for end-to-end examples of running on the local executor.
-
+Now you can run a python Flink Agents job.
+See [quickstart example]({{< ref "docs/get-started/quickstart/workflow_agent" >}}) for end-to-end examples of running on Flink.
```
xintongsong: If we are telling users to always install both packages, we probably want to remove these paragraph, or move it to somewhere else.

docs/content/docs/get-started/installation.md
```diff
@@ -80,68 +79,24 @@ To clone from Git, run:
...
-python -m pip install dist/*.whl
-```
-
-{{< /tab >}}
-{{< /tabs >}}
-
 ### Install Flink Agents to Flink
 
 
 To install the Java dependencies to Flink, run:
```
xintongsong, line 90: Why only installing the java dependencies? What happen to the python?

docs/content/docs/get-started/installation.md
wenjin272: Because we have packaged the flink-agents-dist jar in python package, so actually there is no need to download java package and install it to Flink to run a Python job. I have verified this on standalone cluster.

I agree this is too complicated to user. I will revise this to tell users to always install both.

docs/content/docs/get-started/installation.md
wenjin272: The python package will be installed to the python interpreter. By export `PYTHONPATH` before run python job.
`export PYTHONPATH=$(python -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')`
The standalone cluster can find the python package.

xintongsong: Does it also work for submitting to a remote cluster? If this is the right step, why not include it in the installation doc?

wenjin272: It works when submitting the job to standalone cluster, I haven't test in yarn or k8s cluster.

I think this is not a installation step, but a submit job step. For user must export the PYTHONPATH each time they start cluster and submit the job.

But just install java package to flink, not install python package may confuse the user, I think maybe we can mention this behavior here.

xintongsong: I think we should offer both options: export PYTHONPATH in .bashrc for once, or export it everytime before starting a cluster or submitting the job. This should be mentioned in both the installation and deployment documents.

xintongsong (approved):
LGTM

Issue #227, opened 2025-09-28
[Feature] Introduce built-in CONTEXT_RETRIEVAL_ACTION to enable context retrieval (e.g., RAG)


alnzng:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce built-in CONTEXT_RETRIEVAL_ACTION to enable context retrieval (e.g., RAG)

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #228, opened 2025-09-28
[Feature] Introduce built-in CONTEXT_RETRIEVAL_ACTION to enable context retrieval

alnzng:
Linked issue: #227

### Purpose of change

Introduce built-in CONTEXT_RETRIEVAL_ACTION to enable context retrieval (e.g., RAG)

### Tests

Unit tests and end-to-end RAG agent example.

### API

- Add new events `ContextRetrievalRequestEvent` and `ContextRetrievalResponseEvent`

### Documentation

n/a


xintongsong:
@alnzng,

I think the code changes look good to me in general.

What I'm not entirely sure is to name the action and events with "context_retrieval". I think what the action and events do is essentially semantic search. It's not necessarily for context retrieval. (Admittedly, in the agentic world, it's mainly used for context retrieval). I wonder if we should just name it with "semantic_search" or "similarity_search"?

wenjin272:
LGTM

PR #233, opened 2025-09-28
[docs] Update workflow agent development doc.

wenjin272:
Linked issue: #202

### Purpose of change

Update workflow agent development doc.

### Tests

No

### API

No

### Documentation

Yes


docs/content/docs/development/workflow_agent.md
```diff
@@ -119,20 +117,78 @@ class ReviewAnalysisAgent(Agent):
...
+There are 2 special types of event.
+* `InputEvent`: Generated by the framework, carrying an input data record that arrives at the agent. Actions listening to the `InputEvent` will be the entry points of agent.
+* `OutputEvent`: The framework will listen to `OutputEvent`, and convert its payload into outputs of the agent. By generating `OutputEvent`, actions can emit output data.
+
+User can define owner event by extends `Event`.
+```python
+class MyEvent(Event):
+    value: Any
+```
+Then, user can define action listen or send `MyEvent`
```
xintongsong:
```suggestion
Then, user can define action listen to or send `MyEvent`
```

docs/content/docs/development/workflow_agent.md
```diff
@@ -119,20 +117,78 @@ class ReviewAnalysisAgent(Agent):
...
+    ctx.send_event(ChatRequestEvent(model=xxx, messages=xxx))
+    # output data to downstream
+    ctx.send_event(OutputEvent(output=xxx))
+```
 ## Event
+Events are messages passed between actions. Events may carry payloads. A single event may trigger multiple actions if they are all listening to its type.
+
+There are 2 special types of event.
+* `InputEvent`: Generated by the framework, carrying an input data record that arrives at the agent. Actions listening to the `InputEvent` will be the entry points of agent.
+* `OutputEvent`: The framework will listen to `OutputEvent`, and convert its payload into outputs of the agent. By generating `OutputEvent`, actions can emit output data.
```
xintongsong: I think we need to explicitly mention the field names to be used.

docs/content/docs/development/workflow_agent.md
```diff
@@ -119,20 +117,78 @@ class ReviewAnalysisAgent(Agent):
...
+> **Note**: To support the Event Log of Flink-Agents, the payload of `Event` should be `BaseModel` serializable. 
+## Built-in Events and Actions
 
-{{< hint warning >}}
-**TODO**: How to define, send and handle an event, including InputEvent, OutputEvent, and event between actions.
+There are several built-in `Event` in Flink-Agents:
+* `ChatRequestEvent`: Event carries the chat messages to be sent to llm.
+* `ChatResponseEvent`: Event carries the llm response.
+* `ToolRequestEvent`: Event carries the tool call extracted from llm response.
+* `ToolResponseEvent`: Event carries the tool execution result.
```
xintongsong: If we are not introducing how to use these built-in events here, we don't need to describe each of them. Instead, we should point to the other documents where they will be explained. Like:
* See *link* for how to chat with a LLM leveraging built-in action and events.
* See *link* for how to programmatically use a tool leveraging built-in action and events.

docs/content/docs/development/workflow_agent.md
```diff
@@ -119,20 +117,78 @@ class ReviewAnalysisAgent(Agent):
...
 ## Memory
 
-{{< hint warning >}}
-**TODO**: How to use short-term memory and long-term memory.
-{{< /hint >}}
+Memory is data that will be remembered across actions and agent runs.
+
+### Short-Term Memory
+
+Short-Term Memory is shared across all actions within an agent run, and multiple agent runs with the same input key. This corresponds to Flink's Keyed State, which is visible to processing of multiple records with in the same keyed partition, and is transparent to processing of data in other keyed partitions.
```
xintongsong: Explain agent run.
```suggestion
Short-Term Memory is shared across all actions within an agent run, and multiple agent runs with the same input key. 
Here a *agent run* refers to ...
This corresponds to Flink's Keyed State, which is visible to processing of multiple records with in the same keyed partition, and is transparent to processing of data in other keyed partitions.
```

docs/content/docs/development/workflow_agent.md
```diff
@@ -119,20 +117,78 @@ class ReviewAnalysisAgent(Agent):
...
+* `InputEvent`: Generated by the framework, carrying an input data record that arrives at the agent. Actions listening to the `InputEvent` will be the entry points of agent.
+* `OutputEvent`: The framework will listen to `OutputEvent`, and convert its payload into outputs of the agent. By generating `OutputEvent`, actions can emit output data.
+
+User can define owner event by extends `Event`.
+```python
+class MyEvent(Event):
+    value: Any
+```
+Then, user can define action listen or send `MyEvent`
+> **Note**: To support the Event Log of Flink-Agents, the payload of `Event` should be `BaseModel` serializable.
```
xintongsong: This is not only for supporting Event Log, it's also needed for persisting events in Flink's state and the WAL for per-action consistency, and cross-language transmission in future if we support hybrid agent with both java and python codes.

xintongsong: Just simply say:
```suggestion
> **Note**: The payload of `Event` should be `BaseModel` serializable. 
```

docs/content/docs/development/workflow_agent.md
```diff
@@ -119,20 +117,78 @@ class ReviewAnalysisAgent(Agent):
...
+
+@action(ChatResponseEvent)
+@staticmethod
+def process_chat_response(event: ChatResponseEvent, ctx: RunnerContext) -> None:
+    """Process chat response event and send output event."""
+    ...
+    id=ctx.short_term_memory.get("id"),
+    ...
+
+```
```
xintongsong, line 230: For Short-Term Memory, I think we should also mention that:
1. It's not just a k-v map. You can store nested objects. Detailed APIs ...
2. You can pass reference of memory object, instead of the actual object. When and how to use this.

PR #238, opened 2025-09-29
[hotfix] Refactor codes with latest API changes 

alnzng:
Linked issue: #xxx

### Purpose of change

Refactor codes with latest API changes 

### Tests

Run `rag_agent_example.py` locally

### API

n/a

### Documentation

n/a


xintongsong:
Actually, I've notice that the rag agent is under `examples/`, not `examples/quickstart/`.

I think we should have both quick-start examples which should be the simplest examples to get start with, and also other examples showing how to use the core features or dealing with typical scenarios. That's why we have made `quickstart/` a sub-directory of `examples/`. You may also noticed that we have moved other "examples" from `examples/` to `e2e_tests/`, because those are too detailed and more like test cases to prove that an certain features work.

I think the rag agent might be more suitable for the normal examples. Users probably won't need it in the first step, but would look for it when they need to work on some rag-like stuff. That's why I haven't point it out when reviewing the previous PR.

WDYT?

alnzng:
Good suggestion @xintongsong ! I agree that current RAG implementation is too advanced for quickstart. Perhaps once we have a built-in RAG agent (similar to ReAct), we can add a simpler example to `examples/quickstart`.

I've moved this example to `examples/rag/` and refactored it to be more focused and user-friendly. The test data setup logic has been extracted to a separate` knowledge_base_setup.py` file, making the main example clearer and more concise for learning how to build RAG agents.

Please take a look, thanks.

xintongsong (approved):
LGTM

PR #250, opened 2025-10-06
Support for Ollama Embedding models in Java. (ollama4j)

mcolomerc:
Linked issue: #304 

JAVA API for Embedding models. 

Ollama support fro Embedding models with ollama4j. 

### Purpose of change

Support for Ollama Embedding models in Java. (ollama4j)

### Tests

OllamaEmbeddingModelConnectionTest

Example: WorkflowEmbeddingsAgentExampleJob provided.

### API

Annotations: EmbeddingModelConnection and EmbeddingModelSetup annotations. 

embedding.model: BaseEmbeddingModelConnection and BaseEmbeddingModelSetup 

### Documentation

Integrations: Added Ollama integration for using embedding models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 Apache Flink Agents is an Agentic AI framework based on Apache Flink.
 
+[User Documentation](https://nightlies.apache.org/flink/flink-agents-docs-main/)
+
 ## Building
 
 Prerequisites for building Flink Agents:


integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelSetup.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+     * Generate embeddings for the given text using the configured Ollama model.
+     *
+     * @param text The input text to generate embeddings for
+     * @return An array of floating-point values representing the text embeddings
+     */
+    @Override
+    public float[] embed(String text) {
```
wenjin272: Would it be necessary to override the parent class's method here?

integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelSetup.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+     * Generate embeddings for the given text using a specific model.
+     *
+     * @param text The input text to generate embeddings for
+     * @param model The specific embedding model to use
+     * @return An array of floating-point values representing the text embeddings
+     */
+    public float[] embed(String text, String model) {
```
wenjin272: Would it be possible to define these embed methods in the parent class?

integrations/embedding-models/ollama/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-plan</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>io.github.ollama4j</groupId>
+            <artifactId>ollama4j</artifactId>
+            <version>1.1.0</version>
```
wenjin272: Maybe declare the ollama4j version in _integrations/pom.xml_ and use it in _embedding-models/ollama/pom.xml_ and _chat-models/ollama/pom.xml_ to unify the version of ollama4j.

examples/src/main/java/org/apache/flink/agents/examples/agents/EmbeddingsAgent.java
```diff
@@ -0,0 +1,372 @@
...
+    /**
+     * Generate real embeddings using Ollama embedding model directly. This bypasses the framework
+     * resource system and creates a direct connection.
+     */
+    private static float[] generateRealEmbedding(String text) {
+        try {
+            // Create Ollama connection directly using the same configuration as the resource
+            // descriptor
+            OllamaEmbeddingModelConnection connection =
+                    new OllamaEmbeddingModelConnection(ollamaEmbeddingConnection(), null);
```
wenjin272: Why not use the resource provided by the framework here? like
```
OllamaEmbeddingModelSetup embeddingModel =
                    (OllamaEmbeddingModelSetup) ctx.getResource("embeddingModel", ResourceType.EMBEDDING_MODEL);

float[] embedding = embeddingModel.embed(text);
```

integrations/embedding-models/ollama/pom.xml
mcolomerc: Fixed

integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelSetup.java
mcolomerc: added these methods to the parent class, needed to add them as abstract methods to the base interface, forcing all embedding model implementations to support them. what do you think?

integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelSetup.java
mcolomerc: fixed

integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelSetup.java
wenjin272: It would be fine not to declare it as an abstract method, since most implementing classes may not need to modify the implementation of these methods.

integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelConnection.java
```diff
@@ -0,0 +1,255 @@
...
+                        ? descriptor.getArgument("model")
+                        : "nomic-embed-text";
+
+        this.ollamaAPI = new OllamaAPI(host);
+        // Note: OllamaAPI timeout configuration may vary by version
+        // For ollama4j 1.1.0, timeout is typically configured at request level
+    }
+
+    @Override
+    public float[] embed(String text) {
```
wenjin272: Maybe we could add an additional parameter `Map<String, Object> parameters` here, like `chat` method of `BaseChatModelConnection`.

```
/**
 * Generate embeddings for a single text input.
 *
 * @param text The input text to generate embeddings for
 * @param parameters The additional arguments to configure the embed request.
 * @return An array of floating-point values representing the text embeddings
 */
public abstract float[] embed(String text, Map<String, Object> parameters);
```

Then we can remove the `embed(String text, String model)` and get `model` name from `parameters`.

I recommend this implementation because, in our design, different instances of `EmbeddingModelSetup` can be initialized with different configuration parameters, yet they can share the same `EmbeddingModelConnection`. When an `EmbeddingModelSetup` instance calls the embed method of `EmbeddingModelConnection`, it can pass its own unique configuration parameters, including `model_name`, `encoding_format`, and others.

You can take a look at python implementation as a reference.

wenjin272 (commented):
Sorry for I forgot a comment in last review.

examples/src/main/java/org/apache/flink/agents/examples/agents/EmbeddingsAgent.java
mcolomerc: fixed

integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelConnection.java
mcolomerc: Makes sense. I did a refactor to include the approach and now it's more aligned with BaseChatModelConnection.

wenjin272:
LGTM

api/src/main/java/org/apache/flink/agents/api/embedding/model/BaseEmbeddingModelConnection.java
```diff
@@ -0,0 +1,82 @@
...
+     *     embeddings
+     */
+    public abstract List<float[]> embed(List<String> texts, Map<String, Object> parameters);
+
+    /**
+     * Get the dimension of the embeddings produced by this model.
+     *
+     * @return The embedding dimension
+     */
+    public abstract int getEmbeddingDimension();
```
xintongsong: Why do we need this? I think the dimension should be a parameter that you can set differently for each embedding request.

examples/src/main/java/org/apache/flink/agents/examples/agents/EmbeddingsAgent.java
```diff
@@ -0,0 +1,378 @@
...
+ * An agent that generates embeddings for each row of data using Ollama embedding models.
+ *
+ * <p>This agent receives text data, processes it to generate high-dimensional vector embeddings,
+ * and outputs the results with metadata. It demonstrates how to integrate embedding models into
+ * Flink Agents workflows for vector-based processing and similarity search applications.
+ *
+ * <p>The agent supports various embedding models available in Ollama such as: - nomic-embed-text
+ * (768 dimensions) - mxbai-embed-large (1024 dimensions) - all-minilm (384 dimensions)
+ */
+public class EmbeddingsAgent extends Agent {
```
xintongsong: I think `EmbeddingAgent` and `WorkflowEmbeddingsAgentExampleJob` are more like e2e test for verification purposes, rather than user-facing examples. We probably don't want to recommend users to directly call the embedding models in their actions. Therefore, I'd suggest to move these to the `e2e-test` module.

integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelConnection.java
```diff
@@ -0,0 +1,215 @@
...
+                                    + defaultModel
+                                    + ". Cause: "
+                                    + e.getMessage(),
+                            e);
+            }
+        }
+    }
+
+    /** Check if the specified model is available on the Ollama server. */
+    public boolean isModelAvailable(String model) {
```
xintongsong: This method is never used. I think we should avoid introducing public interfaces unless we see a concrete demand for it.

integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelConnection.java
```diff
@@ -0,0 +1,215 @@
...
+                embed("test", testParams);
+                return true;
+            } catch (Exception testException) {
+                return false;
+            }
+        }
+    }
+
+    /** Get the default embedding model name. */
+    public String getDefaultModel() {
```
xintongsong: Same here. Never used.

integrations/embedding-models/ollama/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/ollama/OllamaEmbeddingModelConnection.java
mcolomerc: [Removed unused](https://github.com/apache/flink-agents/pull/250/commits/f9efd0a79e64bcdcd813a81a5c0d7367949646fc)

api/src/main/java/org/apache/flink/agents/api/embedding/model/BaseEmbeddingModelConnection.java
mcolomerc: Removed. Provides a fallback dimension that the model supports by default

examples/src/main/java/org/apache/flink/agents/examples/agents/EmbeddingsAgent.java
mcolomerc: Moved to e2e-test

api/src/main/java/org/apache/flink/agents/api/embedding/model/BaseEmbeddingModelConnection.java
```diff
@@ -0,0 +1,107 @@
...
+     * what dimension values can be passed in parameters.
+     *
+     * @return An array where [0] is the minimum supported dimension and [1] is the maximum, or null
+     *     if the model only supports fixed dimensions
+     */
+    public int[] getSupportedDimensionRange() {
+        // Default implementation: only supports the default dimension
+        int defaultDim = getDefaultEmbeddingDimension();
+        return new int[] {defaultDim, defaultDim};
+    }
```
xintongsong: I still don't understand why do we need these interfaces.

1. Could you explain the necessity of having these interfaces?  

2. I don't think the embedding model providers are commonly providing this information. Take Ollama as an example, it supports hosting various embedding models and each or them may have a different supported dimension range. In this PR, it just cached the dimension of the first result and assuming it is the default, which is not always true. It also doesn't make sense to assume the default dimension is the only supported range. 

I think we probably don't want to introduce such interfaces that are not necessarily needed and cannot always provide accurate results.

examples/pom.xml
```diff
@@ -69,7 +69,11 @@ under the License.
             <artifactId>flink-agents-integrations-chat-models-ollama</artifactId>
             <version>${project.version}</version>
         </dependency>
-
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-integrations-embedding-models-ollama</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
xintongsong: Since we are moving the example agents to `e2e-test` module, we should not need this change anymore.

integrations/embedding-models/ollama/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-plan</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
xintongsong: What dependencies do we need from `flink-agents-plan`?

api/src/main/java/org/apache/flink/agents/api/embedding/model/BaseEmbeddingModelConnection.java
mcolomerc: _Get the default embedding dimension for this model when no specific dimension is requested.
     * This provides a fallback dimension that the model supports by default._

It's not very common to change vector dimension in your pipeline, without changing the model, so it's quite related to the model used, so the implementation is making it as default by model  

Ollama embedding models: 
nomic-embed-text → outputs vectors of length 768 
mxbai-embed-large → outputs vectors of length 1024 
all-minilm → outputs vectors of length 384

OpenAI 
text-embedding-3-large --> 3072

If you try to use nomic-embed-text (768) but initialize your index or retrieval system with dimension 1024, you’ll get errors such as:

_ValueError: dimension mismatch: expected 1024 but got 768_

Those are the reasons of managing the dimension, the `OllamaEmbeddingModelConnection` is managing and setting dimension for the given Ollama model, avoiding a wrong setting error. 
We can change it and set the dimension in `@EmbeddingModelSetup`, and trust the user it uses the correct dimension for the model, but for me it make sense to set it only once as default, not for every request, because is not going to change. 

Do you have an example of changing the dimension when building a RAG?

xintongsong: I see your point. So you are trying to keep the dimension consistent between the embedding model and the vector storage. In that case:
1. The dimension should be a setup parameter, rather than a connection parameter. This doesn't mean we need to set a different dimension for each embedding request. Imagine your agent needs to retrieve context from 2 different knowledge bases, and their dimension are different. In that case, you may have 2 embedding model setups, with different dimensions, and potentially share the same embedding model connection if the same provider is used.
2. I think we can make `dimension` a property of `EmbeddingModelSetup`. Users may explicitly set the `dimension`, in which case we should pass that value to the model provider to make sure it takes effect. If not explicitly set, no parameter will be passed to the model provider, and the default value is decided on the provider side. 
3. We can also provide a getter for this dimension property. If explicitly set, that value will be returned. If not explicitly set, and the model provider has API that allows us to check the dimension, we can provide it. Otherwise, I think we should just say the dimension is unknown. I'd suggest not to make any further assumptions, like `nomic-embed-text` always uses 768 as the default, because we don't know if there's any custom setups on the provider side, and it's hard to maintain as new models emerging.

mcolomerc: Makes sense. Let me work on that.

mcolomerc: Finally removed the dimension dependency, ollama4j is managing it. I simplified the approach, and the example. 
The method used was deprecated with the last ollama4j version, so now, user does not need to specify dimension or store it as default. 

Having some issues with the GHA now, related to e2e.sh script and pyhton. Need to check that.

mcolomerc: Needed to fix the multi platform python script (tools/e2e.sh) because it was failing the pipeline for different platforms. Now it's working, but it was not the goal of this PR.

xintongsong: 1. It's quite common that during preparing a PR we discover and fix other issues. In such cases, we can simply fix the issue with a separate commit, before the main commits. There's some guidelines on how to organize commits of a PR. You may take a look.
https://flink.apache.org/how-to-contribute/code-style-and-quality-pull-requests/#3-separate-refactoring-cleanup-and-independent-changes

2. Regarding the failure you met, could you explain a bit more about the cause? I looked into the script changes, but find it hard to understand what the changes are trying to fix. In particular, why the `python` directory is not found in CI of this PR, but is fine in previous CI runs? Any changes in this PR caused the problem?

wenjin272: Hi, @mcolomerc, I have verified the latest several commits which modified the `pyproject.toml` and e2e test scripts. Thanks for your work. I think they can be reorganized as two commits. 

1. Because a python dependency `onnxruntime` has released new version which only supports linux platform, and uv will always install the latest version when not specify version. So we should modify the `pyproject.toml` and add version specification for mac & arm platform. 
2. The original e2e test scripts is not robust, and can't reflect the true cause of the test failure when compilation fails. So you add some check and guarantee in the scripts. 

So I think finally you can reorganize the commits of this pr to:
1. [hotfix] Fix dependency onnxruntime  install failure in mac.
2. [infra] Improve the robustness of e2e test.
3. [Feature] Support for Ollama Embedding models in Java.

mcolomerc: Hi, yes, this usually happens when a PR gets extended. The problem I was having with the scripts was the change in the script path; it didn't behave the same on all platforms. Aside from what you mentioned about the onnxruntime, we can also separate it into different PRs, cancel this, and structure it better, along with its associated issues. What do you think?

xintongsong: Thank you both for the explanations.

There's no need to separate into different different PRs. I think it would be nice to keep the codes in this PR, together with all the above discussions. I can reorganize the commits and merge the pr with a rebase command (so the individual commits are preserved).

PR #266, opened 2025-10-09
[Doc] Reorganize installation guide and simplify quickstart preparation sections

alnzng:
Linked issue: #xxx

### Purpose of change

Reorganize installation guide and simplify quickstart preparation sections

### Tests

deploy the doc locally

### API

n/a

### Documentation

n/a

### DIFF
--- a/docs/content/docs/get-started/quickstart/react_agent.md
+++ b/docs/content/docs/get-started/quickstart/react_agent.md
@@ -218,31 +218,15 @@ reviewAnalysisResStream.print();
 
 ### Preparation
 
-#### Prepare Flink
+#### Prepare Flink and Flink Agents
 
-Download a stable release of Flink 1.20.3, then extract the archive:
+Follow the [installation]({{< ref "docs/get-started/installation" >}}) instructions to setup Flink and the Flink Agents.
 
-```bash
-curl -LO https://archive.apache.org/dist/flink/flink-1.20.3/flink-1.20.3-bin-scala_2.12.tgz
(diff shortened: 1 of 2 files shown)


docs/content/docs/get-started/installation.md
```diff
@@ -30,84 +30,185 @@ The sections below show how to install the required dependencies.
...
 
-### Build
-To run on a Flink cluster, we need build the whole project.
+#### From Source
 
-We provide a script to run:
+**Clone the repository:**
 
 ```shell
+git clone https://github.com/apache/flink-agents.git
```
xintongsong, line 117: Nitpicking: We might want users to build the repo from the `release-0.1.0` tag (there will be one once formal released), rather than the latest main branch.

And there are 2 ways of getting the codes:
1. Clone the git repo
2. Download from website. The expected download link is `https://www.apache.org/dyn/closer.lua/flink/flink-agents-0.1.0/flink-agents-0.1.0-src.tgz`, which should be available once formal released.

alnzng: @xintongsong I agree with you that we should suggest the users to download a specific version of source code in each release/stable doc. However, this PR is merging into the main branch first, so in this doc we should suggest the users to use the latest codes due to the same consideration. 

Next step I will create another PR(cherry-pick comment but with customization) which is dedicated for `release-0.1.0` branch, it will cover what you suggested here. 

Let me know if this is not a right approach to make changes on release branch.

xintongsong: That approach makes perfect sense to me. I'll merge this PR as is.

Issue #278, opened 2025-10-18
[Feature][integration][java] Support explicit tool call handling in OllamaChatModel


twosom:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

In PR #162, we integrated Ollama4j as a chat model provider. However, there was a limitation where tool calls were handled internally by Ollama4j, which prevented proper event logging and observability in the flink-agents framework.

As discussed in #162, the flink-agents framework should manage tools and related context (stored in short-memory), with Chat request & Tool call & Tool response all generating `Event` for observability and per-action consistency.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #279, opened 2025-10-18
[Feature][Integrations][Java] Support explicit tool call handling in OllamaChatModel

twosom:
Linked issue: #162 #278

### Purpose of change

This PR enables explicit tool call handling in OllamaChatModel to align with the flink-agents framework's design principles for observability and tool execution lifecycle management.

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
@@ -80,11 +80,7 @@ public class ChatModelAction {
                 toolCallContext.put(initialRequestId, messages);
             }
             List<ChatMessage> messageContext =
-                    (List<ChatMessage>) toolCallContext.get(initialRequestId);
-
-            if (isImmutable(messageContext)) {
-                messageContext = new ArrayList<>(messageContext);
-            }
+                    new ArrayList<>((List<ChatMessage>) toolCallContext.get(initialRequestId));
 
             messageContext.add(response);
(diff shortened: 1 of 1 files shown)


wenjin272:
Hi, @twosom, thanks for your contribution. LGTM.

Additionally, I discover there are outputs like "Asking model: xxx" when run the example, I found that this maybe a log left by mistake in ollama4j, and it has been fixed recently.
image

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -118,6 +120,34 @@ public static void chat(
...
+     * <p>This method determines whether the list is an instance of Java's immutable collections by
+     * checking specific class name patterns used by the JDK implementation.
+     *
+     * <p>Note: This implementation is JDK-specific and may need updates if the JDK's internal
+     * implementation changes in future versions.
+     *
+     * @param maybeImmutableList The list of chat messages to check
+     * @return {@code true} if the list is an immutable collection, {@code false} otherwise
+     */
+    private static boolean isImmutable(List<ChatMessage> maybeImmutableList) {
```
xintongsong: This might be a bit fragile and hard to maintain. Maybe we can simply always create a copy of the message list. It's not perfect in terms of efficiency, which is probably okay since currently the performance bottleneck is in waiting for the model responses.

twosom: @xintongsong 
 Thanks for the feedback! I've updated it to always create a copy of the message list.

Issue #286, opened 2025-10-22
[Feature] Azure LLM Integration for the Agent


tsaiggo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Is Azure LLM integration for the agent on the roadmap? If we're considering this feature, I'd be happy to volunteer to take on this effort

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #290, opened 2025-10-29
[feature][java][chatmodels] Add Azure AI chat model integration

tsaiggo:
Linked issue: https://github.com/apache/flink-agents/issues/286

### Purpose of change

This PR adds support for the Azure AI chat model as a foundational model  for the agent.

### Tests

I added one test to verify this change, but I ran into some issues while running the test cases:

1. When I tried to run the `AgentWithOllamaExample`, the output stream didn't produce any data. Consequently, the same thing happened when I tested my new `AgentWithAzureAIExample`. My assumption is that this isn't related to the chat model implementation itself, but is more likely a higher-level issue (perhaps in the core agent logic or streaming setup).
2. Because of this, I switched to a different test case, `ReActAgentExample`. I modified it locally to use my Azure AI implementation, and it worked perfectly. However, I did not commit this change, as my PR should not introduce breaking changes to existing examples.
3. Finally, since this PR integrates a cloud-hosted model, I cannot check in my personal endpoint or API key. Therefore, I created the new `AgentWithAzureAIExample` and leveraged the `MockChatModel` (from `AgentWithResource`) to add mock tests. The "happy path" for this mock test passes successfully.

### API

No

### Documentation

Yes, exactly. It's because we now offer more model options.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -66,11 +66,6 @@ under the License.
         </dependency>
         <dependency>
             <groupId>org.apache.flink</groupId>
-            <artifactId>flink-agents-integrations-chat-models-azureai</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.flink</groupId>
             <artifactId>flink-agents-integrations-chat-models-ollama</artifactId>
             <version>${project.version}</version>
         </dependency>


integrations/chat-models/azureai/src/main/java/org/apache/flink/agents/integrations/chatmodels/azureai/AzureAIChatModelSetup.java
```diff
@@ -0,0 +1,99 @@
...
+    private final String key;
+    private final String endpoint;
+
+    public AzureAIChatModelSetup(
+            ResourceDescriptor descriptor,
+            java.util.function.BiFunction<String, ResourceType, Resource> getResource) {
+        super(descriptor, getResource);
+        this.model = descriptor.getArgument("model");
+        this.prompt = descriptor.getArgument("prompt");
+        this.tools = (java.util.List<String>) descriptor.getArgument("tools");
```
wenjin272: `prompt` and `tools` will be retrieved in constructor of base class `BaseChatModelSetup`, no need to process here again.

integrations/chat-models/azureai/src/main/java/org/apache/flink/agents/integrations/chatmodels/azureai/AzureAIChatModelSetup.java
```diff
@@ -0,0 +1,99 @@
...
+        this.key = descriptor.getArgument("key");
+        this.endpoint = descriptor.getArgument("endpoint");
```
wenjin272: `key` and `endpoint` should be  the arguments of `AzureAIChatModelConnection`. Different `AzureAIChatModelSetup` can share the same `AzureAIChatModelConnection`.

integrations/chat-models/azureai/src/main/java/org/apache/flink/agents/integrations/chatmodels/azureai/AzureAIChatModelSetup.java
```diff
@@ -0,0 +1,99 @@
...
+                                "endpoint", endpoint)),
+                getResource);
+    }
+
+    @Override
+    public java.util.Map<String, Object> getParameters() {
+        java.util.Map<String, Object> params = new java.util.HashMap<>();
+        params.put("model", model);
+        params.put("prompt", prompt);
+        params.put("tools", tools);
```
wenjin272: Base class `BaseChatModelSetup ` will process the `prompt` and `tools` directly, no need to pass them as parameters here.

integrations/chat-models/azureai/src/main/java/org/apache/flink/agents/integrations/chatmodels/azureai/AzureAIChatModelConnection.java
```diff
@@ -0,0 +1,183 @@
...
+    private ChatRequestMessage convertToChatRequestMessage(ChatMessage message) {
+        final String content = message.getContent();
+        final MessageRole role = message.getRole();
+        switch (role) {
+            case SYSTEM:
+                return new ChatRequestSystemMessage(content);
+            case USER:
+                return new ChatRequestUserMessage(content);
+            case ASSISTANT:
+                return new ChatRequestAssistantMessage(content);
```
wenjin272: Here, only the `content` of `ChatRequestAssistantMessage` is set, but no `toolCalls` are set.

integrations/chat-models/azureai/src/main/java/org/apache/flink/agents/integrations/chatmodels/azureai/AzureAIChatModelConnection.java
```diff
@@ -0,0 +1,183 @@
...
+            case TOOL:
+                return new ChatRequestToolMessage(content);
```
wenjin272: Here, the `content` will be set as `toolCallId` of `ChatRequestToolMessage`, so we should pass the function call id generated by Azure llm.
```
ChatRequestToolMessage msg =
                        new ChatRequestToolMessage(
                                (String) message.getExtraArgs().get("externalId"));
msg.setContent(content);
return msg;
```
and the `externalId` should be set in `convertToAgentsTools`:
```
private List<Map<String, Object>> convertToAgentsTools(
            List<ChatCompletionsToolCall> azureToolCalls) {
    ...
            final Map<String, Object> call =
                    Map.of(
                            "id", functionCall.getId(),
                            "original_id", functionCall.getId(),
                            "type", "function",
                            "function",
                                    Map.of(
                                            "name", functionCall.getFunction().getName(),
                                            "arguments",
                                                    functionCall.getFunction().getArguments()));
            toolCalls.add(call);
    ...
}

```

integrations/chat-models/azureai/src/main/java/org/apache/flink/agents/integrations/chatmodels/azureai/AzureAIChatModelConnection.java
```diff
@@ -0,0 +1,183 @@
...
+            throw new IllegalArgumentException("endpoint should not be null or empty.");
+        }
+        this.client =
+                new ChatCompletionsClientBuilder()
+                        .credential(new AzureKeyCredential(apiKey))
+                        .endpoint(endpoint)
+                        .buildClient();
+    }
+
+    @SuppressWarnings("unchecked")
```
wenjin272: My idea show this is a redundant suppression.

wenjin272 (commented):
Hi, @tsaiggo, thanks for your contribution. I left some minor comments.

tsaiggo:
pushed new changes, please re-review this PR, thx @wenjin272 
btw, there might be something wrong with the python-related test case, can you help retry the failed job? I have no access to do this

integrations/chat-models/azureai/src/main/java/org/apache/flink/agents/integrations/chatmodels/azureai/AzureAIChatModelSetup.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public class AzureAIChatModelSetup extends BaseChatModelSetup {
+
+    public AzureAIChatModelSetup(
+            ResourceDescriptor descriptor,
+            java.util.function.BiFunction<String, ResourceType, Resource> getResource) {
+        super(descriptor, getResource);
+    }
+
+    // For any other specific parameters, please refer to ChatCompletionsOptions    @Override
```
wenjin272: The `@Override` annotation is commented out.

wenjin272 (commented):
Hi, @tsaiggo, thanks for your update. LGTM, only one comment.

The failed test case in CI is caused by the request time out of ollama chat model. I fixed it also in pr #291, you can rebase main branch after it is merged.

integrations/chat-models/azureai/src/main/java/org/apache/flink/agents/integrations/chatmodels/azureai/AzureAIChatModelSetup.java
tsaiggo: Oops, good catch! 😂 Blame it on the late-night brain fog. Will fix it

tsaiggo: resolved

tsaiggo:
Looks like the issue of python-related test case has been self-recovered. 
lease let me know if you have any further comments or feedback—even nitpicks are welcome. If not, is this PR good to go? @wenjin272

examples/pom.xml
```diff
@@ -64,6 +64,11 @@ under the License.
             <artifactId>flink-clients</artifactId>
             <version>${flink.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-integrations-chat-models-azureai</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
xintongsong: Why do we need this dependency in examples/ ?

xintongsong (commented):
@tsaiggo, thanks for your contribution.

The PR overall looks good. I just have one minor comment.

Additionally, I think this new support for Azure chat model needs to be documented. We can either do that in this PR, or create a separate issue for it.

examples/pom.xml
tsaiggo: I prefer to keep documentation updates in a separate PR

tsaiggo: added the `azureaichatmodel` to the `examples` module because I intended it to be our first example of a cloud-hosted model. My original plan was to add specific use cases for cloud-hosted models, but I haven't come up with any good ideas for that yet, so it's temporarily on hold. If you think this addition should be removed for now, I'm happy to take care of that. @xintongsong

tsaiggo: removed

xintongsong: Yes, I think the dependency should be introduced only when it's needed. Thanks for addressing my comments.

xintongsong: FYI, I just created #305 for the documentation work.

xintongsong (approved):
LGTM. Merging.

Issue #310, opened 2025-11-07
[Feature] Introduce Sensory Memory


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce sensory memory to flink-agents. It is similar to short-term memory, but will be cleared when Agent Run finished.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #311, opened 2025-11-07
[api][runtime] Introduce Sensory Memory

wenjin272:
Linked issue: #310 

### Purpose of change

Introduc sensory memory to flink-agents, which is similar to short-term memory, but will be cleared when Agent Run finished.

### Tests

it test

### API

Yes

### Documentation

No, will add doc later

### DIFF
--- a/python/flink_agents/api/memory_reference.py
+++ b/python/flink_agents/api/memory_reference.py
@@ -21,44 +21,55 @@ from typing import TYPE_CHECKING, Any
 
 from pydantic import BaseModel, ConfigDict
 
+from flink_agents.api.memory_object import MemoryType
+
 if TYPE_CHECKING:
-    from flink_agents.api.memory_object import MemoryObject
+    from flink_agents.api.runner_context import RunnerContext
 
 
 class MemoryRef(BaseModel):
     """Reference to a specific data item in the Short-Term Memory."""
 
+    memory_type: MemoryType = MemoryType.SHORT_TERM
     path: str
 
     model_config = ConfigDict(frozen=True)
 
     @staticmethod
-    def create(path: str) -> MemoryRef:
+    def create(memory_type: MemoryType, path: str) -> MemoryRef:
         """Create a new MemoryRef instance based on the given path.
 
         Parameters
         ----------
         path: str
             The absolute path of the data in the Short-Term Memory.
+        memory_type:
+            The type of the memory object this reference points to.
 
         Returns:
         -------
         MemoryRef
             A new MemoryRef instance.
         """
-        return MemoryRef(path=path)
+        return MemoryRef(memory_type=memory_type, path=path)
 
-    def resolve(self, memory: MemoryObject) -> Any:
+    def resolve(self, ctx: RunnerContext) -> Any:
         """Resolve the reference to get the actual data.
 
         Parameters
         ----------
-        memory: MemoryObject
-            The memory object this ref points to.
+        ctx: RunnerContext
+            The current execution context, used to access Short-Term Memory.
 
         Returns:
         -------
         Any
             The deserialized, original data object.
         """
-        return memory.get(self)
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/apache/flink/agents/api/context/MemoryRef.java
```diff
@@ -47,12 +47,12 @@ public static MemoryRef create(String path) {
...
      * Resolves the reference using the provided RunnerContext to get the actual data.
      *
-     * @param ctx The current execution context, used to access Short-Term Memory.
+     * @param memory The memory this ref based on.
      * @return The deserialized, original data object.
      * @throws Exception if the memory cannot be accessed or the data cannot be resolved.
      */
-    public MemoryObject resolve(RunnerContext ctx) throws Exception {
-        return ctx.getShortTermMemory().get(this);
+    public MemoryObject resolve(MemoryObject memory) throws Exception {
```
Sxnan: Where is this method used? Is it only used in test case?

api/src/main/java/org/apache/flink/agents/api/context/RunnerContext.java
```diff
@@ -37,6 +37,14 @@ public interface RunnerContext {
      */
     void sendEvent(Event event);
 
+    /**
+     * Gets the sensory memory.
```
Sxnan, line 41: We should have a detailed explanation about the sensory memory.  When to use sensory memory vs short-term memory

python/flink_agents/api/runner_context.py
```diff
@@ -81,6 +81,17 @@ def get_action_config_value(self, key: str) -> Any:
             The config option value.
         """
 
+    @property
+    @abstractmethod
+    def sensory_memory(self) -> "MemoryObject":
+        """Get the sensory memory.
```
Sxnan, line 87: Need detailed documentation like Java API

python/flink_agents/runtime/local_runner.py
```diff
@@ -122,6 +128,18 @@ def get_action_config_value(self, key: str) -> Any:
             action_name=self.action_name, key=key
         )
 
+    @property
+    @override
+    def sensory_memory(self) -> MemoryObject:
+        """Get the short-term memory object associated with this context.
```
Sxnan:
```suggestion
        """Get the sensory memory object associated with this context.
```

api/src/main/java/org/apache/flink/agents/api/context/MemoryRef.java
wenjin272: This is a syntactic sugar  that user can use it to get the `MemoryObject` the `MemoryRef` referred to. User could use it in action. 

Currently, in example, we use
```
MemoryObject stm = ctx.getShortTermMemory();
ItemData originalData = (ItemData) stm.get(itemRef).getValue();
```
which is equal to
```
MemoryObject stm = ctx.getShortTermMemory();
ItemData originalData = (ItemData) itemRef.resolve(ctx.getShortTermMemory()).getValue();
```

xintongsong: We should store the information that whether a memory object belongs to short-term or long-term memory in the MemoryRef. Otherwise, we are leaving users the chance to make mistake and try to resolve the reference from the wrong memory type.

python/flink_agents/api/memory_reference.py
```diff
@@ -48,17 +48,17 @@ def create(path: str) -> MemoryRef:
         """
         return MemoryRef(path=path)
 
-    def resolve(self, ctx: RunnerContext) -> Any:
+    def resolve(self, memory: MemoryObject) -> Any:
```
xintongsong: ditto

xintongsong:
I accidentally merge this PR while it's still in review. I have reverted the changes. Let's open another PR and continue with the review.

Issue #313, opened 2025-11-12
[Feature] Support the use of Python ChatModel in Java


GreatEugenius:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support the use of Python ChatModel in Java Flink Agents jobs.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #314, opened 2025-11-12
[Feature][runtime] Support the use of Python ChatModel in Java

GreatEugenius:
Linked issue: #313

### Purpose of change
Support the use of Python ChatModel in Java Flink Agents jobs.

### Tests
ut

### API
no

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/python/PythonChatModelConnection.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/python/PythonChatModelConnection.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.api.chat.model.python;
+
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.apache.flink.agents.api.chat.model.BaseChatModelConnection;
+import org.apache.flink.agents.api.resource.Resource;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
+import org.apache.flink.agents.api.resource.ResourceType;
+import org.apache.flink.agents.api.resource.python.PythonResourceAdapter;
+import org.apache.flink.agents.api.resource.python.PythonResourceWrapper;
+import org.apache.flink.agents.api.tools.Tool;
+import pemja.core.object.PyObject;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiFunction;
+
+/**
+ * Python-based implementation of ChatModelConnection that wraps a Python chat model object. This
+ * class serves as a bridge between Java and Python chat model environments, but unlike {@link
+ * PythonChatModelSetup}, it does not provide direct chat functionality in Java.
+ */
+public class PythonChatModelConnection extends BaseChatModelConnection
+        implements PythonResourceWrapper {
+    private final PyObject chatModel;
+    private final PythonResourceAdapter adapter;
+
+    /**
+     * Creates a new PythonChatModelConnection.
+     *
+     * @param adapter The Python resource adapter (required by PythonResourceProvider's
+     *     reflection-based instantiation but not used directly in this implementation)
+     * @param chatModel The Python chat model object
+     * @param descriptor The resource descriptor
+     * @param getResource Function to retrieve resources by name and type
+     */
+    public PythonChatModelConnection(
+            PythonResourceAdapter adapter,
+            PyObject chatModel,
+            ResourceDescriptor descriptor,
+            BiFunction<String, ResourceType, Resource> getResource) {
+        super(descriptor, getResource);
+        this.chatModel = chatModel;
+        this.adapter = adapter;
+    }
+
+    @Override
+    public Object getPythonResource() {
+        return chatModel;
+    }
+
+    @Override
+    public ChatMessage chat(
+            List<ChatMessage> messages, List<Tool> tools, Map<String, Object> arguments) {
+        Map<String, Object> kwargs = new HashMap<>(arguments);
+
+        List<Object> pythonMessages = new ArrayList<>();
+        for (ChatMessage message : messages) {
+            pythonMessages.add(adapter.toPythonChatMessage(message));
+        }
+        kwargs.put("messages", pythonMessages);
+
+        List<Object> pythonTools = new ArrayList<>();
+        for (Tool tool : tools) {
+            pythonTools.add(adapter.convertToPythonTool(tool));
+        }
+        kwargs.put("tools", pythonTools);
+
+        Object pythonMessageResponse = adapter.callMethod(chatModel, "chat", kwargs);
+        return adapter.fromPythonChatMessage(pythonMessageResponse);
+    }
+}
--- a/api/src/main/java/org/apache/flink/agents/api/resource/python/PythonResourceAdapter.java
+++ b/api/src/main/java/org/apache/flink/agents/api/resource/python/PythonResourceAdapter.java
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.resource.python;
+
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.apache.flink.agents.api.tools.Tool;
+import pemja.core.object.PyObject;
+
+import java.util.Map;
+
+/**
+ * Adapter interface for managing Python resources and facilitating Java-Python interoperability.
+ * This interface provides methods to interact with Python objects, invoke Python methods, and
+ * handle data conversion between Java and Python environments.
+ */
+public interface PythonResourceAdapter {
+
+    /**
+     * Retrieves a Python resource by name and type.
+     *
+     * @param resourceName the name of the resource to retrieve
+     * @param resourceType the type of the resource
+     * @return the retrieved resource object
+     */
+    Object getResource(String resourceName, String resourceType);
+
+    /**
+     * Initializes a Python resource instance from the specified module and class.
+     *
+     * @param module the Python module containing the target class
+     * @param clazz the Python class name to instantiate
+     * @param kwargs keyword arguments to pass to the Python class constructor
+     * @return a PyObject representing the initialized Python resource
+     */
+    PyObject initPythonResource(String module, String clazz, Map<String, Object> kwargs);
+
+    /**
+     * Converts a Java ChatMessage object to its Python equivalent.
+     *
+     * @param message the Java ChatMessage to convert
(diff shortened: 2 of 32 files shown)


GreatEugenius:
@Sxnan @wenjin272 please help take a look this PR, thanks.

plan/src/main/java/org/apache/flink/agents/plan/resourceprovider/PythonResourceProvider.java
```diff
@@ -46,6 +54,25 @@ public PythonResourceProvider(
...
+        super(name, type);
+        module = descriptor.getArgument("module");
+        if (module == null || module.isEmpty()) {
+            throw new IllegalArgumentException("module should not be null or empty.");
+        }
+        clazz = descriptor.getArgument("clazz");
+        if (clazz == null || clazz.isEmpty()) {
+            throw new IllegalArgumentException("clazz should not be null or empty.");
+        }
+        this.kwargs = new HashMap<>(descriptor.getInitialArguments());
```
wenjin272: This kwargs may need remove `module` and `clazz`.

Or we can just add a new implementation `PythonResourceDescriptor`, which has additional property `pythonMoudel` and `pythonClass`. User can directly use it when declare a resource in Agent.

plan/src/main/java/org/apache/flink/agents/plan/resourceprovider/PythonResourceProvider.java
```diff
@@ -63,11 +90,22 @@ public Map<String, Object> getKwargs() {
...
     public Resource provide(BiFunction<String, ResourceType, Resource> getResource)
             throws Exception {
-        // TODO: Implement Python resource creation logic
-        // This would typically involve calling into Python runtime to create the
-        // resource
-        throw new UnsupportedOperationException(
-                "Python resource creation not yet implemented in Java runtime");
+        if (pythonResourceAdapter != null) {
+            Class<?> clazz = Class.forName(descriptor.getClazz());
+            Map<String, Object> kwargs = new HashMap<>(descriptor.getInitialArguments());
```
wenjin272: Why not use class data member but create `clazz` and `kwargs` again here?

plan/src/test/resources/resource_providers/python_resource_provider.json
```diff
@@ -3,6 +3,7 @@
   "type" : "chat_model",
   "module" : "flink_agents.plan.tests.test_resource_provider",
   "clazz" : "MockChatModelImpl",
+  "descriptor" :null,
```
wenjin272: This json is serialized from python agent plan, but `PythonResourceProvider` in python won't generate `descriptor` field.

plan/src/main/java/org/apache/flink/agents/plan/resourceprovider/PythonResourceProvider.java
```diff
@@ -46,6 +54,25 @@ public PythonResourceProvider(
         this.module = module;
         this.clazz = clazz;
         this.kwargs = kwargs;
+        this.descriptor = null;
```
wenjin272, line 59: Should we add `descriptor` field for  `PythonResourceProvider` deserialized from agent plan declared using python api?

plan/src/main/java/org/apache/flink/agents/plan/serializer/ResourceProviderJsonDeserializer.java
```diff
@@ -76,6 +76,15 @@ public ResourceProvider deserialize(
     private PythonResourceProvider deserializePythonResourceProvider(JsonNode node) {
         String name = node.get("name").asText();
         String type = node.get("type").asText();
+        try {
+            ResourceDescriptor descriptor =
+                    mapper.treeToValue(node.get("descriptor"), ResourceDescriptor.class);
+            if (descriptor != null) {
```
wenjin272: We can check the `descriptor` field existence rather than if the field is null.

wenjin272 (commented):
Hi, @GreatEugenius, thanks for your work. Overall looks good to me . I left some comments about details.

Besides, I think this pr should add some test cases to show how to declare an agent in java using python chat model, and verify the behavior.

api/src/main/java/org/apache/flink/agents/api/prompt/Prompt.java
```diff
@@ -76,6 +77,14 @@ public String formatString(Map<String, String> kwargs) {
                 });
     }
 
+    public List<ChatMessage> formatMessages(String defaultRoleValue, Map<String, Object> kwargs) {
```
Sxnan: Why do we need to pass the Object type as the value here? Can't we reuse the `formatMessages(MessageRole defaultRole, Map<String, String> kwargs)`?

This method is not used in Java. We should have documentation to say this is intended to be used by Python code.

api/src/main/java/org/apache/flink/agents/api/chat/model/python/PythonChatModelConnection.java
```diff
@@ -0,0 +1,63 @@
...
+ * Python-based implementation of ChatModelConnection that wraps a Python chat model object. This
+ * class serves as a bridge between Java and Python chat model environments, but unlike {@link
+ * PythonChatModelSetup}, it does not provide direct chat functionality in Java.
+ */
+public class PythonChatModelConnection extends BaseChatModelConnection
+        implements PythonResourceWrapper {
+    private PyObject chatModel;
+
+    public PythonChatModelConnection(
+            PythonResourceAdapter adapter,
```
Sxnan, line 56: The adapter is not used. If it must be here for some reason, please document it

api/src/main/java/org/apache/flink/agents/api/resource/python/PythonChatModelSetup.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Python-based implementation of ChatModelSetup that bridges Java and Python chat model
+ * functionality. This class wraps a Python chat model setup object and provides Java interface
+ * compatibility while delegating actual chat operations to the underlying Python implementation.
+ */
+public class PythonChatModelSetup extends BaseChatModelSetup implements PythonResourceWrapper {
+    private final PyObject chatModelSetup;
+    private final PythonResourceAdapter adapter;
+
+    public PythonChatModelSetup(
```
Sxnan: Where is this class used?

api/src/main/java/org/apache/flink/agents/api/resource/python/PythonResourceAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+     */
+    Object callMethod(Object obj, String methodName, Map<String, Object> kwargs);
+
+    /**
+     * Converts a Java ChatMessage object to its Python equivalent.
+     *
+     * @param message the Java ChatMessage to convert
+     * @return the Python representation of the chat message
+     */
+    Object toPythonChatMessage(ChatMessage message);
```
Sxnan, line 59: Does `ChatMessage` only make sense to ChatModel? If so, we should not put the method in this General Resource Adapter interface

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -76,6 +79,8 @@ public class AgentPlan implements Serializable {
 
     private AgentConfiguration config;
 
+    private PythonResourceAdapter pythonResourceAdapter;
```
Sxnan: Is this serializable? Should it be transient?

python/flink_agents/api/prompts/prompt.py
```diff
@@ -22,10 +22,10 @@
 
 from flink_agents.api.chat_message import ChatMessage, MessageRole
 from flink_agents.api.prompts.utils import format_string
-from flink_agents.api.resource import ResourceType, SerializableResource
+from flink_agents.api.resource import Resource, ResourceType, SerializableResource
 
 
-class Prompt(SerializableResource, ABC):
+class Prompt(Resource, ABC):
```
Sxnan: This change also affected  MCPPrompt

python/flink_agents/runtime/java/java_resource_wrapper.py
Sxnan, line 1: Typo in file name, should be java_resource_wrapper.py

Sxnan (commented):
Thanks for the PR. However, the PR lacks tests. For such a complex feature, we need to have unit tests and e2e tests.

python/flink_agents/api/prompts/prompt.py
wenjin272: +1 for this.

If this change is just to pass the serialization check of `SerializableResource ` for `JavaPrompt`, you can annotate the `j_prompt` ignored, like

```
class JavaPrompt(Prompt):
    """Python wrapper for Java's Prompt."""

    j_prompt: Any = Field(exclude=True)
```

api/src/main/java/org/apache/flink/agents/api/prompt/Prompt.java
GreatEugenius: I don't believe we reuse the `formatMessages(MessageRole defaultRole, Map<String, String> kwargs)`. In the Java chat model, the logic is implemented in the `chat` method of BaseChatModelSetup.java. However, for the Python chat model, we need to implement it here.

GreatEugenius:
Hi @Sxnan and @wenjin272, thank you for the review! I have updated this PR and added a PythonChatModel example for the end-to-end test. Please take a look. Thanks!

plan/src/main/java/org/apache/flink/agents/plan/resourceprovider/PythonResourceProvider.java
```diff
@@ -46,6 +54,27 @@ public PythonResourceProvider(
...
+        if (module == null || module.isEmpty()) {
+            throw new IllegalArgumentException("module should not be null or empty.");
+        }
+        clazz = descriptor.getArgument("clazz");
+        if (clazz == null || clazz.isEmpty()) {
+            throw new IllegalArgumentException("clazz should not be null or empty.");
+        }
+        this.kwargs = new HashMap<>(descriptor.getInitialArguments());
+        this.kwargs.remove("module");
+        this.kwargs.remove("clazz");
```
wenjin272: Maybe we can just simplify the code to 
```
this.kwargs = new HashMap<>(descriptor.getInitialArguments());
module = (String) kwargs.remove("module");
if (module == null || module.isEmpty()) {
    throw new IllegalArgumentException("module should not be null or empty.");
}
clazz = (String) kwargs.remove("clazz");
if (clazz == null || clazz.isEmpty()) {
    throw new IllegalArgumentException("clazz should not be null or empty.");
}
```

e2e-test/integration-test/src/main/java/org/apache/flink/agents/integration/test/AgentWithPythonChatModel.java
```diff
@@ -0,0 +1,144 @@
...
+                .addInitialArgument("clazz", "TongyiChatModelConnection")
+                .build();
+    }
+
+    @ChatModelSetup
+    public static ResourceDescriptor pythonChatModel() {
+        return ResourceDescriptor.Builder.newBuilder(PythonChatModelSetup.class.getName())
+                .addInitialArgument("connection", "pythonChatModelConnection")
+                .addInitialArgument(
+                        "module", "flink_agents.integrations.chat_models.tongyi_chat_model")
```
Sxnan: Should we use ollama model for e2e test?

api/src/main/java/org/apache/flink/agents/api/prompt/Prompt.java
Sxnan: Why can't Python chat model call `formatMessages(MessageRole defaultRole, Map<String, String> kwargs)` but `formatMessages(MessageRole defaultRole, Map<String, Object> kwargs)`?

GreatEugenius: The problem was a type mismatch between:
ChatMessage.extra_args which is typed as Dict[str, Any] (equivalent to Map<String, Object> in Java)
Prompt.format_messages() method signature which expects **kwargs: str (equivalent to Map<String, String> in Java)
When we directly passed msg.extra_args to format_messages(**input_variable), we were essentially trying to pass Any typed values where str types were expected, causing the type inconsistency.

GreatEugenius:
Hi @Sxnan, thank you for the review! I have updated this PR; please review it again.

api/src/main/java/org/apache/flink/agents/api/chat/model/python/PythonChatModelConnection.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public Object getPythonResource() {
+        return chatModel;
+    }
+
+    @Override
+    public ChatMessage chat(
+            List<ChatMessage> messages, List<Tool> tools, Map<String, Object> arguments) {
+        throw new UnsupportedOperationException(
```
xintongsong: Why is this not supported?

plan/src/main/java/org/apache/flink/agents/plan/resourceprovider/PythonResourceProvider.java
```diff
@@ -63,11 +90,21 @@ public Map<String, Object> getKwargs() {
...
+                    clazz.getConstructor(
+                            PythonResourceAdapter.class,
+                            PyObject.class,
+                            ResourceDescriptor.class,
+                            BiFunction.class);
+            return (Resource)
+                    constructor.newInstance(
+                            pythonResourceAdapter, pyResource, descriptor, getResource);
+        }
+        throw new UnsupportedOperationException("PythonResourceAdapter is not set.");
```
xintongsong: Use `checkState`

xintongsong (approved):
LGTM

Issue #323, opened 2025-11-19
[Bug] Ollama4j cant't work with ollama v0.12.10 and above


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

If the ollama response contains tool call, the ollama4j will raise exceptions like:
```
io.github.ollama4j.exceptions.OllamaException: Unrecognized field "id" (class io.github.ollama4j.models.chat.OllamaChatToolCalls)
```
See https://github.com/ollama4j/ollama4j/issues/237 for details

### How to reproduce

Run `WorkflowSingleAgentExample` with latest version ollama server

### Version and environment

main branch, but 0.1.0 should has the same question.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #324, opened 2025-11-19
[hotfix] Update ollam4j version to resolve tool call exception.

wenjin272:
Linked issue: #323 

### Purpose of change

Update the ollama4j version to resolve the tool call bug in ollama4j.

### Tests

manually run e2e test

### API

No

### Documentation

- [x] `doc-not-needed`


wenjin272:
The impact of this bug is that if users are using the latest version of Ollama, the ollama4j library will fail to parse the tool call returned by the LLM, but it works with previous version ollama.

We have refactored  ollama chat model after 0.1.0 released, so if we want fix this issue in 0.1.1, we should create an individual pr, rather than pick this pr directly.

Sxnan:
@wenjin272 Can you also create a PR against 0.1.0 to ensure it is not forgotten?

Issue #331, opened 2025-11-20
[Feature] Introduce Long-Term Memory in Python


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce Long-Term Memory in Python

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #332, opened 2025-11-20
[api][runtime] Introduce long-term memory in python

wenjin272:
Linked issue: #331
### Purpose of change

Introduce the long-term memory interface in python, and provide an implementation based on chroma.

This is the first pr of three to introduce long-term memory in python:

1. interface and one implementation
2. support using long-term memory in action
3. async interface and execution

### Tests

Unit test

### API

Yes, add long-term memory related api.

### Documentation

- [x] `doc-needed`

### DIFF
--- a/python/flink_agents/api/memory/long_term_memory.py
+++ b/python/flink_agents/api/memory/long_term_memory.py
@@ -1,0 +1,313 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#################################################################################
+import importlib
+from abc import ABC, abstractmethod
+from datetime import datetime
+from enum import Enum
+from typing import Any, Dict, List, Type
+
+from pydantic import (
+    BaseModel,
+    Field,
+    field_serializer,
+    model_validator,
+)
+from typing_extensions import override
+
+from flink_agents.api.chat_message import ChatMessage
+from flink_agents.api.prompts.prompt import Prompt
+
+ItemType = str | ChatMessage
+
+
+class CompactionStrategyType(Enum):
+    """Strategy for compact memory set."""
+
+    SUMMARIZATION = "summarization"
+
+
+class CompactionStrategy(BaseModel, ABC):
+    """Strategy for compact memory set."""
+
+    @property
+    @abstractmethod
+    def type(self) -> CompactionStrategyType:
+        """Return type of this strategy."""
+
+
+class SummarizationStrategy(CompactionStrategy):
+    """Summarization strategy.
+
+    Attributes:
+        model: The name of the llm used for generating the summarization.
+        prompt: The prompt instance or  name of the prompt used for generating
+            the summarization. Optional.
+        limit: How many summarization should generate for the context. Defaults to 1.
+    """
+
+    model: str
+    prompt: str | Prompt | None = None
+    limit: int = 1
+
+    @property
+    @override
+    def type(self) -> CompactionStrategyType:
+        return CompactionStrategyType.SUMMARIZATION
+
+
+class LongTermMemoryBackend(Enum):
+    """Backend for Long-Term Memory."""
+
+    EXTERNAL_VECTOR_STORE = "external_vector_store"
+
+
+class DatetimeRange(BaseModel):
+    """Represents a datetime range."""
+
+    start: datetime
+    end: datetime
+
+
+class MemorySetItem(BaseModel):
+    """Represents a long term memory item retrieved from vector store.
+
+    Attributes:
+        memory_set_name: The name of the memory set this item belongs to.
+        id: The id of this item.
+        value: The value of this item.
+        compacted: Whether this item has been compacted.
+        created_time: The timestamp this item was added to the memory set.
+        last_accessed_time: The timestamp this item was last accessed.
+        additional_metadata: Additional metadata for this item.
+    """
+
+    memory_set_name: str
+    id: str
+    value: Any
+    compacted: bool = False
+    created_time: datetime | DatetimeRange = None
+    last_accessed_time: datetime
+    additional_metadata: Dict[str, Any] | None = None
+
+
+class MemorySet(BaseModel):
+    """Represents a long term memory set contains memory items.
+
+    Attributes:
+        name: The name of this memory set.
+        item_type: The type of items stored in this set.
+        capacity: The capacity of this memory set.
+        compaction_strategy: Compaction strategy and additional arguments used
+        to compact memory set.
+    """
+
+    name: str
+    item_type: Type[str] | Type[ChatMessage]
+    capacity: int
+    compaction_strategy: CompactionStrategy
+    ltm: "BaseLongTermMemory" = Field(default=None, exclude=True)
+
+    @field_serializer("item_type")
+    def _serialize_item_type(self, item_type: Type) -> Dict[str, str]:
+        return {"module": item_type.__module__, "name": item_type.__name__}
+
+    @field_serializer("compaction_strategy")
+    def _serialize_compaction_strategy(
+        self, compaction_strategy: CompactionStrategy
+    ) -> Dict[str, str]:
+        data = compaction_strategy.model_dump()
+        data.update(
+            {
+                "module": compaction_strategy.__class__.__module__,
+                "name": compaction_strategy.__class__.__name__,
+            }
+        )
+        return data
+
+    @model_validator(mode="before")
+    def _deserialize_item_type(self) -> "MemorySet":
+        if isinstance(self["item_type"], Dict):
+            module = importlib.import_module(self["item_type"]["module"])
+            self["item_type"] = getattr(module, self["item_type"]["name"])
+        if isinstance(self["compaction_strategy"], Dict):
+            module = importlib.import_module(self["compaction_strategy"].pop("module"))
+            clazz = getattr(module, self["compaction_strategy"].pop("name"))
+            self["compaction_strategy"] = clazz.model_validate(
+                self["compaction_strategy"]
+            )
+        return self
+
+    @property
+    def size(self) -> int:
+        """The count of items of this memory set."""
+        return self.ltm.size(memory_set=self)
+
+    def add(
+        self, items: ItemType | List[ItemType], ids: str | List[str] | None = None
+    ) -> None:
+        """Add a memory item to the set, currently only support item with
+        type str or ChatMessage.
+
+        If the capacity of this memory set is reached, will trigger reduce
+        operation to manage the memory set size.
+
+        Args:
+            items: The items to be inserted to this set.
+            ids: The ids of the items to be inserted. Optional.
+        """
+        self.ltm.add(memory_set=self, memory_items=items, ids=ids)
+
+    def get(
+        self, ids: str | List[str] | None = None
+    ) -> MemorySetItem | List[MemorySetItem]:
+        """Retrieve memory items. If no item id provided, will return all items.
+
+        Args:
+            ids: The ids of the items to retrieve.
+
+        Returns:
+            The memory items retrieved.
+        """
+        return self.ltm.get(memory_set=self, ids=ids)
+
+    def search(self, query: str, limit: int, **kwargs: Any) -> List[MemorySetItem]:
+        """Retrieve n memory items related to the query.
+
+        Args:
+            query: The query to search for.
+            limit: The number of items to retrieve.
+            **kwargs: Additional arguments for search.
+        """
+        return self.ltm.search(memory_set=self, query=query, limit=limit, **kwargs)
+
+
+class BaseLongTermMemory(ABC, BaseModel):
+    """Base Abstract class for long term memory."""
+
+    @abstractmethod
+    def get_or_create_memory_set(
+        self,
+        name: str,
+        item_type: str | Type[ChatMessage],
+        capacity: int,
+        compaction_strategy: CompactionStrategy,
+    ) -> MemorySet:
+        """Create a memory set, if the memory set already exists, return it.
+
+        Args:
+            name: The name of the memory set.
+            item_type: The type of the memory item.
+            capacity: The capacity of the memory set.
+            compaction_strategy: The compaction strategy and arguments for
+            storge management.
+
+        Returns:
+            The created memory set.
+        """
+
+    @abstractmethod
+    def get_memory_set(self, name: str) -> MemorySet:
+        """Get the memory set.
+
+        Args:
+            name: The name of the memory set.
+
+        Returns:
+            The memory set.
+        """
+
+    @abstractmethod
+    def delete_memory_set(self, name: str) -> bool:
+        """Delete the memory set.
+
+        Args:
+            name: The name of the memory set.
+
+        Returns:
+            Whether the memory set was deleted.
+        """
+
+    @abstractmethod
+    def size(self, memory_set: MemorySet) -> int:
+        """Get the size of the memory set.
+
+        Args:
+            memory_set: The memory set to count.
+        """
+
+    @abstractmethod
+    def add(
+        self,
+        memory_set: MemorySet,
+        memory_items: ItemType | List[ItemType],
+        ids: str | List[str] | None = None,
+        metadatas: Dict[str, Any] | List[Dict[str, Any]] | None = None,
+    ) -> None:
+        """Add items to the memory set, currently only support items with
+        type str or ChatMessage.
+
+        This method may trigger compaction to manage the memory set size.
+
+        Args:
+            memory_set: The memory set to add to.
+            memory_items: The items to be added to this set.
+            ids: The IDs of items. Will be automatically generated if not provided.
+            Optional.
+            metadatas: The metadata for items. Optional.
+        """
+
+    @abstractmethod
+    def get(
+        self, memory_set: MemorySet, ids: str | List[str] | None = None
+    ) -> MemorySetItem | List[MemorySetItem]:
+        """Retrieve memory items. If no item id provided, return all items.
+
+        Args:
+            memory_set: The set to be retrieved.
+            ids: The ids of the items to retrieve. If not provided, all items will
+            be retrieved. Optional.
+
+        Returns:
+            The memory items retrieved.
+        """
+
+    @abstractmethod
+    def delete(self, memory_set: MemorySet, ids: str | List[str] | None = None) -> None:
+        """Delete memory items. If no item id provided, delete all items.
+
+        Args:
+            memory_set: The memory set to delete from.
+            ids: The ids of items to be deleted, If not provided, all items will be
+            deleted. Optional.
+        """
+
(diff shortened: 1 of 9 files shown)


python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,263 @@
...
+    Field,
+    field_serializer,
+    model_validator,
+)
+
+from flink_agents.api.chat_message import ChatMessage
+from flink_agents.api.prompts.prompt import Prompt
+
+
+class ReduceStrategy(Enum):
```
xintongsong: Not sure about the name `Reduce`. I think `Compact` might be better.

python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,263 @@
...
+    """Strategy for reducing memory set size."""
+
+    TRIM = "trim"
+    SUMMARIZE = "summarize"
+
+
+class ReduceSetup(BaseModel):
```
xintongsong: I'd suggest to name this `CompactionStrategy`, and make it an abstract class that we can provide different implementations, so we can have strict limit on which arguments should be specified for each strategy. We can call the current `ReduceStrategy` `CompactionStrategyType`.

xintongsong: I think `CompactionStrategy.trim(n)` might be more straightforward for users, compared to `ReduceSetup.trim_setup(n)`.

python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,263 @@
...
+        compacted: Whether this item has been compacted.
+        created_time: The timestamp this item was added to the memory set.
+        last_accessed_time: The timestamp this item was last accessed.
+    """
+
+    memory_set_name: str
+    id: str
+    value: Any
+    compacted: bool = False
+    created_time: DatetimeRange
```
xintongsong:
```suggestion
    created_time: DatetimeRange | datetime
```

python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,263 @@
...
+        item_ids: The indices of items stored in this set.
+        reduced: Whether this memory set has been reduced.
+    """
+
+    name: str
+    item_type: Type[str] | Type[ChatMessage]
+    size: int = 0
+    capacity: int
+    reduce_setup: ReduceSetup
+    item_ids: List[str] = Field(default_factory=list)
```
xintongsong: Why do we need to store ids of all the items?

python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,263 @@
...
+    reduced: bool = False
```
xintongsong: And why do we need to know whether a memory set has been reduced/compacted?

python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,263 @@
...
+
+        Args:
+            name: The name of the memory set.
+
+        Returns:
+            The memory set.
+        """
+
+    @abstractmethod
+    def delete_memory_set(self, name: str) -> None:
```
xintongsong: Usually, for a delete method, we return a boolean of the deleted element / null to indicate whether the element has been deleted.

python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,263 @@
...
+        return ReduceSetup(
+            strategy=ReduceStrategy.SUMMARIZE,
+            arguments={"n": n, "model": model, "prompt": prompt},
+        )
+
+
+class LongTermMemoryBackend(Enum):
+    """Backend for Long-Term Memory."""
+
+    CHROMA = "chroma"
```
xintongsong: We probably should either support a embedded backend only, or support any external vector store.

python/flink_agents/runtime/memory/chroma_long_term_memory.py
```diff
@@ -0,0 +1,456 @@
...
+        default=None,
+        description="Custom embedding model used to generate item embedding. "
+        "If not provided, will use the default embedding function of chroma.",
+    )
+    runner_context: RunnerContext = Field(
+        default="The runner context for this long term memory.",
+    )
+
+    # Connection configuration
+    persist_directory: str | None = Field(
```
xintongsong: what is this directory for?

python/flink_agents/runtime/memory/chroma_long_term_memory.py
```diff
@@ -0,0 +1,456 @@
...
+
+    @override
+    def delete_memory_set(self, name: str) -> None:
+        self.client.delete_collection(name=name)
+
+    @override
+    def add(self, memory_set: MemorySet, memory_item: str | ChatMessage) -> None:
+        if memory_set.size >= memory_set.capacity:
+            # trigger reduce operation to manage memory set size.
+            self._reduce(memory_set)
```
xintongsong: This can be extremely slow. We should proactively do the compaction.

python/flink_agents/runtime/memory/chroma_long_term_memory.py
```diff
@@ -0,0 +1,456 @@
...
+        memory_set = MemorySet.model_validate_json(metadata["serialization"])
+        memory_set.ltm = self
+        return memory_set
+
+    @override
+    def delete_memory_set(self, name: str) -> None:
+        self.client.delete_collection(name=name)
+
+    @override
+    def add(self, memory_set: MemorySet, memory_item: str | ChatMessage) -> None:
```
xintongsong: I had a feeling that adding items to long-term memory can take time, for embedding. We probably should also provide async apis.

python/flink_agents/runtime/memory/chroma_long_term_memory.py
```diff
@@ -0,0 +1,456 @@
...
+    def get(self, memory_set: MemorySet) -> List[MemorySetItem]:
+        return self.slice(memory_set=memory_set)
+
+    @override
+    def get_recent(self, memory_set: MemorySet, n: int) -> List[MemorySetItem]:
+        offset = memory_set.size - n if memory_set.size > n else 0
+        return self.slice(memory_set=memory_set, offset=offset, n=n)
+
+    @override
+    def search(
```
xintongsong: Same here.

python/flink_agents/runtime/memory/chroma_long_term_memory.py
```diff
@@ -0,0 +1,456 @@
...
+            raise RuntimeError(msg)
+        memory_set.reduced = True
+        self._update_set_metadata(memory_set)
+
+    def _trim(self, memory_set: MemorySet) -> None:
+        reduce_setup: ReduceSetup = memory_set.reduce_setup
+        n = reduce_setup.arguments.get("n")
+        self.delete(memory_set=memory_set, offset=0, n=n)
+
+    def _summarize(self, memory_set: MemorySet) -> None:
```
xintongsong: Are these methods specialized for this class?

python/flink_agents/api/memory/long_term_memory.py
alnzng: - +1 on that `Compaction` is a better terminology. This `Compaction` term is commonly used in many open source software in the industry today. 
- Should we consider using noun as the name of each strategy. For example, `summarize` -> `summarization`, `trim` -> `truncation`?

alnzng:
> Hi, @alnzng. There's a design issue related to the vector store that I'd appreciate your help reviewing.
> 
> As describe in the design doc #339, long-term memory of flink-agents is also based on vector store. Currently, I provide an implementation based on chroma. In this implementation, I directly use chroma client rather than flink-agents `BaseVectorStore`, because there are some long-term memory needed interface not provided in `BaseVectorStore`.
> 
> @xintongsong believes that we can directly build upon the Flink-Agents `BaseVectorStore`. Thus, we can support using any already supported vector store as the backend for long-term memory.
> 
> I think this make sense, but it requires add some interfaces to `BaseVectorStore`, which maybe look like:
> 
> ```
> def get_or_create_collection(self, name: str, metadata: Dict[str, Any]) -> Dict[str, Any]:
>     """Get a collection, create if it doesn't already exist."""
>     
> def get_collection(self, name: str) -> None:
>     """Get an existing collection."""
>     
> def update_collection(self, name: str, metadata: Dict[str, Any]) -> None:
>     """Update an existing collection."""
> 
> def delete_collection(self, name: str) -> bool:
>     """Delete a collection."""
>     
> def add(self, document: Document, collection_name: str | None = None) -> None:
>     """Add a document to the collection."""
> 
> def update(self, document: Document, collection_name: str | None = None) -> None:
>     """Update a document, can only update metadata."""
> 
> def delete(self, offset: int | None, limit: int | None, ids: List[int] | None = None, **kwargs: Any) -> bool:
>     """Delete documents from collection."""
> 
> def get(self, offset: int | None, limit: int | None, collection_name: str | None = None, **kwargs: Any) -> List[Document]:
>     """Get documents from collection."""
> ```
> 
> These interface may not be achievable for each vector store, I will conduct research and refinement afterward. WDTY?

Thanks for raising this design question @wenjin272! I agree with @xintongsong that building long-term memory on top of `BaseVectorStore` would be beneficial for consistency and reusability.

The current `BaseVectorStore` is intentionally minimal, focusing on read-only semantic search. This design works well for RAG retrieval use cases but indeed lacks the CRUD operations needed for long-term memory management. Adding document-level CRUD operations to `BaseVectorStore` makes sense, as these operations are generally supported across all major vector stores.
 
A few thoughts on the proposed methods:
  - add(), delete(), get() - Widely supported, should be included
  - update() - I'm not certain about common use cases for updating existing long-term memory entries, but it shouldn't be
  problematic to include.

However, I have concerns about including collection management in the base interface. This concept is vector
store specific and may cause integration issues. "Collection" is most common term (e.g. Chroma, Milvus, Qdrant, etc), but Pinecone and Weaviate use different terms(e.g. Class, Namespace, etc).

My suggestion: Keep collection management implementation specific rather than in the base interface. Each vector store integration can expose collection management using its native terminology and patterns.

This approach aligns with how `LlamaIndex` and `LangChain` handle vector stores. I'd recommend you reviewing these implementations, they provide good precedents for balancing abstraction with flexibility.
- LlamaIndex BasePydanticVectorStore:
  https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/vector_stores/types.py#L334
- LangChain VectorStore: https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/vectorstores/base.py#L43

python/flink_agents/runtime/memory/tests/start_chroma_server.sh
```diff
@@ -0,0 +1,23 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+
+path=$1
+chroma run --path $path
```
alnzng: is this script must to have? 

IIUC, You can just add "import chromadb" in the test class which will "start" the chroma automatically. I did this in `test_chroma_vector_store`.

xintongsong:
Maybe we can have two separate interfaces, something like: `vector_store` & `collection_manageable` (or `collection_manageable_vector_store` who extends `vector_store`). 
- In order to be used as a backend of LTM, the integration must implements both interfaces.
- For integration that doesn't implement `collection_manageable` (incompatible, or simply haven't done it), it can still be used for rag-like use cases as long as it implements `vector_store`.

python/flink_agents/runtime/memory/tests/start_chroma_server.sh
wenjin272: IIUC, the `import chromadb` will start the chroma in In-Memory mode. For long-term memory, we recommend Server-Client mode or Cloud mode for data persistence, so I start a chroma server here.

python/flink_agents/api/vector_stores/vector_store.py
```diff
@@ -176,3 +250,74 @@ def query_embedding(self, embedding: List[float], limit: int = 10, **kwargs: Any
         Returns:
             List of documents matching the search criteria
         """
+
+    @abstractmethod
+    def add_embedding(
```
xintongsong: `add_embedding` and `query_embedding` are not meant for users to call. We probably should mark them as protected.

python/flink_agents/api/vector_stores/vector_store.py
```diff
@@ -176,3 +250,74 @@ def query_embedding(self, embedding: List[float], limit: int = 10, **kwargs: Any
...
+
+        Returns:
+            List of document IDs that were added to the vector store
+        """
+
+
+class Collection(BaseModel):
+    """Represents a collection of documents."""
+    name: str
+    size: int
```
xintongsong: What is this size used for and how do we keep it consistent?

python/flink_agents/api/vector_stores/vector_store.py
```diff
@@ -176,3 +250,74 @@ def query_embedding(self, embedding: List[float], limit: int = 10, **kwargs: Any
...
+    def delete_collection(self, name: str) -> Collection:
+        """Delete a collection.
+
+        Args:
+            name: Name of the collection
+        Returns:
+            The deleted collection
+        """
+
+def maybe_cast_to_list(value: Any | List[Any]) -> List[Any] | None:
```
xintongsong: Should this be private?

python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,294 @@
...
+    @property
+    @override
+    def type(self) -> CompactionStrategyType:
+        return CompactionStrategyType.SUMMARIZATION
+
+
+class LongTermMemoryBackend(Enum):
+    """Backend for Long-Term Memory."""
+
+    VectorStore = "vectorstore"
```
xintongsong:
```suggestion
    EXTERNAL_VECTOR_STORE = "external_vector_store"
```

python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,294 @@
...
+        compaction_strategy: Compaction strategy and additional arguments used
+        to compact memory set.
+        size: The size of this memory set.
+    """
+
+    name: str
+    item_type: Type[str] | Type[ChatMessage]
+    capacity: int
+    compaction_strategy: CompactionStrategy
+    size: int = Field(default=0, exclude=True)
```
xintongsong: It's fragile to maintain the size separately from the underlying actual collection. We might consider get the size from the store.

python/flink_agents/runtime/memory/compaction_functions.py
```diff
@@ -0,0 +1,149 @@
...
+)
+from flink_agents.api.resource import ResourceType
+from flink_agents.api.runner_context import RunnerContext
+
+if TYPE_CHECKING:
+    from flink_agents.api.chat_models.chat_model import BaseChatModelSetup
+    from flink_agents.api.prompts.prompt import Prompt
+
+
+def summarize(
```
xintongsong, line 59: It doesn't make sense to always compact the entire long term memory into 1 message. I think we should only merge similar messages, and discard meaning less one. As a result, we should get a smaller set of messages with higher information density.

xintongsong: It might require more efforts than we can take in this PR to tune the compaction strategy in order to get an ideal performance. But at least in the abstraction we should not assume this only returns one message.

python/flink_agents/api/memory/long_term_memory.py
```diff
@@ -0,0 +1,294 @@
...
+
+        Args:
+            name: The name of the memory set.
+
+        Returns:
+            The memory set.
+        """
+
+    @abstractmethod
+    def delete_memory_set(self, name: str) -> MemorySet:
```
xintongsong: It probably make sense to just return a boolean. The memory set should be no longer bond with the underlying store.

python/flink_agents/runtime/memory/tests/start_chroma_server.sh
alnzng: To enable PersistentClient[1], specify `persist_directory` parameter while building the `ChromaVectorStore` should help.  IIUC, this is similar to what "--path $path" parameter offer in chroma CLI.

[1] https://docs.trychroma.com/docs/run-chroma/persistent-client

python/flink_agents/api/vector_stores/vector_store.py
```diff
@@ -85,11 +91,18 @@ class Document(BaseModel):
...
     content: str = Field(description="The actual text content of the document.")
-    metadata: Dict[str, Any] = Field(default_factory=dict, description="Document metadata such as source, author, timestamp, etc.")
-    id: str | None = Field(default=None, description="Unique identifier of the document.")
+    metadata: Dict[str, Any] = Field(
+        default_factory=dict,
+        description="Document metadata such as source, author, timestamp, etc.",
+    )
+    id: str | None = Field(
+        default=None, description="Unique identifier of the document."
+    )
```
alnzng, line 105: Out of curiosity, did we change the code format stype? seems there are many diffs like this

python/flink_agents/api/vector_stores/vector_store.py
```diff
@@ -135,6 +152,40 @@ def store_kwargs(self) -> Dict[str, Any]:
         when performing vector search operations.
         """
 
+    def add(
+        self, documents: Document | List[Document], collection_name: str | None = None, **kwargs: Any
```
alnzng: The newly added CRUD methods look flexible enough to handle operations across multiple collections, which is good. However, the current vector store implementation is limited to a single collection. To keep the behavior consistent, we should update the existing implementation to support multiple collections as well. For example, For example, the `query`implementation needs to be extended to support retrieval from multiple collections to align with the new CRUD methods.

python/flink_agents/api/vector_stores/vector_store.py
```diff
@@ -135,6 +152,40 @@ def store_kwargs(self) -> Dict[str, Any]:
...
+        # Normalize to list
+        documents = maybe_cast_to_list(documents)
+
+        # Generate embeddings for all documents
+        embedding_model = self.get_resource(
+            self.embedding_model, ResourceType.EMBEDDING_MODEL
+        )
+
+        # Generate embeddings for each document
+        embeddings = [embedding_model.embed(doc.content) for doc in documents]
```
alnzng: Chroma seems to have a size limitation on each request when persisting embeddings, batches should contain fewer than 41,666 embeddings. This has been discussed in the Chroma community (see [GitHub issue #1049](https://github.com/chroma-core/chroma/issues/1049)). I’m not sure if this restriction has changed recently, but to be safe we should add a check for requests exceeding 41,666 embeddings, similar to how LlamaIndex handles it today: https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/llama_index/vector_stores/chroma/base.py#L97C1-L97C72

That being said, it may be better to abstract the embedding function at the document level and allow each vector store implementation to override it as needed.

python/flink_agents/api/vector_stores/vector_store.py
wenjin272: I just run `ruff format` before commit the codes. Looks like `ruff check` only checks the lint rule, while `ruff format` only format the code style.

python/flink_agents/runtime/memory/compaction_functions.py
```diff
@@ -0,0 +1,173 @@
...
+    strategy: SummarizationStrategy = cast(
+        "SummarizationStrategy", memory_set.compaction_strategy
+    )
+
+    # retrieve all items
+    items: List[MemorySetItem] = ltm.get(memory_set=memory_set, ids=ids)
+
+    response: ChatMessage = _generate_summarization(
+        items, memory_set.item_type, strategy, ctx
+    )
```
xintongsong, line 86: The amount of memory to be compacted can be large and may exceed the context window of the model. We should take that into consideration. Let's add a todo here and create a follow-up issue.

Issue #321, opened 2025-11-19
[Feature][integration][Java] Java Implementation of Vector Store with Elasticsearch Support


twosom:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Following the design proposal in #143, we need to implement Java version of Vector Store functionality to enable RAG (Retrieval-Augmented Generation) capabilities in Flink Agents. This issue focuses on implementing the core Vector Store infrastructure and Elasticsearch-based vector store as the first implementation.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #341, opened 2025-11-25
[Feature][Integration][Java] add ElasticsearchVectorStore in Java

twosom:
Linked issue: #321

### Purpose of change

This PR implements the Java version of Vector Store functionality for Flink Agents, following the design proposal in #143. This implementation enables RAG (Retrieval-Augmented Generation) capabilities by providing vector-based context retrieval.

- Added Vector Store API with query support and document handling
- Added `@VectorStore` annotation for agent plan resource management
- Added context retrieval request and response events
- Implemented context retrieval action with vector store support
- Implemented Elasticsearch vector store integration
- Added ElasticsearchVectorStore RAG example with Ollama integration

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed`

### DIFF
--- a/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/VectorStoreIntegrationTest.java
+++ b/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/VectorStoreIntegrationTest.java
@@ -83,7 +83,6 @@ public class VectorStoreIntegrationTest {
 
     @SuppressWarnings("unchecked")
     private void checkResult(CloseableIterator<Object> results) {
-        // 단일 입력에 대한 결과 1건 검증 및 벤더 비의존 최소 계약 확인
         Assertions.assertTrue(
                 results.hasNext(), "No output received from VectorStoreIntegrationAgent");
 


wenjin272:
Hi, @twosom, I'll start reviewing this PR as soon as possible. Besides, this PR should be labeled 'doc-needed'.

api/src/main/java/org/apache/flink/agents/api/vectorstores/BaseVectorStore.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Performs vector search using structured query object. Converts text query to embeddings and
+     * returns structured query result.
+     *
+     * @param query VectorStoreQuery object containing query parameters
+     * @return VectorStoreQueryResult containing the retrieved documents
+     */
+    public VectorStoreQueryResult<ContentT> query(VectorStoreQuery query) {
+        final BaseEmbeddingModelConnection embeddingModel =
```
wenjin272: Should be `BaseEmbeddingModelSetup`

plan/src/main/java/org/apache/flink/agents/plan/actions/ContextRetrievalAction.java
```diff
@@ -0,0 +1,75 @@
...
+    public static Action getContextRetrievalAction() throws Exception {
+        return new Action(
+                "context_retrieval_action",
+                new JavaFunction(
+                        ContextRetrievalAction.class,
+                        "processContextRetrievalRequest",
+                        new Class[] {Event.class, RunnerContext.class}),
+                List.of(
+                        ContextRetrievalRequestEvent.class.getName(),
+                        ContextRetrievalResponseEvent.class.getName()));
```
wenjin272: This action expect receive `ContextRetrievalRequestEvent` and send `ContextRetrievalResponseEvent `, so the listen event types here should only be `ContextRetrievalRequestEvent.class.getName()`

plan/src/main/java/org/apache/flink/agents/plan/actions/ContextRetrievalAction.java
```diff
@@ -0,0 +1,75 @@
...
+            final BaseVectorStore<DocumentT> vectorStore =
+                    (BaseVectorStore<DocumentT>)
+                            ctx.getResource(
+                                    contextRetrievalRequestEvent.getVectorStore(),
+                                    ResourceType.VECTOR_STORE);
+
+            final VectorStoreQuery vectorStoreQuery =
+                    new VectorStoreQuery(
+                            contextRetrievalRequestEvent.getQuery(),
+                            ((ContextRetrievalRequestEvent) event).getMaxResults());
```
wenjin272: Could use `contextRetrievalRequestEvent` directly.

integrations/vector-stores/elasticsearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/elasticsearch/ElasticsearchVectorStore.java
```diff
@@ -0,0 +1,308 @@
...
+            creds.setCredentials(
+                    AuthScope.ANY, new UsernamePasswordCredentials(username, password));
+            builder.setHttpClientConfigCallback(hcb -> hcb.setDefaultCredentialsProvider(creds));
+        }
+
+        // Build the REST client and the transport layer used by the high-level client
+        final RestClient restClient = builder.build();
+        final ElasticsearchTransport transport =
+                new RestClientTransport(restClient, new JacksonJsonpMapper());
+        this.client = new ElasticsearchClient(transport);
```
wenjin272, line 205: Here create a client but no where close it. This may cause connection leak. But this is a common issue for resources using persistent connection, and need framework provide clean up mechanism. I will create a dedicated issue for this.

integrations/vector-stores/elasticsearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/elasticsearch/ElasticsearchVectorStore.java
```diff
@@ -0,0 +1,308 @@
...
+    public List<Document<Map<String, Object>>> queryEmbedding(
+            float[] embedding, int limit, Map<String, Object> args) {
+        try {
+            int k = (int) args.getOrDefault("k", Math.max(1, limit));
+
+            int numCandidates = (int) args.getOrDefault("num_candidates", Math.max(100, k * 2));
+            String filter = (String) args.get("filter_query");
+
+            List<Float> queryVector = new ArrayList<>(embedding.length);
+            for (float v : embedding) queryVector.add(v);
```
wenjin272, line 264: Maybe could simplify to 
```
List<Float> queryVector = Arrays.stream(embedding)
                                .boxed()
                                .collect(Collectors.toList());
```

integrations/vector-stores/elasticsearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/elasticsearch/ElasticsearchVectorStore.java
```diff
@@ -0,0 +1,308 @@
...
+                                                    .queryVector(queryVector)
+                                                    .k(k)
+                                                    .numCandidates(numCandidates));
+
+            if (filter != null) {
+                builder = builder.postFilter(f -> f.withJson(new StringReader(filter)));
+            }
+
+            final SearchResponse<Map<String, Object>> searchResponse =
+                    (SearchResponse) this.client.search(builder.build(), Map.class);
```
wenjin272, line 280: Here we specify the `TDocument` to Map, could it be other types?

wenjin272 (commented):
Hi, @twosom, thanks for your contribution. Overall looks good to me. I left some minor comments about details.

Besides, due to the example requires es cluster, I will verify the example next week.

And some test cases failed, I think it is caused by the newly added built-in action, which break the agent plan verification.

integrations/vector-stores/elasticsearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/elasticsearch/ElasticsearchVectorStore.java
twosom: @wenjin272 
Thank you for your feedback.
But there is no float support in `Arrays.stream()` method.

image

integrations/vector-stores/elasticsearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/elasticsearch/ElasticsearchVectorStore.java
twosom: @wenjin272 
Thank you for your feedback.
Good point! While Python's Document class fixes content as str, I believe this is an over-generalization. Different vector store vendors may have different document structures.

Therefore, keeping `ContentT` as a generic type and delegating the responsibility to each vendor implementation would provide better flexibility. This allows each integration to define its own document structure rather than forcing all vendors into a single String-based content model.

api/src/main/java/org/apache/flink/agents/api/vectorstores/BaseVectorStore.java
```diff
@@ -0,0 +1,96 @@
...
+     * @return VectorStoreQueryResult containing the retrieved documents
+     */
+    public VectorStoreQueryResult<ContentT> query(VectorStoreQuery query) {
+        final BaseEmbeddingModelSetup embeddingModel =
+                (BaseEmbeddingModelSetup)
+                        this.getResource.apply(this.embeddingModel, ResourceType.EMBEDDING_MODEL);
+
+        // TODO
+        // for now, we don't need to use additional parameters.
+        final float[] queryEmbedding = embeddingModel.embed(query.getQueryText(), Map.of());
```
wenjin272: Could directly use `embeddingModel.embed(query.getQueryText())` here.

integrations/vector-stores/elasticsearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/elasticsearch/ElasticsearchVectorStore.java
wenjin272: I think this make sense to some extent. But if different implementation return document in different structure, the user application must be aware of the implementation they used. If they changed the vector store used in their  code, they may also change the code which process the returned documents.
WDTY @alnzng ?

examples/src/main/java/org/apache/flink/agents/examples/rag/ElasticsearchRagExample.java
```diff
@@ -0,0 +1,431 @@
...
+            final String userQuery = event.getQuery();
+            final List<Document<Map<String, Object>>> docs = event.getDocuments();
+
+            // Build context text from retrieved documents
+            List<String> items = new ArrayList<>();
+            for (int i = 0; i < docs.size(); i++) {
+                Object content = docs.get(i).getContent();
+                items.add(String.format("%d. %s", i + 1, content));
+            }
+            String contextText = String.join("\n\n", items);
```
wenjin272: Here, the content is a `Map<String, Object>`, we just convert it to `String`, so the `contextText` is actually
```
1. {metadata={source=official, topic=flink}, content_vector=[0.015702553, 0.09362559, -0.16092904, ..., 0.007019141, 0.058410715, 0.009363564, -0.046617374, 0.0066441908, 0.010014975, 0.02780144, -0.07676514, -0.035377625, -0.04142186, -0.0014243895], 
content=Apache Flink is a framework and distributed processing engine for stateful computations over unbounded and bounded data streams.}
```

The `metadata` and `content_vector` may be not needed to pass to llm.

This also made me wonder: do we really need to use a Map-type document content, i.e. `ContentT`, or is String sufficient?

wenjin272 (commented):
Hi, @twosom, thanks for your update. I left some comments mainly about the document type.

Additionally, when run the example locally, I encountered an issue related to SSL certificate authentication. I didn't seem to find where we can configure the `http_ca.crt`, so I manually disabled security authentication for the es cluster.

integrations/vector-stores/elasticsearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/elasticsearch/ElasticsearchVectorStore.java
wenjin272: I think this make sense to some extent. But if different implementation return document with different `ContentT`, the user application must be aware of the implementation they used. If they changed the vector store used in their  code, they may also change the code which process the returned content.
WDTY @alnzng, @xintongsong ?

integrations/vector-stores/elasticsearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/elasticsearch/ElasticsearchVectorStore.java
wenjin272: You are right, sorry for this suggestion.

examples/src/main/java/org/apache/flink/agents/examples/rag/ElasticsearchRagExample.java
twosom: @wenjin272 

Thank you for your feedback.

I believe this should be left entirely to the user's discretion.

In the example, I converted it to String for simplicity. However, downstream users might want to extract only specific properties from the Map<String, Object> ContentT to use as context for the LLM.

For example:
```java
Map<String, Object> content = (Map) docs.get(i).getContent();
String someProperty = (String) content.get("<some-property>");
// and now we can use 'someProperty' for llm context
```

Of course, this means users need to be aware of the ContentT type returned by the specific vector database implementation, which makes the code more complex.

However, it's important to note that retrieved documents are not always String type. By default, Elasticsearch maintains documents in JSON format.

I think the current generic approach provides flexibility while keeping the example simple. Users who need String-only content can easily convert it, while users who need structured access to document fields can work with the Map directly.

WDYT?

wenjin272:
> @wenjin272 Thank you for your review.
> 
> I didn't consider SSL configuration while testing locally. Thank you for pointing this out.
> 
> Would it be better to add SSL certificate configuration support in a separate PR rather than including it in this one? I'm happy to follow your guidance on this.
> 
> Thank you.

I think ssl support can be done in separate PR.

examples/src/main/java/org/apache/flink/agents/examples/rag/ElasticsearchRagExample.java
```diff
@@ -0,0 +1,431 @@
...
+ *
+ * <p>Running the example will:
+ *
+ * <ol>
+ *   <li>Optionally populate the Elasticsearch index with sample documents and stored vectors,
+ *   <li>Create a simple agent pipeline that retrieves context from Elasticsearch, and
+ *   <li>Print the model's answers for a set of example queries.
+ * </ol>
+ */
+public class ElasticsearchRagExample {
```
xintongsong: Not sure about introducing an ES-specific example in the `examples` module.
 This module should only contain a minimal set of representative, recommended, quick-start examples.

I'd suggest:
1. Add a `VectorStoreIntegrationTest` in the `flink-agents-end-to-end-tests-integration` module, which should be parameterized and can verify not only ES but also future vector store implementations.
2. Add a built-in `RagAgent` (just like `ReActAgent`), and create an example based on it.

xintongsong: And it's not necessary to include 2) in this PR. We can make it a separate follow-up task.

examples/src/main/java/org/apache/flink/agents/examples/rag/ElasticsearchRagExample.java
xintongsong: @twosom 
I see your point that users may store structured documents in the vector store and only want to access part of the document. However, in that case users can always deserialized / parse the content by themselves.

In your current implementation, it's the `ElasticsearchVectorStore` that decides the content type should be a map. That means different vector store implementation would have different content types, and users need to be aware of that. I don't think this is a good idea.

Ideally, if we want to support structured types, we should support them in a unified way for all vector stores. E.g., we may ask users to specify the content type when creating a vector store resource. We probably don't need to support arbitrary types. Just a few types (string, json, bytes) should be flexible enough.

TBH, I'm not entirely sure about the necessity of supporting structured access. Yes, people MAY want to do it. But is that a must? How bad is it if we just let users to parse the string by themselves? Not saying it's not important. I just really don't know at this moment. And considering the complexity of supporting it, I'd suggest to keep it simple and only support strings in this PR. We can add more supports like json later if we see a concrete demand for it.

WDYT?

twosom: @xintongsong

Thank you for your feedback.
I completely agree with your point.

If we use strings, I think there are two approaches:

Serialize the entire Map to a string
Specify which field(s) to use for LLM context (e.g., title or content)
While option 1 (serializing the entire Map) isn't necessarily bad, unless we explicitly specify Elasticsearch's _source fields, all content would be used as LLM context, which could lead to significant token waste.

With option 2, we can provide only the necessary property (or properties) to the LLM, which would result in more appropriate token usage. However, this would add more initialization parameters to ElasticsearchVectorStore, increasing complexity.

I find it difficult to make this decision alone.

WDYT?
@xintongsong @wenjin272

xintongsong: I think retrieving the entire map from ES does not necessarily mean feeding everything to LLM. We can apply transformations on the retrieved data before sending them to the LLM. This can be done in several ways:

- The user can add a custom Action that listens to the context retrieval response.
- We can introduce a built-in RagAgent that accepts a function which will be applied on the retrieved documents
- We can add a function parameter in the general vector store interface which will be applied on the retrieved data

With either of the above options, we don't need to do this specifically for each vector store implementation.

twosom:
@xintongsong @wenjin272 

Hello.
As you mentioned, I removed the generic type and changed it to a string type.
I also removed the existing examples and added e2e tests.
I would appreciate it if you could review it when you have time.
Thank you.

examples/pom.xml
```diff
@@ -69,6 +69,16 @@ under the License.
...
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-integrations-embedding-models-ollama</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-integrations-vector-stores-elasticsearch</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
xintongsong: This can be removed.

integrations/vector-stores/elasticsearch/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-agents-plan</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
xintongsong: I think we should not need to depends on `flink-agents-plan`?

e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/VectorStoreIntegrationTest.java
```diff
@@ -0,0 +1,122 @@
...
+        final CloseableIterator<Object> results = outputStream.collectAsync();
+
+        agentEnv.execute();
+
+        checkResult(results);
+    }
+
+    @SuppressWarnings("unchecked")
+    private void checkResult(CloseableIterator<Object> results) {
+        // 단일 입력에 대한 결과 1건 검증 및 벤더 비의존 최소 계약 확인
```
wenjin272: comments not in English.

twosom: @wenjin272 
Sorry, that's my mistake 😭

PR #343, opened 2025-11-27
[hotfix] Fix the type loss issue in the deserialization of ResourceProvider

GreatEugenius:
### Purpose of change
Fix the type loss issue in the deserialization of ResourceProvider and correct the improper use of the descriptor.

### Tests
ut

### Documentation
The commit already includes the changes to the Docs.

- [ ] `doc-needed` 
- [x] `doc-not-needed`

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/serializer/ResourceProviderJsonDeserializer.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/serializer/ResourceProviderJsonDeserializer.java
@@ -33,9 +33,7 @@ import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.ObjectMap
 import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.deser.std.StdDeserializer;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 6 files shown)


wenjin272:
What result will the type loss  cause?

GreatEugenius:
> What result will the type loss cause?

Currently, it does not affect user usage because a forced type conversion has been performed on the Python side through the `BaseModel`. However, to support cross-language features, we need to address this issue first.

plan/src/main/java/org/apache/flink/agents/plan/serializer/ResourceProviderJsonDeserializer.java
```diff
@@ -82,7 +82,7 @@ private PythonResourceProvider deserializePythonResourceProvider(JsonNode node)
         JsonNode kwargsNode = node.get("kwargs");
         Map<String, Object> kwargs = new HashMap<>();
         if (kwargsNode != null && kwargsNode.isObject()) {
-            kwargs = (Map<String, Object>) parseJsonNode(kwargsNode);
```
Sxnan, line 85: Can we remove the `parseJsonNode` method, as it is not used anymore?

Sxnan (commented):
Thanks for the PR! Just left one small comment.

Sxnan (approved):
LGTM!

Issue #334, opened 2025-11-21
[Feature] [api][Java] Add built-in mcp support for Java


xli1996:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Flink agents python have bulit-in mcp support in https://github.com/apache/flink-agents/pull/168, we should support built-in support in java api as well.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

xintongsong:
Created #411 as a follow-up of this.

PR #356, opened 2025-12-01
[Feature][Java] Add MCP support in Java

yanand0909:
Linked issue: https://github.com/apache/flink-agents/issues/334

### Purpose of change

- This PR adds StreamableHttp transport based support for MCP Servers which includes MCPTools and MCPPrompts. 
- This PR also updates Prompt structure to be in line with python Prompt, LocalPrompt and MCPPrompt Structure.
- MCP SDK is compiled in Java 17 so to support it we need to update the project java version to 17 as well.

### Tests

Added Unit tests for it

### API

Yes, Prompt

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed`

### DIFF
--- a/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
+++ b/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
@@ -18,6 +18,10 @@
 
 package org.apache.flink.agents.integrations.mcp;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.modelcontextprotocol.client.McpClient;
 import io.modelcontextprotocol.client.McpSyncClient;
 import io.modelcontextprotocol.client.transport.HttpClientStreamableHttpTransport;
@@ -31,10 +35,6 @@ import org.apache.flink.agents.integrations.mcp.auth.ApiKeyAuth;
 import org.apache.flink.agents.integrations.mcp.auth.Auth;
 import org.apache.flink.agents.integrations.mcp.auth.BasicAuth;
 import org.apache.flink.agents.integrations.mcp.auth.BearerTokenAuth;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonCreator;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonIgnore;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonProperty;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.ObjectMapper;
 
 import java.net.URI;
 import java.net.http.HttpRequest;
--- a/integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPSerializationTest.java
+++ b/integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPSerializationTest.java
@@ -18,8 +18,8 @@
 
 package org.apache.flink.agents.integrations.mcp;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.flink.agents.api.tools.ToolMetadata;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledOnJre;
@@ -29,7 +29,7 @@ import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 
-import static org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES;
+import static com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES;
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
--- a/integrations/mcp/pom.xml
+++ b/integrations/mcp/pom.xml
@@ -37,6 +37,10 @@ under the License.
             <artifactId>flink-agents-api</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
     </dependencies>
 
     <profiles>
--- a/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPPrompt.java
+++ b/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPPrompt.java
@@ -18,12 +18,12 @@
 
 package org.apache.flink.agents.integrations.mcp;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.flink.agents.api.chat.messages.ChatMessage;
 import org.apache.flink.agents.api.chat.messages.MessageRole;
 import org.apache.flink.agents.api.prompt.Prompt;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonCreator;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonIgnore;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonProperty;
 
 import java.util.HashMap;
 import java.util.List;
--- a/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPTool.java
+++ b/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPTool.java
@@ -18,14 +18,14 @@
 
 package org.apache.flink.agents.integrations.mcp;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.flink.agents.api.tools.Tool;
 import org.apache.flink.agents.api.tools.ToolMetadata;
 import org.apache.flink.agents.api.tools.ToolParameters;
 import org.apache.flink.agents.api.tools.ToolResponse;
 import org.apache.flink.agents.api.tools.ToolType;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonCreator;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonIgnore;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonProperty;
 
 import java.util.HashMap;
 import java.util.List;
--- a/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/auth/ApiKeyAuth.java
+++ b/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/auth/ApiKeyAuth.java
@@ -18,8 +18,8 @@
 
 package org.apache.flink.agents.integrations.mcp.auth;
 
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonCreator;
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
 
 import java.net.http.HttpRequest;
 import java.util.Objects;
--- a/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/auth/Auth.java
+++ b/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/auth/Auth.java
@@ -18,8 +18,8 @@
 
 package org.apache.flink.agents.integrations.mcp.auth;
 
-import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.annotation.JsonSubTypes;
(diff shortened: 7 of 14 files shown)


integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
```diff
@@ -0,0 +1,424 @@
...
+        McpSyncClient mcpClient = getClient();
+        McpSchema.GetPromptRequest request =
+                new McpSchema.GetPromptRequest(
+                        name, arguments != null ? arguments : new HashMap<>());
+        McpSchema.GetPromptResult result = mcpClient.getPrompt(request);
+
+        List<ChatMessage> chatMessages = new ArrayList<>();
+        for (var message : result.messages()) {
+            if (message.content() instanceof McpSchema.TextContent) {
+                var textContent = (McpSchema.TextContent) message.content();
```
wenjin272, line 369: The idea show the variable 'textContent' can be replaced with pattern variable
```
if (message.content() instanceof McpSchema.TextContent textContent) {
    MessageRole role = MessageRole.valueOf(message.role().name().toUpperCase());
    chatMessages.add(new ChatMessage(role, textContent.text()));
}
```

api/src/main/java/org/apache/flink/agents/api/mcp/MCPPrompt.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /**
+     * Format the prompt as a list of chat messages with the given arguments. This overloaded
+     * version accepts Object values for flexibility.
+     *
+     * @param arguments Arguments to pass to the prompt (Object values)
+     * @return List of formatted chat messages
+     */
+    public List<ChatMessage> formatMessages(Map<String, Object> arguments) {
```
wenjin272: Maybe this method can be private

api/src/main/java/org/apache/flink/agents/api/mcp/MCPServer.java
```diff
@@ -0,0 +1,424 @@
...
+ * }</pre>
+ *
+ * <p>Reference: <a href="https://modelcontextprotocol.io/sdk/java/mcp-client">MCP Java Client</a>
+ */
+public class MCPServer extends SerializableResource {
+
+    private static final String FIELD_ENDPOINT = "endpoint";
+    private static final String FIELD_HEADERS = "headers";
+    private static final String FIELD_TIMEOUT_SECONDS = "timeoutSeconds";
+    private static final String FIELD_SSE_READ_TIMEOUT_SECONDS = "sseReadTimeoutSeconds";
```
wenjin272: This field is never used.

api/src/test/java/org/apache/flink/agents/api/mcp/MCPServerTest.java
```diff
@@ -0,0 +1,262 @@
...
+        assertThat(server.getHeaders()).containsEntry("X-Custom-Header", "value");
+        assertThat(server.getTimeoutSeconds()).isEqualTo(30);
+        assertThat(server.getAuth()).isInstanceOf(BearerTokenAuth.class);
+    }
+
+    @Test
+    @DisabledOnJre(JRE.JAVA_11)
+    @DisplayName("Create MCPServer with simple constructor")
+    void testSimpleConstructor() {
+        MCPServer server = new MCPServer("http://localhost:8000/mcp");
```
wenjin272: The endpoint `http://localhost:8000/mcp` is used in each test case, maybe we can extract it as a literal.

integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPSerializationTest.java
```diff
@@ -0,0 +1,151 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for MCP classes serialization with Flink's Kryo serializer.
+ *
+ * <p>This test ensures that MCP objects (MCPServer, MCPTool, MCPPrompt) can be properly serialized
+ * and deserialized by Flink's serialization framework, which uses Kryo under the hood. This is
+ * critical for distributed execution where these objects need to be sent across the network.
+ */
+class MCPSerializationTest {
```
wenjin272, line 41: Looks like this test doesn't depend on any runtime class, maybe we can move it to api module.

This test looks like for testing the json serialization and deserialization for MCP, not Kryo serialization. The comments is confused.

runtime/pom.xml
```diff
@@ -134,4 +134,56 @@ under the License.
...
+                    <argLine>
+                        --add-opens java.base/java.util=ALL-UNNAMED
+                        --add-opens java.base/java.lang=ALL-UNNAMED
+                    </argLine>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
 </project>
```
wenjin272, line 145: If we move the mcp test to api, this profile can be removed. 
Besides, the project root pom has add surefire options, maybe we don't need add again here.

api/src/main/java/org/apache/flink/agents/api/agents/ReActAgent.java
```diff
@@ -91,7 +92,7 @@ public ReActAgent(
                         "Output schema must be RowTypeInfo or Pojo class.");
             }
             Prompt schemaPrompt =
-                    new Prompt(
+                    new LocalPrompt(
```
wenjin272: Maybe we can use `Prompt.fromText` here, to avoid user be aware of `MCPPrompt` and `LocalPrompt`.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -63,6 +63,8 @@
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 
+import static org.apache.flink.agents.api.resource.ResourceType.*;
```
wenjin272: In flink, we recommend avoid star import.

pom.xml
```diff
@@ -53,6 +53,7 @@ under the License.
         <mockito.version>5.8.0</mockito.version>
         <gpg.useagent>true</gpg.useagent>
         <arguments />
+        <mcp.enabled>true</mcp.enabled>
```
Sxnan: I cannot find where it is used.

api/pom.xml
```diff
@@ -65,4 +65,53 @@ under the License.
         </dependency>
     </dependencies>
 
+    <profiles>
+        <profile>
+            <id>mcp-enabled</id>
```
Sxnan: It is better to name the profile as `java-17`, as it is activated automatically when the JDK version is 17+.

Same for the other poms

pom.xml
```diff
@@ -87,6 +88,32 @@ under the License.
...
+            <properties>
+                <target.java.version>11</target.java.version>
+                <maven.compiler.source>11</maven.compiler.source>
+                <maven.compiler.target>11</maven.compiler.target>
+                <mcp.enabled>false</mcp.enabled>
+            </properties>
+        </profile>
+
+        <profile>
+            <id>java-17-plus</id>
```
Sxnan: I think `java-17` should be enough. Otherwise, when we add the profile for java21, we need to update the id of this profile as well.

pom.xml
```diff
@@ -95,7 +122,7 @@ under the License.
                 </property>
             </activation>
             <properties>
-                <target.java.version>11</target.java.version>
+                <target.java.version>17</target.java.version>
```
Sxnan: I think we should target for jdk11 by default for better compatibility.

api/pom.xml
Sxnan, line 1: We should use the classifier for the java17 jar, such as `java-17`. 

So that after releasing, user can depend on the java17 jar by specifying the classifier in their pom, for example:

```
<dependency>
    <groupId>org.apache.flink</groupId>
    <artifactId>flink-agents-api</artifactId>
    <version>0.2.0</version>
    <classifier>java-17</classifier>
</dependency>
```

Sxnan (commented):
Thanks for the PR! I just left some comments regarding the multi-JDK version support

api/src/main/java/org/apache/flink/agents/api/mcp/MCPServer.java
yanand0909: Yeah I forgot to remove this after refactoring, Thanks for pointing this out

integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
yanand0909: Spotless fails for Java11 if we use pattern variable here

pom.xml
yanand0909: Yeah this was part of my initial implementation but its not being used now. Thanks for catching this

api/pom.xml
yanand0909: Yeah we should definitely do this but I am not very familiar with the deployment process and how we will be able to release 2 jars one built on java 11 and one on java 17 with classifier. May be I can raise another PR for this and continue discussion there?

Sxnan: Sure, it doesn't have to block this PR. We can create an issue to keep track of this.

Sxnan: https://github.com/apache/flink-agents/issues/389 was created to keep track of the deployment aspect.

integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
```diff
@@ -0,0 +1,423 @@
...
+ *     .timeout(Duration.ofSeconds(30))
+ *     .build();
+ *
+ * List<MCPTool> tools = server.listTools();
+ * server.close();
+ * }</pre>
+ *
+ * <p>Reference: <a href="https://modelcontextprotocol.io/sdk/java/mcp-client">MCP Java Client</a>
+ */
+public class MCPServer extends SerializableResource {
```
xintongsong, line 76: Currently, the entire implementation of MCPServer and related tools/prompts are in the `api` module, which is not ideal. We should keep the `api` module as lightweight as possible, including only a minimum set of interfaces that users need to depend on when writing their agent codes. I think we can move the entire `api/mcp` directory into the the `integration` module. The only think users need to see in the `api` module is the annotation.

xintongsong: @wenjin272, we should also fix this for the python api.

api/src/main/java/org/apache/flink/agents/api/prompt/LocalPrompt.java
```diff
@@ -0,0 +1,207 @@
...
+ *     new ChatMessage(MessageRole.USER, "Help me with {task}")
+ * );
+ * LocalPrompt prompt2 = new LocalPrompt(messages);
+ * List<ChatMessage> result2 = prompt2.formatMessages(
+ *     MessageRole.SYSTEM,
+ *     Map.of("role", "an assistant", "task", "coding")
+ * );
+ * }</pre>
+ */
+public class LocalPrompt extends Prompt {
```
xintongsong: I'd suggest to make this a private class in `Prompt.java`. Users should not be aware of the different implementations of `Prompt`. The should always use `Prompt.fromText/Messages()` to create a prompt, rather than `new LocalPrompt()`.

xintongsong: And we should also use `Prompt.fromText/Messages()` in the doc sample codes, examples and tests.

.github/workflows/ci.yml
```diff
@@ -63,8 +63,24 @@ jobs:
...
+        run: |
+          cd python
+          uv sync --all-extras
       - name: Run Java Tests
-        run: tools/ut.sh -j
+        run: |
+          # Add Python venv to PATH so Java tests can find MCP dependencies
+          export PATH="${{ github.workspace }}/python/.venv/bin:$PATH"
+          export PYTHONPATH="${{ github.workspace }}/python/.venv/lib/python3.11/site-packages:$PYTHONPATH"
+          tools/ut.sh -j
```
xintongsong, line 83: IIUC, we need this to start an external mcp server, so that we can access from the test cases? If that is the case, I'd suggest to move the test cases that depends on external mcp server to the `it-java` pipeline. In theory, any test case that interacts with an external system should be considered as integration test or e2e test.

yanand0909: This is for AgentPlanDeclareMCPServerTest to test whether MCP Tools and Prompts from MCP Server is properly discoverable, we need a running MCP server for this test. Do you think we should move this test to integration test?

api/src/main/java/org/apache/flink/agents/api/prompt/LocalPrompt.java
yanand0909: I can move localPrompt to be a subclass of prompt but I cannot make it private as PythonPrompt extends it

xintongsong: I see. Can we move `LocalPrompt` to the `plan` module, and use reflection to create it in `Prompt.fromText/Messages`? My point is to hide it from the user-facing APIs if possible.

.github/workflows/ci.yml
xintongsong: Yes, I think we can move that test to integration test.

api/src/main/java/org/apache/flink/agents/api/prompt/LocalPrompt.java
yanand0909: If we move LocalPrompt to Plan module that will create circular dependency as we would still need LocalPrompt on the classpath at runtime for the reflection call, we can either move this to integration module but that does not seems appropriate looking at the design or we can keep it in API module under some internal package and discourage users to use it directly.

WDYT @xintongsong ?

xintongsong: > If we move LocalPrompt to Plan module that will create circular dependency as we would still need LocalPrompt on the classpath at runtime for the reflection call

That should be fine. We are using the same approach for `AgentsExecutionEnvironment` and it works fine. User programs only need to depend on the `api` module, with `provided` scope. They don't need to explicitly depend on the `plan` / `runtime` modules. Users must first install flink-agents to their flink clusters in order to execute the programs, which is specified in our installation documentations. So we can just assume the `plan` / `runtime` modules exist in the classpath in runtime.

yanand0909: There are multiple tests in API module like `BaseChatModelTest` that needs LocalPrompt and if we don't add plan module dependency those tests fails and adding the dependency even for test scope creates circular dependency

xintongsong: I think you are right. It seems we don't have a good way to hide `LocalPrompt` from users. Let's keep it as is.

xintongsong (approved):
@yanand0909 Thanks for addressing my comments. LGTM.

Issue #340, opened 2025-11-25
[Bug] Python async execution can not work with flink checkpoint.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

When run `integrate_datastream_with_agent_example.py`, the flink job does not generate any output.

If remove the checkpoint configs
```
config.set_string("state.backend.type", "rocksdb")
config.set_string("checkpointing.interval", "1s")
config.set_string("restart-strategy.type", "disable")
```
Or disable async execution in Agent action
`log_success = yield from ctx.execute_async(log_to_stdout, input_data, total)`

The job can output messages normally.

### How to reproduce

`python python/flink_agents/e2e_tests/integrate_datastream_with_agent_example.py`

### Version and environment

Run test in main branch on mac os.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #365, opened 2025-12-02
[runtime] Implement hashCode and equals for ActionTask.

wenjin272:
Linked issue: #340 

### Purpose of change

Implement `hashCode` and `equals` for ActionTask, to resolve the issue that python action async execution can not work with rocksdb.

### Tests

it

### API

No

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed`


runtime/src/main/java/org/apache/flink/agents/runtime/python/event/PythonEvent.java
```diff
@@ -55,4 +57,17 @@ public byte[] getEvent() {
     public String getEventType() {
         return eventType;
     }
+
+    @Override
+    public boolean equals(Object o) {
```
Sxnan, line 62: We should also invoke the `super().equals()`

Same for the hashcode

Sxnan (commented):
Thanks for the PR! I left one comment below

Sxnan (approved):
LGTM

Issue #354, opened 2025-12-01
[Feature] Support the use of Java ChatModel in Python


GreatEugenius:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support the use of Java ChatModel in Python Flink Agents jobs.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #373, opened 2025-12-04
[Feature][runtime] Support the use of Java ChatModel in Python

GreatEugenius:
Linked issue: #354 

### Purpose of change
Support the use of the Java ChatModel in Python Flink Agent jobs. 

### Tests
Unit tests and end-to-end test

### API
no

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/resource/ResourceDescriptor.java
+++ b/api/src/main/java/org/apache/flink/agents/api/resource/ResourceDescriptor.java
@@ -27,19 +27,15 @@ import java.util.Objects;
 
 /** Helper class to describe a {@link Resource} */
 public class ResourceDescriptor {
-    private static final String FIELD_CLAZZ = "java_clazz";
-    private static final String FIELD_PYTHON_CLAZZ = "python_clazz";
-    private static final String FIELD_PYTHON_MODULE = "python_module";
+    private static final String FIELD_CLAZZ = "target_clazz";
+    private static final String FIELD_MODULE = "target_module";
     private static final String FIELD_INITIAL_ARGUMENTS = "arguments";
 
     @JsonProperty(FIELD_CLAZZ)
     private final String clazz;
 
-    @JsonProperty(FIELD_PYTHON_CLAZZ)
-    private final String pythonClazz;
-
-    @JsonProperty(FIELD_PYTHON_MODULE)
-    private final String pythonModule;
+    @JsonProperty(FIELD_MODULE)
+    private final String module;
 
     @JsonProperty(FIELD_INITIAL_ARGUMENTS)
     private final Map<String, Object> initialArguments;
@@ -51,43 +47,40 @@ public class ResourceDescriptor {
      * arguments. This constructor supports cross-platform compatibility between Java and Python
      * resources.
      *
-     * @param clazz The Java class full path for the resource type to create a descriptor for.
-     *     Example: "com.example.YourJavaClass"
-     * @param pythonModule The Python module path for cross-platform compatibility. **REQUIRED when
-     *     declaring Python resources in Java.** Defaults to empty string for Java-only resources.
-     *     Example: "your_module.submodule"
-     * @param pythonClazz The Python class name for cross-platform compatibility. **REQUIRED when
-     *     declaring Python resources in Java.** Defaults to empty string for Java-only resources.
-     *     Example: "YourPythonClass"
+     * @param clazz The class identifier for the resource. Its meaning depends on the resource type:
+     *     <ul>
+     *       <li><b>For Java resources:</b> The fully qualified Java class name (e.g.,
+     *           "com.example.YourJavaClass"). The {@code module} parameter should be empty or null.
+     *       <li><b>For Python resources (when declaring from Java):</b> The Python class name
+     *           (simple name, not module path, e.g., "YourPythonClass"). The Python module path
+     *           must be specified in the {@code module} parameter (e.g., "your_module.submodule").
+     *     </ul>
+     *
+     * @param module The Python module path for cross-platform compatibility. Defaults to empty
+     *     string for Java resources. Example: "your_module.submodule"
      * @param initialArguments Additional arguments for resource initialization. Can be null or
      *     empty map if no initial arguments are needed.
      */
     @JsonCreator
     public ResourceDescriptor(
+            @JsonProperty(FIELD_MODULE) String module,
             @JsonProperty(FIELD_CLAZZ) String clazz,
-            @JsonProperty(FIELD_PYTHON_MODULE) String pythonModule,
-            @JsonProperty(FIELD_PYTHON_CLAZZ) String pythonClazz,
             @JsonProperty(FIELD_INITIAL_ARGUMENTS) Map<String, Object> initialArguments) {
         this.clazz = clazz;
-        this.pythonClazz = pythonClazz;
-        this.pythonModule = pythonModule;
+        this.module = module;
         this.initialArguments = initialArguments;
     }
 
     public ResourceDescriptor(String clazz, Map<String, Object> initialArguments) {
-        this(clazz, "", "", initialArguments);
+        this("", clazz, initialArguments);
     }
 
     public String getClazz() {
         return clazz;
     }
 
-    public String getPythonClazz() {
-        return pythonClazz;
-    }
-
-    public String getPythonModule() {
-        return pythonModule;
+    public String getModule() {
+        return module;
     }
 
     public Map<String, Object> getInitialArguments() {
@@ -116,22 +109,18 @@ public class ResourceDescriptor {
 
         ResourceDescriptor that = (ResourceDescriptor) o;
         return Objects.equals(this.clazz, that.clazz)
-                && Objects.equals(this.pythonClazz, that.pythonClazz)
-                && Objects.equals(this.pythonModule, that.pythonModule)
+                && Objects.equals(this.module, that.module)
                 && Objects.equals(this.initialArguments, that.initialArguments);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(clazz, pythonClazz, pythonModule, initialArguments);
+        return Objects.hash(clazz, module, initialArguments);
     }
 
     public static class Builder {
         private final String clazz;
         private final Map<String, Object> initialArguments;
-
-        private String pythonClazz;
-        private String pythonModule;
 
         public static Builder newBuilder(String clazz) {
             return new Builder(clazz);
@@ -147,16 +136,7 @@ public class ResourceDescriptor {
             return this;
         }
 
-        public Builder setPythonResourceClass(String pythonModule, String pythonClazz) {
-            this.pythonClazz = pythonClazz;
-            this.pythonModule = pythonModule;
-            return this;
-        }
-
         public ResourceDescriptor build() {
-            if (pythonClazz != null && pythonModule != null) {
-                return new ResourceDescriptor(clazz, pythonModule, pythonClazz, initialArguments);
-            }
             return new ResourceDescriptor(clazz, initialArguments);
         }
     }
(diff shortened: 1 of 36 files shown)


GreatEugenius:
Hi, @Sxnan and @wenjin272. Could you please review this PR? Thank you!

python/flink_agents/plan/resource_provider.py
```diff
@@ -159,20 +159,61 @@ def provide(self, get_resource: Callable, config: AgentConfiguration) -> Resourc
...
+
+        j_resource = self._j_resource_adapter.getResource(self.name, self.type.value)
+
+        from flink_agents.runtime.java.java_chat_model import (
+            JavaChatModelConnectionImpl,
+            JavaChatModelSetupImpl,
         )
-        raise NotImplementedError(err_msg)
+
+        JAVA_RESOURCE_MAPPING: dict[ResourceType, Type[Resource]] = {
```
wenjin272: `JAVA_RESOURCE_MAPPING` is independent of this function and class, can be moved outside.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/JavaResourceAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Retrieves a Java resource by name and type value.
+     *
+     * @param name the name of the resource to retrieve
+     * @param typeValue the type value of the resource
+     * @return the resource
+     * @throws Exception if the resource cannot be retrieved
+     */
+    public Resource getResource(String name, String typeValue) throws Exception {
```
wenjin272, line 58: This method is never used in java side, but will be called in python side. Should add description.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/JavaResourceAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+        return resource;
+    }
+
+    /**
+     * Convert a Python chat message to a Java chat message.
+     *
+     * @param pythonChatMessage the Python chat message
+     * @return the Java chat message
+     */
+    public ChatMessage fromPythonChatMessage(Object pythonChatMessage) {
```
wenjin272, line 109: same here

plan/src/main/java/org/apache/flink/agents/plan/serializer/ResourceProviderJsonDeserializer.java
```diff
@@ -105,6 +114,17 @@ private PythonSerializableResourceProvider deserializePythonSerializableResource
...
+        if (node.has("clazz")) {
+            String clazz = node.get("clazz").asText();
+
+            JsonNode kwargsNode = node.get("kwargs");
+            Map<String, Object> kwargs = new HashMap<>();
+            if (kwargsNode != null && kwargsNode.isObject()) {
+                ObjectMapper mapper = new ObjectMapper();
+                kwargs = mapper.convertValue(kwargsNode, Map.class);
+            }
+            return new JavaResourceProvider(name, ResourceType.fromValue(type), clazz, kwargs);
```
wenjin272: IIUC, because `JavaResourceProvider` created in python contains field `clazz` and `kwargs` rather than `descriptor`, so here we should deserialize `clazz` and `kwargs` from json, and reconstruct `Descriptor` in `JavaResourceProvider` in java according these.

Due to `ResourceDescriptor` is only a data class, could we refactor the `JavaResourceProvider` in python, to make it serialize `descriptor` also? Then, the `JavaResourceProvider` serialization and deserialization behavior can be same.

The `PythonResourceProvider` may also can be refactored in this way.

wenjin272 (commented):
Hi, @GreatEugenius, thanks for your work. Overall looks good to me. 

A major comment is could we unify the representation of `ResourceProvider` in both Python and Java to reduce the complexity for its serialization and deserialization.

GreatEugenius:
Hi @wenjin272, I have addressed all the comments and supported ResourceDescriptor serialization and deserialization. Please take a look.

python/flink_agents/plan/resource_provider.py
```diff
@@ -89,34 +90,29 @@ class PythonResourceProvider(ResourceProvider):
...
-    module: str
-    clazz: str
-    kwargs: Dict[str, Any]
+    descriptor: ResourceDescriptor
 
     @staticmethod
     def get(name: str, descriptor: ResourceDescriptor) -> "PythonResourceProvider":
         """Create PythonResourceProvider instance."""
-        clazz = descriptor.clazz
+        clazz = descriptor.clazz  # 获取实际的类对象
```
wenjin272: Chinese note

python/flink_agents/plan/resource_provider.py
```diff
@@ -158,21 +154,60 @@ def provide(self, get_resource: Callable, config: AgentConfiguration) -> Resourc
...
+        j_resource = self._j_resource_adapter.getResource(self.name, self.type.value)
+
+        class_path = JAVA_RESOURCE_MAPPING.get(self.type)
+        if not class_path:
+            err_msg = f"No Java resource mapping found for {self.type.value}"
+            raise ValueError(err_msg)
+        module_path, class_name = class_path.rsplit(".", 1)
+        cls = get_resource_class(module_path, class_name)
+        kwargs = self.descriptor.arguments
+        print(kwargs)
```
wenjin272: test print

python/flink_agents/api/resource.py
```diff
@@ -73,23 +74,177 @@ def validate_serializable(self) -> "SerializableResource":
...
+    _clazz: Type[Resource] | None = PrivateAttr(default=None)
+
+    def __init__(self, /,
+                 python_clazz: str | None = None,
+                 python_module: str | None = None,
+                 java_clazz: str | None = "",
+                 arguments: Dict[str, Any] | None = None,
+                 *,
+                 clazz: Type[Resource] | None = None,
+                 **kwargs: Dict[str, Any]) -> None:
```
wenjin272: The keyed arguments should be `**kwargs: Any`

Besides, if we need both `python_clazz` and `java_clazz`? For it won't be set  in the same time, and the type of the `ResourceProvider` can help distinguish the `clazz` is python or java class. Maybe just `target_module` and `target_clazz` is enough.

python/flink_agents/api/resource.py
```diff
@@ -73,23 +74,177 @@ def validate_serializable(self) -> "SerializableResource":
...
+            python_clazz: The Python class name (not the full path).
+                         Usually auto-generated from clazz, users typically don't need
+                         to set this manually.
+            python_module: The Python module path (not including the class name).
+                          Usually auto-generated from clazz, users typically don't need
+                          to set this manually.
+            java_clazz: The Java class full path for cross-platform compatibility.
+                       Defaults to empty string for Python-only resources.
+                       Users typically don't need to set this manually.
+            arguments: Dictionary containing resource initialization parameters.
```
wenjin272: The doc string is not explicit. If users typically don't need to set this manually, the arguments should not set in the init function.

Actually, when user want to declare a java resource in python, they must set `java_clazz`.

python/flink_agents/api/resource.py
GreatEugenius: I believe we need `python_clazz` and `java_clazz` for implementing cross-language resources. For instance, when using a Java resource in Python, the user registers the resource that corresponds to the wrapper's Class (Python Class) along with the actual Java resource being called (Java Class).

api/src/main/java/org/apache/flink/agents/api/resource/ResourceDescriptor.java
```diff
@@ -20,36 +20,59 @@
...
 
+    @JsonProperty(FIELD_PYTHON_CLAZZ)
+    private final String pythonClazz;
+
+    @JsonProperty(FIELD_PYTHON_MODULE)
+    private final String pythonModule;
+
     // TODO: support nested map/list with non primitive value.
     @JsonProperty(FIELD_INITIAL_ARGUMENTS)
-    @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS)
```
wenjin272, line 38: If remove this annotation, the jackson can't reconstruct object instance for non primitive class due to the lack of   class information. For example, `initialArguments` contains a key-value pair `"myObject": new CustomClass()`

Currently, the initialArguments are always primitive types, or map/list of primitive types, so to keep the consistency with serialization behavior in python side, we can remove this, but we need revise the TODO.

api/src/main/java/org/apache/flink/agents/api/resource/ResourceDescriptor.java
```diff
@@ -20,36 +20,59 @@
...
     private final Map<String, Object> initialArguments;
 
     @JsonCreator
     public ResourceDescriptor(
             @JsonProperty(FIELD_CLAZZ) String clazz,
+            @JsonProperty(FIELD_PYTHON_MODULE) String pythonModule,
+            @JsonProperty(FIELD_PYTHON_CLAZZ) String pythonClazz,
```
wenjin272: Should add doc strings like python side

wenjin272 (commented):
LGTM.

python/flink_agents/api/resource.py
```diff
@@ -73,23 +74,111 @@ def validate_serializable(self) -> "SerializableResource":
...
-        """Get the initialize arguments of the resource."""
-        return self._arguments
+    Attributes:
+        clazz: The Python Resource class name.
+        java_clazz: The Java class full path (e.g., 'com.example.YourJavaClass').
+                   Empty string for Python-only resources.
+        arguments: Dictionary containing resource initialization parameters.
+    """
+    clazz: Type[Resource] | None = None
+    java_clazz: str
```
xintongsong: Why not just use the arguments?

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
```diff
@@ -59,11 +59,22 @@ public class PythonActionExecutor {
...
     private final PythonInterpreter interpreter;
     private final String agentPlanJson;
+    private final JavaResourceAdapter javaResourceAdapter;
     private Object pythonAsyncThreadPool;
 
     public PythonActionExecutor(PythonInterpreter interpreter, String agentPlanJson) {
         this.interpreter = interpreter;
         this.agentPlanJson = agentPlanJson;
+        this.javaResourceAdapter = null;
+    }
```
xintongsong: Where is this constructor used?

python/flink_agents/e2e_tests/chat_model_integration_test.py
```diff
@@ -82,3 +93,59 @@ def test_chat_model_integration(model_provider: str) -> None:  # noqa: D103
 
     assert "3" in output_list[0]["0001"]
     assert "cat" in output_list[1]["0002"]
+
+@pytest.mark.skipif(client is None, reason="Ollama client is not available or test model is missing.")
```
xintongsong: Should not mix cross-language tests with integration e2e tests.

python/flink_agents/e2e_tests/chat_model_integration_test.py
```diff
@@ -82,3 +93,59 @@ def test_chat_model_integration(model_provider: str) -> None:  # noqa: D103
...
+def test_java_chat_model_integration(tmp_path: Path) -> None:  # noqa: D103
+    os.environ["MODEL_PROVIDER"] = "JAVA"
```
xintongsong: How is this variable set?

api/src/main/java/org/apache/flink/agents/api/resource/ResourceDescriptor.java
```diff
@@ -20,36 +20,76 @@
...
 
     @JsonProperty(FIELD_CLAZZ)
     private final String clazz;
 
-    // TODO: support nested map/list with non primitive value.
+    @JsonProperty(FIELD_PYTHON_CLAZZ)
+    private final String pythonClazz;
+
+    @JsonProperty(FIELD_PYTHON_MODULE)
+    private final String pythonModule;
```
xintongsong: Why do we need these?

GreatEugenius:
Hi @xintongsong, I have addressed all the comments. Please take a look.

PR #388, opened 2025-12-22
[api][runtime] Support use long-term memory in action and async compaction.

wenjin272:
Linked issue: #331 

### Purpose of change

Support use long-term memory in action and async compaction.

### Tests

Add it test to verify the async execution of base operations and compaction.

### API

Yes, modify `RunnerContext`

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
@@ -90,6 +90,7 @@ import java.util.Map;
 import java.util.Optional;
 
 import static org.apache.flink.agents.api.configuration.AgentConfigOptions.ACTION_STATE_STORE_BACKEND;
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.JOB_IDENTIFIER;
 import static org.apache.flink.agents.runtime.actionstate.ActionStateStore.BackendType.KAFKA;
 import static org.apache.flink.agents.runtime.utils.StateUtil.*;
 import static org.apache.flink.util.Preconditions.checkState;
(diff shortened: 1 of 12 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/operator/StateUtils.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.flink.runtime.state.StateInitializationContext;
+import org.apache.flink.shaded.guava31.com.google.common.collect.Lists;
+import org.apache.flink.util.Preconditions;
+
+import javax.annotation.Nullable;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Utility class for sink state manipulation. Copy from Apache Paimon. */
```
xintongsong: In our case, this is not related to sink.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/StateUtils.java
```diff
@@ -0,0 +1,73 @@
...
+            String stateName,
+            Class<T> valueClass,
+            T defaultValue)
+            throws Exception {
+        ListState<T> state =
+                context.getOperatorStateStore()
+                        .getUnionListState(new ListStateDescriptor<>(stateName, valueClass));
+
+        List<T> values = new ArrayList<>();
+        state.get().forEach(values::add);
```
xintongsong, line 46: Why would there be multiple values?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -635,6 +644,18 @@ public void initializeState(StateInitializationContext context) throws Exception
...
+        if (jobIdentifier == null) {
+            String initialJobIdentifier =
+                    agentPlan
+                            .getConfig()
+                            .getStr(
+                                    JOB_IDENTIFIER.getKey(),
+                                    getRuntimeContext().getJobInfo().getJobId().toString());
+            jobIdentifier =
+                    StateUtils.getSingleValueFromState(
+                            context, "identifier_state", String.class, initialJobIdentifier);
```
xintongsong: We should respect the configured value, and only read from state when it's not configured.

Issue #385, opened 2025-12-17
[Feature][Doc] Add document for Elasticsearch VectorStore integration


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Add document for Elasticsearch VectorStore integration

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

xintongsong:
@twosom,
Sure. You are assigned. Please move ahead.

PR #397, opened 2025-12-26
doc : add ElasticsearchVectorStore doc

twosom:
Linked issue: #385, #321

### Purpose of change

Added documentation for ElasticsearchVectorStore.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed`

### DIFF
--- a/docs/content/docs/development/vector_stores.md
+++ b/docs/content/docs/development/vector_stores.md
@@ -77,7 +77,6 @@ Vector stores use structured query objects for consistent interfaces:
 ```python
 # Create a semantic search query
 query = VectorStoreQuery(
-    mode=VectorStoreQueryMode.SEMANTIC,
     query_text="What is Apache Flink Agents?",
     limit=3
 )
@@ -180,7 +179,6 @@ class MyAgent(Agent):
         # Create a semantic search query
         user_query = str(event.input)
         query = VectorStoreQuery(
-            mode=VectorStoreQueryMode.SEMANTIC,
             query_text=user_query,
             limit=3
         )
@@ -230,7 +228,7 @@ public class MyAgent extends Agent {
         // Option 1: Manual search via the vector store
         VectorStore vectorStore = (VectorStore) ctx.getResource("vectorStore", ResourceType.VECTOR_STORE);
         String queryText = (String) event.getInput();
-        VectorStoreQuery query = new VectorStoreQuery(VectorStoreQueryMode.SEMANTIC, queryText, 3);
+        VectorStoreQuery query = new VectorStoreQuery(queryText, 3);
         VectorStoreQueryResult result = vectorStore.query(query);
 
         // Option 2: Request context retrieval via built-in events


docs/content/docs/development/vector_stores.md
```diff
@@ -25,7 +25,7 @@ under the License.
 # Vector Stores
 
 {{< hint info >}}
-Vector stores are currently supported in the Python API only. Java API support is planned for future releases.
+Vector stores are supported in both Python and Java APIs.
```
xintongsong: Maybe we can just remove this hint?

docs/content/docs/development/vector_stores.md
```diff
@@ -265,6 +285,105 @@ def chroma_store() -> ResourceDescriptor:
...
+        // Request context retrieval via the vector store
+        ctx.sendEvent(new ContextRetrievalRequestEvent(query, "vectorStore"));
+    }
+
+    @Action(listenEvents = ContextRetrievalResponseEvent.class)
+    public static void onSearchResponse(ContextRetrievalResponseEvent event, RunnerContext ctx) {
+        List<Document> documents = event.getDocuments();
+        // Process the retrieved documents...
+    }
+}
```
xintongsong, line 441: Part of this example (i.e., how to search in the vector store with built-in events) is not specific to ES. We may consider modifying the `Getting Started` section and describe how to do the steps in Java. Then in the `Elasticsearch` section example, we can keep only the ES specific contents (i.e. creating the resource descriptor). Imagine that in future we may support many vector stores, so it might be better to avoid repeated contents as much as possible. WDYT?

docs/content/docs/development/vector_stores.md
```diff
@@ -265,6 +285,105 @@ def chroma_store() -> ResourceDescriptor:
...
+```
+
+{{< /tab >}}
+
+{{< /tabs >}}
+
```
xintongsong, line 447: Shall we also add instructions for building custom vector stores in Java?

docs/content/docs/development/vector_stores.md
```diff
@@ -67,26 +83,64 @@ query = VectorStoreQuery(
...
-```python
-# Execute the query
-result = vector_store.query(query)
+```java
+// Create a semantic search query
+VectorStoreQuery query = new VectorStoreQuery(
+        VectorStoreQueryMode.SEMANTIC,
+        "What is Apache Flink Agents?",
+        3
+);
```
xintongsong, line 94: Unlike in python where we have keyword arguments, in java we'd better have some comments to help users understand the arguments, especially when the arguments are literals rather than variables with meaningful names.
```suggestion
VectorStoreQuery query = new VectorStoreQuery(
        VectorStoreQueryMode.SEMANTIC, // mode
        "What is Apache Flink Agents?", // query text
        3 // limit
);
```

docs/content/docs/development/vector_stores.md
```diff
@@ -153,6 +268,10 @@ class MyAgent(Agent):
 
 #### ChromaVectorStore Parameters
 
+{{< tabs "ChromaVectorStore Parameters" >}}
+
+{{< tab "Python" >}}
```
xintongsong: We probably don't need tabs for the parameters. Even later we support Chroma in Java, I think the parameters should be language-independent.

docs/content/docs/development/vector_stores.md
```diff
@@ -265,6 +396,74 @@ def chroma_store() -> ResourceDescriptor:
...
+#### Prerequisites
+
+1. An Elasticsearch cluster (version 8.0 or later for KNN support).
+2. An index with a `dense_vector` field.
+
+#### ElasticsearchVectorStore Parameters
+
+{{< tabs "ElasticsearchVectorStore Parameters" >}}
+
+{{< tab "Java" >}}
```
xintongsong: Same here for the parameter tabs.

xintongsong (commented):
Thanks for addressing my comments, @twosom. The PR looks quite nice. I just have a few minor comments. Please take a look.

docs/content/docs/development/vector_stores.md
twosom: @xintongsong 

Thank you for the review!

I apologize - I realize now that this constructor doesn't actually exist in the current implementation. I will add it to support this cleaner API.

I'll also apply your suggested inline comments.

Will update shortly.

xintongsong: Currently, `SEMANTIC` is the only supported mode. That's why `KEYWORD` & `HYBRID` are commented out in Python. I think in Java we probably should also hide the modes from user for the moment. That means, we can also comment out the unsupported modes, and use the constructor without the mode argument in the document. WDYT?

twosom: @xintongsong 
That makes sense! I'll update the documentation to use the constructor without the mode argument, keeping it consistent with the Python implementation. Will also comment out the unsupported modes in the Java enum.

xintongsong (approved):

Issue #393, opened 2025-12-24
[Feature] Add document for how to add dependency in pom for flink-agents project.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Add document to tell user how to add dependency in pom for Flink-Agents project.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #399, opened 2025-12-26
Add documentation for adding Maven dependencies

Shekharrajak:
Ref https://github.com/apache/flink-agents/issues/393

- [ ] `doc-needed` 
- [x] `doc-not-needed`


Shekharrajak:
Thanks @wenjin272  for the review. Check the new commit : https://github.com/apache/flink-agents/pull/399/commits/891f52f47e88f595522ee0ea41462452f96ce570

docs/content/docs/get-started/installation.md
```diff
@@ -140,6 +140,70 @@ After building:
...
 
+## Maven Dependencies (Java Development)
+
+For developing Flink Agents applications in Java, add the following dependencies to your `pom.xml`:
+
+### Basic Dependencies for Agent Development
+
+**Required for developing Agent applications:**
+- `flink-agents-api` - Core API interfaces and classes
+- Integration modules - Add the specific integrations you need (e.g., `flink-agents-integrations-chat-models-openai`)
```
wenjin272, line 151: In #433, we add constant strings to point a resource implementation, so integration modules is not required for developing. But they are still need for running in IDE

docs/content/docs/get-started/installation.md
```diff
@@ -140,6 +140,70 @@ After building:
...
+        <artifactId>flink-agents-integrations-chat-models-openai</artifactId>
+        <version>${flink-agents.version}</version>
+    </dependency>
+
+    <!-- Runtime (required for IDE/local execution) -->
+    <dependency>
+        <groupId>org.apache.flink</groupId>
+        <artifactId>flink-agents-runtime</artifactId>
+        <version>${flink-agents.version}</version>
+    </dependency>
```
xintongsong, line 185: Should this also be provided?

docs/content/docs/get-started/installation.md
```diff
@@ -140,6 +140,70 @@ After building:
...
+        <artifactId>flink-clients</artifactId>
+        <version>${flink.version}</version>
+        <scope>provided</scope>
+    </dependency>
+</dependencies>
+```
+
+{{< hint info >}}
+**Note:** Replace `0.2-SNAPSHOT` with the actual release version when available. For SNAPSHOT versions, you may need to add the Apache snapshot repository to your `pom.xml`.
+{{< /hint >}}
```
xintongsong, line 206: I think we can remove this hint and just use `0.2` w/o the `-SNAPSHOT` suffix. The version will soon be available in 1-2 weeks.

docs/content/docs/get-started/installation.md
```diff
@@ -140,6 +140,70 @@ After building:
...
+    </dependency>
+
+    <dependency>
+        <groupId>org.apache.flink</groupId>
+        <artifactId>flink-clients</artifactId>
+        <version>${flink.version}</version>
+        <scope>provided</scope>
+    </dependency>
+</dependencies>
+```
```
xintongsong, line 202: I have a feeling that there are too many dependencies need to be added for execution in an IDE. This not only harms the user experience, but also makes this document hard to maintain as the dependencies need to be added may change in future and this document can easily become inconsistent. I wonder if we can introduce a pom like `flink-agents-ide-supports`, which includes all dependencies needed for the local execution. Of course this can be a separate issue and should not block this PR.
cc @wenjin272

Issue #381, opened 2025-12-11
[Feature] Support async execution for built-in actions


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, we has already support async execution for action in python, and will support it in java in 0.2.0. The built-in actions like chat, tool call process, which may cost much time, should benefit from it.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #400, opened 2025-12-26
Add async execution support for built-in actions

Shekharrajak:
Linked issue: #381

Purpose of change:

Adds async execution support for built-in actions (chat_model_action and tool_call_action)
Improves performance by making LLM calls and tool executions non-blocking
Maintains backward compatibility. 

Tests:

Added test_built_in_actions_async_execution() unit test
Existing tests still pass (backward compatibility verified)
Verified memory/resource access constraints

API:

No public API changes
Internal implementation detail only
Function signatures unchanged
### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed`

### DIFF
--- a/python/flink_agents/plan/actions/chat_model_action.py
+++ b/python/flink_agents/plan/actions/chat_model_action.py
@@ -40,21 +40,33 @@ def chat(
     model: str,
     messages: List[ChatMessage],
     ctx: RunnerContext,
-) -> None:
+):
     """Chat with llm.
 
     If there is no tool call generated, we return the chat response event directly,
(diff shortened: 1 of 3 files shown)


wenjin272:
Hi, @Shekharrajak, thanks for your contribution.

The basic implementation looks good to me. But I wonder if we should provide config options to enable/disable the async execution of built-in actions, for user may want decrease complexity when debug program.

Besides, there are some check failed.

wenjin272:
Hi @Shekharrajak , in #409 @Sxnan refactor the python async api. Could you update this pr?

Issue #349, opened 2025-12-01
[Feature] Introduce Long-Term Memory in Java


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Introduce long-term memory interface and an implementation in java.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #416, opened 2026-01-07
[api][integration][java] Enrich interface for vector store and implement these for elasticsearch.

wenjin272:
Linked issue: #349 

### Purpose of change

This is the first pr to introduce long-term memory in java. 
This pr enrich the interface for vector store due to the long-term memory needed, and implement these for elasticsearch.

### Tests

ut

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/vectorstores/CollectionManageableVectorStore.java
+++ b/api/src/main/java/org/apache/flink/agents/api/vectorstores/CollectionManageableVectorStore.java
@@ -17,21 +17,12 @@
  */
 package org.apache.flink.agents.api.vectorstores;
 
-import org.apache.flink.agents.api.resource.Resource;
-import org.apache.flink.agents.api.resource.ResourceDescriptor;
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/apache/flink/agents/api/vectorstores/CollectionManageableVectorStore.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.apache.flink.agents.api.resource.Resource;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
+import org.apache.flink.agents.api.resource.ResourceType;
+
+import java.util.Map;
+import java.util.function.BiFunction;
+
+/** Base abstract class for vector store which support collection management. */
+public abstract class CollectionManageableVectorStore extends BaseVectorStore {
```
xintongsong: Should this be an interface? Otherwise, if we want to support more optional features for vector stores, we might run into inherit conflicts.

PR #422, opened 2026-01-09
[runtime] Support Fine-grain Durable Execution

Sxnan:
Linked issue: #423

### Purpose of change
- Introduce `CallRecord` into `ActionState` and persist/restore call records across durable execution recovery.
- Expose `execute` API on `runner_context.py` and wire it through local/flink runner contexts.
- Extend Python/Java runtimes to emit and replay call records, aligning durable execution semantics and ensuring operator context restoration.

### Tests
- Added/updated unit tests for `ActionState`/`CallRecord` serde and store, runner context execute path, and durable execution flows.
- Added e2e integration tests for runner context `execute` (basic, multiple calls, async) with ground-truth fixtures.

### API
- Added `execute` to Python `runner_context` (new public API surface); no breaking changes identified.

### Documentation
- [x] `doc-needed`
- [ ] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/runtime/flink_runner_context.py
+++ b/python/flink_agents/runtime/flink_runner_context.py
@@ -15,6 +15,8 @@
 #  See the License for the specific language governing permissions and
 # limitations under the License.
 #################################################################################
+import hashlib
+import logging
 import os
 from concurrent.futures import ThreadPoolExecutor
 from typing import Any, Callable, Dict
@@ -39,11 +41,136 @@ from flink_agents.runtime.memory.vector_store_long_term_memory import (
     VectorStoreLongTermMemory,
 )
 
+logger = logging.getLogger(__name__)
+
+
+class _DurableExecutionResult:
+    """Wrapper that holds result and triggers recording when unwrapped."""
+
+    def __init__(
+        self,
+        func: Callable,
+        args: tuple,
+        kwargs: dict,
+        result: Any,
+        record_callback: Callable,
+    ) -> None:
+        self.func = func
+        self.args = args
+        self.kwargs = kwargs
+        self.result = result
+        self.record_callback = record_callback
+        self._recorded = False
+
+    def get_result(self) -> Any:
+        """Get the result and record completion if not already recorded."""
+        if not self._recorded:
+            self.record_callback(self.func, self.args, self.kwargs, self.result, None)
+            self._recorded = True
+        return self.result
+
+
+class _DurableExecutionException(Exception):
+    """Wrapper exception that holds exception info and triggers recording."""
+
+    def __init__(
+        self,
+        func: Callable,
+        args: tuple,
+        kwargs: dict,
+        result: Any,
+        exception: BaseException,
+        record_callback: Callable,
+    ) -> None:
+        super().__init__(str(exception))
+        self.func = func
+        self.args = args
+        self.kwargs = kwargs
+        self.original_exception = exception
+        self.record_callback = record_callback
+        self._recorded = False
+
+    def record_and_raise(self) -> None:
+        """Record completion and raise the original exception."""
+        if not self._recorded:
+            self.record_callback(
+                self.func, self.args, self.kwargs, None, self.original_exception
+            )
+            self._recorded = True
+        raise self.original_exception from None
+
+
+class _CachedAsyncExecutionResult(AsyncExecutionResult):
+    """An AsyncExecutionResult that returns a cached value immediately."""
+
+    def __init__(self, cached_result: Any) -> None:
+        # Don't call super().__init__ as we don't need executor/func/args/kwargs
+        self._cached_result = cached_result
+
+    def __await__(self) -> Any:
+        """Return the cached result immediately.
+
+        This is a generator that yields nothing and returns the cached result.
+        """
+        if False:
+            yield  # Make this a generator function
+        return self._cached_result
+
+
+class _DurableAsyncExecutionResult(AsyncExecutionResult):
+    """An AsyncExecutionResult that records completion after execution."""
+
+    def __init__(
+        self, executor: Any, func: Callable, args: tuple, kwargs: dict
+    ) -> None:
+        super().__init__(executor, func, args, kwargs)
+
+    def __await__(self) -> Any:
+        """Execute and record completion when awaited."""
+        future = self._executor.submit(self._func, *self._args, **self._kwargs)
+        while not future.done():
+            yield
+
+        result = future.result()
+
+        # Handle the wrapped result/exception
+        if isinstance(result, _DurableExecutionResult):
+            return result.get_result()
+        elif isinstance(result, _DurableExecutionException):
+            result.record_and_raise()
+        else:
+            return result
+
+
+def _compute_function_id(func: Callable) -> str:
+    """Compute a stable function identifier from a callable.
+
+    Returns module.qualname for functions/methods.
+    """
+    module = getattr(func, "__module__", "<unknown>")
+    qualname = getattr(func, "__qualname__", getattr(func, "__name__", "<unknown>"))
+    return f"{module}.{qualname}"
+
+
+def _compute_args_digest(args: tuple, kwargs: dict) -> str:
+    """Compute a stable digest of the serialized arguments.
+
+    The digest is used to validate that the same arguments are passed
+    during recovery as during the original execution.
+    """
+    try:
+        serialized = cloudpickle.dumps((args, kwargs))
+        return hashlib.sha256(serialized).hexdigest()[:16]
+    except Exception:
+        # If serialization fails, return a fallback digest
+        return hashlib.sha256(str((args, kwargs)).encode()).hexdigest()[:16]
+
 
 class FlinkRunnerContext(RunnerContext):
     """Providing context for agent execution in Flink Environment.
 
-    This context allows access to event handling.
+    This context allows access to event handling and provides fine-grained
+    durable execution support through execute() and execute_async() methods.
     """
 
     __agent_plan: AgentPlan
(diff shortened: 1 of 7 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/CallRecord.java
```diff
@@ -0,0 +1,184 @@
...
+ * Represents a record of a function call execution for fine-grained durable execution.
+ *
+ * <p>This class stores the execution result of a single {@code execute} or {@code execute_async}
+ * call, enabling recovery without re-execution when the same call is encountered during job
+ * recovery.
+ *
+ * <p>During recovery, the success or failure of the original call is determined by checking whether
+ * {@code exceptionPayload} is null.
+ */
+public class CallRecord {
```
xintongsong: I'd suggest to name this `CallResult`, which makes it explicit that this class represents a result of call execution.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/CallRecord.java
```diff
@@ -0,0 +1,184 @@
...
+        this.resultPayload = resultPayload;
+    }
+
+    public byte[] getExceptionPayload() {
+        return exceptionPayload;
+    }
+
+    public void setExceptionPayload(byte[] exceptionPayload) {
+        this.exceptionPayload = exceptionPayload;
+    }
```
xintongsong: Why do we need to set the payloads after construction? Can these fields be `final`?

python/flink_agents/api/runner_context.py
```diff
@@ -186,6 +186,48 @@ def action_metric_group(self) -> MetricGroup:
             The individual metric group specific to the current action.
         """
 
+    @abstractmethod
+    def execute(
```
xintongsong: I'd suggest the name `durable_execute()`

python/flink_agents/api/runner_context.py
```diff
@@ -196,6 +238,14 @@ def execute_async(
...
         is prohibited within the function.
 
+        The result of the function will be stored and returned when the same
+        execute_async call is made again during job recovery. The arguments
+        and the result must be serializable.
+
+        The action that calls this API should be deterministic, meaning that it
+        will always make the execute_async call with the same arguments and in
+        the same order during job recovery. Otherwise, the behavior is undefined.
+
```
xintongsong, line 248: I'd suggest to name this method `durable_execute_async`.

xintongsong: I don't see the necessity, but it's still possible that we support non-durable async execution in future, which simply execute in a separate thread/coroutine but the result should not be reused for replaying.

E.g., getting a token from remote, which might be already expired during the replay.

python/flink_agents/runtime/local_runner.py
```diff
@@ -179,6 +179,22 @@ def action_metric_group(self) -> MetricGroup:
...
+        *args: Any,
+        **kwargs: Any,
+    ) -> Any:
+        """Synchronously execute the provided function. Access to memory
+        is prohibited within the function.
+
+        Note: Local runner does not support durable execution, so recovery
+        is not available.
+        """
+        return func(*args, **kwargs)
```
xintongsong, line 198: If we are renaming this to `durable_execute`, we also need to print a warning log here. There's no need to fail the call, as we don't want users to change their codes when switching between local execution & remote execution.

runtime/src/main/java/org/apache/flink/agents/runtime/python/context/PythonRunnerContextImpl.java
```diff
@@ -48,4 +78,120 @@ public void sendEvent(String type, byte[] event, String eventString) {
...
+
+        if (currentActionState != null && actionStatePersister != null) {
+            currentActionState.addCallRecord(callRecord);
+            actionStatePersister.run();
+            LOG.debug(
+                    "Recorded and persisted CallRecord at index {}: functionId={}, argsDigest={}",
+                    currentCallIndex,
+                    functionId,
+                    argsDigest);
+        }
```
xintongsong: Why would `currentActionState` and `actionStatePersister` be `null` when this method is called? Shall we use assertion here?

runtime/src/main/java/org/apache/flink/agents/runtime/python/context/PythonRunnerContextImpl.java
```diff
@@ -48,4 +78,120 @@ public void sendEvent(String type, byte[] event, String eventString) {
...
+     *
+     * <p>This method is called by Python's execute/execute_async to check if a previous result
+     * exists. If found and validated, the cached result is returned.
+     *
+     * @param functionId the function identifier
+     * @param argsDigest the digest of serialized arguments
+     * @return array containing [isHit (boolean), resultPayload (byte[]), exceptionPayload
+     *     (byte[])], or null if miss
+     */
+    public Object[] tryGetCachedCallRecord(String functionId, String argsDigest) {
```
xintongsong: I'd suggest the name `matchNextOrClearSubsequentCallRecord`. Otherwise, the clearing is super implicit.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -459,7 +463,19 @@ private void processActionTaskForKey(Object key) throws Exception {
...
+                maybeInitActionState(key, sequenceNumber, actionTask.action, actionTask.event);
+                actionState =
+                        maybeGetActionState(
+                                key, sequenceNumber, actionTask.action, actionTask.event);
+            }
+
+            // Set up recovery state for fine-grained durable execution
+            if (actionTask instanceof PythonActionTask) {
+                setupRecoveryStateForPythonAction(actionTask, actionState);
+            }
```
xintongsong: Why doing this every time executing the action task? This should be done once when creating the context.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -888,7 +904,59 @@ private void maybePersistTaskResult(
...
+        final Event event = actionTask.event;
+
+        // Initialize recovery state with ActionState and a callback to persist after each code
+        // block
+        pythonContext.initRecoveryState(
+                actionState,
+                () -> {
+                    try {
+                        actionStateStore.put(
+                                key, sequenceNumberKState.value(), action, event, actionState);
```
xintongsong: `sequenceNumberKState` might change when the callback is called.

runtime/src/main/java/org/apache/flink/agents/runtime/python/context/PythonRunnerContextImpl.java
```diff
@@ -20,21 +20,51 @@
...
+    private int currentCallIndex = 0;
+
+    /** List of existing CallRecords loaded during recovery. */
+    private List<CallRecord> recoveryCallRecords;
+
+    /** The current ActionState being built during action execution. */
+    @Nullable private ActionState currentActionState;
+
+    /** Callback to persist ActionState after each code block completion. */
+    @Nullable private Runnable actionStatePersister;
```
xintongsong: It's kind of weird that these are maintained in `PythonRunnerContextImpl`.
1. These are not only needed by python. We will also support this for Java.
2. These are actually ActionTask-specific. I think we should handle this in `RunnerContextImpl.switchActionContext()`. We might want to actually introduce an `ActionTaskContext`.

python/flink_agents/api/runner_context.py
yanand0909: Just another suggestion if it makes sense, we can separate durable specific operations

```
  runner_context.durable.execute(...) 
  runner_context.durable.execute_async(...) 
```

python/flink_agents/runtime/flink_runner_context.py
```diff
@@ -194,8 +436,37 @@ def execute_async(
...
+        # Create a wrapper function that records completion
+        def wrapped_func(*a: Any, **kw: Any) -> Any:
+            exception = None
+            result = None
+            try:
+                result = func(*a, **kw)
+            except BaseException as e:
+                exception = e
+
+            # Note: This runs in a thread pool, so we need to be careful
```
yanand0909, line 465: We should add this in documentation that the recording will happen when the result is awaited and will not work for fire and forget function calls

runtime/src/main/java/org/apache/flink/agents/runtime/python/context/PythonRunnerContextImpl.java
```diff
@@ -48,4 +78,120 @@ public void sendEvent(String type, byte[] event, String eventString) {
...
+                LOG.debug(
+                        "CallRecord hit at index {}: functionId={}, argsDigest={}",
+                        currentCallIndex,
+                        functionId,
+                        argsDigest);
+                currentCallIndex++;
+                return new Object[] {true, record.getResultPayload(), record.getExceptionPayload()};
+            } else {
+                // Non-deterministic call detected, clear subsequent records
+                LOG.warn(
```
yanand0909: Should we give an option to fail fast here? some config like strict_mode?

Sxnan: I think strict_mode could be useful for development/testing to catch non-deterministic issues early.
For now, I'd prefer to keep the current lenient behavior as the default since it's more resilient in production. We can add this as a follow-up enhancement. WDYT?

xintongsong: @yanand0909 , is there any concrete use case you see that makes failing fast necessary? I'm asking because I just can't think of why a user may want to fail the agent simply because it behaves differently before / after the recovery.

xintongsong (approved):
Thanks for addressing my comments. This looks a lot better after the refactor. My only concern is that `ActionExecutionOperator#processActionTaskForKey` is becoming a bit too complex and may need a further refactor. But let's take it as a follow-up and unblock this PR.

runtime/src/main/java/org/apache/flink/agents/runtime/python/context/PythonRunnerContextImpl.java
yanand0909: Mostly I had Testing Environments (it would be easy to test non deterministic behavior) , use case like Financial Agents where retry is completely unwanted even during recovery. Having fail fast options provides more flexibility. We can add this as follow up enhancement if you feel it will be useful @Sxnan  @xintongsong

yanand0909 (approved):
Thanks for addressing my comments @Sxnan! This looks good.

PR #124, opened 2025-08-20
[hotfix]Enrich test cases of Short-Term Memory in java

KeGu-069:
Linked issue: #123 

### Purpose of change

Add test cases of Short-Term Memory in java.

PR #425, opened 2026-01-09
[api][runtime][java] Introduce Long-Term Memory.

wenjin272:
Linked issue: #349

### Purpose of change

Introduce Long-Term Memory in Java.

### Tests

ut & e2e

### API

Yes, add Long-Term Memory interfaces in java.

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/VectorStoreLongTermMemoryTest.java
+++ b/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/VectorStoreLongTermMemoryTest.java
@@ -1,0 +1,333 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.integration.test;
+
+import org.apache.flink.agents.api.AgentsExecutionEnvironment;
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.apache.flink.agents.api.chat.messages.MessageRole;
+import org.apache.flink.agents.api.configuration.AgentConfigOptions;
+import org.apache.flink.agents.api.context.RunnerContext;
+import org.apache.flink.agents.api.memory.LongTermMemoryOptions;
+import org.apache.flink.agents.api.memory.MemorySet;
+import org.apache.flink.agents.api.memory.MemorySetItem;
+import org.apache.flink.agents.api.memory.compaction.SummarizationStrategy;
+import org.apache.flink.agents.api.resource.Resource;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
+import org.apache.flink.agents.api.resource.ResourceType;
+import org.apache.flink.agents.integrations.chatmodels.ollama.OllamaChatModelConnection;
+import org.apache.flink.agents.integrations.chatmodels.ollama.OllamaChatModelSetup;
+import org.apache.flink.agents.integrations.embeddingmodels.ollama.OllamaEmbeddingModelConnection;
+import org.apache.flink.agents.integrations.embeddingmodels.ollama.OllamaEmbeddingModelSetup;
+import org.apache.flink.agents.integrations.vectorstores.elasticsearch.ElasticsearchVectorStore;
+import org.apache.flink.agents.plan.AgentConfiguration;
+import org.apache.flink.agents.runtime.memory.VectorStoreLongTermMemory;
+import org.apache.flink.api.common.eventtime.WatermarkStrategy;
+import org.apache.flink.connector.file.src.FileSource;
+import org.apache.flink.connector.file.src.reader.TextLineInputFormat;
+import org.apache.flink.core.fs.Path;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+import org.mockito.Mockito;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Test for {@link VectorStoreLongTermMemory}
+ *
+ * <p>We use {@link ElasticsearchVectorStore} as the backend of Long-Term Memory, so need setup
+ * Elasticsearch server to run this test. Look <a
+ * href="https://www.elastic.co/docs/deploy-manage/deploy/self-managed/install-elasticsearch-docker-basic">Start
+ * a single-node cluster in Docker</a> for details.
+ *
+ * <p>For {@link ElasticsearchVectorStore} doesn't support security check yet, when start the
+ * container, should add "-e xpack.security.enabled=false" option.
+ */
+@Disabled("Should setup Elasticsearch server.")
+public class VectorStoreLongTermMemoryTest {
+    private static final Logger LOG = LoggerFactory.getLogger(VectorStoreLongTermMemoryTest.class);
+
+    private static final String NAME = "chat-history";
+    private final VectorStoreLongTermMemory ltm;
+    private MemorySet memorySet;
+    private List<ChatMessage> messages;
+
+    public static Resource getResource(String name, ResourceType type) {
+        if (type == ResourceType.CHAT_MODEL_CONNECTION) {
+            return new OllamaChatModelConnection(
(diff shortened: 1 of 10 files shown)


e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/VectorStoreLongTermMemoryTest.java
```diff
@@ -0,0 +1,329 @@
...
+                            .toDataStream();
+
+            // Print the analysis results to stdout.
+            outputStream.print();
+
+            // Execute the Flink pipeline.
+            agentsEnv.execute();
+
+            // check async compaction
+            System.out.print(es.get(null, "ltm_test_job-2-test-ltm", Collections.emptyMap()));
```
yanand0909: Was this for debugging? Remove this?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/CompactionFunctions.java
```diff
@@ -0,0 +1,224 @@
...
+                    created_times.add((LocalDateTime) items.get(index).getCreatedTime());
+                }
+
+                lastAccessedTimes.add(items.get(index).getLastAccessedTime());
+
+                itemIds.add(items.get(index).getId());
+            }
+
+            Preconditions.checkArgument(!created_times.isEmpty());
+            String start =
```
yanand0909: Can we calculate global min at line #124 instead of adding to list and iterating again?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/CompactionFunctions.java
```diff
@@ -0,0 +1,224 @@
...
+                    created_times.stream().min(LocalDateTime::compareTo).get().format(formatter);
+            String end =
```
yanand0909: Similar for end and lastAccessedTime?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/CompactionFunctions.java
```diff
@@ -0,0 +1,224 @@
...
+                            Map.of(
+                                    "compacted",
+                                    true,
+                                    "created_time_start",
+                                    start,
+                                    "created_time_end",
+                                    end,
+                                    "last_accessed_time",
+                                    lastAccessedTime)));
+            System.out.print("summarize");
```
yanand0909: Remove this?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
```diff
@@ -0,0 +1,307 @@
...
+                        .exceptionally(
+                                e -> {
+                                    throw new RuntimeException(
+                                            String.format(
+                                                    "Compaction for %s failed",
+                                                    this.nameMangling(memorySet.getName())),
+                                            e);
+                                });
+            } else {
+                this.compact(memorySet);
```
yanand0909, line 213: When `asyncCompaction = false`, the entire add() call blocks during compaction this blocks the Flink operator and could cause backpressure?

Should we document this clearly or make async the default?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
```diff
@@ -0,0 +1,307 @@
...
+            }
+        }
+
+        return itemIds;
+    }
+
+    @Override
+    public List<MemorySetItem> get(MemorySet memorySet, @Nullable List<String> ids)
```
yanand0909, line 222: Should we add pagination parameters or a separate `getAll()` method with warnings since for a large memory set this could cause OOM?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
```diff
@@ -0,0 +1,307 @@
...
+        if (strategy.type() == CompactionStrategy.Type.SUMMARIZATION) {
+            summarize(this, memorySet, ctx, null);
+        } else {
+            throw new RuntimeException(
+                    String.format("Unknown compaction strategy: %s", strategy.type()));
+        }
+    }
+
+    private ExecutorService workerExecutor() {
+        // TODO: shutdown executor when close.
```
yanand0909, line 307: Do you want to address this TODO in this PR?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
```diff
@@ -0,0 +1,307 @@
...
+        if (lazyCompactExecutor == null) {
+            lazyCompactExecutor =
+                    Executors.newFixedThreadPool(
+                            2,
```
yanand0909: Should we make number of threads configurable?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
```diff
@@ -0,0 +1,307 @@
...
+
+        List<String> itemIds =
+                this.store()
+                        .add(
+                                documents,
+                                this.nameMangling(memorySet.getName()),
+                                Collections.emptyMap());
+
+        if (memorySet.size() >= memorySet.getCapacity()) {
+            if (this.asyncCompaction) {
```
yanand0909, line 189: QQ: Could this cause race condition?  If multiple threads add items simultaneously, compaction could be triggered multiple times before any completes

api/src/main/java/org/apache/flink/agents/api/memory/BaseLongTermMemory.java
```diff
@@ -0,0 +1,141 @@
...
+            MemorySet memorySet, String query, int limit, Map<String, Object> extraArgs)
+            throws Exception;
+
+    /**
+     * Switches the context for the long-term memory operations. This allows the same memory
+     * instance to be used for different key by isolating data based on the provided job ID and key.
+     *
+     * @param key the context key
+     */
+    void switchContext(String key);
```
xintongsong: This should not be exposed to users.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
wenjin272: Actually I addressed this TODO in this PR, but I forgot remove this TODO.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
```diff
@@ -0,0 +1,307 @@
...
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import static org.apache.flink.agents.runtime.memory.CompactionFunctions.summarize;
+
+public class VectorStoreLongTermMemory implements BaseLongTermMemory {
+    private static final Logger LOG = LoggerFactory.getLogger(VectorStoreLongTermMemory.class);
```
xintongsong: Never used.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
```diff
@@ -0,0 +1,307 @@
...
+        this.ctx = ctx;
+        this.vectorStore = vectorStore;
+        this.jobId = jobId;
+        this.key = key;
+        this.asyncCompaction = ctx.getConfig().get(LongTermMemoryOptions.ASYNC_COMPACTION);
+    }
+
+    @Override
+    public void switchContext(String key) {
+        this.key = key;
```
xintongsong, line 89: How is this handled in python?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/CompactionFunctions.java
```diff
@@ -0,0 +1,224 @@
...
+     * in summarization.
+     *
+     * @param ltm The long term memory the memory set belongs to.
+     * @param memorySet The memory set to be summarized.
+     * @param ctx The runner context used to retrieve needed resources.
+     * @param ids The ids of items to be summarized. If not provided, all items will be involved in
+     *     summarization. Optional.
+     */
+    @SuppressWarnings("unchecked")
+    public static void summarize(
```
xintongsong, line 87: It's weird that we need to duplicate this for python & java.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
wenjin272: I will clarify this in documents.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
wenjin272: Yes, when user retrieve too many items at once, this may cause OOM, not only get all items, but also when provide a large list of ids. I will remind users of this in the documentation.

Supporting pagination parameters and returning a page of the entire set each time should help resolve this issue. I'll add a TODO here and implement it in the next version (0.3).

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
wenjin272: Well spotted! I add a `AtomicBoolean` to avoid trigger multiple times compaction for same memory set.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/VectorStoreLongTermMemory.java
wenjin272: Currently, python will create a new LTM object each time when switch context. Because the underlying resource is reused, it won't cause too much cost. 

After we unify the implementation of LTM to java, this should be also removed in python.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/CompactionFunctions.java
wenjin272: As we discuss offline, we will unify the implementation in next version: #426

Issue #362, opened 2025-12-02
[Feature] Support the use of Java EmbeddingModels in Python


GreatEugenius:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support the use of Java EmbeddingModels in Python Flink Agents jobs.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #427, opened 2026-01-12
[Feature][runtime] Support the use of Java EembeddingModel in Python

GreatEugenius:
Linked issue: #362 

### Purpose of change
- Support using JavaChatModelConnection directly in Python
- Support Python resource retrieval through the Java resource using the `getResource` function
    The previous implementation had the following issues:
    
    - **PythonResourceProvider couldn't directly create PythonResource**: In the old design, `PythonResourceProvider` relied on `PythonResourceAdapter` to instantiate `PythonResource`. However, `PythonResourceAdapter` was only created when a Java component invoked a Python resource.
    
    **Changes made:**
    
    - Rewrote the creation logic for `PythonResourceAdapter` so that it is automatically instantiated whenever a Python resource exists.
    - Aligned the `getResource` logic in `JavaResourceAdapter` with that of `AgentPlan`, eliminating redundant code by removing the duplicated implementation.

- Support Java resource retrieval through the Python resource using the `get_resource` function
- Support the use of the Java EembeddingModel in Python

### Tests
E2E test

### API
No

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
@@ -28,6 +28,7 @@ import org.apache.flink.agents.api.logger.EventLogger;
 import org.apache.flink.agents.api.logger.EventLoggerConfig;
 import org.apache.flink.agents.api.logger.EventLoggerFactory;
 import org.apache.flink.agents.api.logger.EventLoggerOpenParams;
+import org.apache.flink.agents.api.resource.Resource;
 import org.apache.flink.agents.api.resource.ResourceType;
 import org.apache.flink.agents.plan.AgentPlan;
 import org.apache.flink.agents.plan.JavaFunction;
@@ -141,6 +142,9 @@ public class ActionExecutionOperator<IN, OUT> extends AbstractStreamOperator<OUT
 
     // PythonResourceAdapter for Python resources in Java actions
     private transient PythonResourceAdapterImpl pythonResourceAdapter;
+
+    // PythonResourceAdapter for Java resources in Python actions or Python resources
+    private transient JavaResourceAdapter javaResourceAdapter;
 
     private transient FlinkAgentsMetricGroupImpl metricGroup;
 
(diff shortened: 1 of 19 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/JavaResourceAdapter.java
```diff
@@ -121,4 +122,15 @@ public ChatMessage fromPythonChatMessage(Object pythonChatMessage) {
...
     }
+
+
+    /**
+     * Retrieves a Java tool by name. This method is intended for use by the Python interpreter.
+     *
+     * @param name the name of the tool to retrieve
+     * @return the tool
+     */
+    public Tool fromToolName(String name) throws Exception {
```
xintongsong: I'd suggest the name `getTool`, similar to `getResource` except that the type is already known.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -576,17 +579,25 @@ private void initPythonEnvironment() throws Exception {
...
+
+            javaResourceAdapter =
+                    new JavaResourceAdapter(
+                            (String anotherName, ResourceType anotherType) -> {
+                                try {
+                                    return agentPlan.getResource(anotherName, anotherType);
+                                } catch (Exception e) {
+                                    throw new RuntimeException(e);
+                                }
+                            },
```
xintongsong: Minor: Le't make this a method of ActionExecutionOperator, and use `this::getResource` here.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/JavaResourceAdapter.java
xintongsong: Why not just use `getResource`?

GreatEugenius:
Hi @xintongsong, I have addressed all your comments. Please take a look when you have a moment.

xintongsong (approved):
LGTM

Issue #345, opened 2025-11-28
[Tech Debt] Support close method for Resource


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

In Flink, when job failed, canceled or finished, the close method of operator will be called to clean up resources link persistence connection. 

Currently, some resource in flink-agents, like vector store, may also contains persistence connection, so we should also add close interface for `Resource` and close them when close operator.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #428, opened 2026-01-12
[api][runtime] Clean up resource when close.

wenjin272:
Linked issue: #345 

### Purpose of change

Some resource may keep socket connection or other resources, we should clean up these resources when job failed/finished/cancelled.

### Tests

e2e test

### API

No

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/python/PythonChatModelConnection.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/python/PythonChatModelConnection.java
@@ -87,4 +87,9 @@ public class PythonChatModelConnection extends BaseChatModelConnection
         Object pythonMessageResponse = adapter.callMethod(chatModel, "chat", kwargs);
         return adapter.fromPythonChatMessage(pythonMessageResponse);
     }
+
+    @Override
+    public void close() throws Exception {
+        this.chatModel.close();
+    }
 }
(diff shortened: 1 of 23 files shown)


api/src/main/java/org/apache/flink/agents/api/chat/model/python/PythonChatModelConnection.java
```diff
@@ -40,7 +41,7 @@
  */
 public class PythonChatModelConnection extends BaseChatModelConnection
         implements PythonResourceWrapper {
-    private final PyObject chatModel;
+    private PyObject chatModel;
```
xintongsong: I don't see a strong reason making this non-final. Better keep it final if possible.

api/src/main/java/org/apache/flink/agents/api/embedding/model/python/PythonEmbeddingModelConnection.java
```diff
@@ -43,7 +44,7 @@
 public class PythonEmbeddingModelConnection extends BaseEmbeddingModelConnection
         implements PythonResourceWrapper {
 
-    private final PyObject embeddingModel;
+    private PyObject embeddingModel;
```
xintongsong: ditto

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
```diff
@@ -105,6 +105,9 @@ public ResourceType getResourceType() {
         return ResourceType.CHAT_MODEL;
     }
 
+    @Override
+    public void close() throws Exception {}
```
xintongsong: We can add an empty default implementation in the root abstract class. In this way, we won't need to add this for all the resources. We only need to override it in resources that actually need to close something.

wenjin272:
The failed test case `ChatModelCrossLanguageTest.testChatModeIntegration` is caused by the ollama server update, for the CI always install the latest version. See #435 for details.

Therefore, this failure case is unrelated to this PR and should not block its merge.

Issue #432, opened 2026-01-14
[Feature] Use class name to declare resources and provide constant for built-in resources.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, when user want to declare a resource in agent, they should write code as fllow:
java:
```
ResourceDescriptor.Builder.newBuilder(OllamaEmbeddingModelSetup.class.getName())
                    .addInitialArgument("connection", "embeddingConnection")
                    .addInitialArgument("model", OLLAMA_MODEL)
                    .build();
```
python:
```
ResourceDescriptor(
            clazz=Constant.OLLAMA_CHAT_MODEL_CONNECTION, request_timeout=240.0
        )
```
which requires importing the specific integration module.

To make python also point a resource by class name, and provide constant strings for built-in resource, we can move the integration importing.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #433, opened 2026-01-14
 [api ]Provide constant to point a resource to avoid importing integration package.

wenjin272:
Linked issue: #432 

### Purpose of change

 Provide constant to point a resource to avoid importing integration package.

### Tests

existing tests

### API

Yes, python `ResourceDescriptor` use class name(including module) rather than class to point a resource implementation.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/chat_models.md
+++ b/docs/content/docs/development/chat_models.md
@@ -65,7 +65,7 @@ class MyAgent(Agent):
     @staticmethod
     def ollama_connection() -> ResourceDescriptor:
         return ResourceDescriptor(
-            clazz=OllamaChatModelConnection,
+            clazz=Constant.OLLAMA_CHAT_MODEL_CONNECTION,
             base_url="http://localhost:11434",
             request_timeout=30.0
         )
@@ -74,7 +74,7 @@ class MyAgent(Agent):
     @staticmethod
     def ollama_chat_model() -> ResourceDescriptor:
         return ResourceDescriptor(
-            clazz=OllamaChatModelSetup,
+            clazz=Constant.OLLAMA_CHAT_MODEL_SETUP,
             connection="ollama_connection",
             model="qwen3:8b",
             temperature=0.7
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/apache/flink/agents/api/resource/Constant.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * <p>TODO: Could use service provider interface to simplify the definition of constant in the
+ * future.
+ */
+public class Constant {
+    // Built-in ChatModel
+    // ollama
+    public static String OLLAMA_CHAT_MODEL_CONNECTION =
+            "org.apache.flink.agents.integrations.chatmodels.ollama.OllamaChatModelConnection";
+    public static String OLLAMA_CHAT_MODEL =
```
xintongsong: Let's call it `OLLAMA_CHAT_MODEL_SETUP`.

python/flink_agents/api/resource.py
```diff
@@ -198,3 +209,47 @@ def get_resource_class(module_path: str, class_name: str) -> Type[Resource]:
...
     module = importlib.import_module(module_path)
     return getattr(module, class_name)
+
+
+class Constant:
+    """Constant strings for pointing a built-in resource implementation."""
+    # Built-in ChatModel
+    # java wrapper
+    JAVA_CHAT_MODEL_CONNECTION = "flink_agents.api.chat_models.java_chat_model.JavaChatModelConnection"
+    JAVA_CHAT_MODEL = "flink_agents.api.chat_models.java_chat_model.JavaChatModelSetup"
```
xintongsong: ditto

xintongsong: Why don't we include python resources in the java Constants?

xintongsong (approved):
LGTM

Issue #412, opened 2026-01-06
[Bug] Failed to connect to remote mcp server


qingfei1994:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

I customized a ReactAgent with a remote mcp server something like this.

```
class ReActAgentWithMCP(ReActAgent):
    """ReAct Agent with MCP integration."""
    @mcp_server
    @staticmethod
    def my_mcp_server() -> MCPServer:
        """Define MCP server connection."""
        return MCPServer(endpoint="https://xxxx/mcp/")
````

But fail to use MCP server. 

```
2026-01-06 14:31:37,332 - INFO - Using Any for unsupported type: typing.Sequence[~T]
2026-01-06 14:31:37,400 - INFO - No module named google.cloud.bigquery_storage_v1. As a result, the ReadFromBigQuery transform *CANNOT* be used with `method=DIRECT_READ`.
2026-01-06 14:31:37,619 - INFO - HTTP Request: POST https://test.work/mcp/ "HTTP/1.1 200 OK"
2026-01-06 14:31:37,619 - INFO - Negotiated protocol version: 2025-03-26
2026-01-06 14:31:37,661 - INFO - HTTP Request: POST https://test.work/mcp/ "HTTP/1.1 202 Accepted"
2026-01-06 14:31:37,841 - INFO - HTTP Request: POST https://test.work/mcp/ "HTTP/1.1 200 OK"
  + Exception Group Traceback (most recent call last):
  |   File "/flink-agents/python/flink_agents/examples/react_agent_diagnose.py", line 140, in <module>
  |     main()
  |   File "/flink-agents/python/flink_agents/examples/react_agent_diagnose.py", line 128, in main
  |     .apply(log_analysis_react_agent)
  |   File "/flink-1.20.3/bin/test_venv/lib/python3.10/site-packages/flink_agents/runtime/remote_execution_environment.py", line 97, in apply
  |     self.__agent_plan = AgentPlan.from_agent(agent, self.__config)
  |   File "/flink-1.20.3/bin/test_venv/lib/python3.10/site-packages/flink_agents/plan/agent_plan.py", line 137, in from_agent
  |     for provider in _get_resource_providers(agent):
  |   File "/flink-1.20.3/bin/test_venv/lib/python3.10/site-packages/flink_agents/plan/agent_plan.py", line 330, in _get_resource_providers
[...]

PR #434, opened 2026-01-14
Support connecting to remote MCP with predefined prompts

qingfei1994:
Linked issue: #412 
- for MCP server without list_prompts interface, we could return the empty list from list_prompts interface
- For MCPTool, remove the exclude=True parameter from the mcpServer field to ensure it is serialized

### Purpose of change

- Make sure remote MCPServer could be defined and called successfully

### Tests

Tests with remote MCP

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/context/RunnerContext.java
+++ b/api/src/main/java/org/apache/flink/agents/api/context/RunnerContext.java
@@ -19,6 +19,7 @@ package org.apache.flink.agents.api.context;
 
 import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.api.configuration.ReadableConfiguration;
+import org.apache.flink.agents.api.memory.BaseLongTermMemory;
 import org.apache.flink.agents.api.metrics.FlinkAgentsMetricGroup;
 import org.apache.flink.agents.api.resource.Resource;
 import org.apache.flink.agents.api.resource.ResourceType;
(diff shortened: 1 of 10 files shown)


python/flink_agents/integrations/mcp/mcp.py
```diff
@@ -79,7 +79,7 @@ class MCPPrompt(Prompt):
     description: str | None = None
     prompt_arguments: list[PromptArgument] = Field(default_factory=list)
     mcp_server: "MCPServer" = Field(default=None, exclude=True)
```
wenjin272, line 81: Should we also remove the exclude here?

python/flink_agents/integrations/mcp/mcp.py
```diff
@@ -256,6 +267,8 @@ def get_tool_metadata(self, name: str) -> ToolMetadata:
 
     def list_prompts(self) -> List[MCPPrompt]:
         """List available prompts from the MCP server."""
+        if self.prompts is not None:
+            return self.prompts
         return asyncio.run(self._list_prompts_async())
```
wenjin272, line 259: If self.prompts is None, mcp will still list prompts from server. In this case, if server doesn't support list prompt, it will cause the same issue this pr want to resolve.

python/flink_agents/integrations/mcp/mcp.py
qingfei1994: sure, let me fix this

PR #439, opened 2026-01-15
[Feature][runtime] Support the use of Java VectorStore in Python

GreatEugenius:
Linked issue: #363 

### Purpose of change
Support the use of Java VectorStore in Python.

### Tests
E2E test

### API
No

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,6 +165,21 @@ jobs:
   cross_language_tests:
     name: cross-language [${{ matrix.os }}] [python-${{ matrix.python-version}}] [java-${{ matrix.java-version}}]
     runs-on: ${{ matrix.os }}
+    services:
+      elasticsearch:
+        image: docker.elastic.co/elasticsearch/elasticsearch:8.19.0
+        env:
+          discovery.type: single-node
+          xpack.security.enabled: false
+          ES_JAVA_OPTS: "-Xms512m -Xmx512m"
+        ports:
+          - 9200:9200
+        options: >-
+          --health-cmd "curl -f http://localhost:9200/_cluster/health || exit 1"
+          --health-interval 10s
+          --health-timeout 5s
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -173,6 +188,15 @@ jobs:
...
       - uses: actions/checkout@v4
+      - name: Free disk space
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be
+        with:
+          tool-cache: true
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: true
+          docker-images: true
```
xintongsong, line 201: Why do we need this step?

python/flink_agents/e2e_tests/e2e_tests_resource_cross_language/vector_store_cross_language_agent.py
```diff
@@ -0,0 +1,207 @@
...
+
+    @vector_store
+    @staticmethod
+    def vector_store() -> ResourceDescriptor:
+        """Vector store setup for knowledge base."""
+        return ResourceDescriptor(
+            clazz=JavaCollectionManageableVectorStore,
+            java_clazz="org.apache.flink.agents.integrations.vectorstores.elasticsearch.ElasticsearchVectorStore",
+            embedding_model="embedding_model",
+            host=os.environ.get("ES_HOST", "http://localhost:9200"),
```
xintongsong: The test case should not assume that it's always executed in the GHA pipeline and the ES service always exist. We can set the ES_HOST env in the ci pipeline. And if the env is not set, the test should be skipped.

GreatEugenius:
Hi @xintongsong, I have addressed all your comments. Please take a look when you have a moment.

PR #440, opened 2026-01-15
[plan][python] Built-in actions support async execution.

wenjin272:
Linked issue: #381 

### Purpose of change
Python built-in actions support async execution.

### Tests

e2e

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

PR #441, opened 2026-01-16
[fix][runtime] Fix not switch memory context for async action task.

wenjin272:
### Purpose of change

This is the first commit of #440 to resolve the bug that not switching memory context for async action task. For #440 occurs some other bugs, I create this separate PR.

### Tests

e2e in #440 

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`


wenjin272:
This is related to async execution, could you please help take a check @Sxnan.

Sxnan:
Thanks for the fix. I have the same fix in my PR https://github.com/apache/flink-agents/pull/438/changes#diff-2d5bb3af3f20e3592ac192b27364fbdf9c883e40bb856b161f1ed085082cf386L828

To avoid conflict, how about we fix it with my PR

PR #446, opened 2026-01-16
[plan][java] Built-in actions support async execution.

wenjin272:
Linked issue: #381 

### Purpose of change

Support async execution for java built-in actions

### Tests

e2e

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
@@ -215,30 +215,30 @@ public class ChatModelAction {
 
         ChatMessage response = null;
 
+        DurableCallable<ChatMessage> callable =
+                new DurableCallable<>() {
+                    @Override
+                    public String getId() {
+                        return "chat";
+                    }
+
+                    @Override
+                    public Class<ChatMessage> getResultClass() {
+                        return ChatMessage.class;
+                    }
+
+                    @Override
+                    public ChatMessage call() throws Exception {
+                        return chatModel.chat(messages, Map.of());
+                    }
+                };
+
         for (int attempt = 0; attempt < numRetries + 1; attempt++) {
             try {
-                if (chatAsync) {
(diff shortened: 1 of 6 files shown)


wenjin272:
[hs_err_pid84909.log](https://github.com/user-attachments/files/24674348/hs_err_pid84909.log)
Async execution for `PythonChatModelSetup` will cause jvm crash, so the async execution for cross language resources is disabled.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -210,7 +213,28 @@ public static void chat(
...
+                                            return ChatMessage.class;
+                                        }
+
+                                        @Override
+                                        public ChatMessage call() throws Exception {
+                                            return chatModel.chat(messages, Map.of());
+                                        }
+                                    });
+                } else {
+                    response = chatModel.chat(messages, Map.of());
```
xintongsong: Shall we also use `durableExecute` for the synchronized model calls? Same for the other resources and in python.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -198,6 +199,8 @@ public static void chat(
                 (BaseChatModelSetup) ctx.getResource(model, ResourceType.CHAT_MODEL);
 
         boolean chatAsync = ctx.getConfig().get(AgentExecutionOptions.CHAT_ASYNC);
+        // TODO: python chat model doesn't support async execution yet.
+        chatAsync = chatAsync && !(chatModel instanceof PythonChatModelSetup);
```
xintongsong, line 204: It would be better to point to an issue that describes and tracks the problem blocking python chat model from supporting async execution.

xintongsong (approved):
LGTM

PR #447, opened 2026-01-17
support mcp without list_prompts interface

qingfei1994:
Linked issue: https://github.com/apache/flink-agents/issues/412

For MCP server without list_prompts interface, we could return the empty list from list_prompts interface
For MCPTool, remove the exclude=True parameter from the mcpServer field to ensure it is serialized

### Tests

Test connecting to a McpServer without list_prompts interface

### API

Yes

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/python/flink_agents/e2e_tests/e2e_tests_integration/e2e_tests_mcp/mcp_test.py
+++ b/python/flink_agents/e2e_tests/e2e_tests_integration/e2e_tests_mcp/mcp_test.py
@@ -72,7 +72,7 @@ class MyMCPAgent(Agent):
     @mcp_server
     @staticmethod
     def my_mcp_server() -> ResourceDescriptor:
-        """Define MCP server connection based on MCP_SERVER_MODE environment variable."""
+        """Define MCP server connection based on MCP_SERVER_MODE env variable."""
         mcp_mode = os.environ.get("MCP_SERVER_MODE", "with_prompts")
         if mcp_mode == "without_prompts":
             endpoint = MCP_SERVER_ENDPOINT_WITHOUT_PROMPTS
@@ -151,7 +151,7 @@ client = pull_model(OLLAMA_MODEL)
 
 
 @pytest.mark.parametrize(
-    "mcp_server_mode,server_file,server_endpoint",
+    ("mcp_server_mode", "server_file", "server_endpoint"),
     [
         ("with_prompts", "mcp_server.py", MCP_SERVER_ENDPOINT),
         ("without_prompts", "mcp_server_without_prompts.py", MCP_SERVER_ENDPOINT_WITHOUT_PROMPTS),


docs/content/docs/development/mcp.md
```diff
@@ -187,4 +187,5 @@ public class ReviewAnalysisAgent extends Agent {
 
 **Key points:**
 - All tools and prompts from the MCP server are automatically registered.
-- Reference MCP prompts and tools by their names, like reference [local prompt]({{< ref "docs/development/prompts#using-prompts-in-agents" >}}) and [function tool]({{< ref "docs/development/tool_use#define-tool-as-static-method-in-agent-class" >}}) .
\ No newline at end of file
+- Reference MCP prompts and tools by their names, like reference [local prompt]({{< ref "docs/development/prompts/#using-prompts-in-agents" >}}) and [function tool]({{< ref "docs/development/tool_use/#define-tool-as-static-method-in-agent-class" >}}) .
+- For MCP servers that do not support the list_prompts interface, the list_prompts method will return an empty list. In such cases, you need to define a LocalPrompt to call the tools.
```
wenjin272: I think prompt is not required for calling MCP tools. The LLM understand the tools according to their metadata.

python/flink_agents/integrations/mcp/mcp.py
```diff
@@ -271,7 +271,10 @@ def get_tool_metadata(self, name: str) -> ToolMetadata:
 
     def list_prompts(self) -> List[MCPPrompt]:
         """List available prompts from the MCP server."""
-        return asyncio.run(self._list_prompts_async())
+        try:
+            asyncio.run(self._list_prompts_async())
+        except Exception:
```
wenjin272: Return empty list for any exception may be too broad. Can we just return empty list for the exception that tells the mcp server doesn't support list prompt?

docs/content/docs/development/mcp.md
qingfei1994: so you mean we actually doesn't need to document on this?

python/flink_agents/integrations/mcp/mcp.py
qingfei1994: I've updated to "return empty list when mcp server doesn't support list_prompts", please take a look

docs/content/docs/development/mcp.md
wenjin272: Yes.

python/flink_agents/e2e_tests/e2e_tests_integration/e2e_tests_mcp/mcp_test.py
```diff
@@ -162,3 +226,36 @@ def test_mcp() -> None:  # noqa:D103
             print(f"{key}: {value}")
 
     server_process.kill()
+
+
+@pytest.mark.skipif(
+    client is None, reason="Ollama client is not available or test model is missing"
+)
+def test_mcp_without_prompts() -> None:  # noqa:D103
```
wenjin272: Could we merge `test_mcp_without_prompts` and `test_mcp`? I think the main logic is same, the difference can be apply according to an environment variable. You can see `test_chat_model_integration` as a reference.

wenjin272 (commented):
Hi, @qingfei1994. Overall looks good to me. I left a minor comment.

python/flink_agents/e2e_tests/e2e_tests_integration/e2e_tests_mcp/mcp_test.py
qingfei1994: I've merge test_mcp_without_prompts and test_mcp, and also I found that mcp_server.py doesn't provide a explicit prompt for mcp to call the tool which may eventually output empty result, so I also update the prompt ask_sum provided by mcp_server.py

PR #452, opened 2026-01-19
[Feature][runtime] Support the use of Python MCP in Java

GreatEugenius:
Linked issue: #361 

### Purpose of change
Support the use of Python MCP in Java.

### Tests
E2E test.

### API
No

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`


api/src/main/java/org/apache/flink/agents/api/tools/Tool.java
```diff
@@ -29,12 +29,16 @@
  */
 public abstract class Tool extends SerializableResource {
 
-    protected final ToolMetadata metadata;
+    protected ToolMetadata metadata;
```
xintongsong: Should not make this non-final.

api/src/main/java/org/apache/flink/agents/api/annotation/MCPServerPython.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Annotation to mark a field as a chat model resource that should be managed by the agent plan.
+ *
+ * <p>Fields annotated with @ChatModel will be scanned during agent plan creation and corresponding
+ * resource providers will be created to manage the chat model instances.
+ */
+@Target({ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface MCPServerPython {}
```
xintongsong: Why do we need this dedicated annotation?

e2e-test/flink-agents-end-to-end-tests-resource-cross-language/src/test/java/org/apache/flink/agents/resource/test/MCPCrossLanguageAgent.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class MCPCrossLanguageAgent extends Agent {
+    @MCPServerPython
+    public static ResourceDescriptor pythonMCPServer() {
+        return ResourceDescriptor.Builder.newBuilder(
+                        org.apache.flink.agents.plan.resource.python.PythonMCPServer.class
+                                .getName())
+                .addInitialArgument("endpoint", "http://127.0.0.1:8000/mcp")
+                .build();
+    }
```
xintongsong: I think the major complexity of this PR comes from that we require returning a `ResourceDescriptor` for python mcp servers, while for java mcp servers we require returning a `MCPServer` object.

I think it would be better to migrate the java mcp server to also use resource descriptor. This brings several benefits: user codes won't depend on the integration module, and we can automatically fallback to the python mcp server for java api users with lower jdk versions.

api/src/main/java/org/apache/flink/agents/api/annotation/MCPServer.java
```diff
@@ -58,11 +58,15 @@
...
 @Retention(RetentionPolicy.RUNTIME)
-public @interface MCPServer {}
+public @interface MCPServer {
+    /**
+     * The version of MCPServer to be used. Default is "python" for JDK 16 and below, and "java" for
+     * JDK 17+.
+     *
+     * @return the version value
+     */
+    String version() default "java";
```
xintongsong:
```suggestion
    String lang() default "java";
```

api/src/main/java/org/apache/flink/agents/api/resource/Constant.java
```diff
@@ -81,5 +81,5 @@ public class Constant {
             "org.apache.flink.agents.integrations.vectorstores.elasticsearch.ElasticsearchVectorStore";
 
     // MCP
-    public static String MCP_SERVER = "org.apache.flink.agents.integrations.mcp.MCPServer";
+    public static String MCP_SERVER = "unified.class.path.MCPServer";
```
xintongsong:
```suggestion
    public static String MCP_SERVER = "DECIDE_IN_RUNTIME_MCPServer";
```

api/src/main/java/org/apache/flink/agents/api/annotation/MCPServer.java
xintongsong: The default value should be `null` or something like `auto`, so that we can differentiate whether it's explicitly configured by users, or automatically decided by the framework. If explicitly configured `java` and the jdk version doesn't meet the requirements, we should fail.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -448,18 +500,18 @@ private void extractResourceProvidersFromAgent(Agent agent) throws Exception {
...
-                    }
-                } catch (ClassNotFoundException e) {
-                    // MCP annotation not available (Java 11 build), skip MCP processing
+            } else if (method.isAnnotationPresent(MCPServer.class)) {
+                // Check the MCPServer annotation version to determine which version to use.
+                MCPServer MCPServerAnnotation = method.getAnnotation(MCPServer.class);
+                String version = MCPServerAnnotation.version();
+                int javaVersion = Runtime.version().feature();
+                if (version.equalsIgnoreCase("java") && javaVersion >= 17) {
+                    extractMCPServer(method);
```
xintongsong: `extractMCPServer` is confusing. I'd suggest the name `extractJavaMCPServer.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -304,6 +344,13 @@ private void extractResource(ResourceType type, Method method) throws Exception
         String name = method.getName();
         ResourceProvider provider;
         ResourceDescriptor descriptor = (ResourceDescriptor) method.invoke(null);
+        if (type == MCP_SERVER) {
+            descriptor =
+                    new ResourceDescriptor(
+                            descriptor.getModule(),
+                            PythonMCPServer.class.getName(),
```
xintongsong: It's against intuition that `extractResource` would assume using python mcp server. I think we can introduce a descriptor decorator for this method.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -133,8 +141,40 @@ public AgentPlan(Agent agent, AgentConfiguration config) throws Exception {
...
+
+                    List<PythonMCPPrompt> pythonMCPPrompts = server.listPrompts();
+                    for (PythonMCPPrompt pythonMCPPrompt : pythonMCPPrompts) {
+                        resourceCache
+                                .computeIfAbsent(PROMPT, k -> new ConcurrentHashMap<>())
+                                .put(pythonMCPPrompt.getName(), pythonMCPPrompt);
+                    }
+                }
+            }
+        }
```
xintongsong, line 148: Minor: unnecessary deep indention

PR #401, opened 2025-12-28
Add sensory memory documentation

Shekharrajak:
Linked issue: #350

### Purpose of change

This PR adds comprehensive documentation for the Sensory Memory feature in Flink Agents. The documentation covers:

- Overview and key characteristics of Sensory Memory
- When to use Sensory Memory vs other memory types
- Access methods and API usage (Python and Java)
- Practical usage examples
- Auto-cleanup behavior and lifecycle
- Comparison with Short-Term and Long-Term Memory
- Best practices and anti-patterns

The documentation follows the existing documentation style and provides clear guidance for developers using Sensory Memory in their agents.

### Tests

Documentation changes do not require code tests. The documentation was verified by:
- Reviewing existing documentation style and format
- Ensuring code examples match the actual API implementation
- Validating that all referenced concepts align with the codebase

### API

This change does not touch any public APIs. It is documentation-only.

### Documentation

- [x] `doc-needed` 

- [ ] `doc-not-needed`

Issue #351, opened 2025-12-01
[doc] Add documents for long-term memory


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Add correspond documents for long-term memory

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

Issue #350, opened 2025-12-01
[doc] Add documents for sensory memory


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Add correspond documents for sensory memory

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #466, opened 2026-01-22
[doc] Add documents for memory of flink-agents.

wenjin272:
Linked issue: #350 and #351

### Purpose of change

Refactor the doc organization for flink-agents memory, and add docs for sensory and long-term memory.  

This pr is based on #401 and #402.

### Tests

No

### API

No

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`


docs/content/docs/development/memory/_index.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Memory
+bold: true
+bookCollapseSection: true
+weight: 9
```
xintongsong: I'd suggest to move this to between Vector Stores and MCP

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,85 @@
...
+under the License.
+-->
+
+## Introduction
+
+Memory has emerged, and will continue to remain, a core capability of foundation model-based agents.
+It underpins long-horizon reasoning, continual adaptation, and effective interaction with complex
+environments.<sup>[1](https://arxiv.org/pdf/2512.13564)</sup>
+
+To address the critical limitations of current AI agents—lack of persistent memory, poor long-term relationship building—we propose a distributed memory system for flink-agents that mirrors human cognitive processes.
```
xintongsong: Let's just describe what ai agents need. We'd better avoid criticizing "current AI agents", as it can easily become outdated and not true.

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,85 @@
...
+  * In flink agents, inputs of the agent are partitioned by their keys. This corresponds to how data are partitioned by keys in Flink's Keyed DataStream.
+* **retention**: The lifecycle of the memory, i.e., when it will be cleaned up, Single-Run or Multiple-Runs.
+  * An agent run refers to a complete execution of an agent to process an input event. Each record from upstream will trigger a new agent run.
+* **accuracy**: The completeness and accuracy of information stored in memory, i.e., whether uses can retrieve the complete original information.
+
+|                            | **Single-Key**    | **Cross-Key** |
+|----------------------------|-------------------|---------------|
+| **Single Run**             | Sensory Memory    | Unsupported   |
+| **Multiple Runs(precise)** | Short-Term Memory | Unsupported   |
+| **Multiple Runs(vague)**   | Long-Term Memory  | Knowledge     |
```
xintongsong: 1. We use "precise" and "vague" in the table, while these two words are never mentioned / explained in the above "accuracy" part.
2. It's unclear from the table whether Sensory Memory is precise or vague.

I'd suggest to use a 2D coordinate graph, with different colors suggesting the 3rd dimension.

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Sensory Memory
+* **Use Case**: Store temporary data generated during the agent execution which is only needed for a single run. For example, tool call context, or the data users want to pass to other Agent actions.
+* **Characteristics**: Automatically cleaned up by the framework after an agent run finished.
+
```
xintongsong, line 52: Better to link to the sensory memory page.

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,85 @@
...
+### Short-Term Memory
+* **Use Case**: Store data generated during the agent execution, but
+compared to Sensory Memory, the lifecycle of stored data can across multiple Runs, compared to Long-Term Memory, the stored data is small, can be stored in Flink State, and requires precise information retrieval.
+* **Characteristics**: The lifecycle of the stored data can across multiple Runs. Small and precise information retrieval.
+
+### Long-Term Memory
+* **Use Case**: Store data generated during the agent execution, but compared to Short-Term Memory
+the stored data may expand rapidly as the agent execution, which requires compaction to control the storage usage.
+When retrieving data from a large amount of memory in Long-Term Memory, it may not be possible to recall complete data, and the data itself may be compacted.
```
xintongsong: I think we should present LTM not only as managing the storage space, but also to provide concise and highly related context. In terms of that, we need to support similarity search and compact similar items in advance.

In that sense, we might want to present information loss as derived (vs. original) information, which might be exactly same as how it was written into the memory.

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,85 @@
...
+* **Characteristics**: The ownership is independent of the agent and flink job, so the Flink-Agents framework will not manage the data stored in it; The stored data is visible to any flink job and agent.
+
+## Comparison
+
+| Feature        | Sensory Memory         | Short-Term Memory      | Long-Term Memory               | Knowledge                          |
+|----------------|------------------------|------------------------|--------------------------------|------------------------------------|
+| **Retention**  | Single Run             | Multiple Runs          | Multiple Runs                  | Multiple Runs                      |
+| **Visibility** | Single-Key             | Single-Key             | Single-Key                     | Cross-Key                          |
+| **Accuracy**   | Precise                | Precise                | Vague (May be Compacted)       | Vague (May be recall incompletely) |
+| **Storage**    | Flink MapState         | Flink MapState         | Vector Store                   | Vector Store                       |
```
xintongsong:
```suggestion
| **Storage**    | Flink State         | Flink State         | External Vector Store                 | External Vector Store                       |
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,85 @@
...
+| **Search**     | No                     | No                     | Semantic Search                | Sematic Search                     |
```
xintongsong:
```suggestion
| **Access Pattern**     | Key-Value / Nested Objects                    | Key-Value / Nested Objects                     | Semantic Search                | Sematic Search                     |
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,85 @@
...
+| **Capacity**   | Limited by Flink state | Limited by Flink state | Configurable (with compaction) | Unlimited                          |
```
xintongsong: There's no explicit limit from Flink state. Let's remove this line for now.

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,85 @@
...
+| **Use Case**   | Temporary data         | Small persistent data  | Large searchable data          | Shareable Knowledge                |
```
xintongsong: Let's avoid discussing small vs. large, but focus on different access pattern.

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,85 @@
...
+| **Item Types** | Any primitive/object   | Any primitive/object   | String, ChatMessage            | Unsupported                        |
+
+{{< hint info >}}
+Knowledge has not been implemented yet.
```
xintongsong: The shall we simply don't expose this at the moment?

docs/content/docs/development/memory/sensory_memory.md
```diff
@@ -0,0 +1,211 @@
...
+specific language governing permissions and limitations
+under the License.
+-->
+
+## Overview
+
+Sensory Memory is a temporary storage mechanism in Flink Agents designed for data that only needs to persist during a single agent run. It provides a convenient way to store intermediate results, tool call contexts, and other temporary data without the overhead of persistence across multiple runs.
+
+{{< hint info >}}
+Sensory Memory is automatically cleared after each agent run completes, making it ideal for temporary data that doesn't need to persist across runs.
```
xintongsong: Auto-clean also means isolation between agent runs, which is convenient.

docs/content/docs/development/memory/sensory_memory.md
```diff
@@ -0,0 +1,211 @@
...
+{{< /hint >}}
+
+### Key Characteristics
+
+- **Single Key Visible**: Based on Flink MapState, and is visible within the same keyed partition.
+- **Single-Run Lifecycle**: Data is automatically cleared when an agent run finishes.
```
xintongsong: This feels a redundant.

docs/content/docs/development/memory/sensory_memory.md
```diff
@@ -0,0 +1,211 @@
...
+
+## When to Use Sensory Memory
+
+Sensory Memory is ideal for:
+
+- **Intermediate Results**: Temporary computation results that don't need persistence.
+- **Action-to-Action Communication**: Passing data between actions within a single run.
+- **Temporary State**: Any data that should be discarded after the run completes.
```
xintongsong: - Intermediate results and temporary information, which will be used later by other actions in the same agent run.
- Passing data through multiple actions, reduce unnecessary data copy and serialization.

docs/content/docs/development/memory/sensory_memory.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Data Types & Operations
+
+Sensory Memory supports a hierarchical key-value structure:
+
+### Supported Value Types
+
+- **Primitive Types**: `str`, `int`, `float`, `bool`, `None`
+- **Collections**: `list`, `dict` (maps)
+- **General Class Types**: API and custom
```
xintongsong: What does this mean?

docs/content/docs/development/memory/sensory_memory.md
```diff
@@ -0,0 +1,211 @@
...
+    MemoryObject sensoryMemory = ctx.getSensoryMemory();
+    // write value to sensory memory
+    sensoryMemory.set("key", 123);
+    // read value from sensory memory
+    int value = (int) sensoryMemory.get("key").getValue();
+}
+```
+{{< /tab >}}
+
+{{< /tabs >}}
```
xintongsong: This is example of only primitive types.

docs/content/docs/development/memory/sensory_memory.md
```diff
@@ -0,0 +1,211 @@
...
+
+#### Nested Object
+
+Fields fo nested object can be read and wrote using dot-separated paths:
```
xintongsong: We need to first explain what is a MemoryObject, and then the two ways of accessing them (as an object, or as map values keyed by nested path).

docs/content/docs/development/memory/sensory_memory.md
```diff
@@ -0,0 +1,211 @@
...
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+### Memory Reference
+
+The `set` method of sensory term memory will return a `MemoryRef`, which can be treated as a reference to the stored object.
+
+If user want to pass the stored object between actions, they can pass the reference instead, which can reduce the payload size of Event.
```
xintongsong: 1. What is a MemoryRef - a reference of objects stored in memory
2. It is useful for - passing data across actions via memory
3. We record always use MemoryRef in events, rather than original data. Advantages: event size (affects orchestration, observability, fault tolerance), passing through multiple actions (redundant data copy & SerDe)

docs/content/docs/development/memory/sensory_memory.md
```diff
@@ -0,0 +1,211 @@
...
+
+{{< /tabs >}}
+
+## Auto-Cleanup Behavior
+
+Sensory Memory is automatically cleared by the framework after each agent run completes. This cleanup happens:
+
+- **When**: After the agent run finishes processing all events trigger by one input event.
+- **What**: All data stored in sensory memory is cleared.
+- **Why**: Ensures no temporary data persists unnecessarily.
```
xintongsong: Why: isolation across agent runs

docs/content/docs/development/memory/short_term_memory.md
```diff
@@ -0,0 +1,47 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+## Overview
```
xintongsong: I think we should combine sensory & short-term into one page.

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,515 @@
...
+- **Conversation History**: Maintaining long conversation histories with semantic search.
+- **Context Retrieval**: Finding relevant context from past interactions.
+
+{{< hint warning >}}
+Long-Term Memory is designed for large-scale data. For small, precise data that needs fast access, consider using [Short-Term Memory]({{< ref "docs/development/memory/short_term_memory" >}}) instead.
+{{< /hint >}}
+
+## Data Structure
+
+### Memory Item
```
xintongsong, line 47: What is a memory item? A piece of text, a chat message, a java or python object / semi-structured document / image / audio / video (in future)

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,515 @@
...
+The item stored in the long-term memory will be represented by `MemorySetItem`, which contains:
+
+* **memory_set_name**: The name of the memory set this item belong to.
+* **id**: The unique identifier of the memory item.
+* **value**: The value of the memory item.
+* **compacted**: Whether this item has been compacted.
+* **created_time**: Timestamp or timestamp range for when this memory item was created.
+* **last_accessed_time**: Timestamp for the last time this memory item was accessed.
+
+### Memory Set
```
xintongsong, line 63: What is a memory set? A set of memory items, that can be maintained and searched separately.

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,515 @@
...
+Currently, Flink Agents only supports Summarization Strategy.
+{{< /hint >}}
+
+### Summarization Strategy
+
+The summarization strategy uses an LLM to summarize multiple items into fewer items:
+
+- **Trigger**: Automatically triggered when `memory_set.size >= memory_set.capacity`
+- **Process**: Groups items into topics and generates summaries
+- **Result**: Original items are deleted and replaced with summarized versions
```
xintongsong: These are internal and too detail. We should just say that the summarization strategy reduces the capacity by summarizing and combining similar items.

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,515 @@
...
+
+{{< /tabs >}}
+
+## Compaction
+
+Compaction is the process of reducing memory set size when capacity is reached.
+
+{{< hint info >}}
+Currently, Flink Agents only supports Summarization Strategy.
+{{< /hint >}}
```
xintongsong: Then can we not expose strategy to users and always use summary for now?

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,515 @@
...
+### Async Compaction
+
+Compaction can be executed asynchronously to avoid blocking operations:
+
+{{< tabs "Async Compaction" >}}
+
+{{< tab "Python" >}}
+```python
+# Enable async compaction in configuration
+agents_config.set(LongTermMemoryOptions.ASYNC_COMPACTION, True)
```
xintongsong: Why not enabled by default?

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,515 @@
...
+When async compaction is enabled, compaction won't block user adding items to the memory set. The size of the memory set may exceed capacity temporarily.
+{{< /hint >}}
+
+## Limitations and Future Work
+
+### Current Limitations
+
+- **Item Types**: Currently supports only `str` and `ChatMessage`.
+- **Compaction Strategies**: Only summarization strategy is implemented.
+- **Recent Items**: `get_recent(n)` method not yet implemented. User can use short-term memory to store the recent items instead.
```
xintongsong: not a limitation

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,515 @@
...
+{{< hint info >}}
+When async compaction is enabled, compaction won't block user adding items to the memory set. The size of the memory set may exceed capacity temporarily.
+{{< /hint >}}
+
+## Limitations and Future Work
+
+### Current Limitations
+
+- **Item Types**: Currently supports only `str` and `ChatMessage`.
+- **Compaction Strategies**: Only summarization strategy is implemented.
```
xintongsong: not a limitation if we don't expose the strategies

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,515 @@
...
+- **Recent Items**: `get_recent(n)` method not yet implemented. User can use short-term memory to store the recent items instead.
+
+### Future Enhancements
+
+- **User-Defined Strategies**: Custom compaction strategies.
+- **Extended Item Types**: Support for any JSON-serializable objects.
+- **Extended Backend**: Support other long-term memory backend, e.g. [Mem0](https://github.com/mem0ai/mem0), [ReMe](https://github.com/agentscope-ai/ReMe).
```
xintongsong: This might be too uncertain to be mentioned in user docs.

api/src/main/java/org/apache/flink/agents/api/memory/compaction/CompactionSetup.java
```diff
@@ -25,7 +25,7 @@
 
 import java.util.Objects;
 
-public class SummarizationStrategy implements CompactionStrategy {
+public class CompactionSetup {
```
xintongsong: I'd suggest to name this `CompactionConfig`.

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+
+Memory is a system that remembers information about previous interactions. Memory has emerged, and will continue to remain, a core capability of foundation model-based agents.
+It underpins long-horizon reasoning, continual adaptation, and effective interaction with complex
+environments.<sup>[1](https://arxiv.org/pdf/2512.13564)</sup>
+
+We propose a distributed memory system for flink-agents that mirrors human cognitive processes. The solution implements four memory types: Sensory Memory, Short-term Memory, Long-term Memory and Knowledge, based on their visibility, retention and data origin.
+
+## Memory Types
+
+Flink-Agents classifies memory types based on visibility, retention and information accuracy.
```
xintongsong:
```suggestion
Flink Agents classifies memory types based on visibility, retention and derivation.
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+
+* **visibility**: The recall scope of the memory, i.e., whether it is visible across keys.
+  * In flink agents, inputs of the agent are partitioned by their keys. This corresponds to how data are partitioned by keys in Flink's Keyed DataStream.
```
xintongsong:
```suggestion
  * In Flink Agents, inputs of the agent are partitioned by their keys. This corresponds to how data are partitioned by keys in Flink's Keyed DataStream.
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+* **retention**: The lifecycle of the memory, i.e., when it will be cleaned up, Single-Run or Multiple-Runs.
```
xintongsong:
```suggestion
* **retention**: How long the memory will be remembered, within a single run or across multiple runs.
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+  * An agent run refers to a complete execution of an agent to process an input event. Each record from upstream will trigger a new agent run.
+* **data origin**: Whether uses can retrieve the complete original data from the memory.
```
xintongsong:
```suggestion
* **derivation**: Whether the retrieved information can be a derived version of the original one.
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+  * original: User can retrieve all the original data.
```
xintongsong:
```suggestion
  * original: The retrieved information is exactly the same as how it was stored.
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+  * derived: The data retrieved may be derived from original data.
```
xintongsong:
```suggestion
  * derived: The retrieved information might be derived from the original information.
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+
+{{< img src="/fig/memory/memory_types.png" alt="Memory Types" >}}
```
xintongsong, line 45: Let's also update the image.

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+
+### [Sensory Memory]({{< ref "docs/development/memory/sensory_and_short_term_memory" >}})
+* **Use Case**: Store temporary data generated during the agent execution which is only needed for a single run. For example, tool call context, or the data users want to pass to other agent actions.
+* **Characteristics**: Isolation between agent runs, which means will be automatically cleaned up by the framework after an agent run finished.
```
xintongsong:
```suggestion
* **Characteristics**: Isolation between agent runs, which means the memory will be automatically cleaned up by the framework when an agent run is completed.
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+
+### [Short-Term Memory]({{< ref "docs/development/memory/sensory_and_short_term_memory" >}})
```
xintongsong: The link in the title can easily be ignored. Let's add a sentence in the section like "for more information see ...".

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+  * Support for compacting similar items.
+  * Support for semantic search.
+
+## Comparison
+
+| Feature            | Sensory Memory             | Short-Term Memory                | Long-Term Memory                             |
+|--------------------|----------------------------|----------------------------------|----------------------------------------------|
+| **Retention**      | Single Run                 | Multiple Runs                    | Multiple Runs                                |
+| **Visibility**     | Single-Key                 | Single-Key                       | Single-Key                                   |
+| **Accuracy**       | Precise                    | Precise                          | Vague (May be Compacted)                     |
```
xintongsong:
```suggestion
| ** Derivation**       | Original                    | Original                          | Derived                     |
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+| **Storage**        | Flink State                | Flink State                      | External Vector Store                        |
+| **Access Pattern** | Key-Value / Nested Objects | Key-Value / Nested Objects       | Semantic Search                              |
+| **Use Case**       | Temporary data             | Complete original data retrieval | Concise and highly related context retrieval |
```
xintongsong:
```suggestion
| **Use Case**       | Temporary execution states of an agent run             | Exact information shared across multiple runs | Large, compactable and searchable information |
```

docs/content/docs/development/memory/overview.md
```diff
@@ -0,0 +1,76 @@
...
+| **Item Types**     | Any primitive/object       | Any primitive/object             | String, ChatMessage                          |
```
xintongsong:
```suggestion
| **Types of Contents**     | Any primitive/object       | Any primitive/object             | String, ChatMessage                          |
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+```python
+memory_object.set("key1", value1)
+memory_object.set("key2", value2)
+```
+{{< /tab >}}
+
+{{< tab "Java" >}}
+```java
+memoryObject.set("key1", value1)
+memoryObject.set("key2", value2)
```
xintongsong: This seems redundant, as we have a section to discuss how to read and write fields of a memory object.

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+def process_event(event: InputEvent, ctx: RunnerContext) -> None:
+    memory: MemoryObject = ctx.sensory_memory # or ctx.short_term_memory
+    # write values to memory
+    memory.set("primitive",  123)
+    memory.set("collection", [1, 2, 3])
+    memory.set("object", ChatMessage(role=MessageRole.USER, content="test"))
+    # read values from memory
+    value1: int = memory.get("primitive")
+    value2: List[int] = memory.get("collection")
+    value3: ChatMessage = memory.get("object")
```
xintongsong: The example doesn't seem to cover all supported types as mentioned above.

- Primitives and collections are covered.
- What is a ChatMessage? Does it fall into the general class types category?
- There's no example of pojo.
- There's no example of memory object. (I know it's covered in the next section, but we'd better have a simple line here to avoid confusion)
- Better use comments to explain which line(s) corresponds to which type.

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+}
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+#### Nested Object
+
+Nested objects can be read and write in two ways:
+* Use `new_object` to manually create nested field.
```
xintongsong: We need to emphasize that all memory objects must be created via another memory object with `new_object`. Because this might be not that straightforward for users.

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+* Use dot-separated path as key to crated nested field.
```
xintongsong:
```suggestion
* Use dot-separated path as key to created nested field.
```

xintongsong: This is not necessarily true. There are two ways to access (get & set) fields of s nested object. Users can choose whichever they like. We don't recommend one over another.

1.
obj_a = sensory_mem.new_object("a") 
obj_b = obj_a.new_object("b")
obj_b.set("value", 123)

2.
sensory_mem.set("a.b.value", 123)

xintongsong: I'd suggest to organize the structure of the Read & Write section as follows.

```
# Read & Write
memory.set/get("primitive")
memory.set/get("collection")
memory.set/get("pojo")
memory.set/get("general class")
memory.new/get_object("mem_obj")

## Access nested object fields quickly with path
memory.new/get_object("a.b")
memory.set/get("a.b")
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+name = (String) sensoryMemory.get("user.name").getValue();
+age = (int) sensoryMemory.get("user.age").getValue();
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+### Memory Reference
+
+`MemoryRef` is a reference of the objects stored in memory. The `set` method of sensory or short-term memory will return a `MemoryRef`.
```
xintongsong:
```suggestion
`MemoryRef` is a reference of the objects stored in memory. The `set` method of memory object will return a `MemoryRef`.
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+
+#### When to use
+`MemoryRef` is useful for passing data across multiple actions via memory. We recommend user to always use `MemoryRef` in events rather than original data. It can bring follow benefit:
+* **Reduce the event payload size**: The size of `MemoryRef` is usually typically smaller than that of the original data. For event will be used not only for orchestration, but also for observability and fault tolerance. The reduction of event size can help reduce some cost.
```
xintongsong:
```suggestion
* **Reduce the event payload size**: The size of `MemoryRef` is usually typically smaller than that of the original data. Events are widely used in Flink Agents - action orchestration, observability and fault tolerance, etc. Therefore, reducing the event size can help avoid unnecessary overheads.
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+{{< /tab >}}
+
+{{< /tabs >}}
+
+### Memory Reference
+
+`MemoryRef` is a reference of the objects stored in memory. The `set` method of sensory or short-term memory will return a `MemoryRef`.
+
+#### When to use
+`MemoryRef` is useful for passing data across multiple actions via memory. We recommend user to always use `MemoryRef` in events rather than original data. It can bring follow benefit:
```
xintongsong:
```suggestion
`MemoryRef` is useful for passing data across multiple actions via memory. We recommend user to use `MemoryRef` for large data in events rather than original data. It can bring follow benefit:
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+* **Reduce the event payload size**: The size of `MemoryRef` is usually typically smaller than that of the original data. For event will be used not only for orchestration, but also for observability and fault tolerance. The reduction of event size can help reduce some cost.
+* **Redundant data copy & SerDe**: To argue this point, we first need to clarify what copy and SerDe occur when using original data versus memory reference, respectively.
```
xintongsong:
```suggestion
* **Avoid unnecessary data copy & SerDe**: When traveling through actions, only the reference of data needs to be copied and serialized.
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+  * Use `MemoryRef`:
+    * Write/read original data to/from memory will cause kryo SerDe.
+  * Use original data
+    * For python actions, write/read data to/from event will cause pickle SerDe.
+    * For java actions, write/read to/from event won't cause any SerDe.
+    * Json SerDe for event log and per-action state.
+  
+    Thus, using `MemoryRef` can avoid the data copy and SerDe for event log and per-action state.
```
xintongsong: These are too detailed.

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,262 @@
...
+    ...
+
+@action(MyEvent)
+@staticmethod
+def second_action(event: Event, ctx: RunnerContext):
+    ...
+    sensory_memory = ctx.sensory_memory
+    
+    content_ref: MemoryRef = event.value
+    processed_data: ProcessedData = sensory_memory.get(content_ref)
```
xintongsong: Two ways to get the data.
```suggestion
    processed_data: ProcessedData = ctx.sensory_memory.get(event.value)
    processed_data: ProcessedData = event.value.resolve(ctx)
```

xintongsong: The original design was that, our built-in actions should support passing memory ref instead of values. E.g., in `ChatRequestEvent`, users can pass in either a list of chat messages or a memory ref for the argument `messages`, and if a ref is passed in the framework should automatically get the actual data from the memory and use if for the argument.

It seems Ke Gu has not completed this before her internship ends. And I also forgot about this.

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,483 @@
...
+);
+
+// Enable async compaction 
+agentsConfig.set(LongTermMemoryOptions.ASYNC_COMPACTION, true);
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+### Prerequisites
```
xintongsong: Let's move this to before Configuration

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,483 @@
...
+    new CompactionSetup("my_chat_model", 1)
+);
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+### Async Compaction
+
+Compaction can be executed asynchronously to avoid blocking operations:
```
xintongsong: Compactions are by default asynchronously performed, to avoid blocking the agent execution. You can also explicitly disable this, so that the agent execution will be paused during the compaction.

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,483 @@
...
+
+{{< tab "Python" >}}
+```python
+# Enable async compaction in configuration
+# Default is True, we only show how to configure it here.
+agents_config.set(LongTermMemoryOptions.ASYNC_COMPACTION, True)
+
+# When enabled, compaction runs in background
+# Operations continue without waiting for compaction to complete
+memory_set.add("New item")  # If capacity reached, compaction runs async
```
xintongsong:
```suggestion
# Explicitly disable async compaction in configuration
agents_config.set(LongTermMemoryOptions.ASYNC_COMPACTION, False)
```

xintongsong: Let's not expose when exactly the compaction is triggered (reaching the capacity), and how it blocks the operations. They are internal and too detailed.

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,483 @@
...
+
+{{< tab "Java" >}}
+```java
+import org.apache.flink.agents.api.memory.long_term_memory.SummarizationStrategy;
+
+// Create memory set with summarization strategy
+MemorySet memorySet = ltm.getOrCreateMemorySet(
+    "conversations",
+    String.class,
+    10,  // Compaction triggers when size >= 10
```
xintongsong:
```suggestion
    10,  // The framework will automatically trigger compactions and try to maintain the size of the memory set not exceeding the given capacity with best efforts
```

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -0,0 +1,483 @@
...
+    compaction_setup=CompactionSetup(model="my_chat_model", limit=1)
+)
+```
+{{< /tab >}}
+
+{{< tab "Java" >}}
+```java
+import org.apache.flink.agents.api.memory.long_term_memory.SummarizationStrategy;
+
+// Create memory set with summarization strategy
```
xintongsong:
```suggestion
// Create memory set with compaction configs
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
wenjin272: Ok, I will create a follow issue for this.

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+```python
+@action(InputEvent)
+def process_event(event: InputEvent, ctx: RunnerContext) -> None:
+    memory: MemoryObject = ctx.sensory_memory # or ctx.short_term_memory
+    # store primitive
+    memory.set("primitive",  123)
+    # store collection
+    memory.set("collection", [1, 2, 3])
+    # store general class types
+    memory.set("object", LocalPrompt("the test {content}"))
```
xintongsong:
```suggestion
    memory.set("object", Prompt.from_text("the test {content}"))
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+    # store memory object
+    memory.new_object("memory_object")
```
xintongsong:
```suggestion
    obj1: MemoryObject = memory.new_object("obj1")
    obj1.set("foo", "bar")
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+    
+    # read values from memory
+    value1: int = memory.get("primitive")
+    value2: List[int] = memory.get("collection")
+    value3: Prompt = memory.get("object")
+    value4: MemoryObject = memory.get("memory_object")
```
xintongsong:
```suggestion
    value4: MemoryObject = memory.get("obj1")
    value5: str = value4.get("foo")
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+    // store primitive
+    memory.set("primitive", 123);
+    // store collection
+    memory.set("collection", List.of(1, 2, 3));
+    // store pojo
+    memory.set("pojo", new WordWithCount("hello", "1"))
+    // store general class types 
+    memory.set("object", new LocalPrompt("the test {content}"));
+    // store memory object
+    memory.newObject("memory_object", true); // The second argument means whether to overwrite existing field if it's not a nested object
```
xintongsong: I think we should add an overload method that does not require the second argument and use `false` for default (aligned with python), and use that for the example.

xintongsong:
```suggestion
    MemoryObject obj1 = memory.newObject("obj1", true); // The second argument means whether to overwrite existing field if it's not a nested object 
    obj1.set("foo", "bar");
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+    memory.set("object", new LocalPrompt("the test {content}"));
+    // store memory object
+    memory.newObject("memory_object", true); // The second argument means whether to overwrite existing field if it's not a nested object 
+    
+    // read values from memory
+    int value1 = (int) memory.get("primitive").getValue();
+    List<Integer> value2 = (List<Integer>) memory.get("collection").getValue();
+    WordWithCount value3 = (WordWithCount) memory.get("pojo").getValue();
+    Prompt value4 = (Prompt) memory.get("object").getValue();
+    MemoryObject value5 = memory.get("memory_object");
```
xintongsong:
```suggestion
    MemoryObject value5 = memory.get("obj1");
    String value6 = value5.get("foo");
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+}
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+#### Nested Object
+
+There are two ways to access (get & set) fields of s nested object. Users can choose whichever they like. We don't recommend one over another:
```
xintongsong:
```suggestion
There are two ways to access (get & set) fields of s nested object. Users can choose whichever they like.
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+* Use `new_object` to manually create nested field.
+* Use dot-separated path as key to created nested field.
```
xintongsong: This doesn't seem right. "Always using `new_object` for creating memory objects" is something we need to emphasis, but it's not a way of accessing the fields.

xintongsong:
```suggestion
* Use `new_object` to manually create nested field.
* Use dot-separated path as key to access nested field.
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+
+{{< tabs "Nested Object Access" >}}
+
+{{< tab "Python" >}}
+```python
+# Use new_object
```
xintongsong:
```suggestion
# access fields from the innermost memory object with field names
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+user: MemoryObject = memory.new_object("user")
+user.set("name", "john")
+user.set("age", 13)
+
+user: MemoryObject = memory.get_object("user")
+name: str = user.get("name")
+age: int = user.get("age")
+
+# Use dot-separated path
```
xintongsong:
```suggestion
# access fields from an outer memory object with paths to the fields
```

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -0,0 +1,253 @@
...
+name = (String) sensoryMemory.get("user.name").getValue();
+age = (int) sensoryMemory.get("user.age").getValue();
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+### Memory Reference
+
+`MemoryRef` is a reference of the objects stored in memory. The `set` method memory object will return a `MemoryRef`.
```
xintongsong:
```suggestion
`MemoryRef` is a reference of the objects stored in memory. The `set` method of memory object will return a `MemoryRef`.
```

Issue #374, opened 2025-12-08
[Feature] [doc] Add document for built-in anthropic chat model support in java


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

This is a follow-up of #326 

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #468, opened 2026-01-23
[doc] Add java Anthropic chat model docs

xli1996:
### Purpose of change

https://github.com/apache/flink-agents/issues/374
- Add docs for java based Anthropic ChatModel
- Minor changes on the python based Anthropic ChatModel to consistent with the Openai one.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/chat_models.md
+++ b/docs/content/docs/development/chat_models.md
@@ -235,15 +235,16 @@ Model availability and specifications may change. Always check the official Azur
 
 Anthropic provides cloud-based chat models featuring the Claude family, known for their strong reasoning, coding, and safety capabilities.
 
-{{< hint info >}}
-Anthropic is only supported in Python currently. To use Anthropic from Java agents, see [Using Cross-Language Providers](#using-cross-language-providers).
-{{< /hint >}}
-
 #### Prerequisites
 
-1. Get an API key from [Anthropic Console](https://console.anthropic.com/)
+1. Create an account at [Anthropic Console](https://console.anthropic.com/)
+2. Navigate to [API Keys](https://console.anthropic.com/settings/keys) and create a new secret key
 
 #### AnthropicChatModelConnection Parameters
+
+{{< tabs "AnthropicChatModelConnection Parameters" >}}
+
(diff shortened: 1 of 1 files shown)


docs/content/docs/development/chat_models.md
```diff
@@ -235,24 +235,44 @@ Model availability and specifications may change. Always check the official Azur
...
-{{< hint info >}}
-Anthropic is only supported in Python currently. To use Anthropic from Java agents, see [Using Cross-Language Providers](#using-cross-language-providers).
-{{< /hint >}}
-
 #### Prerequisites
 
-1. Get an API key from [Anthropic Console](https://console.anthropic.com/)
+1. Create an account at [Anthropic Console](https://console.anthropic.com/)
+2. Navigate to [API Keys](https://console.anthropic.com/settings/keys) and create a new secret key
+3. Set the API key as an environment variable: `export ANTHROPIC_API_KEY=your-api-key`
```
xintongsong: This only works when running Flink locally, but will not work when running the job in a remote, distributed Flink cluster. This is same for OpenAI, sorry I just realized this.

There are several ways to securely passing passwords/keys into Flink, e.g., via an K8s secrets. I'd suggest not discuss this in detail here, and just use `<your-api-key>` in the doc.

docs/content/docs/development/chat_models.md
```diff
@@ -235,24 +235,44 @@ Model availability and specifications may change. Always check the official Azur
...
 #### AnthropicChatModelConnection Parameters
 
+{{< tabs "AnthropicChatModelConnection Parameters" >}}
+
+{{< tab "Python" >}}
+
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
-| `api_key` | str | Required | Anthropic API key for authentication |
+| `api_key` | str | `$ANTHROPIC_API_KEY` | Anthropic API key for authentication |
```
xintongsong: Le't keep using `Required`, indicating that there's no default value and users are required to explicitly set this.

PR #470, opened 2026-01-23
[doc] Update document for vector store.

wenjin272:
No

### Purpose of change

Update document for vector store.

### Tests

No

### API

No

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/vector_stores.md
+++ b/docs/content/docs/development/vector_stores.md
@@ -30,9 +30,9 @@ This page covers semantic search using vector stores. Additional query modes (ke
 
 ## Overview
 
-### Use Case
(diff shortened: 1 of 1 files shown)


docs/content/docs/development/vector_stores.md
```diff
@@ -470,14 +613,84 @@ class MyVectorStore(BaseVectorStore):
...
         return {"index": "my_index", ...}
-
+    
+    @override
+    def size(self, collection_name: str | None = None) -> int:
+        """Get the size of the collection in vector store.
+
+        Args:
+            collection_name: The target collection. If not provided, use defualt collection.
+        """
```
xintongsong, line 765: This code snippet is an example of `MyVectorStore` that implements `BaseVectorStore`. It's not an interface declaration. In that sense, implementation should be provided for all these methods. 

The implementation can be a simple demonstration, like:
```
def size(self, ...):
    size = ...
    return size
```
or
```
def delete(self, ...):
    // delete the document
    pass
```

Issue #473, opened 2026-01-23
[Bug] Serialize DurableExecutionException object exception


choucmei:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Env: flink1.20.3 jdk17 
The following error occurred when executing WorkflowSingleAgentExample:
`rg.apache.flink.client.program.ProgramInvocationException: The main method caused an error: org.apache.flink.client.program.ProgramInvocationException: Job failed (JobID: c8f07189a69282c095d1c03ef9cd57f2)
	at org.apache.flink.client.program.PackagedProgram.callMainMethod(PackagedProgram.java:373)
	at org.apache.flink.client.program.PackagedProgram.invokeInteractiveModeForExecution(PackagedProgram.java:223)
	at org.apache.flink.client.ClientUtils.executeProgram(ClientUtils.java:113)
	at org.apache.flink.client.cli.CliFrontend.executeProgram(CliFrontend.java:1026)
	at org.apache.flink.client.cli.CliFrontend.run(CliFrontend.java:247)
	at org.apache.flink.client.cli.CliFrontend.parseAndRun(CliFrontend.java:1270)
	at org.apache.flink.client.cli.CliFrontend.lambda$mainInternal$10(CliFrontend.java:1367)
	at org.apache.flink.runtime.security.contexts.NoOpSecurityContext.runSecured(NoOpSecurityContext.java:28)
	at org.apache.flink.client.cli.CliFrontend.mainInternal(CliFrontend.java:1367)
	at org.apache.flink.client.cli.CliFrontend.main(CliFrontend.java:1335)
Caused by: java.util.concurrent.ExecutionException: org.apache.flink.client.program.ProgramInvocationException: Job failed (JobID: c8f07189a69282c095d1c03ef9cd57f2)
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
	at org.apache.flink.client.program.StreamContextEnvironment.getJobExecutionResult(StreamContextEnvironment.java:174)
	at org.apache.flink.client.program.StreamContextEnvironment.execute(StreamContextEnvironment.java:121)
	at org.apache.flink.streaming.api.environment.StreamExecutionEnvironment.execute(StreamExecutionEnvironment.java:2325)
[...]

PR #474, opened 2026-01-23
[fix] Fixed the DurableExecutionException error that could not be serialized

choucmei:
Linked issue: [#473](https://github.com/apache/flink-agents/issues/473)

### Purpose of change

Fixed the DurableExecutionException error that could not be serialized. 

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
@@ -17,6 +17,7 @@
  */
 package org.apache.flink.agents.runtime.context;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -359,7 +360,13 @@ protected static class DurableExecutionRuntimeException extends RuntimeException
 
     /** Serializable exception info for durable execution persistence. */
     public static class DurableExecutionException {
+        private static final String FIELD_MESSAGE = "message";
+        private static final String FIELD_EXCEPTION_CLASS = "exceptionClass";
+
+        @JsonProperty(FIELD_MESSAGE)
```
Sxnan, line 366: It seems that the `@JsonProperty` values are swapped between the two fields. `exceptionClass` field should map to "exceptionClass" in JSON, and `message` should map to "message".

Issue #477, opened 2026-01-26
[Feature] [Integration][Python] Add built-in support for Azure OpenAI Chat Model


alnzng:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Add built-in support for Azure OpenAI Chat Model in Python SDK

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #478, opened 2026-01-26
[Feature] [Integration][Python] Add built-in support for Azure OpenAI Chat Model

alnzng:
Linked issue: #477

### Purpose of change

This PR introduces support for Azure OpenAI chat models in the Flink Agents Python API, enabling users to leverage Azure-hosted OpenAI LLMs in their agent applications. 

### Tests

Unit tests and E2E tests, verified them locally with a real Azure OpenAI API key.

### API

- Added `AzureOpenAIChatModelConnection` and `AzureOpenAIChatModelSetup` for Azure OpenAI chat model

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/python/flink_agents/integrations/chat_models/azure/azure_openai_chat_model.py
+++ b/python/flink_agents/integrations/chat_models/azure/azure_openai_chat_model.py
@@ -128,12 +128,12 @@ class AzureOpenAIChatModelConnection(BaseChatModelConnection):
         if tools is not None:
             tool_specs = [to_openai_tool(metadata=tool.metadata) for tool in tools]
 
-        # Extract azure_deployment and model from kwargs for special handling
-        azure_deployment = kwargs.pop("azure_deployment", "")
+        # Extract model (azure_deployment) and model_of_azure_deployment from kwargs
+        azure_deployment = kwargs.pop("model", "")
         if not azure_deployment:
-            msg = "azure_deployment is required for Azure OpenAI API calls"
+            msg = "model is required for Azure OpenAI API calls"
             raise ValueError(msg)
-        model_name = kwargs.pop("model", None)
+        model_of_azure_deployment = kwargs.pop("model_of_azure_deployment", None)
 
         response = self.client.chat.completions.create(
             # Azure OpenAI APIs use Azure deployment name as the model parameter
@@ -144,9 +144,9 @@ class AzureOpenAIChatModelConnection(BaseChatModelConnection):
         )
 
         extra_args = {}
-        # Record token metrics only if model name is provided
-        if model_name and response.usage:
-            extra_args["model_name"] = model_name
+        # Record token metrics only if model_of_azure_deployment is provided
+        if model_of_azure_deployment and response.usage:
+            extra_args["model_name"] = model_of_azure_deployment
             extra_args["promptTokens"] = response.usage.prompt_tokens
             extra_args["completionTokens"] = response.usage.completion_tokens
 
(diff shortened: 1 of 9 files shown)


python/flink_agents/integrations/chat_models/azure/azure_openai_chat_model.py
```diff
@@ -0,0 +1,260 @@
...
+        """Get Azure OpenAI client."""
+        self._client = AzureOpenAI(
+            azure_endpoint=self.azure_endpoint,
+            api_key=self.api_key,
+            api_version=self.api_version,
+            timeout=self.timeout,
+            max_retries=self.max_retries,
+        )
+        if self._client is None:
+            pass
```
wenjin272: Should we throw exception if client is not created as expected?

python/flink_agents/integrations/chat_models/azure/azure_openai_chat_model.py
```diff
@@ -0,0 +1,260 @@
...
+    ----------
+    connection : str
+        Name of the referenced connection. (Inherited from BaseChatModelSetup)
+    prompt : Optional[Union[Prompt, str]
+        Prompt template or string for the model. (Inherited from BaseChatModelSetup)
+    tools : Optional[List[str]]
+        List of available tools to use in the chat. (Inherited from BaseChatModelSetup)
+    azure_deployment : str
+        Name of Azure OpenAI deployment to use.
+    model : Optional[str]
```
wenjin272: Looks like the attribute `azure_deployment` in Azure is correspond to `model` in other model providers, which may    cause some confusion. Could we still use `model`? 

For token usage track, we can just also use the same attribute `model` or provide an optional config like `token_usage_identifier`.

wenjin272 (commented):
Hi, @alnzng. The pr overall looks good to me, I left two minor comments. Besides, due to I don't have the azure account, I don't verify the test.

python/flink_agents/integrations/chat_models/azure/azure_openai_chat_model.py
alnzng: Good catch, thanks! I should use below logic to initialize the client instance, let me update the codes:

```
if self._client is None:
  self._client = AzureOpenAI(
      azure_endpoint=self.azure_endpoint,
      api_key=self.api_key,
      api_version=self.api_version,
      timeout=self.timeout,
      max_retries=self.max_retries,
  )
```

python/flink_agents/integrations/chat_models/azure/azure_openai_chat_model.py
alnzng: Good question, and you raised a good point about consistency!                                                   
                                                                                                                  
The reason I used azure_deployment is that Azure OpenAI has a different concept: users deploy models to         
 `deployments` with custom names, and the API requires the deployment name, not the model name directly. Using `azure_deployment` accurately reflects Azure's terminology and avoids confusion for Azure users who are familiar with this concept.                                                                                              
                                                                                                                  
I totally understand the consideration for API consistency, and I am open to either approach. I can use `model` as the name to represent the Azure deployment name, but I won't prefer using it for tracking token usage / cost because:   
  - Azure deployment names are user-defined identifiers that don't indicate the underlying model                  
  - Users can recreate a deployment with the same name but a different model                                      
  - Token pricing/cost calculation requires knowing the actual model (e.g., gpt-4 vs gpt-35-turbo)                
                                                                                                                  
I would prefer to use `underlying_model` as the field name for tracking token usage / cost. WDYT?

xintongsong: We might consider using `deployment` as an alias of `model` here, means user can choose whichever he likes. In case of both keys are set and values are inconsistent, I don't really have a preferences on which should take effects, as long as it's well documented.

xintongsong: To be specific, my suggestion is:
- azure_deployment / model - corresponds to model in other providers
- model_of_deployement - for metrics counting

wenjin272: > To be specific, my suggestion is:
> 
> * azure_deployment / model - corresponds to model in other providers
> * model_of_deployement - for metrics counting

+1

alnzng: Sounds good @xintongsong @wenjin272 !

Let me keep `azure_deployment` and add a new field `model` as the alias of `azure_deployment`, use `model_of_deployement` for token usage tracing.

alnzng: I decided to have `model` and `model_of_deployment` fields only,  because I thought keep both `azure_deployment` and `model` for same purpose is confusing to the users.

xintongsong (approved):
@alnzng We also need to add the resource names on the java side, which I addressed with an extra commit. PTAL.

Other than that, LGTM.

PR #480, opened 2026-01-26
[doc] Refactor the doc for configuration.

wenjin272:
### Purpose of change

Refactor and update the doc for configuration

### Tests

No

### API

No

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/operations/configuration.md
+++ b/docs/content/docs/operations/configuration.md
@@ -24,79 +24,16 @@ under the License.
 
 ## How to configure Flink Agents
 
-There are three ways to configure Flink Agents, listed in order of **priority from high to low**:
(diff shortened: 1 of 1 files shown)


python/flink_agents/api/core_options.py
```diff
@@ -90,6 +90,10 @@ class AgentConfigOptions(metaclass=AgentConfigOptionsMeta):
         default=None,
     )
 
+
+class AgentExecutionOptions(metaclass=AgentConfigOptionsMeta):
```
GreatEugenius: The purpose of AgentConfigOptionsMeta is to read the ConfigOption directly from AgentExecutionOptions.java. If we have a ConfigOption from each side, we can remove AgentConfigOptionsMeta.

Issue #508, opened 2026-01-29
[Bug] [runtime] Python awaitable lost during checkpoint restore causes NoneType error


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

When a Python async action fails mid-execution (e.g., LLM API timeout), the job attempts to restart from the last checkpoint. During restore, the Python awaitable (coroutine) reference is `None`, causing:                                             
                                                                                        
 

>  AttributeError: 'NoneType' object has no attribute 'send'           

                  
                                                                                        
This happens because Python coroutines/generators **cannot be serialized**. The checkpoint saves a reference to the awaitable, but not the coroutine object itself. On restore, the reference exists but points to `None`.                                  
                                                         
### How to reproduce

 1. Run `react_agent_example.py` with Ollama:                                          
  ```bash                                                                               
  bin/flink run -py python/flink_agents/examples/quickstart/react_agent_example.py      
                                                                                        
  2. Wait for the job to process some records (checkpoints 1 & 2 complete successfully) 
  3. Trigger an LLM timeout (slow network, overloaded Ollama, etc.)                     
  4. Job fails with httpx.ReadTimeout, attempts to restart from checkpoint              
  5. Observe repeated NoneType errors during restore                                    
                                                                                        
  Error sequence                                                                        
  ┌──────────┬──────────────────────────────────────────────────────────────────┐       
  │   Time   │                              Event                               │       
  ├──────────┼──────────────────────────────────────────────────────────────────┤       
  │ 00:02:49 │ Checkpoint 1 completed ✅                                        │       
  ├──────────┼──────────────────────────────────────────────────────────────────┤       
  │ 00:03:49 │ Checkpoint 2 completed ✅ (9.28 MB state)                        │       
  ├──────────┼──────────────────────────────────────────────────────────────────┤       
  │ 00:04:38 │ Initial failure: httpx.ReadTimeout: timed out                    │       
  ├──────────┼──────────────────────────────────────────────────────────────────┤       
  │ 00:04:49 │ Checkpoint 3 failed - "Checkpoint Coordinator is suspending"     │       
  ├──────────┼──────────────────────────────────────────────────────────────────┤       
  │ 00:06:38 │ Job restarts from checkpoint 2                                   │       
  ├──────────┼──────────────────────────────────────────────────────────────────┤       
[...]

PR #509, opened 2026-01-29
[runtime] Fix Python awaitable lost during checkpoint restore

weiqingy:
Linked issue: #508 

### Purpose of change

  Fix a crash that occurs when restoring a Python async action from checkpoint.                        
                                                                                                       
  **Root cause:** Python coroutines cannot be serialized. When a checkpoint captures state while an async action is in progress (e.g., waiting for LLM response), the coroutine object is lost. On restore, the awaitable reference is `None`, causing:                                                 
  AttributeError: 'NoneType' object has no attribute 'send'                                            
                                                                                                       
  **Fix:**                                                                                             
  - Detect `None` awaitable in `PythonActionExecutor.callPythonAwaitable()`                            
  - Throw `AwaitableLostException` to signal the awaitable was lost                                    
  - `PythonGeneratorActionTask` catches this and re-executes the action from the beginning             
  - Durable execution cache ensures already-completed calls are skipped   

### Tests

  - Manual testing: Run `react_agent_example.py`, trigger LLM timeout, verify job recovers instead of crashing                                                                                             
  - No unit test added - these classes depend on Pemja (Python interpreter) which is difficult to mock; the fix is better validated via e2e testing    
### API

No public API changes. 
### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
@@ -203,6 +203,8 @@ public class ActionExecutionOperator<IN, OUT> extends AbstractStreamOperator<OUT
 
     private final transient Map<ActionTask, ContinuationContext> continuationContexts;
 
+    private final transient Map<ActionTask, String> pythonAwaitableRefs;
+
     // Each job can only have one identifier and this identifier must be consistent across restarts.
     // We cannot use job id as the identifier here because user may change job id by
     // creating a savepoint, stop the job and then resume from savepoint.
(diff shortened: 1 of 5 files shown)


weiqingy:
Hey @Sxnan, wondering if there’s a way to re-run the failed CI? It seems I don’t have access to trigger it. This looks like a pre-existing flaky test, not related to the PR.

weiqingy:
> @weiqingy , Thanks for identifying this bug and submitting the fix! The issue you found is real and important. After reviewing the PR, I made some adjustments to align the Python async handling with how we handle Java continuations. The current approach uses a transient map in the operator (similar to continuationContexts for Java), which makes the checkpoint restore detection more explicit and consistent across both languages. I've pushed a fixup commit with these changes. Could you please take a look and let me know what you think?

Thanks for the quick review and changes, @Sxnan! It LGTM.

xintongsong (approved):
LGTM

PR #510, opened 2026-01-31
[hotfix] Fix the problems using Kafka as action state store

Sxnan:
Linked issue: #xxx

This PR fixes issues with the Kafka ActionStateStore that prevented actions from being correctly skipped after checkpoint recovery:

1. **Fix ActionState deserialization failure**: Removed `@JsonIgnore` from `PythonEvent.getEvent()` so that Python event bytes are persisted in ActionState. Without this, recovered ActionState had `null` event bytes, causing `TypeError: a bytes-like object is required, not 'NoneType'` after recovery.

2. **Fix ActionStateStore not initialized during state recovery**: Moved ActionStateStore initialization to also run in `initializeState()`, ensuring it's available when rebuilding state from recovery markers.

3. **Fix state key mismatch after recovery**: Changed Python `Event.id` from random UUID to deterministic content-based UUID (MD5 hash). This ensures the same event produces the same ActionState key across restarts, enabling proper state lookup and divergence detection.

4. **Fix Kafka consumer partition assignment**: Use `consumer.assign()` instead of `subscribe()` for explicit partition control during state rebuild.

### Tests

Manually verified with local Kafka instance. E2E tests will be added in a follow-up PR.

### API

No public API changes.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/runtime/src/test/java/org/apache/flink/agents/runtime/python/event/PythonEventTest.java
+++ b/runtime/src/test/java/org/apache/flink/agents/runtime/python/event/PythonEventTest.java
@@ -83,8 +83,8 @@ class PythonEventTest {
         assertThat(jsonNode.has("eventType")).isTrue();
         assertThat(jsonNode.has("eventJsonStr")).isTrue();
         assertThat(jsonNode.has("attributes")).isTrue();
-        // event bytes should not be serialized
-        assertThat(jsonNode.has("event")).isFalse();
+        // event bytes should be serialized for ActionState persistence
+        assertThat(jsonNode.has("event")).isTrue();
         assertThat(jsonNode.get("eventType").asText()).isEqualTo(eventType);
         assertThat(jsonNode.get("eventJsonStr").asText()).isEqualTo(eventJsonStr);
         assertThat(jsonNode.get("attributes").get("testKey").asText()).isEqualTo("testValue");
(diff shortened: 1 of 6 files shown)


runtime/src/test/java/org/apache/flink/agents/runtime/python/event/PythonEventTest.java
```diff
@@ -117,7 +117,7 @@ void testEventLogRecordSerializationWithEventJsonStr() throws Exception {
         JsonNode eventNode = jsonNode.get("event");
         assertThat(eventNode.get("eventType").asText()).isEqualTo(eventType);
         assertThat(eventNode.get("id").asText()).isEqualTo(eventId.toString());
-        // Byte array should not be in the log
-        assertThat(eventNode.has("event")).isFalse();
+        // Byte array should be serialized for ActionState persistence
+        assertThat(eventNode.has("event")).isTrue();
```
xintongsong: This test case should not be changed. It verified the serialization for event logs, rather than for action states. It's expected not changed. This is the cause of the CI failures.

xintongsong (approved):
There's one minor issue, which I've fixed with an extra commit. Other than that, LGTM.

PR #514, opened 2026-02-03
[Bugfix] Use Thread Context ClassLoader for user class loading

klaudworks:
Linked issue: #515 

## Summary

This PR fixes `ClassNotFoundException` when loading user-defined resource classes (e.g., custom `ChatModel` implementations like for me `AzureOpenAIChatModelSetup`) from user JARs uploaded via the REST API.

## Problem

When `flink-agents-dist.jar` is deployed in `/opt/flink/lib` (which is required), it is loaded by the **System ClassLoader**. User JARs uploaded at runtime are loaded by Flink's **User ClassLoader**.

The existing code uses `Class.forName(className)` which uses the caller's classloader (System ClassLoader). The System ClassLoader cannot see classes in its child classloaders, resulting in `ClassNotFoundException`.

## Solution

Use the **Thread Context ClassLoader (TCCL)** instead:

```java
Class.forName(className, true, Thread.currentThread().getContextClassLoader())
```

Flink sets the TCCL to the User ClassLoader before executing user code, making user-defined classes accessible to framework code.

## Testing

Validated in a Flink 1.20 cluster with flink-agents, successfully loading custom `AzureOpenAIChatModelSetup` classes from user JARs uploaded via the REST API.

## Docs

- [ ] `doc-needed` 
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecordJsonDeserializer.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecordJsonDeserializer.java
@@ -84,8 +84,7 @@ public class EventLogRecordJsonDeserializer extends JsonDeserializer<EventLogRec
         try {
             // Load the concrete event class
             Class<?> eventClass =
-                    Class.forName(
-                            eventType, true, Thread.currentThread().getContextClassLoader());
+                    Class.forName(eventType, true, Thread.currentThread().getContextClassLoader());
 
             // Verify it's actually an Event subclass
             if (!Event.class.isAssignableFrom(eventClass)) {


xintongsong:
@klaudworks, thank you for reporting and fixing this bug. This is indeed a valid problem, and the fix sounds good. @wenjin272, could you please help check the details?

The 0.2.0 version is already in RC vote, and will soon be releases. Given that there's a workaround to this issue (placing user-defined resource in `flink/lib`), I'd suggest not to withdraw the RC and fix this in the next patch release. WDYT?

Additionally, I think we should:
- Create an issue describing the problem, workaround and track the fix version. This is helpful for other people who run into the same problem, because people usually tend to search in issues rather than PRs when they run into problems. @klaudworks, would you like to do that?
- We are planning to add some e2e tests that actually submit jobs into a flink standalone cluster, instead of currently running in a mini-cluster. I think user-defined resource should be included as one of the test cases. @wenjin272

klaudworks:
Thanks for looking into this!

@xintongsong I'll write a proper issue for this in the next few minutes.

@xintongsong For my personal case the current workaround is quite inconvenient for proper deployment scenarios. I'm providing the platform running flink and can't anticipate what users will try run. Putting flink-agents into the lib folder to load it with the system classloader feels acceptable but rebuilding the flink Docker image whenever someone modifies a custom resource would involve some effort.

klaudworks:
@wenjin272 thanks for further validating the fix!

Issue #521, opened 2026-02-08
[Feature] Add retries for remote calls in MCPServer, MCPTool and MCPPrompt


yanand0909:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently we don't have any retries for remote calls in MCPServer, MCPTool and MCPPrompt, we should add backoff based retries for all the calls to avoid retryable errors to fail the job, we can keep these values user configurable

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #531, opened 2026-02-10
[Feature][Java] Added retries to remote calls in MCP server

yanand0909:
Linked issue: [521](https://github.com/apache/flink-agents/issues/521)

### Purpose of change

Add user configurable retries for MCPServer remote calls

### Tests

Added retry test and existing UT and IT

### API

Yes

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
+++ b/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
@@ -87,9 +87,6 @@ public class MCPServer extends Resource {
     private static final String FIELD_MAX_BACKOFF_MS = "maxBackoffMs";
 
     private static final long DEFAULT_TIMEOUT_VALUE = 30L;
-    private static final int MAX_RETRIES_VALUE = 3;
-    private static final int INITIAL_BACKOFF_MS_VALUE = 100;
-    private static final int MAX_BACKOFF_MS_VALUE = 10000;
 
     @JsonProperty(FIELD_ENDPOINT)
     private final String endpoint;
@@ -103,14 +100,11 @@ public class MCPServer extends Resource {
     @JsonProperty(FIELD_AUTH)
     private final Auth auth;
 
-    @JsonProperty(FIELD_MAX_RETRIES)
-    private final int maxRetries;
+    private final Integer maxRetries;
 
-    @JsonProperty(FIELD_INITIAL_BACKOFF_MS)
-    private final long initialBackoffMs;
+    private final Long initialBackoffMs;
 
-    @JsonProperty(FIELD_MAX_BACKOFF_MS)
-    private final long maxBackoffMs;
+    private final Long maxBackoffMs;
 
     @JsonIgnore private transient RetryExecutor retryExecutor;
 
@@ -122,9 +116,9 @@ public class MCPServer extends Resource {
         private final Map<String, String> headers = new HashMap<>();
         private long timeoutSeconds = DEFAULT_TIMEOUT_VALUE;
         private Auth auth = null;
-        private int maxRetries = MAX_RETRIES_VALUE;
-        private long initialBackoffMs = INITIAL_BACKOFF_MS_VALUE;
-        private long maxBackoffMs = MAX_BACKOFF_MS_VALUE;
+        private Integer maxRetries;
+        private Long initialBackoffMs;
+        private Long maxBackoffMs;
 
         public Builder endpoint(String endpoint) {
             this.endpoint = endpoint;
@@ -195,21 +189,17 @@ public class MCPServer extends Resource {
 
         Object maxRetriesArg = descriptor.getArgument(FIELD_MAX_RETRIES);
         this.maxRetries =
-                maxRetriesArg instanceof Number
-                        ? ((Number) maxRetriesArg).intValue()
-                        : MAX_RETRIES_VALUE;
+                maxRetriesArg instanceof Number ? ((Number) maxRetriesArg).intValue() : null;
 
         Object initialBackoffArg = descriptor.getArgument(FIELD_INITIAL_BACKOFF_MS);
         this.initialBackoffMs =
                 initialBackoffArg instanceof Number
                         ? ((Number) initialBackoffArg).longValue()
-                        : INITIAL_BACKOFF_MS_VALUE;
+                        : null;
 
         Object maxBackoffArg = descriptor.getArgument(FIELD_MAX_BACKOFF_MS);
         this.maxBackoffMs =
-                maxBackoffArg instanceof Number
-                        ? ((Number) maxBackoffArg).longValue()
-                        : MAX_BACKOFF_MS_VALUE;
+                maxBackoffArg instanceof Number ? ((Number) maxBackoffArg).longValue() : null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


weiqingy:
PR title has typo: "retires" should be "retries".

integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
```diff
@@ -257,28 +347,117 @@ private void validateHttpUrl() {
...
+     * @return true if the operation should be retried
+     */
+    private boolean isRetryable(Exception e) {
+        // Network-related errors are retryable
+        if (e instanceof SocketTimeoutException || e instanceof ConnectException) {
+            return true;
+        }
+        String message = e.getMessage();
+        if (message != null) {
+            if (message.contains("503")) {
```
weiqingy: The string-contains check for "503" and "429" could match unrelated messages - e.g., a tool named "room-503-monitor" or an error like "processed 429 items". Would a more specific pattern like "HTTP 503" or a word-boundary check work here? Alternatively, if the MCP SDK throws typed exceptions for HTTP errors, checking the exception type would be more robust.

integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
```diff
@@ -77,10 +81,21 @@
  */
 public class MCPServer extends Resource {
 
+    private static final Random RANDOM = new Random();
+
     private static final String FIELD_ENDPOINT = "endpoint";
     private static final String FIELD_HEADERS = "headers";
     private static final String FIELD_TIMEOUT_SECONDS = "timeoutSeconds";
+    private static final String FIELD_TIMEOUT = "timeout";
```
weiqingy: nit: There seems to be an inconsistency between FIELD_TIMEOUT ("timeout") and the existing FIELD_TIMEOUT_SECONDS ("timeoutSeconds"). The @JsonCreator constructor uses FIELD_TIMEOUT_SECONDS, but the descriptor-based constructor reads FIELD_TIMEOUT. This means the field name used depends on how the MCPServer is constructed (JSON deserialization vs descriptor), which could confuse users. Would it make sense to unify these to a single field name?

integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
```diff
@@ -257,28 +347,117 @@ private void validateHttpUrl() {
...
+                    long jitter = RANDOM.nextInt((int) (backoff / 10) + 1);
+                    long sleepTime = backoff + jitter;
+                    Thread.sleep(sleepTime);
+                } catch (InterruptedException ie) {
+                    Thread.currentThread().interrupt();
+                    throw new RuntimeException(
+                            "Interrupted while retrying MCP operation: " + operationName, ie);
+                }
+
+                backoff = Math.min(backoff * 2, maxBackoffMs);
```
xintongsong: Binary Exponential Backoff Algorithm is typically as follows:
- There's a window size `W`, initially `W = W_min`, and each round `W = min(W * 2, W_max)`
- In each round, we randomly pick a value from the range `[0, W]`, and wait for that time.

Compared to the 10% jitter approach, the BEB algorithm brings more randomness, and reduce unnecessary long wait. I'd suggest to stick with the BEB algorithm.

integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
```diff
@@ -257,28 +347,117 @@ private void validateHttpUrl() {
...
 
+    /**
+     * Execute an operation with retry logic.
+     *
+     * @param operation The operation to execute
+     * @param operationName Name of the operation for error messages
+     * @return The result of the operation
+     * @throws RuntimeException if all retries fail
+     */
+    private <T> T executeWithRetry(Callable<T> operation, String operationName) {
```
xintongsong: I'd suggest to make `executeWithRetry` a method of a util class, so that it can be shared by other remote calls. E.g., calling an LLM or querying a vector store. The util class may take optional parameters during initialization, or use the default values.

xintongsong: This also helps decouple the retrying logics from `MCPServer`.

integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPRetryTest.java
```diff
@@ -0,0 +1,342 @@
...
+                .hasMessageContaining("failed after 2 retries");
+
+        // Should try initial attempt + 2 retries
+        assertThat(attempts.get()).isEqualTo(3);
+    }
+
+    @Test
+    @DisabledOnJre(JRE.JAVA_11)
+    @DisplayName("Exponential backoff timing")
+    void testExponentialBackoff() throws Exception {
```
xintongsong: Shall we remove this test case? Seems it doesn't verify anything more than other test cases already do.

integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPRetryTest.java
```diff
@@ -0,0 +1,342 @@
...
+        String result = invokeExecuteWithRetry(server, operation, "testOperation");
+
+        assertThat(result).isEqualTo("success");
+        assertThat(attempts.get()).isEqualTo(4);
+    }
+
+    @Test
+    @DisabledOnJre(JRE.JAVA_11)
+    @DisplayName("Max backoff limit respected")
+    void testMaxBackoffLimit() throws Exception {
```
xintongsong: ditto

integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPRetryTest.java
```diff
@@ -0,0 +1,342 @@
...
+            return (T) method.invoke(server, operation, operationName);
+        } catch (java.lang.reflect.InvocationTargetException e) {
+            Throwable cause = e.getCause();
+            if (cause instanceof Exception) {
+                throw (Exception) cause;
+            }
+            throw e;
+        }
+    }
+}
```
xintongsong, line 277: I'd suggest to add another test case that verifies non-retryable exception will not be retried..

integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
```diff
@@ -81,6 +82,14 @@ public class MCPServer extends Resource {
...
     private static final String FIELD_TIMEOUT_SECONDS = "timeoutSeconds";
     private static final String FIELD_AUTH = "auth";
+    private static final String FIELD_MAX_RETRIES = "maxRetries";
+    private static final String FIELD_INITIAL_BACKOFF_MS = "initialBackoffMs";
+    private static final String FIELD_MAX_BACKOFF_MS = "maxBackoffMs";
+
+    private static final long DEFAULT_TIMEOUT_VALUE = 30L;
+    private static final int MAX_RETRIES_VALUE = 3;
+    private static final int INITIAL_BACKOFF_MS_VALUE = 100;
+    private static final int MAX_BACKOFF_MS_VALUE = 10000;
```
xintongsong: Do we still need these default values? They are already included in `RetryExecutor`.

yanand0909: Yes actually it is redundant in MCPServer now, thanks for pointing out @xintongsong

PR #532, opened 2026-02-10
[hotfix][mcp] Fix MCPTool Jackson deserialization by ignoring unknown properties

avichaym:

`MCPTool` deserialization fails with `UnrecognizedPropertyException` when a default `ObjectMapper` encounters the `name` and `description` fields serialized from the parent `Tool` class. These fields are not declared in `MCPTool`'s `@JsonCreator` constructor.

This breaks checkpoint/restore and any JSON round-trip of `MCPTool`.

## Brief change log

- Add `@JsonIgnoreProperties(ignoreUnknown = true)` to `MCPTool`
- Add regression test `testJsonDeserializationWithDefaultMapper` that uses a default `ObjectMapper` (the existing `testJsonSerialization` masked the bug by disabling `FAIL_ON_UNKNOWN_PROPERTIES` globally)

- Runtime: yes (checkpoint/restore of MCP tool state)
- Integration: yes (MCP module)

## Documentation

N/A — bug fix only

## Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/tools/Tool.java
+++ b/api/src/main/java/org/apache/flink/agents/api/tools/Tool.java
@@ -18,6 +18,7 @@
 
 package org.apache.flink.agents.api.tools;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import org.apache.flink.agents.api.resource.ResourceType;
 import org.apache.flink.agents.api.resource.SerializableResource;
 
@@ -41,6 +42,7 @@ public abstract class Tool extends SerializableResource {
     }
 
     /** Return resource type of class. */
+    @JsonIgnore
     @Override
     public final ResourceType getResourceType() {
         return ResourceType.TOOL;
@@ -53,11 +55,13 @@ public abstract class Tool extends SerializableResource {
     public abstract ToolResponse call(ToolParameters parameters);
 
     /** Get the tool name from metadata. */
+    @JsonIgnore
     public final String getName() {
         return metadata.getName();
     }
 
     /** Get the tool description from metadata. */
+    @JsonIgnore
     public final String getDescription() {
         return metadata.getDescription();
     }
--- a/integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPToolTest.java
+++ b/integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPToolTest.java
@@ -36,6 +36,35 @@ class MCPToolTest {
 
     @Test
     @DisabledOnJre(JRE.JAVA_11)
+    @DisplayName(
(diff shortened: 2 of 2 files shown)


integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPTool.java
```diff
@@ -38,6 +39,7 @@
  * <p>This represents a single tool from an MCP server. It extends the base Tool class and delegates
  * actual execution to the MCP server.
  */
+@JsonIgnoreProperties(ignoreUnknown = true)
```
xintongsong: Call me old-fashioned, but I'd suggest to explicitly list the properties to be ignored rather than just ignore everything unknown. This forces us to explicitly think about whether a field should be ignored when adding a new one (or the test case will fail), preventing accidentally adding a new field without updating the serde.

wenjin272:
Hi, @avichaym, IIUC, this deserialization failure is because the `name`, `description` and `resourceType` field in json string, like
```
{
  "metadata" : {
    "name" : "add",
    "description" : "Add two numbers",
    "inputSchema" : "{\"type\":\"object\"}"
  },
  "mcpServer" : {
    "endpoint" : "http://localhost:8000/mcp",
    "headers" : { },
    "timeoutSeconds" : 30,
    "auth" : null
  },
  "name" : "add",
  "description" : "Add two numbers",
  "resourceType" : "TOOL"
}
```
but the `MCPTool` constructor only accepts `metadata` and `mcpServer`. 

I think we can just avoid serializing `name`, `description` and `resourceType` to the json string, because `name` and `description` can be retrieved from `metadata` and `resourceType` is always `TOOL`. 

This can be done by add `@JsonIgnore` in the get methods of `Tool`:
```
    /** Get the tool description from metadata. */
    @JsonIgnore
    public final String getDescription() {
        return metadata.getDescription();
    }
```

integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPTool.java
avichaym: Thanks  for the feedback. I've adopted wenjin272's approach — added @JsonIgnore on Tool.getName(), Tool.getDescription(), and Tool.getResourceType() so these redundant fields
are never serialized in the first place. This removes the need for @JsonIgnoreProperties on MCPTool entirely.

The fix is in Tool.java (api module) rather than MCPTool.java, so it benefits all Tool subclasses, not just MCPTool.

wenjin272:
The code looks good to me. There are some test failures due to code style.

xintongsong (approved):
@avichaym, thanks for addressing the comments. LGTM. Merging.

Issue #523, opened 2026-02-09
[Feature] Add Amazon Bedrock integration for chat models and embedding models


avichaym:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Motivation

Flink Agents currently supports Ollama, OpenAI, Anthropic (direct), and Azure AI as chat model providers, and Ollama/OpenAI for embeddings. There is no integration for Amazon Bedrock, which is the primary LLM gateway for AWS customers.

## Proposed Changes

Add two new integration modules:

- **Chat model** (`integrations/chat-models/bedrock/`) — Uses the Bedrock Converse API with native tool calling support. SigV4 auth via `DefaultCredentialsProvider`. Supports all Bedrock models accessible via Converse API (Claude, Llama, Mistral, Titan, etc.)

- **Embedding model** (`integrations/embedding-models/bedrock/`) — Uses Titan Text Embeddings V2 via InvokeModel. Batch `embed(List<String>)` parallelizes via configurable thread pool (`embed_concurrency` parameter, default 4).

I have a working implementation with unit tests ready to submit as a PR.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

buenzlidominik:
Thank you very much for this work @xintongsong @avichaym! I'm super excited to use it.
Is the documentation still missing? As I really need this feature, I could help if there's no time to write it on your side?

avichaym:
@buenzlidominik , @xintongsong    

Hi , sorry for the very delayed response (current geo-political situation in Israel shifted my focus and i forgot about the docs part , so thanks for the reminder) . 

I have a few working demo examples and can add the docs in the coming days .  @buenzlidominik you can reach me in the flink slack , my user is 'marciano avichay',  i can send you some working examples to unblock you in the meantime.

buenzlidominik:
Hi @avichaym 

Sorry to hear that. I hope everything improves soon and thanks for following up despite the difficult circumstances.
That would be fantastic, I'll write you in Slack.

xintongsong:
@avichaym, sorry to hear that. I hope you stay safe.

PR #406, opened 2025-12-29
[plan] Support error handling strategy for built-in chat action.

wenjin272:
Linked issue: #346 

### Purpose of change

Support retry/fail/ignore error handling strategy when chat occurs exception or response schema is not expected.

### Tests

ut & it

### API

No

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed`

PR #534, opened 2026-02-10
 [FLINK-AGENTS-523] Add Amazon Bedrock chat model and embedding model integrations

avichaym:
Linked issue: #523

### Purpose of change

Add Amazon Bedrock as a chat model and embedding model provider for Flink Agents.

- `BedrockChatModelConnection` — Converse API with native tool calling, SigV4 auth
- `BedrockChatModelSetup` — model, temperature, max_tokens configuration
- `BedrockEmbeddingModelConnection` — Titan Text Embeddings V2 with parallel batch embedding
- `BedrockEmbeddingModelSetup` — model, dimensions configuration

New modules: `integrations/chat-models/bedrock/`, `integrations/embedding-models/bedrock/`

### Tests

- Unit tests: `BedrockChatModelConnectionTest` (4), `BedrockChatModelSetupTest` (3), `BedrockEmbeddingModelTest` (4)
- End-to-end validated against real Bedrock with 6 demo applications (single-agent, ReAct tool calling, multi-agent pipeline, RAG, long-term memory, MCP)

### API

No public API changes. New integration modules only.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
+++ b/integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
@@ -1,0 +1,427 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.integrations.chatmodels.bedrock;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.flink.agents.api.RetryExecutor;
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.apache.flink.agents.api.chat.messages.MessageRole;
+import org.apache.flink.agents.api.chat.model.BaseChatModelConnection;
+import org.apache.flink.agents.api.resource.Resource;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
+import org.apache.flink.agents.api.resource.ResourceType;
+import org.apache.flink.agents.api.tools.Tool;
+import org.apache.flink.agents.api.tools.ToolMetadata;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+import software.amazon.awssdk.core.SdkNumber;
+import software.amazon.awssdk.core.document.Document;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.bedrockruntime.BedrockRuntimeClient;
+import software.amazon.awssdk.services.bedrockruntime.model.ContentBlock;
+import software.amazon.awssdk.services.bedrockruntime.model.ConversationRole;
+import software.amazon.awssdk.services.bedrockruntime.model.ConverseRequest;
+import software.amazon.awssdk.services.bedrockruntime.model.ConverseResponse;
+import software.amazon.awssdk.services.bedrockruntime.model.InferenceConfiguration;
+import software.amazon.awssdk.services.bedrockruntime.model.Message;
+import software.amazon.awssdk.services.bedrockruntime.model.SystemContentBlock;
+import software.amazon.awssdk.services.bedrockruntime.model.ToolConfiguration;
+import software.amazon.awssdk.services.bedrockruntime.model.ToolInputSchema;
+import software.amazon.awssdk.services.bedrockruntime.model.ToolResultBlock;
+import software.amazon.awssdk.services.bedrockruntime.model.ToolResultContentBlock;
+import software.amazon.awssdk.services.bedrockruntime.model.ToolSpecification;
+import software.amazon.awssdk.services.bedrockruntime.model.ToolUseBlock;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+/**
+ * Bedrock Converse API chat model connection for flink-agents.
+ *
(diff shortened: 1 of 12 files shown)


integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
```diff
@@ -0,0 +1,413 @@
...
+            int firstNewline = trimmed.indexOf('\n');
+            if (firstNewline >= 0) {
+                trimmed = trimmed.substring(firstNewline + 1);
+            }
+            if (trimmed.endsWith("```")) {
+                trimmed = trimmed.substring(0, trimmed.length() - 3).trim();
+            }
+            return trimmed;
+        }
+        int start = trimmed.indexOf('{');
```
weiqingy: This fallback path extracts the substring between the first '{'and last '}' from any response. If the model returns a normal text response that happens to contain braces (e.g., "Use the format {key: value} for config"), this would discard everything outside the braces.
Should this extraction only apply when structured output is expected (i.e., when an outputSchema is provided), rather than on every response? The other integrations (OpenAI, Anthropic) don't modify the text content at the connection layer.

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
```diff
@@ -0,0 +1,413 @@
...
+
+            ConverseResponse response = client.converse(requestBuilder.build());
+
+            if (response.usage() != null) {
+                recordTokenMetrics(
+                        modelId, response.usage().inputTokens(), response.usage().outputTokens());
+            }
+
+            return convertResponse(response);
+        } catch (Exception e) {
```
weiqingy: Bedrock frequently returns ThrottlingException under load. The current code wraps all exceptions into a RuntimeException without retry. It might be worth handling ThrottlingException and                                                                    ServiceUnavailableException with retry here - or at minimum, preserving the original exception type so callers can distinguish transient from permanent failures. (PR #531 is adding similar retry logic to MCPServer for the same reason.)

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
avichaym: Good catch. Removed the {...} extraction fallback entirely — stripMarkdownFences now only strips markdown code fences ( json... ), which is a lossless operation. Additionally, fence stripping is now 
only applied to non-tool-call responses, since tool-call responses should pass through unmodified.

The connection layer doesn't have visibility into whether outputSchema is set , so scoping it that way would require changing the BaseChatModelConnection interface. 
Removing the extraction altogether addresses the core concern — normal text responses containing braces are no longer corrupted.

Also Added 5 unit tests for stripMarkdownFences covering: text with braces preserved, clean JSON passthrough,  json fence stripping, plain fence stripping, and null handling.

> Fixed in 6d50c47

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
avichaym: Added exponential backoff retry around client.converse(), consistent with the retry already in BedrockEmbeddingModelConnection in this same PR. 
MAX_RETRIES=5, backoff with jitter (2^attempt  200ms  (0.5 + random  0.5)), covering ThrottlingException, ServiceUnavailableException, ModelErrorException, 429, and 503. Non-retryable exceptions still fail immediately with the original exception preserved as the cause.

Fixed in 6d50c47

wenjin272: Actually we have provided an unified retry mechanism in built-in action `ChatModelAction`, and user can configure `error-handling-strategy` and `max-retries`.  You can see #406 for details.

But the unified implementation doesn't consider whether the exception is retryable, and only take effort for chat model. I think an unified retry mechanism is better than implement it for each integration, which can decrease the complexity for adding new integrations. We can complete and improve it in the future prs, like throwing `RetryableException` in the integration, and retry in the corresponding built-in action, and supporting configure different retry times for different model. But I think it is not necessary to modify this pr. WDYT @xintongsong

xintongsong: We are adding a unified `RetryExecutor` in PR #531, which is almost mergeable. It also supports defining retryable exceptions. I think we can also leverage it in this PR.

avichaym: Thank you @wenjin272  and @xintongsong , Adopted the unified RetryExecutor from PR #531. Both BedrockChatModelConnection and BedrockEmbeddingModelConnection now use RetryExecutor.builder() with a Bedrock-specific retryablePredicate covering ThrottlingException, ServiceUnavailableException, ModelErrorException, 429, and 503. The manual retry loops are removed.

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
```diff
@@ -0,0 +1,428 @@
...
+        this.client =
+                BedrockRuntimeClient.builder()
+                        .region(Region.of(region))
+                        .credentialsProvider(DefaultCredentialsProvider.create())
+                        .build();
+
+        this.defaultModel = descriptor.getArgument("model");
+        this.retryExecutor =
+                RetryExecutor.builder()
+                        .maxRetries(5)
```
wenjin272: Should this be configurable?

integrations/embedding-models/bedrock/src/main/java/org/apache/flink/agents/integrations/embeddingmodels/bedrock/BedrockEmbeddingModelConnection.java
```diff
@@ -0,0 +1,187 @@
...
+        String model = descriptor.getArgument("model");
+        this.defaultModel = (model != null && !model.isBlank()) ? model : DEFAULT_MODEL;
+
+        Integer concurrency = descriptor.getArgument("embed_concurrency");
+        int threads = concurrency != null ? concurrency : 4;
+        this.embedPool = Executors.newFixedThreadPool(threads);
+
+        this.retryExecutor =
+                RetryExecutor.builder()
+                        .maxRetries(5)
```
wenjin272: ditto

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
avichaym: Made max_retries configurable via the descriptor's initialArguments map, defaulting to 5. Applied to both BedrockChatModelConnection and BedrockEmbeddingModelConnection. See 007b8f3.

Issue #538, opened 2026-02-14
[Bug] Java agents fail to connect to MCP servers that do not support prompts


avichaym:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

PR #447 (linked to #412) fixed this for Python agents, but the same bug exists in the Java path.

When using @MCPServer in a Java agent with an MCP server that only supports tools (not prompts), AgentPlan.extractJavaMCPServer fails during agent plan creation with:

io.modelcontextprotocol.spec.McpError: Method not found
JSONRPCError[code=-32601, message=Method not found, data=null]

Root cause: extractJavaMCPServer (line 436) calls listPrompts() unconditionally via reflection, ignoring the server's ServerCapabilities which declares prompts=null.

Impact: Any Java agent using @MCPServer with a tools-only MCP server (e.g., Amazon Bedrock AgentCore Gateway) cannot create an agent plan. Tools are discovered successfully, but the unconditional listPrompts call crashes plan creation.

Steps to Reproduce:
1. Create a Java agent with @MCPServer pointing to an MCP server that only supports tools
2. Run the agent — fails during plan creation with McpError: Method not found

Expected: Agent plan creation succeeds, prompts are simply not registered (matching the Python behavior after #447).

### How to reproduce

Steps to Reproduce:
1. Create a Java agent with @MCPServer pointing to an MCP server that only supports tools
2. Run the agent — fails during plan creation with McpError: Method not found

### Version and environment

- Flink Agents: 0.3-SNAPSHOT (main branch, commit 7db1157)
- MCP Java SDK: 0.16.0 (bundled in flink-agents-dist)
- Flink: 1.20.0
- Java: 11+ (reproduced on 17 and 23)
- OS: macOS (also reproducible on Linux CI)
- MCP Server: Amazon Bedrock AgentCore Gateway (any tools-only MCP server triggers this)

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #539, opened 2026-02-14
[FLINK-AGENTS] Fix Java agents failing to connect to MCP servers without prompt support (#538)

avichaym:
Linked issue: #538

Fix Java agents failing with `McpError: Method not found` when connecting to MCP servers that don't support prompts (e.g., AgentCore Gateway). PR #447 fixed the Python path (#412) but the Java path in `AgentPlan.extractJavaMCPServer` was not addressed.

## Brief change log

- Add `MCPServer.supportsPrompts()` — checks `McpSyncClient.getServerCapabilities().prompts() != null`
- Guard `listPrompts` call in `AgentPlan.extractJavaMCPServer` with `supportsPrompts()` via reflection
- Add unit test verifying `supportsPrompts` method exists and is reflection-callable

- Runtime / Operators: no
- Plan: yes (AgentPlan.extractJavaMCPServer)
- Integrations / MCP: yes (MCPServer.supportsPrompts)

## How was this patch tested?

- Unit test: `MCPServerTest.testSupportsPromptsMethodExists`
- End-to-end: Tested with Amazon Bedrock AgentCore Gateway (tools-only MCP server, `capabilities.prompts=null`). Agent plan creation succeeds, tools are discovered, Flink job completes successfully.

### Documentation

- [x] `doc-not-needed`

### DIFF
--- a/integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPServerTest.java
+++ b/integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPServerTest.java
@@ -30,6 +30,7 @@ import org.junit.jupiter.api.condition.JRE;
 
 import java.time.Duration;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import static com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES;
@@ -269,5 +270,15 @@ class MCPServerTest {
         assertThat(server.getMaxRetries()).isEqualTo(5);
         assertThat(server.getInitialBackoffMs()).isEqualTo(200);
         assertThat(server.getMaxBackoffMs()).isEqualTo(5000);
+    }
+
+    @Test
+    @DisabledOnJre(JRE.JAVA_11)
+    @DisplayName("listPrompts returns empty list when server does not support prompts")
+    void testListPromptsReturnsEmptyWhenNotSupported() {
+        MCPServer server = MCPServer.builder(DEFAULT_ENDPOINT).build();
+
+        List<MCPPrompt> prompts = server.listPrompts();
+        assertThat(prompts).isEmpty();
     }
 }
--- a/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
+++ b/integrations/mcp/src/main/java/org/apache/flink/agents/integrations/mcp/MCPServer.java
@@ -42,6 +42,7 @@ import java.net.URI;
 import java.net.http.HttpRequest;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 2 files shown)


plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -432,18 +432,22 @@ private void extractJavaMCPServer(Method method) throws Exception {
...
-        for (SerializableResource prompt : prompts) {
-            Method getNameMethod = prompt.getClass().getMethod("getName");
-            String promptName = (String) getNameMethod.invoke(prompt);
-            addResourceProvider(
-                    JavaSerializableResourceProvider.createResourceProvider(
-                            promptName, PROMPT, prompt));
+        // Call listPrompts() only if the server supports prompts (optional per MCP spec).
+        // MCP servers like AgentCore Gateway only support tools, not prompts.
+        Method supportsPromptsMethod = mcpServer.getClass().getMethod("supportsPrompts");
+        if ((boolean) supportsPromptsMethod.invoke(mcpServer)) {
```
xintongsong: I think it might be better to check `supportsPrompts` inside `listPrompts`, so that `listPrompts` returns an empty list rather than fail when the server doesn't support prompts. `AgentPlan` doesn't need to be aware of this complexity.

WDYT?

integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPServerTest.java
```diff
@@ -243,4 +243,16 @@ void testClose() {
...
+    @DisabledOnJre(JRE.JAVA_11)
+    @DisplayName("supportsPrompts method exists and is callable via reflection")
+    void testSupportsPromptsMethodExists() throws Exception {
+        MCPServer server = new MCPServer(DEFAULT_ENDPOINT);
+
+        // Verify the method exists (AgentPlan calls it via reflection)
+        java.lang.reflect.Method method = server.getClass().getMethod("supportsPrompts");
+        assertThat(method).isNotNull();
+        assertThat(method.getReturnType()).isEqualTo(boolean.class);
+    }
```
xintongsong, line 283: This test case just verifies a method exist and can be called via reflection. It seems a bit unnecessary.

I think what we need to verify is that, when the mcp server does not support prompts, the framework should not fail. This would become easier if we move the `supportsPrompt` check into `listPrompts`, as we just need to verify the behavior of `MCPServer`, without having to involve `AgentPlan` for the test.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
wenjin272: > I think it might be better to check `supportsPrompts` inside `listPrompts`, so that `listPrompts` returns an empty list rather than fail when the server doesn't support prompts. `AgentPlan` doesn't need to be aware of this complexity.
> 
> WDYT?

+1

avichaym: Done. Moved the supportsPrompts check inside MCPServer.listPrompts() — it returns an empty list when the server has no prompt capability. AgentPlan calls listPrompts() unconditionally and doesn't need 
to be aware of prompt support.

integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPServerTest.java
avichaym: Replaced with testListPromptsReturnsEmptyWhenNotSupported — calls listPrompts() on a server without prompt support and verifies it returns an empty list without throwing. Tests the actual behavior 
rather than checking a method exists via reflection.

Issue #545, opened 2026-02-21
[Tech Debt] Refactor ActionExecutionOperator into focused manager classes


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## What is currently not ideal?

`ActionExecutionOperator` is a 1,166-line "God class" that manages 15 state objects across 6 unrelated concerns: event routing, Python bridge management, durable execution, state management, context management, and action lifecycle coordination. This makes it difficult to:
  - Test individual concerns in isolation
  - Reason about state ownership and lifecycle
  - Safely add future features (multi-agent, better async) without risk of regressions

  ## How should it be fixed or improved?
  Decompose `ActionExecutionOperator` into 5 package-private manager classes, each owning a single concern:

  | New Class | Responsibility |
  |-----------|---------------|
  | `PythonBridgeManager` | Python interpreter, PythonActionExecutor, resource adapters lifecycle |
  | `DurableExecutionManager` | ActionStateStore, action state persistence, recovery, checkpoint pruning (implements `ActionStatePersister`) |
  | `ActionTaskContextManager` | Runner context creation, 4 transient maps, context transfer across async invocations |
  | `EventRouter` | Event wrapping/routing, notification (EventLogger + EventListeners), watermark management via SegmentedQueue |
  | `OperatorStateManager` | 7 Flink state objects initialization, sequence number management, snapshot helpers |

  After extraction, `ActionExecutionOperator` shrinks to ~562 lines of pure coordination logic.

  ### Constraints
  - All new classes are **package-private** — no public API changes
  - Flink state descriptor names/types unchanged — **savepoint compatible**
  - New managers are `transient`, initialized in `open()` — same serialization pattern
  - All existing tests must pass unchanged (reflection paths in tests updated to reach through managers)

  ### Verification
  - `mvn test -pl runtime` — all existing operator tests pass
  - `./tools/lint.sh -c` — formatting compliance
  - `./tools/ut.sh -j` — full Java test suite

### References     
- Originally proposed in [Discussion #516](https://github.com/apache/flink-agents/discussions/516) under "Tech Debt"

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #546, opened 2026-02-22
[runtime] Refactor ActionExecutionOperator into focused manager classes

weiqingy:
Linked issue: #545

### Purpose of change

Decomposes `ActionExecutionOperator` (1,166 lines) into 5 package-private manager classes, each owning a single concern:

  | New Class | Lines | Responsibility |
  |-----------|-------|---------------|
  | `OperatorStateManager` | 225 | Flink state handles (7 fields), sequence numbers, key-group filtering |
  | `DurableExecutionManager` | 285 | ActionStateStore, persistence, recovery, checkpoint pruning (implements `ActionStatePersister`) |
  | `ActionTaskContextManager` | 215 | Runner context creation, memory contexts, continuation executor, context transfer |
  | `EventRouter` | 175 | Event wrapping/routing, notification (logger + listeners), watermark management |
  | `PythonBridgeManager` | 177 | Python interpreter, action executor, resource adapters lifecycle |

  The operator shrinks from **1,166 to 562 lines** (52% reduction) of pure coordination logic.

  **Key design decisions:**
  - No manager-to-manager references — all cross-cutting dependencies flow through method parameters with the operator as mediator
  - `sequenceNumberKState` placed in OperatorStateManager (not DurableExecutionManager) because it's used in non-durable mode
  - `ActionStatePersister` interface moved from the operator to `DurableExecutionManager`
  - `ResourceCache` stays operator-owned — closed first before Python stack
  - Close order preserved: resourceCache → contextManager → pythonBridge → eventRouter → durableExecManager → super

  **Backward compatibility:**
  - All new classes are package-private — no public API changes
  - Flink state descriptor names/types unchanged — savepoint compatible
  - `ActionExecutionOperatorFactory` constructor signature preserved

### Tests

  - All 21 existing `ActionExecutionOperatorTest` tests pass as regression gates
  - `./tools/lint.sh -c` — formatting compliance verified
  - `./tools/ut.sh -j` — full Java test suite passes
  - `./tools/build.sh -j` — full Java build passes

  Test reflection accesses to `actionStateStore` (6 tests) and `eventLogger` (1 test) updated to use `@VisibleForTesting` getter chains instead of reflection on private fields.

  No new tests added. This is a pure structural refactoring — every code path flows through `ActionExecutionOperator` which the existing integration tests exercise via `KeyedOneInputStreamOperatorTestHarness`. The managers are package-private and not independently consumable APIs.

### API

  No public API changes. All new classes are package-private.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`


runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
```diff
@@ -0,0 +1,215 @@
...
+        if (context instanceof PythonRunnerContextImpl) {
+            // Get the awaitable ref from the transient map. After checkpoint restore, this will
+            // be null, signaling that the awaitable was lost and needs re-execution.
+            String awaitableRef = durableExecManager.getPythonAwaitableRef(actionTask);
+            ((PythonRunnerContextImpl) context).setPythonAwaitableRef(awaitableRef);
+        }
+        actionTask.setRunnerContext(context);
+    }
+
+    Resource getResource(String name, ResourceType type, ResourceCache resourceCache) {
```
wenjin272: Looks like this method is never invoked.

wenjin272 (commented):
Hi, @weiqingy, LGTM, only one comment. Please take a look at your convenience @xintongsong.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
weiqingy: Good catch — removed in 300a0fa. The method was dead code (callers go through RunnerContextImpl.getResource or ResourceCache directly), so I deleted it along with the now-unused Resource/ResourceType imports.

wenjin272:
LGTM. Please take a look at your convenience @xintongsong.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -253,82 +132,42 @@ public void setup(
...
+        pythonBridge.open(
+                agentPlan,
+                resourceCache,
+                getExecutionConfig(),
+                getRuntimeContext().getDistributedCache(),
+                getContainingTask().getEnvironment().getTaskManagerInfo().getTmpDirectories(),
+                getRuntimeContext().getJobInfo().getJobId(),
+                metricGroup,
+                this::checkMailboxThread,
+                stateManager.getJobIdentifier());
```
xintongsong: It's a bit against intuitive that  job id is fetched from the state manager. I think it should be maintained in the operator.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
```diff
@@ -0,0 +1,285 @@
...
+            Iterable<Object> recoveryMarkers = markerState.get();
+            if (recoveryMarkers != null) {
+                recoveryMarkers.forEach(markers::add);
+            }
+            LOG.info("Rebuilding action state from {} recovery markers", markers.size());
+            actionStateStore.rebuildState(markers);
+        }
+    }
+
+    ActionState maybeGetActionState(Object key, long sequenceNum, Action action, Event event)
```
xintongsong, line 201: I'd suggest to return an `Optional` or mark `@Nullable` for methods that may return `null`.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -121,54 +71,27 @@
...
 
-    private static final String RECOVERY_MARKER_STATE_NAME = "recoveryMarker";
-    private static final String MESSAGE_SEQUENCE_NUMBER_STATE_NAME = "messageSequenceNumber";
-    private static final String PENDING_INPUT_EVENT_STATE_NAME = "pendingInputEvents";
-
     private final AgentPlan agentPlan;
 
     private transient ResourceCache resourceCache;
 
     private final Boolean inputIsJava;
```
xintongsong: Never used.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
```diff
@@ -0,0 +1,285 @@
...
+    private static final String RECOVERY_MARKER_STATE_NAME = "recoveryMarker";
+
+    private ActionStateStore actionStateStore;
+    private transient ListState<Object> recoveryMarkerOpState;
+    private final Map<Long, Map<Object, Long>> checkpointIdToSeqNums;
+
+    private final Map<ActionTask, RunnerContextImpl.DurableExecutionContext>
+            actionTaskDurableContexts;
+    private final Map<ActionTask, ContinuationContext> continuationContexts;
+    private final Map<ActionTask, String> pythonAwaitableRefs;
```
xintongsong: These 2 maps might belong to `ActionTaskContextManager`. They are not related to durable execution. In this class, they are only initialized, and used in some one-line getter/setters which are all used by `ActionTaskContextManager`.

xintongsong: This also leads to the violation of "no manager-to-manager references" design goal.

xintongsong (commented):
@weiqingy Thanks for working on this. The refactor looks good to me in general. I just left a few comments.

In addition to the inline comments, I think we need more JavaDocs to explicitly describe the contract of each component interfaces, and unit tests to verify them. Currently, they are a bit vague relying on the names and implementations to understand.

weiqingy:
Thanks for reviewing, @xintongsong . Pushed `300a0fa..e2e4e03` to address all the comments:

- *"job id should be maintained in the operator"* — moved `jobIdentifier` from `OperatorStateManager` onto `ActionExecutionOperator`. Savepoint name unchanged. → `1139b9c`
- *"`@Nullable` for methods that may return `null`"* — applied `@Nullable` (consistent with the existing `@Nullable ActionStateStore`) across all 5 managers, 16 methods. → `85df684`
- *"`inputIsJava` never used"* — removed the field; constructor param is forwarded directly to `EventRouter`. → `1139b9c`
- *"these 2 maps belong to `ActionTaskContextManager` / no manager-to-manager references"* — moved `continuationContexts` and `pythonAwaitableRefs` to `ActionTaskContextManager`. `actionTaskDurableContexts` stays in `DurableExecutionManager` (consumed there). `createAndSetRunnerContext` no longer takes `DurableExecutionManager`. → `ab0c581`
- *"more JavaDocs to describe the contract of each component"* — added class-level Javadoc on each manager covering responsibilities, owned state, lifecycle, and the no-cross-reference constraint. → `6991b99`
- *"unit tests to verify them"* — added focused contract tests per manager (4 files, 7 tests). Heavier coverage (Flink-harness `OperatorStateManager` tests, full snapshot/recovery, watermark draining, listener notification) tracked as a follow-up to keep this PR reviewable. → `e2e4e03`

PTAL when you have a moment.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
```diff
@@ -0,0 +1,351 @@
...
+     * @param checkpointId the id of the completed checkpoint.
+     */
+    void notifyCheckpointComplete(long checkpointId) {
+        if (actionStateStore != null) {
+            Map<Object, Long> keyToSeqNum =
+                    checkpointIdToSeqNums.getOrDefault(checkpointId, new HashMap<>());
+            for (Map.Entry<Object, Long> entry : keyToSeqNum.entrySet()) {
+                actionStateStore.pruneState(entry.getKey(), entry.getValue());
+            }
+            checkpointIdToSeqNums.remove(checkpointId);
```
wenjin272, line 338: It appears that `checkpointIdToSeqNums` is only cleaned up when `actionStateStore` is not null, but there is no check for whether `actionStateStore` is null in `recordCheckpointSequenceNumbers`. This seems likely to cause a memory leak. 

I checked, and this issue existed before the refactoring. We can open a separate PR to fix it.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/EventRouter.java
```diff
@@ -0,0 +1,260 @@
...
+    void processEligibleWatermarks(WatermarkEmitter watermarkEmitter) throws Exception {
+        Watermark mark = keySegmentQueue.popOldestWatermark();
+        while (mark != null) {
+            watermarkEmitter.emit(mark);
+            mark = keySegmentQueue.popOldestWatermark();
+        }
+    }
+
+    @Nullable
+    SegmentedQueue getKeySegmentQueue() {
```
wenjin272, line 222: Is the `@Nullable` annotation here necessary?

For methods like `maybeGetActionState` and `pollNextPendingInputEvent`, I think the Nullable annotation is appropriate: It may return null when called, and the caller will perform a check. But for methods like `getKeySegmentQueue` and `getSensoryMemState`, they are initialized in the `open` method of `ActionExecutionOperator`, and all accesses occur after `open`.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -121,54 +72,25 @@
  * and the resulting output event is collected for further processing.
  */
 public class ActionExecutionOperator<IN, OUT> extends AbstractStreamOperator<OUT>
```
wenjin272, line 74: They are initialized in the Open method of ActionExecutionOperator, and all calls occur after Open.

wenjin272 (commented):
Hi, @weiqingy. I left a few minor comments, mainly regarding code style and doc.

Additionally, I found a bug that existed before the refactoring during my review. I noticed that you plan to add integration tests in a separate PR; we can fix this issue and add test verification in that PR.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
weiqingy: Good catch — confirmed, `checkpointIdToSeqNums` entries will accumulate without cleanup when durable execution is disabled. This is pre-existing and orthogonal to the current refactoring, so we'll track it in a separate PR as suggested. Thanks for the thorough review.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/EventRouter.java
weiqingy: Removed `@Nullable` from `getKeySegmentQueue()`, `getReusedStreamRecord()` (EventRouter), `getSensoryMemState()`, and `getShortTermMemState()` (OperatorStateManager) in 3699f14. These fields are unconditionally initialized in `open()` before any caller can reach them. Kept `@Nullable` where the return value genuinely can be null: `getEventLogger()` (config-dependent) and the `pollNext*` methods (poll from list-state returns null when empty).

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
weiqingy: Updated the PR.

weiqingy:
Created separate issues to follow up on the items deferred from this PR:

- #646 — Fix memory leak in `DurableExecutionManager.checkpointIdToSeqNums`
- #645 — Add comprehensive tests for the manager classes

Will follow up on them separately.

wenjin272:
LGTM. Please take a look at your convenience @xintongsong

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
```diff
@@ -0,0 +1,327 @@
...
+            // Get the awaitable ref from the transient map. After checkpoint restore, this will
+            // be null, signaling that the awaitable was lost and needs re-execution.
+            String awaitableRef = this.getPythonAwaitableRef(actionTask);
+            ((PythonRunnerContextImpl) context).setPythonAwaitableRef(awaitableRef);
+        }
+        actionTask.setRunnerContext(context);
+    }
+
+    @Nullable
+    RunnerContextImpl.MemoryContext getMemoryContext(ActionTask actionTask) {
```
xintongsong: Never used.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
```diff
@@ -0,0 +1,327 @@
...
+        return actionTaskMemoryContexts.get(actionTask);
+    }
+
+    void putMemoryContext(ActionTask actionTask, RunnerContextImpl.MemoryContext memoryContext) {
```
xintongsong: Should be private.

xintongsong (approved):
Thanks for addressing my comments, @weiqingy . This PR already looks quite good to me. I think it can be merged after resolving the conflicts and fixing the following minor issues.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
weiqingy: Removed in `1e04629`. Confirmed no callers anywhere - the accessor was a leftover from earlier iterations.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
weiqingy: Changed to `private` in `1e04629`. The only caller is `transferContexts` in the same class.

weiqingy:
Thanks @xintongsong @wenjin272 . I’ve updated the PR to address the comments and resolve the branch conflicts.

wenjin272 (approved):
LGTM, merged. Ty @weiqingy.

Issue #547, opened 2026-02-22
[Tech Debt]  Refactor AgentPlan & ResourceCache


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

 `AgentPlan` (624 lines) mixes plan definition, resource caching/resolution, Python bridge wiring, and serialization
  into one class. This makes the plan mutable after construction, prevents testing resource resolution independently,
  and couples Python MCP discovery logic to the plan definition.

  ### What is currently not ideal?

  - `AgentPlan` is **mutable** after construction — `setPythonResourceAdapter()` writes to the transient `resourceCache`
  - Resource resolution (`getResource()`) and cleanup (`close()`) are runtime concerns living on a serializable plan
  object
  - Python MCP tool/prompt discovery is entangled with the plan — AgentPlan directly references `PythonMCPServer`,
  `PythonResourceAdapter`
  - Cannot test resource resolution without constructing a full AgentPlan
  - Blocks runtime MCP reloading — the plan owns the cache, so refreshing tools requires mutating the plan

  ### How should it be fixed?

  Extract two classes:

  1. **`ResourceCache`** (~80 lines) — Lazy resource resolution, caching, dependency injection, and close/cleanup.
  Created by the operator in `open()`, owned by the operator lifecycle.
  2. **`PythonResourceBridge`** (~55 lines) — Static method that discovers tools/prompts from Python MCP servers and
  registers them in the cache. Called from `ActionExecutionOperator.initPythonResourceAdapter()`.

  After extraction, `AgentPlan` becomes an immutable data object (~490 lines) holding actions, resource providers, and
  config. The removed public methods are `getResource()`, `close()`, and `setPythonResourceAdapter()`.

  ### Files affected

  - `plan/.../AgentPlan.java` — remove extracted methods and transient fields
  - `plan/.../ResourceCache.java` — **new**
  - `plan/.../PythonResourceBridge.java` — **new**
  - `runtime/.../RunnerContextImpl.java` — delegate to `ResourceCache` instead of `AgentPlan`
  - `runtime/.../JavaRunnerContextImpl.java` — constructor change
  - `runtime/.../PythonRunnerContextImpl.java` — constructor change
  - `runtime/.../ActionExecutionOperator.java` — create and own `ResourceCache`
  - 6 test files updated to use `ResourceCache` directly

### References     
[...]

PR #548, opened 2026-02-22
[plan] Extract ResourceCache and PythonResourceBridge from AgentPlan

weiqingy:
Linked issue: #547 

### Purpose of change

 `AgentPlan` (624 lines) mixes plan definition, resource caching/resolution, Python bridge wiring, and serialization.
  This PR extracts two classes to separate concerns:

  - **`ResourceCache`** — lazy resource resolution, caching, and cleanup. Created by the operator in `open()`, owned by
  the operator lifecycle.
  - **`PythonResourceBridge`** — static `discoverPythonMCPResources()` for Python MCP tool/prompt discovery. Called
  during operator init.

  After extraction, `AgentPlan` becomes immutable after construction (~490 lines, down from 624). The removed public
  methods are `getResource()`, `close()`, and `setPythonResourceAdapter()`.

### Tests

  - `mvn test -pl plan` — all plan module tests pass
  - `mvn test -pl runtime` — all runtime module tests pass
  - `./tools/lint.sh -c` — formatting check passed
  - `./tools/ut.sh -j` — full Java test suite passed

### API

 Yes. Three public methods removed from `AgentPlan`:
  - `getResource(String, ResourceType)` — replaced by `ResourceCache.getResource()`
  - `close()` — replaced by `ResourceCache.close()`
  - `setPythonResourceAdapter(PythonResourceAdapter)` — replaced by `PythonResourceBridge.discoverPythonMCPResources()`

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/runtime/flink_runner_context.py
+++ b/python/flink_agents/runtime/flink_runner_context.py
@@ -36,7 +36,6 @@ from flink_agents.api.metric_group import MetricGroup
 from flink_agents.api.resource import Resource, ResourceType
 from flink_agents.api.runner_context import AsyncExecutionResult, RunnerContext
 from flink_agents.plan.agent_plan import AgentPlan
-from flink_agents.runtime.resource_cache import ResourceCache
 from flink_agents.runtime.flink_memory_object import FlinkMemoryObject
 from flink_agents.runtime.flink_metric_group import FlinkMetricGroup
 from flink_agents.runtime.memory.internal_base_long_term_memory import (
@@ -46,6 +45,7 @@ from flink_agents.runtime.memory.vector_store_long_term_memory import (
     VectorStoreLongTermMemory,
 )
 from flink_agents.runtime.python_java_utils import _build_event_log_string
+from flink_agents.runtime.resource_cache import ResourceCache
 
 logger = logging.getLogger(__name__)
 
--- a/python/flink_agents/runtime/local_runner.py
+++ b/python/flink_agents/runtime/local_runner.py
@@ -33,9 +33,9 @@ from flink_agents.api.resource import Resource, ResourceType
 from flink_agents.api.runner_context import AsyncExecutionResult, RunnerContext
 from flink_agents.plan.agent_plan import AgentPlan
 from flink_agents.plan.configuration import AgentConfiguration
-from flink_agents.runtime.resource_cache import ResourceCache
 from flink_agents.runtime.agent_runner import AgentRunner
 from flink_agents.runtime.local_memory_object import LocalMemoryObject
+from flink_agents.runtime.resource_cache import ResourceCache
 
 logging.basicConfig(
     level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s"


plan/src/main/java/org/apache/flink/agents/plan/ResourceCache.java
```diff
@@ -0,0 +1,131 @@
...
+ * <p>Resources are created on first access via their provider's {@code provide()} method and cached
+ * for subsequent lookups. Supports recursive dependency resolution — a resource can depend on other
+ * resources.
+ *
+ * <p>This class is designed for single-threaded access within Flink's mailbox execution model.
+ */
+public class ResourceCache implements AutoCloseable {
+
+    private final Map<ResourceType, Map<String, ResourceProvider>> resourceProviders;
+    private final Map<ResourceType, Map<String, Resource>> cache = new HashMap<>();
```
wenjin272: Should we use `ConcurrentHashMap` here? For task submit by `ctx.durableExecuteAsync` may read/write this hashmap parallel.

weiqingy: Good question! This is intentional. After the refactoring, `ResourceCache` is created and owned by `ActionExecutionOperator.open()`, so it's scoped to a single operator subtask. In Flink's execution model, all access (processElement, open, close) runs on the same mailbox thread. `durableExecuteAsync` dispatches work to external threads, but resource resolution from the cache happens on the operator thread before dispatch. So `HashMap` is sufficient and avoids the overhead of `ConcurrentHashMap`. Let me know if this matches your understanding.

wenjin272: > but resource resolution from the cache happens on the operator thread before dispatch

I think the resource resolution may not always happens before dispatch. 

Take the built in chat action as an example. In chat action, we submit chat task to external threads, and the chat task will call `chat` method of `BaseChatModelSetup`, which occurs in an asynchronous thread. In the `chat` method of chat model setup, it will resolves the correspond connection, prompt and tools, which I think may lead to concurrent access to the cache.

weiqingy: You're right, thanks for pushing back on this. I traced the code path more carefully:

ChatModelAction → durableExecuteAsync(callable) → async pool thread runs chatModel.chat() → BaseChatModelSetup.chat() calls  this.getResource.apply() to resolve connection, prompt, and tools → which hits ResourceCache.getResource().

So resource resolution does happen on async threads, not just the mailbox thread. I'll switch back to ConcurrentHashMap. Good catch!

weiqingy: Updated the PR.

wenjin272 (approved):
LGTM. Could you take a look at your convenience @xintongsong ?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -723,8 +731,8 @@ public void close() throws Exception {
         if (actionStateStore != null) {
             actionStateStore.close();
         }
-        if (runnerContext != null) {
-            runnerContext.close();
+        if (resourceCache != null) {
+            resourceCache.close();
```
wenjin272: I investigated the cross-language test issues locally and found that `resourceCache.close()` must be called before `pythonInterpreter.close()`. After moving it to the very beginning of the `close()` method, the tests passed successfully.

The issue didn't appear earlier because I missed something while resolving a merge conflict, which caused runnerContext to be closed twice in the `close()` method. In reality, only the first close operation took effect.

weiqingy: @wenjin272 Thanks for the review and for catching the close() ordering issue! Applied your fix — moved resourceCache.close() to the top of ActionExecutionOperator.close(), before pythonInterpreter.close(). Added a comment explaining the ordering constraint so it doesn't get accidentally re-ordered in the future.

weiqingy:
@wenjin272 The cross-language test (ChatModelCrossLanguageTest) has been consistently failing with httpx.ReadTimeout on this PR. After digging into it, I found:

On JDK 21, BaseChatModelSetup.chat() resolves resources (connection, prompt, tools) via getResource() inside the async thread (durableExecuteAsync). In cross-language scenarios, this ends up triggering Pemja calls into the Python interpreter from the async pool thread, which could lead to issues.

It looks like your PR #571 addresses this by moving resource resolution into open() on the main thread, so chat() no longer calls getResource() from the async thread. WDYT?

Issue #132, opened 2025-08-28
[Feature] Support OpenAI latest Response API


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

According to OpenAI documents

> The [Responses API](https://platform.openai.com/docs/api-reference/responses) is a core agentic API primitive, combining the simplicity of [Chat Completions](https://platform.openai.com/docs/api-reference/chat) with the ability to do more agentic tasks and express powerful reasoning.
> While Chat Completions remains supported, Responses is recommended for all new projects.

Compared with `Chat Completions API`, `Response API` supports more built-in tools, mcp and reasoning summaries. 

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #556, opened 2026-02-28
[Feature][Java] Add OpenAI Responses API integration

addu390:
Linked issue: #132

### Purpose of change

Adds support for OpenAI's Responses API as a new chat model integration alongside the existing Chat Completions integration.

The Responses API is OpenAI's recommended API. This initial implementation covers just the core functionality, similar to the existing Chat Completions integration.

The Setup class shares several parameters with `OpenAIChatModelSetup`. A shared base class could reduce duplication, happy to refactor in a follow-up if that's preferred.

### Tests

Verified via the existing `ChatModelIntegrationTest` e2e integration test with a new `OPENAI_RESPONSES` parameterized variant. All tests pass.

### API

Yes, adds two new public classes (`OpenAIResponseModelConnection`, `OpenAIResponseModelSetup`) and two new resource name constants. No changes to existing public APIs (intentionally left it as-is).

### Documentation

- [x] `doc-needed`
- [ ] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/resource/ResourceName.java
+++ b/api/src/main/java/org/apache/flink/agents/api/resource/ResourceName.java
@@ -67,9 +67,9 @@ public final class ResourceName {
 
         // OpenAI
         public static final String OPENAI_COMPLETIONS_CONNECTION =
-                "org.apache.flink.agents.integrations.chatmodels.openai.OpenAIChatModelConnection";
+                "org.apache.flink.agents.integrations.chatmodels.openai.OpenAICompletionsConnection";
         public static final String OPENAI_COMPLETIONS_SETUP =
-                "org.apache.flink.agents.integrations.chatmodels.openai.OpenAIChatModelSetup";
+                "org.apache.flink.agents.integrations.chatmodels.openai.OpenAICompletionsSetup";
 
         // OpenAI Responses API
         public static final String OPENAI_RESPONSES_CONNECTION =
--- a/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIResponsesModelConnection.java
+++ b/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIResponsesModelConnection.java
@@ -48,12 +48,12 @@ import java.util.function.BiFunction;
 /**
  * A <b>dedicated</b> OpenAI chat model integration using the Responses API.
  *
- * <p>Unlike {@link OpenAIChatModelConnection} which uses the Chat Completions API and works with
+ * <p>Unlike {@link OpenAICompletionsConnection} which uses the Chat Completions API and works with
  * any OpenAI-compatible provider (DeepSeek, DashScope, etc.), this implementation uses OpenAI's
  * Responses API which is specific to OpenAI.
  *
  * <p>For OpenAI-compatible providers that only support the Chat Completions API, use {@link
- * OpenAIChatModelConnection} instead.
+ * OpenAICompletionsConnection} instead.
  *
  * <p>Supported connection parameters:
  *
--- a/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
+++ b/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
@@ -80,7 +80,7 @@ import java.util.stream.Collectors;
  * public class MyAgent extends Agent {
  *   @ChatModelConnection
  *   public static ResourceDesc openAI() {
- *     return ResourceDescriptor.Builder.newBuilder(OpenAIChatModelConnection.class.getName())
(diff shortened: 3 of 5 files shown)


integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIResponsesModelConnection.java
```diff
@@ -0,0 +1,470 @@
...
+ * }
+ * }</pre>
+ */
+public class OpenAIResponseModelConnection extends BaseChatModelConnection {
+
+    private static final TypeReference<Map<String, Object>> MAP_TYPE = new TypeReference<>() {};
+
+    private final ObjectMapper mapper = new ObjectMapper();
+    private final OpenAIClient client;
+    private final String defaultModel;
```
wenjin272, line 90: This field maybe not needed. The `OpenAIResponseModelSetup` will always pass a model name when calling chat.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIResponseModelConnection.java
```diff
@@ -0,0 +1,470 @@
...
+ *             .build();
+ *   }
+ * }
+ * }</pre>
+ */
+public class OpenAIResponseModelConnection extends BaseChatModelConnection {
+
+    private static final TypeReference<Map<String, Object>> MAP_TYPE = new TypeReference<>() {};
+
+    private final ObjectMapper mapper = new ObjectMapper();
```
wenjin272: Because ObjectMapper is thread safe, this could be a static field.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIResponsesModelConnection.java
addu390: True. Was following an existing pattern, as seen in `AnthropicChatModelConnection` and `OpenAIChatModelConnection`.

It could be useful, if the intention was to choose the default at connection and not set-up.

e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/ChatModelIntegrationAgent.java
```diff
@@ -84,6 +84,11 @@ public static ResourceDescriptor chatModelConnection() {
             return ResourceDescriptor.Builder.newBuilder(ResourceName.ChatModel.OPENAI_CONNECTION)
                     .addInitialArgument("api_key", apiKey)
                     .build();
+        } else if (provider.equals("OPENAI_RESPONSE")) {
```
xintongsong:
```suggestion
        } else if (provider.equals("OPENAI_RESPONSES")) {
```

api/src/main/java/org/apache/flink/agents/api/resource/ResourceName.java
```diff
@@ -71,6 +71,12 @@ public static final class ChatModel {
         public static final String OPENAI_SETUP =
                 "org.apache.flink.agents.integrations.chatmodels.openai.OpenAIChatModelSetup";
 
+        // OpenAI Responses API
+        public static final String OPENAI_RESPONSE_CONNECTION =
+                "org.apache.flink.agents.integrations.chatmodels.openai.OpenAIResponseModelConnection";
+        public static final String OPENAI_RESPONSE_SETUP =
+                "org.apache.flink.agents.integrations.chatmodels.openai.OpenAIResponseModelSetup";
+
```
xintongsong, line 79: 1. According to the [documents](https://developers.openai.com/api/reference/responses/overview), the formal name of the API is "Responses API". Let's make it consistent and use RESPONSES (rather than RESPONSE) for the constants and class names.

2. I'd suggest to also change `OPENAI_CONNECTION`  to `OPENAI_COMPLETIONS_CONNECTION`, as well as the class names, to avoid confusion.

3. There are some string constants for referencing java integrations from python codes in `resource.py`, which should also be updated.

addu390: Missed those in `resource.py`, I made the changes to fix naming across the code-base.

Should we also rename `OpenAIChatModelConnection` to `OpenAICompletionsConnection`, `OpenAIChatModelSetup` to `OpenAICompletionsSetup` for consistency in naming?

wenjin272: > Should we also rename `OpenAIChatModelConnection` to `OpenAICompletionsConnection`, `OpenAIChatModelSetup` to `OpenAICompletionsSetup` for consistency in naming?

Yes, I think we should also change the class names to avoid confusion, as well as the `resource.py`.

addu390: I think so too! Updated the class names and the references.

xintongsong (approved):
Thanks for addressing my comments, @addu390. LGTM. Merging.

Issue #519, opened 2026-02-06
[Tech Debt] Reduce size of flink-agents wheel


Sxnan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The size of the wheel increased to 400M. There are a couple optimization we can do to reduce the size of the wheel. For example, remove the duplicated contents in the bundled jars, etc.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #557, opened 2026-03-02
[infra] Deduplicate common jar dependency to reduce wheel size

Sxnan:
Linked issue: #519

### Purpose of change

Deduplicate common jar dependency to reduce wheel size

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/dist/flink-1.20/pom.xml
+++ b/dist/flink-1.20/pom.xml
@@ -138,7 +138,7 @@ under the License.
                 </executions>
             </plugin>
 
-            <!-- Extend parent's shade plugin with flink-1.20 specific exclusions -->
+            <!-- Extend parent's shade plugin with flink-1.20 specific configurations -->
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


dist/flink-1.20/pom.xml
```diff
@@ -159,6 +159,46 @@ under the License.
...
+                            <filters>
+                                <!-- Only include flink-agents classes -->
+                                <filter>
+                                    <artifact>*:*</artifact>
+                                    <includes>
+                                        <include>org/apache/flink/agents/**</include>
+                                    </includes>
+                                    <!-- Exclude the conflict class for flink-1.20 -->
+                                    <excludes>
+                                        <exclude>org/apache/flink/agents/runtime/operator/OperatorUtils.class</exclude>
```
GreatEugenius: I believe this would exclude OperatorUtils for all versions. For Flink 1.20, we cannot find OperatorUtils.class. Should we exclude OperatorUtils only in the runtime module?

GreatEugenius:
For the whl package size, I wonder if we should include all integrations in the distribution, or if we should only add some default integrations, such as the Ollama ChatModel. The idea is quite similar to how Flink handles connectors in the distribution.

xintongsong:
The CI failure seems to be related.

GreatEugenius (approved):
LGTM

dist/flink-1.20/pom.xml
```diff
@@ -159,6 +159,49 @@ under the License.
...
+                                    <artifact>*:*</artifact>
+                                    <includes>
+                                        <include>org/apache/flink/agents/**</include>
+                                    </includes>
+                                </filter>
+                                <!-- Exclude the conflict class in runtime module for flink-1.20 -->
+                                <filter>
+                                    <artifact>org.apache.flink:flink-agents-runtime</artifact>
+                                    <excludes>
+                                        <exclude>org/apache/flink/agents/runtime/operator/OperatorUtils.class</exclude>
```
xintongsong: Do we really need this excluding? I think classes of the current module should have higher priority compared to classes introduced from dependency.

Issue #424, opened 2026-01-09
[Feature] Simplify the custom event definition.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, if user want to send/receive custom events, they must define a new event type, like
```
class MyEvent(Event):
    """User custom event."""
    field1: str
    field2: int
```
This can be somewhat cumbersome at times. 

To simplify the definition of Agent, we want to support user only declare the identifier, like:
`ctx.send_event(identifier="MyEvent", field1="test", field2=1)`

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wenjin272:
Hi, @addu390, thank you for your contribution! The code PR has been merged.

After reviewing the relevant documentation, I found that the introduction to `Event` needs to be updated and supplemented with additional content before we close this issue. The current introduction to `Event` is located in [docs/content/docs/development/workflow_agent.md](https://github.com/apache/flink-agents/blob/77a2ca4d200ebc95ffd1af681e674301dd49c71b/docs/content/docs/development/workflow_agent.md?plain=1#L493), and it still describes the original typed event.

I think we need introduce:
1. Unified event
```
# send event in one action
ctx.send_event(Event(type="my_event", attributes={"field1": "test", "field2": 42}))
# consume event in another action
field1: str = event.get_attr("field1")
```
2.  Json SerDe
```
input_data = ItemData.model_validate(InputEvent.from_event(event).input)
```
The field of a non-trivial type loses its type information after JSON serialization and deserialization, requiring users to manually reconstruct the object of the specific type. (This is only required for Python actions, not for Java actions, because event passing between Java actions does not require serialization.)

3. For subclass event, user need implement `from_event` method, and do validation and deserialization.

addu390:
@wenjin272 I just open a PR for it: https://github.com/apache/flink-agents/pull/651/changes

PR #561, opened 2026-03-05
Cross Language Unified Event

addu390:
Linked issue: #424

### Purpose of change

Add unified event support. Users can create events with a `type` string and `attributes` map instead of defining subclasses. Enables string-based event routing and JSON-based cross-language transport. Fully backward compatible with existing subclassed events.

### Tests

New unit tests in Java (`EventTest`, `EventLogRecordJsonSerdeTest`) and Python (`test_event`, `test_decorators`, `test_agent_plan`, `test_local_execution_environment`) covering unified event creation, serialization, routing, and backward compatibility.

### API

Yes. `Event` class (Java/Python) gains `type`, `attributes`, `getType()`/`get_type()`. `@Action` annotation gains `listenEventTypes()`. `@action` decorator accepts string identifiers. `RunnerContext.send_event()` accepts string identifier with kwargs.

More details about the design in the issue #424

### Documentation

- [x] `doc-needed`
- [ ] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/api/decorators.py
+++ b/python/flink_agents/api/decorators.py
@@ -15,12 +15,12 @@
 #  See the License for the specific language governing permissions and
 # limitations under the License.
 #################################################################################
-from typing import Callable, Type, Union
+from typing import Callable, Type
 
 from flink_agents.api.events.event import Event
 
 
-def action(*listen_events: Union[Type[Event], str]) -> Callable:
+def action(*listen_events: Type[Event] | str) -> Callable:
     """Decorator for marking a function as an agent action.
 
     Each argument can be either an :class:`Event` subclass (class-based routing)
@@ -31,12 +31,12 @@ def action(*listen_events: Union[Type[Event], str]) -> Callable:
     listen_events : Type[Event] | str
         Event classes or type-identifier strings that this action responds to.
 
-    Returns
+    Returns:
     -------
     Callable
         Decorator function that marks the target function with event listeners.
 
-    Raises
+    Raises:
     ------
     AssertionError
         If no events are provided, or if an argument is neither a string nor
--- a/python/flink_agents/api/decorators.py
+++ b/python/flink_agents/api/decorators.py
@@ -15,35 +15,41 @@
 #  See the License for the specific language governing permissions and
 # limitations under the License.
 #################################################################################
-from typing import Callable, Type
+from typing import Callable, Type, Union
 
 from flink_agents.api.events.event import Event
 
 
-def action(*listen_events: Type[Event]) -> Callable:
+def action(*listen_events: Union[Type[Event], str]) -> Callable:
     """Decorator for marking a function as an agent action.
+
+    Each argument can be either an :class:`Event` subclass (class-based routing)
+    or a plain ``str`` (unified-event routing by type identifier).
 
     Parameters
     ----------
-    listen_events : list[Type[Event]]
-        List of event types that this action should respond to.
+    listen_events : Type[Event] | str
+        Event classes or type-identifier strings that this action responds to.
 
-    Returns:
+    Returns
     -------
     Callable
         Decorator function that marks the target function with event listeners.
 
-    Raises:
+    Raises
     ------
     AssertionError
-        If no events are provided to listen to.
+        If no events are provided, or if an argument is neither a string nor
+        an ``Event`` subclass.
     """
     assert len(listen_events) > 0, (
         "action must have at least one event type to listen to"
     )
 
-    for event in listen_events:
-        assert issubclass(event, Event), "action must only listen to event types."
+    for evt in listen_events:
+        assert isinstance(evt, str) or (
+            isinstance(evt, type) and issubclass(evt, Event)
+        ), f"action must listen to Event subclasses or string identifiers, got {evt!r}"
 
     def decorator(func: Callable) -> Callable:
         func._listen_events = listen_events
--- a/python/flink_agents/api/runner_context.py
+++ b/python/flink_agents/api/runner_context.py
@@ -24,6 +24,8 @@ from flink_agents.api.memory.long_term_memory import BaseLongTermMemory
 from flink_agents.api.metric_group import MetricGroup
 from flink_agents.api.resource import Resource, ResourceType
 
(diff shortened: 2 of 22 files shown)


python/flink_agents/api/runner_context.py
```diff
@@ -79,15 +81,51 @@ class RunnerContext(ABC):
...
 
+        Can be called in two ways:
+
+        1. **Object form** — pass a pre-built :class:`Event` (or subclass)::
+
+               ctx.send_event(OutputEvent(output="hi"))
+
+        2. **Shorthand form** — provide ``identifier`` and key-value attrs::
+
+               ctx.send_event(identifier="MyEvent", field1="test", field2=1)
```
wenjin272, line 101: In #424, I described that I would like to support this shorthand form. But after the offline discuss with @xintongsong, we consider this way to be somewhat too casual. 

I think we can keep support only for the first approach, having `send_event` accept only pre-built event object. This ensures consistency of the `send_event` API between Python and Java, and eliminates the need to introduce a `_send_event` method.

python/flink_agents/api/decorators.py
```diff
@@ -20,13 +20,16 @@
 from flink_agents.api.events.event import Event
 
 
-def action(*listen_events: Type[Event]) -> Callable:
+def action(*listen_events: Type[Event] | str) -> Callable:
```
wenjin272, line 23: I think we can just remove support for `Type[Event]`, and only support declaring type-identifier strings in the action decorator.

Regarding type safety and ease of use, I will describe them in another comment.

python/flink_agents/api/tests/test_event.py
```diff
@@ -123,5 +123,72 @@ def test_event_with_mixed_serializable_types() -> None:
     assert parsed["input"]["nested_row"]["inner"]["type"] == "Row"
 
 
-def test_input_event_ignore_row_unserializable() -> None:  # noqa D103
+def test_input_event_ignore_row_unserializable_duplicate() -> None:  # noqa D103
```
wenjin272, line 126: Why we need modify this method name.

python/flink_agents/runtime/flink_runner_context.py
```diff
@@ -218,22 +218,45 @@ def set_long_term_memory(self, ltm: InternalBaseLongTermMemory) -> None:
         self.__ltm = ltm
 
     @override
-    def send_event(self, event: Event) -> None:
+    def _send_event(self, event: Event) -> None:
         """Send an event to the agent for processing.
 
+        Unified events (where ``event.type`` is set) are serialized as JSON and
+        sent via ``sendUnifiedEvent`` so that Java can reconstruct them without
+        cloudpickle.  Subclassed events use the legacy cloudpickle path.
```
wenjin272, line 226: In our plan, we will support cross-language usage of events and actions. To support call actions in another language, we should always serialize events to json, including unified events and subclassed events, so that another language can reconstruct them. The cloudpickle path should be removed.

python/flink_agents/api/tests/test_event.py
```diff
@@ -123,5 +123,72 @@ def test_event_with_mixed_serializable_types() -> None:
...
+
+    with pytest.raises(ValueError, match="type"):
+        Event.from_json(json.dumps({"attributes": {}}))
+
+
+def test_unified_event_serialization_roundtrip() -> None:
+    """Test that unified events survive JSON serialization/deserialization."""
+    import json
+
+    original = Event(type="RoundTrip", attributes={"a": 1, "b": "two"})
```
wenjin272, line 187: I think `Row` field should also be tested, for when user using Table api, the `input` of `InputEvent` will be `Row` type.

wenjin272 (commented):
HI, @addu390, thanks for your work. Due to some urgent tasks that came up, I apologize for not responding in a timely manner.

I think the intent of this PR aligns with our expectations, but there are some disagreements regarding the specific design.

I think we can keep the subclassed events, but there's no need to retain the support for declaring actions and sending events according to FQCNs and the cloudpickle SerDe . We can standardize on using the type string and json SerDe.

For the type safety, we can validate the data field when receive the event. Take internal `ChatRequestEvent` as an example.

```
class ChatRequestEvent(Event):
    """Event representing a request to chat model.

    Attributes:
    ----------
    model : str
        The name of the chat model to be chatted with.
    messages : List[ChatMessage]
        The input to the chat model.
    output_schema: OutputSchema | None
        The expected output schema of the chat model final response. Optional.
    """
    def __init__(self, model: str, messages: List[ChatMessage], output_schema: OutputSchema):
        super().__init__(type="_chat_request_event", attributes={"model": model,
                                                                 "messages": messages,
                                                                 "output_schema": output_schema})
    @classmethod
    @override
    def from_event(cls, event: Event) -> "ChatRequestEvent":
        assert "model" in event.attributes
        assert "messages" in event.attributes
        assert "output_schema" in event.attributes
        return ChatRequestEvent(event.attributes["messages"], event.attributes["output_schema"])
    
    @property
    def model(self) -> str:
        return self.attributes["model"]
    
    @property
    def messages(self) -> List[ChatMessage]:
        return self.attributes["messages"]
    
    @property
    def output_schema(self) -> OutputSchema:
        return self.attributes["output_schema"]
```

When a action received the chat request event, they can call `ChatRequestEvent.from_event(event)` to validate the event, and then use it as before.

wenjin272:
Hi, @addu390, do you have time to handle this PR recently?

addu390:
Hi @wenjin272 
Lost the flow of PRs and left it out for too long, sorry about that.

I had it in draft. Removing backward compatibility and migrating has quite a bit of changes, although, nothing major.
Created a new branch to show the diff: https://github.com/addu390/flink-agents/pull/7/changes

Would you rather have me update this branch with the changes or raise a new PR and close this one out?

wenjin272:
> Would you rather have me update this branch with the changes or raise a new PR and close this one out?

Hi, @addu390, both options work for me; whichever is more convenient for you.

addu390:
Closing this PR, since the new changes affects a lot more files and was easier to re-write. The new PR can be tracked here: https://github.com/apache/flink-agents/pull/631/changes

PR #567, opened 2026-03-13
[infra] Download JARs from Maven Central during pip install

Sxnan:
Linked issue: #519 

### Purpose of change

The `flink-agents` Python wheel includes JAR files (~110 MB), which exceeds PyPI's default 100 MB upload limit. This PR adds a [PEP 517](https://peps.python.org/pep-0517/) custom build backend so that JARs are downloaded from Maven Central during `pip install` (from sdist), keeping the published package well under the limit.

**Changes:**

- **Custom build backend** (`python/_build_backend/backend.py`): Wraps `setuptools.build_meta`, overrides `build_wheel()` to download JARs per `jar_manifest.json` before the standard build, with SHA-256 verification.
- **`jar_manifest.json`**: Declares Maven coordinates, destination paths, and checksums for 5 JARs (1 common + 4 Flink-version-specific thin JARs).
- **`pyproject.toml`**: Points `build-backend` to the custom backend.
- **Maven POM changes**:
  - `dist/common/pom.xml`: Remove `maven-deploy-plugin` skip to enable Maven Central deployment.
  - `dist/flink-{1.20,2.0,2.1,2.2}/pom.xml`: Set `shadedArtifactAttached=true` with `thin` classifier so thin JARs are published as attached artifacts.
- **Release scripts**: `create_release_branch.sh` and `update_branch_version.sh` now update `jar_manifest.json` version alongside `pyproject.toml`.
- **`tools/ut.sh`**: Run `_build_backend/tests/` before the per-Flink-version test loop.
- **`.gitignore`**: Ignore transient `python/flink_agents/lib/`.

**Environment variables:**

| Variable | Description |
|----------|-------------|
| `FLINK_AGENTS_SKIP_JAR_DOWNLOAD` | Skip JAR download (for local dev with `tools/build.sh`) |
| `FLINK_AGENTS_MAVEN_MIRROR` | Override Maven repository base URL |

### Tests

- 11 unit/integration tests in `python/_build_backend/tests/test_backend.py` covering manifest loading, checksum verification, skip logic, HTTP download with a local test server, checksum mismatch detection, and end-to-end `build_wheel` producing a wheel with JARs.

### API

No public API changes. The custom build backend is an internal build infrastructure component.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/get-started/installation.md
+++ b/docs/content/docs/get-started/installation.md
@@ -176,7 +176,7 @@ The following environment variables can be used to control how JARs are resolved
 
 | Variable | Description |
 |----------|-------------|
-| `FLINK_AGENTS_SKIP_JAR_DOWNLOAD` | Set to `1`, `true`, `yes`, or `on` to skip downloading JARs from Maven Central. Useful when building from source with `tools/build.sh`, which copies JARs from the local Maven build. |
+| `FLINK_AGENTS_SKIP_JAR_DOWNLOAD` | Set to `1`, `true`, `yes`, or `on` (case-insensitive) to skip downloading JARs from Maven Central. Useful when building from source with `tools/build.sh`, which copies JARs from the local Maven build. |
 | `FLINK_AGENTS_MAVEN_MIRROR` | Override the Maven repository base URL for JAR downloads. Defaults to `https://repo1.maven.org/maven2`. Useful for environments with restricted network access or internal mirrors. |
 
 ## Maven Dependencies (For Java)


docs/content/docs/get-started/installation.md
```diff
@@ -170,6 +170,15 @@ After building:
 - The Python package is installed and ready to use
 - The distribution JAR is located at: `dist/flink-${FLINK_VERSION%.*}/target/flink-agents-dist-*.jar`
 
+### Build Environment Variables
+
+The following environment variables can be used to control how JARs are resolved during `pip install flink-agents` (from sdist) or `python -m build`:
+
+| Variable | Description |
+|----------|-------------|
+| `FLINK_AGENTS_SKIP_JAR_DOWNLOAD` | Set to `1`, `true`, `yes`, or `on` to skip downloading JARs from Maven Central. Useful when building from source with `tools/build.sh`, which copies JARs from the local Maven build. |
```
alnzng: Are these values case sensitive?

python/jar_manifest.json
```diff
@@ -0,0 +1,36 @@
+{
+  "maven_base_url": "https://repo1.maven.org/maven2",
+  "group_id": "org.apache.flink",
+  "version": "0.3-SNAPSHOT",
```
alnzng, line 4: Just want to confirm - We have a tool to maintain this version in `main` branch, right?

python/_build_backend/backend.py
```diff
@@ -0,0 +1,183 @@
...
+        )
+        return
+
+    manifest = _load_manifest(manifest_path)
+    maven_base_url = os.environ.get(_MIRROR_ENV_VAR) or manifest["maven_base_url"]
+    maven_base_url = maven_base_url.rstrip("/")
+    group_path = manifest["group_id"].replace(".", "/")
+    version = manifest["version"]
+
+    for jar_entry in manifest["jars"]:
```
alnzng, line 122: Technically, when a user run `pip install`, they should already have the Flink version decided. Ideally, the users only need to download a jar. But I guess "download all 5 jars" is not a big deal, since this only happen once during each build time.

python/_build_backend/backend.py
```diff
@@ -0,0 +1,183 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#################################################################################
+
+"""Custom PEP 517 build backend that downloads JARs from Maven Central.
+
+Wraps ``setuptools.build_meta`` and overrides ``build_wheel()`` to download
+JAR files before the standard setuptools build runs.  This ensures that
+``pip install flink-agents`` (from sdist) produces a wheel that already
```
alnzng, line 23: To make sure I fully understand the UX impact for both devs and users, I have a few clarifying questions:

- Local Development Flow: How does ./tools/build.sh behave on the local main branch now? Since this change seems to expect JARs to be available in a Maven repo, how should a developer test local Java changes before they are published?
- Distribution: Just to confirm, will we only be providing the sdist (.tar.gz) on PyPI moving forward, with no wheel files available?
- Dependency Management: Since dependencies will now be resolved on the user's side at install time, they may end up with different library versions than what we tested at release time. If a newer dependency breaks the framework, is there a way for us to control or 'pin' these versions from the framework side, or is the burden now on the user to manage those conflicts?

docs/content/docs/get-started/installation.md
Sxnan: These values are case-insensitive. So TRUE, True, true, 1, etc., all work correctly.

python/jar_manifest.json
Sxnan: Yes, we have scripts that automatically maintain the version in `jar_manifest.json`:

  - create_release_branch.sh:69-70: Updates version when creating a release branch
  - update_branch_version.sh:59-60: Updates version when updating the main branch version

python/_build_backend/backend.py
Sxnan: I agree with your assessment. The current design downloads all 5 JARs (1 common ~110MB + 4 thin JARs ~1MB each), but this only happens once during build time. The total size is reasonable (~114MB), and it simplifies the implementation while ensuring users have all JARs available for any Flink version they
  choose at runtime.

python/_build_backend/backend.py
Sxnan: 1. **Local Development Flow:** When using `./tools/build.sh` on a local branch (including `-SNAPSHOT` versions):
   - Maven builds the JARs locally first
   - `build.sh` copies them to `python/flink_agents/lib/`
   - When `build_wheel` runs, `_download_jar` detects the JARs already exist and skips the download (see `backend.py:152-154`)
   - For `-SNAPSHOT` versions, Maven Central doesn't have them, but since local JARs exist, no download is attempted

2. **Distribution:** Correct, we will only publish **sdist** (`.tar.gz`) to PyPI, not wheel files. The wheel is built on the user's machine during `pip install` from sdist, and includes the downloaded JARs.

3. **Dependency Management:** There's no runtime dependency resolution for JARs. The JARs are downloaded at **build time** with **fixed SHA-256 checksums** specified in `jar_manifest.json` included in the sdist. This ensures users get the exact same JARs that were tested at release time - no version drift is possible. Python dependencies are managed normally through `pyproject.toml`.

docs/content/docs/get-started/installation.md
alnzng: Great! It would be better if we call out the value is not case senstive in the doc.

python/_build_backend/backend.py
alnzng: 1. Ah, this makes sense.
2. Will we upload wheel file once PyPI team approved our request? Or we only keep source distrbution in PyPI in the future?
3. I should be more clear. Actually my question was for Python dependency management, the dependencies are not pined in our `pyproject.toml`, so the actual dependency version resolved could be different during each `pip install` process. But the wheel file pre-installed all the dependencies, so their versions are kindly pinned. That being said, the users might have to manage the versions for the dependencies pulled by Flink-Agents framework expliactly. Ideally, we should keep both wheel file and sdist if it is possible, IMO.

Sxnan: Actually, I think wheel files work similarly in terms of dependency resolution. Even with a wheel file, Python dependencies listed in pyproject.toml are still resolved and downloaded by pip at installation time.

So whether we distribute via sdist or wheel, the Python dependency resolution behavior would be the same.

alnzng:
> > I don't have concerns on downloading jars on the fly, especially we give the options that the users have the ability to skip this process. And as you mentioned, this could address future size constraints since the jar size might become larger and larger.
> > I was trying to know if we have plan to support wheel file (keep both of sdist and wheel) once we got PyPI approved on increasing the size. I know most of Python supports both, so I was thinking we should support both if we can in the future. I understand there is some complexity on maintaining both of them because of our design.
> 
> @alnzng That's a valid point. For most Python packages, the main benefits of wheel are faster installation (no build step) and not requiring a build environment. However, for flink-agents specifically:
> 
> * Our package is pure Python + JAR files, with no C extensions to compile
> * The installation time would be similar since the total download size is the same (larger wheel vs smaller sdist + JARs from Maven)
> * The current sdist approach works well for our use case
> 
> Additionally, supporting both sdist and wheel would add some extra maintenance overhead for the build process. For users in environments with restricted access to Maven Central, they can use `FLINK_AGENTS_SKIP_JAR_DOWNLOAD=1` and manually provide the JAR files.
> 
> So I'm inclined to stick with sdist-only for now, unless we find compelling reasons to add wheel support. We can always revisit this discussion if needed. WDYT?

Sounds good @Sxnan! Let's start with source distribution only.

python/_build_backend/backend.py
alnzng: I re-confirmed this, I think you are right on this!

Issue #560, opened 2026-03-04
[Feature] Add support for custom job name in Flink Agents Job execution


GreatEugenius:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description
When submitting a Flink Agents Job to a Flink cluster following the quickstart guide, the job appears in the Flink Web UI with the default name "Flink Streaming Job". This makes it difficult to identify and manage multiple agent jobs running in the cluster.

#### Proposed Solution
Add a `job_name` parameter to the `execute()` method to allow users to specify a custom job name programmatically.

**Current Implementation:**
```python
def execute(self) -> None:
    """Execute agent individually."""
```

**Proposed Implementation:**
```python
def execute(self, job_name: str = None) -> None:
    """Execute agent individually.
    
    Args:
        job_name: Optional custom name for the Flink job. 
    """
```

#### Use Case Example
```python
agent_env.execute(job_name="Diagnosis Agent")
```

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #570, opened 2026-03-13
Feat job name config

qingfei1994:
Linked issue: #560 

### Purpose of change

Users can customize Job Name of Flink Agents.

### Tests

UnitTest

### API

Yes

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/runtime/src/test/java/org/apache/flink/agents/runtime/env/RemoteExecutionEnvironmentTest.java
+++ b/runtime/src/test/java/org/apache/flink/agents/runtime/env/RemoteExecutionEnvironmentTest.java
@@ -17,7 +17,10 @@
  */
 package org.apache.flink.agents.runtime.env;
 
+import org.apache.flink.agents.api.AgentsExecutionEnvironment;
 import org.apache.flink.agents.plan.AgentConfiguration;
+import org.apache.flink.api.common.JobExecutionResult;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
```diff
@@ -197,6 +197,20 @@ public AgentBuilder fromTable(Table input) {
...
+     *
+     * <p>This method triggers the execution of the configured agent pipeline with the specified job
+     * name. For remote environments, the job name is passed to the underlying Flink execution
+     * environment. The default implementation ignores the job name and delegates to {@link
+     * #execute()}.
+     *
+     * @param jobName The name for the Flink job.
+     */
+    public void execute(String jobName) throws Exception {
+        execute();
```
wenjin272: The job name is not passed to the underlying Flink execution environment.

qingfei1994: My first thought is to providing an optional execute with jobName method for other execution environment like LocalExecutionEnvironment which doesn't support executing with jobName, but since then I think I would just provide an abstract method here.

xintongsong (approved):
Thank you for working on this, @qingfei1994 and @wenjin272 . LGTM. Merging.

Issue #448, opened 2026-01-17
[Tech Debt] Cross language resource can't support async execution.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Flink-Agents provide `ctx.durable_execute_async` interface, and support built-in actions async execution in python and java.  However, due to some unresolved issues, invoking a resource from another language in async execution causes the JVM to crash—whether it's Java using a Python resource or Python using a Java resource. S

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

xintongsong:
@wenjin272  The PR is marked as `doc-needed`. Let's update the docs before closing this issue.

wenjin272:
We have updated the documents in #745

PR #571, opened 2026-03-17
[runtime] Support async execution for cross language resources.

wenjin272:
Linked issue: #448 

### Purpose of change

When invoking a Java method via Pemja from a Python async thread, if that Java method creates objects using reflection, a `ClassNotFoundException` may occur because the Python async thread lacks the necessary Java environment context. 

In Flink agents, this issue occurs when the Python API calls a Java resource that depends on other Java resources in async thread. For example, the python chat model action call `chat` of a java chat model setup, and the setup will get the correspond chat model connection, which will construct the connection object by reflection.

To resolve this issue, this pr moves the construction of resources from `chat`/`embed` method to `open`, and the `open` will invoked in flink mailbox thread. Additionally, we have added a check similar to the one for memory: users are now prohibited from calling get_resource outside of a mailbox thread.

This patch depends on https://github.com/alibaba/pemja/pull/95 in pemja. Therefore, we need to wait for the fix versions for Pemja and Flink 1.20, 2.0, 2.1, and 2.2.

### Tests

existed e2e test for cross language resources

### API

Yes. Add `open` for `Resource`

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
@@ -26,6 +26,9 @@ import org.apache.flink.agents.api.resource.ResourceDescriptor;
 import org.apache.flink.agents.api.resource.ResourceType;
 import org.apache.flink.agents.api.tools.Tool;
 import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.util.Preconditions;
+
+import javax.annotation.Nullable;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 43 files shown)


wenjin272:
After the code review is complete with no issues, I will add the corresponding documentation updates to this PR.

tools/ut.sh
```diff
@@ -170,7 +170,7 @@ java_tests() {
 
         exclude_list="!e2e-test/flink-agents-end-to-end-tests-integration,!e2e-test/flink-agents-end-to-end-tests-resource-cross-language"
 
-        mvn -T16 --batch-mode --no-transfer-progress test -pl "${exclude_list}"
+        mvn -T16 --batch-mode --no-transfer-progress test -fn -pl "${exclude_list}"
```
xintongsong: Should we use `-fae`?

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -259,7 +259,7 @@ public Resource getResource(String name, ResourceType type) throws Exception {
 
         // Cache the resource
         resourceCache.computeIfAbsent(type, k -> new ConcurrentHashMap<>()).put(name, resource);
-
+        resource.open();
```
xintongsong: `open()` is called after putting the resource into the cache, which means if `open()` fails, the corrupted resource might be left in the cache.

plan/src/main/java/org/apache/flink/agents/plan/actions/Utils.java
```diff
@@ -0,0 +1,97 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+import java.util.StringTokenizer;
+
+public final class Utils {
+    private static final Logger LOG = LoggerFactory.getLogger(Utils.class);
+    private static final String defaultValue = "<unknown>";
```
xintongsong: DEFAULT_VALUE

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
```diff
@@ -35,18 +35,44 @@
...
 
 public abstract class BaseChatModelSetup extends Resource {
-    protected final String connection;
+    protected final String connectionName;
     protected String model;
     protected Object prompt;
-    protected List<String> tools;
+    protected List<String> toolNames;
+
+    protected BaseChatModelConnection connection;
```
xintongsong: We should mark this `@Nullable`, and always check that it's not null before using it. Same for the other resources.

python/flink_agents/api/chat_models/chat_model.py
```diff
@@ -133,9 +137,11 @@ class BaseChatModelSetup(Resource):
     different chat configurations.
     """
 
-    connection: str = Field(description="Name of the referenced connection.")
+    connection: str | BaseChatModelConnection = Field(
```
xintongsong, line 140: We should always check that `connection` is a `BaseChatModelConnection` before using. Maybe overridding the getter? Same for other resources.r

PR #581, opened 2026-03-20
[runtime] Filter resumed currentProcessingKeys by subtask ownership a…

joeyutong:
…fter rescaling

### Purpose of change

`ActionExecutionOperator` persists `currentProcessingKeysOpState` as union operator state so all subtasks can see in-flight keys after restore/rescaling. However, `tryResumeProcessActionTasks()` resumed every restored key without checking whether the key belonged to the current subtask.

That behavior is incorrect after scale-out: multiple subtasks can receive the same union state entry, so non-owner subtasks may also enqueue resume work for keys they do not own. This can lead to duplicated processing after recovery.

This change fixes the restore path by filtering `currentProcessingKeysOpState` with subtask ownership before resubmitting work:
   - compute the key group for each restored key
  - compute the owning subtask for the current maxParallelism and parallelism
  - only resume keys owned by the current subtask

The ownership check is encapsulated in `isKeyOwnedByCurrentSubtask(...)` in `ActionExecutionOperator.java.
`
### Tests

Added an operator-level restore/rescaling regression test in `ActionExecutionOperatorTest.java`:

  - snapshot state from parallelism=1
  - repartition it to parallelism=2
  - restore both subtasks
  - verify only the owner subtask gets a resume mail, while the non-owner subtask does not

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
@@ -68,6 +68,7 @@ import org.apache.flink.api.common.state.ValueState;
 import org.apache.flink.api.common.state.ValueStateDescriptor;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.python.env.PythonDependencyInfo;
+import org.apache.flink.runtime.state.KeyGroupRange;
 import org.apache.flink.runtime.state.KeyGroupRangeAssignment;
 import org.apache.flink.runtime.state.StateInitializationContext;
 import org.apache.flink.runtime.state.StateSnapshotContext;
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -1125,6 +1126,18 @@ private void maybeInitActionStateStore() {
...
 
+    private boolean isKeyOwnedByCurrentSubtask(Object key) {
+        int maxParallelism = getRuntimeContext().getTaskInfo().getMaxNumberOfParallelSubtasks();
+        int parallelism = getRuntimeContext().getTaskInfo().getNumberOfParallelSubtasks();
+        int subtaskIndex = getRuntimeContext().getTaskInfo().getIndexOfThisSubtask();
+
+        int keyGroup = KeyGroupRangeAssignment.assignToKeyGroup(key, maxParallelism);
+        int owner =
+                KeyGroupRangeAssignment.computeOperatorIndexForKeyGroup(
+                        maxParallelism, parallelism, keyGroup);
```
xintongsong: This computation here is repeated for every single key in the list state. This is unnecessary because all the inputs (maxParallelism, parallelism, keyGroup) won't change during the lifecycle of the operator. Therefore, we should calculate `owner` once and use it for comparing with all the keys.

joeyutong: Thanks, good point. I updated this to compute the current subtask’s key-group range once and reuse it for all restored keys, instead of recomputing ownership for each key.

PR #584, opened 2026-03-24
[ci] Remove action astral-sh/setup-uv@v4 in ci due to legal issue.

wenjin272:
### Purpose of change

The  action `astral-sh/setup-uv@v4` is not allowed used in this repo. I think this may be the apache requirement, for I can still use it in my fork repo.

image

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`


xintongsong (approved):
LGTM

Issue #590, opened 2026-03-26
[api][python] Support Agent Skills in python


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support using agnet skills in python api

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #596, opened 2026-03-26
[api][integration] Support agent skills in python

wenjin272:
Linked issue: #590 

### Purpose of change

This PR provides basic support for using agent skills in the Flink Agents Python API. We will further enhance usability in subsequent PRs.

### Tests

ut & e2e

### API

Yes, add abstraction for agnet skill, and modify the chat model api.

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/resource/ResourceType.java
+++ b/api/src/main/java/org/apache/flink/agents/api/resource/ResourceType.java
@@ -31,7 +31,8 @@ public enum ResourceType {
     VECTOR_STORE("vector_store"),
     PROMPT("prompt"),
     TOOL("tool"),
-    MCP_SERVER("mcp_server");
+    MCP_SERVER("mcp_server"),
+    SKILLS("skills");
 
     private final String value;
 
--- a/python/flink_agents/api/decorators.py
+++ b/python/flink_agents/api/decorators.py
@@ -190,6 +190,23 @@ def vector_store(func: Callable) -> Callable:
     return func
 
 
+def skills(func: Callable) -> Callable:
+    """Decorator for marking a function declaring skills.
+
+    Parameters
+    ----------
+    func : Callable
+        Function to be decorated.
+
+    Returns:
+    -------
+    Callable
+        Decorator function that marks the target function declare skills.
+    """
+    func._is_skills = True
+    return func
+
+
 def java_resource(cls: Type) -> Type:
     """Decorator to mark a class as Java resource."""
     cls._is_java_resource = True
--- a/python/flink_agents/api/resource.py
+++ b/python/flink_agents/api/resource.py
@@ -32,7 +32,7 @@ class ResourceType(Enum):
     """Type enum of resource.
 
     Currently, support chat_model, chat_model_server, tool, embedding_model,
-    vector_store, prompt, mcp_server.
+    vector_store, prompt, mcp_server, skills.
     """
 
     CHAT_MODEL = "chat_model"
@@ -43,6 +43,7 @@ class ResourceType(Enum):
     VECTOR_STORE = "vector_store"
     PROMPT = "prompt"
     MCP_SERVER = "mcp_server"
+    SKILLS = "skills"
 
 
 class Resource(BaseModel, ABC):
--- a/python/flink_agents/api/resource.py
+++ b/python/flink_agents/api/resource.py
@@ -18,9 +18,11 @@
 import importlib
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, Dict, Type
+from typing import TYPE_CHECKING, Any, Dict, Type
 
-from pydantic import BaseModel, Field, PrivateAttr, model_validator
+from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, model_validator
+
+from flink_agents.api.resource_context import ResourceContext
 
 if TYPE_CHECKING:
     from flink_agents.api.metric_group import MetricGroup
@@ -30,7 +32,7 @@ class ResourceType(Enum):
     """Type enum of resource.
 
     Currently, support chat_model, chat_model_server, tool, embedding_model,
-    vector_store and prompt.
+    vector_store, prompt, mcp_server.
     """
 
     CHAT_MODEL = "chat_model"
@@ -52,14 +54,13 @@ class Resource(BaseModel, ABC):
 
     Attributes:
     ----------
-    get_resource : Callable[[str, ResourceType], "Resource"]
+    resource_context : ResourceContext
         Get other resource object declared in the same Agent. The first argument is
         resource name and the second argument is resource type.
     """
 
-    get_resource: Callable[[str, ResourceType], "Resource"] = Field(
-        exclude=True, default=None
-    )
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    resource_context: ResourceContext | None = Field(exclude=True, default=None)
 
     # The metric group bound to this resource, injected in RunnerContext#get_resource
     _metric_group: "MetricGroup | None" = PrivateAttr(default=None)
--- a/python/flink_agents/api/resource_context.py
+++ b/python/flink_agents/api/resource_context.py
@@ -1,0 +1,36 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+"""Public ResourceContext interface.
+
+Defines the capabilities available to a Resource during execution.
+The concrete implementation lives in :mod:`flink_agents.runtime.resource_context`.
+"""
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from flink_agents.api.resource import Resource, ResourceType
+
+
+class ResourceContext(ABC):
+    """Base abstract class for Resource Context."""
+
+    @abstractmethod
+    def get_resource(self, name: str, resource_type: "ResourceType") -> "Resource":
+        """Get another resource declared in the same Agent."""
--- a/python/flink_agents/api/resource_context.py
+++ b/python/flink_agents/api/resource_context.py
@@ -22,7 +22,7 @@ The concrete implementation lives in :mod:`flink_agents.runtime.resource_context
 """
 
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, List
 
 if TYPE_CHECKING:
     from flink_agents.api.resource import Resource, ResourceType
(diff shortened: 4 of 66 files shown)


python/flink_agents/runtime/skill/repository/url_repository.py
```diff
@@ -0,0 +1,46 @@
...
+
+from typing_extensions import override
+
+from flink_agents.runtime.skill.agent_skill import AgentSkill
+from flink_agents.runtime.skill.skill_repository import (
+    SkillRepository,
+)
+
+
+# TODO: Implement
```
alnzng: I would suggest to remove this unimplemented class to avoid confusion.

python/flink_agents/runtime/skill/skill_manager.py
```diff
@@ -0,0 +1,285 @@
...
+        if any(script_path.endswith(ext) for ext in self._allowed_extensions):
+            return None
+        return (
+            f"Script '{script_path}' has an unsupported type. "
+            f"Allowed extensions: {sorted(self._allowed_extensions)}."
+        )
+
+    @staticmethod
+    def _split_commands(command: str) -> List[str]:
+        """Split a command string by shell operators, respecting quotes.
```
alnzng: IIUC, The intention of this method is to extract individual executable commands from a compound command string, not to perform a full shell/bash tokenization.  Shell has many operators (e.g. redirection >, >>, <), but this method only targets command-separating operators, those that connect multiple executables. Can you confirm? If yes, let's make it clear in the comment.

Also can we confirm all command-separating operators are covered? For example, are && and || handled correctly?

Out of curiosity, is there any native Python lib or thirdparty lib could provide this kind of functionality? I am not sure if we covered all the necessary shell / bash operators, so I'm thinking if there is a such kind of lib then we could just use it.

python/flink_agents/runtime/skill/repository/filesystem_repository.py
```diff
@@ -0,0 +1,210 @@
...
+            return None
+
+        try:
+            skill_md_content = skill_md_path.read_text()
+
+            skill = SkillParser.parse_skill(skill_md_content)
+
+            if skill.name != skill_dir.name:
+                logger.warning(
+                    f"The skill name {skill.name}is different from the base directory {skill_dir.name}."
```
alnzng: `{skill.name}is` -> `{skill.name} is`

python/flink_agents/runtime/skill/skill_parser.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # Pattern to match frontmatter: starts with ---, ends with ---
+    # Group 1: frontmatter content (non-greedy)
+    # Group 2: remaining content
+    FRONTMATTER_PATTERN = re.compile(
+        r"^---\s*[\r\n]+(.*?)[\r\n]*---(?:\s*[\r\n]+)?(.*)", re.DOTALL
+    )
+
+    # Pattern to match key: value format
+    KEY_VALUE_PATTERN = re.compile(r"^([a-zA-Z_][a-zA-Z0-9_-]*)\s*:\s*(.*)$")
```
alnzng: Looks like we can remove this? I didn't see it is used in any other place.

python/flink_agents/runtime/skill/skill_parser.py
```diff
@@ -0,0 +1,146 @@
...
+            return ParsedMarkdown(metadata={}, content=markdown)
+
+        yaml_content = matcher.group(1).strip()
+        markdown_content = matcher.group(2)
+
+        if not yaml_content:
+            return ParsedMarkdown(metadata={}, content=markdown_content)
+
+        try:
+            metadata = yaml.safe_load(yaml_content)
```
alnzng, line 86: does markdown frontmatter always follow YAML syntax?

python/flink_agents/runtime/skill/skill_parser.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+
+    metadata: Dict[str, str]
+    content: str
+
+
+class MarkdownSkillParser:
+    """Utility for parsing Markdown files with YAML frontmatter."""
+
+    # Pattern to match frontmatter: starts with ---, ends with ---
```
alnzng, line 48: is below a valid frontmatter?

```
 ---
  name: tricky
  description: |
    This has a
    --- inside
  ---
```

python/flink_agents/runtime/skill/skill_repository.py
```diff
@@ -0,0 +1,84 @@
...
+    """Abstract interface for skill repositories.
+
+    A SkillRepository is responsible for loading and optionally storing skills
+    from a specific source (filesystem, classpath, URL, etc.).
+
+    Each skill is stored in its own subdirectory containing a SKILL.md file
+    and optional resource files:
+
+    baseDir/
+    ├── skill-name-1/
```
alnzng, line 54: What will happen if the users define folder with other different names? For example, this Claude official skill: https://github.com/anthropics/skills/tree/main/skills/skill-creator

alnzng (commented):
Left a few comments, will continue the review. BTW, for future PRs, I'd suggest separating code formatting / refactoring changes from actual logic changes into separate PRs.

python/flink_agents/api/decorators.py
```diff
@@ -189,6 +189,24 @@ def vector_store(func: Callable) -> Callable:
     func._is_vector_store = True
     return func
 
+
+def skills(func: Callable) -> Callable:
```
xintongsong, line 193: wrong commit

python/flink_agents/e2e_tests/e2e_tests_integration/long_term_memory_test.py
```diff
@@ -218,7 +218,7 @@ async def retrieve_items(event: Event, ctx: RunnerContext):  # noqa D102
     "flink-agent doesn't allow get resource in async thread. We will deprecate VectorStoreLongTermMemory in 0.3.0,"
     "so we will not fix this issue for now."
 )
-def test_long_term_memory_async_execution_in_action(tmp_path: Path) -> None:  # noqa: D103
+def test_long_term_memory_async_execution_in_action(tmp_path: Path) -> None:
```
xintongsong, line 205: wrong commit

python/flink_agents/api/skills.py
```diff
@@ -0,0 +1,83 @@
...
+        resources: List of Python package resources to load skills from.
+        allowed_script_types: Script types that are allowed to execute
+            from skill directories. Defaults to ``["shell", "python"]``.
+            Supported types: ``"shell"`` (.sh, .bash), ``"python"`` (.py).
+        allowed_commands: Whitelist of executable command names that skills
+            are permitted to run when the command is not a skill script.
+            Only the first token of a command is checked against this list.
+    """
+
+    paths: List[str] = Field(default_factory=list)
```
xintongsong, line 52: I'd suggest to provide some factory methods, like `Skills.from_local_dir()` or `Skills.from_url()`. This aligns with `Prompt.from_text() / from_messages()`, and is more easy-to-use.

xintongsong: And maybe hide the implementation of `Skills` from the users.

python/flink_agents/api/resource_context.py
```diff
@@ -33,3 +33,10 @@ class ResourceContext(ABC):
     @abstractmethod
     def get_resource(self, name: str, resource_type: "ResourceType") -> "Resource":
         """Get another resource declared in the same Agent."""
+
+    @abstractmethod
+    def generate_skill_discovery_prompt(self, *skill_names: str) -> str:
```
xintongsong: I'd suggest the name `generate_available_skills_prompt`

python/flink_agents/api/resource.py
```diff
@@ -42,6 +42,7 @@ class ResourceType(Enum):
     VECTOR_STORE = "vector_store"
     PROMPT = "prompt"
     MCP_SERVER = "mcp_server"
+    SKILL = "skill"
```
xintongsong: `SKILLS`

api/src/main/java/org/apache/flink/agents/api/resource/ResourceType.java
```diff
@@ -31,7 +31,8 @@ public enum ResourceType {
     VECTOR_STORE("vector_store"),
     PROMPT("prompt"),
     TOOL("tool"),
-    MCP_SERVER("mcp_server");
+    MCP_SERVER("mcp_server"),
+    SKILL("skill");
```
xintongsong: `SKILLS`

python/flink_agents/runtime/skill/skill_manager.py
```diff
@@ -0,0 +1,285 @@
...
+from flink_agents.api.skills import Skills
+from flink_agents.runtime.skill.agent_skill import AgentSkill
+from flink_agents.runtime.skill.repository.filesystem_repository import (
+    FileSystemSkillRepository,
+)
+from flink_agents.runtime.skill.skill_prompt_provider import SkillPromptProvider
+from flink_agents.runtime.skill.skill_repository import SkillRepository
+
+
+class RegisteredSkill:
```
xintongsong: Why do we need both `AgentSkill` and `RegisteredSkill` ?

python/flink_agents/api/skills.py
```diff
@@ -0,0 +1,83 @@
...
+    urls: List[str] = Field(default_factory=list)
+    resources: List[str] = Field(default_factory=list)
+    allowed_script_types: List[str] = Field(
+        default_factory=lambda: ["shell", "python"]
+    )
+    allowed_commands: List[str] = Field(default_factory=list)
```
xintongsong: This doesn't sounds right. How is allowed script types / commands an attribute of the skills? If the model decide to execute something, how do we know it's following the instructions of a skill or not?

python/flink_agents/runtime/skill/skill_tools.py
```diff
@@ -0,0 +1,253 @@
...
+        ctx = self.resource_context
+        if isinstance(ctx, ResourceContextImpl):
+            return ctx.get_skill_manager()
+        return None
+
+
+class ExecuteCommandArgs(BaseModel):
+    """Arguments for ExecuteCommandTool."""
+
+    skill_name: str = Field(
```
xintongsong: Not sure about requiring a skill name for executing a command. Skills are not executable programs like tools, but are a set of knowledges, scripts and resource for doing something. It's the LLM/agent that learns a skill and perform actions. You cannot execute a skill, nor saying an action belongs to a certain skill.

python/flink_agents/runtime/skill/skill_parser.py
wenjin272: According to the Agent Skills specification: https://agentskills.io/specification#skill-md-format, The `SKILL.md` file must contain YAML frontmatter followed by Markdown content.

python/flink_agents/runtime/skill/skill_manager.py
wenjin272: Yes, the functionality of this function is indeed as you described. 

I could not find an existing third-party library that provides this functionality, so I implemented one myself by referencing [agentscope-java](https://github.com/agentscope-ai/agentscope-java/blob/8b2f4513f6f5b200d08d1df40774507a19522182/agentscope-core/src/main/java/io/agentscope/core/tool/coding/UnixCommandValidator.java#L170).

I added some tests to confirm that delimiters such as `&&`, `|`, and `;` are handled correctly. Additional tests are needed for `||`.

Actually, I currently feel that the security validation for our execute_shell_command might be implemented too simply. I mainly referenced agentscope-java, but the bash tool in Claude Code's implementation includes extensive security validation rules. Therefore, I may refactor it by referencing Claude Code.

python/flink_agents/runtime/skill/skill_parser.py
wenjin272: This is a valid frontmatter, and the parsed result is `{'description': 'This has a', 'name': 'tricky'}`, the left content will be regarded as markdown content `inside\n---\n`.

python/flink_agents/runtime/skill/skill_repository.py
wenjin272: I don't quite understand this question. Regarding skill-creator, then the `baseDir` can be any dir the flink-agents job can access, like `/flink/usrlib/skills`, and the `skill-name-1` will be `skill-creator`.

python/flink_agents/runtime/skill/skill_parser.py
wenjin272: Initially, the coding agent generated code that used regular expressions to parse YAML frontmatter. I later switched to using a YAML library but forgot to remove these two PATTERNs.
I will remove this.

wenjin272:
> Regarding the skill e2e testing, I evaluated qwen3:8b, qwen3.5:4b, and qwen3.5:9b. I found that only qwen3.5:9b with thinking enabled could consistently follow the prompts. However, in the GitHub CI environment, the lack of GPU resources results in very slow model inference, causing tests to take over 30 minutes to complete. Consequently, I have temporarily skipped the skill e2e tests.
> 
> I plan to dedicate time later to investigate and resolve this issue.

To resolve this issue, I replace the chat model from ollama to openai, and create a api key. The key is saved as github action secrets. The secrets will exist in encrypted form and are not passed to the runner when a workflow is triggered from a forked repository.

I created an API key on DashScope using my personal account. Developers can also replace it with any base_url and api_key compatible with the OpenAI v1 protocol during testing.

xintongsong (approved):
LGTM

Issue #586, opened 2026-03-25
[Feature]  Import Wizard for Installation Setup


JinkunLiu:
### Motivation
Enhance the quickstart experience. - Currently, there are too many manual steps, thus chances for users to make mistakes and run into problems.

#### 1) Ask user whether to install Flink

If user selects **Yes**:
- Download the Flink binary archive to a temporary directory.
- Extract Flink into the target installation directory and remove the downloaded archive.
- Set and export `FLINK_HOME`.

If user selects **No**:
- Validate that `FLINK_HOME` is already set and points to a valid Flink directory .

---

#### 2) Ask user whether to enable PyFlink

If user selects **Yes**:
- Copy `flink-python` JAR from `$FLINK_HOME/opt/` to `$FLINK_HOME/lib/`.
- Create and activate a Python virtual environment.
- Install `flink-agents` and pinned `apache-flink` via pip.
- Copy Flink Agents Java artifacts into `$FLINK_HOME/lib/`.

If user selects **No**:
- Skip all Python/PyFlink-related steps.

---

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wenjin272:
Hi, @JinkunLiu, Ty for contributing the install script.  

The code pr has been merged, but before we close this issue, I think we need refactor the installation documentation for flink-agents. I have created sub-issue #699 for this. Are you interested in taking this issue?

JinkunLiu:
Hi @wenjin272, Thanks for your help in getting this issue done together, and thanks for setting up the follow-up sub-issue!

I'd be glad to take on [#699](https://github.com/apache/flink-agents/issues/699). Please feel free to assign it to me. I'll provide a refactored documentation outline as soon as possible so we can align on the structure first.

PR #599, opened 2026-04-01
[tools]Import Wizard for Installation Setup

JinkunLiu:
Linked issue: #586 

### Purpose of change

Import Wizard for Installation Setup 

Install Flink & Install PyFlink Environment

### Tests

I have test shell on Ubuntu 22.0 and macOS 15.0

### API

not

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`


tools/install.sh
```diff
@@ -0,0 +1,838 @@
...
+
+configure_verbose() {
+    if [[ "$VERBOSE" != "1" ]]; then
+        return 0
+    fi
+    set -x
+}
+
+FLINK_VERSION="${FLINK_VERSION:-2.2.0}"
+FLINK_AGENTS_VERSION="${FLINK_AGENTS_VERSION:-0.2.1}"
```
wenjin272, line 448: The default version `0.2.1` mismatch the default version `0.3.0` in help message.

tools/install.sh
```diff
@@ -0,0 +1,838 @@
...
+print_usage() {
+    cat <<EOF
+Apache Flink Agents Installer
+
+Usage:
+  bash install.sh [options]
+  curl -fsSL <url>/install.sh | bash -s -- [options]
+
+Options:
+  --yes, -y               Non-interactive mode (accept all defaults)
```
wenjin272: I think `--yes` is a bit semantically unclear. It might be better to change it to something like `--non-interactive`.

tools/install.sh
```diff
@@ -0,0 +1,838 @@
+#!/bin/bash
```
wenjin272: Shall we use `#!/usr/bin/env bash`.

tools/install.sh
```diff
@@ -0,0 +1,838 @@
...
+    fi
+    if command -v wget &> /dev/null; then
+        DOWNLOADER="wget"
+        return 0
+    fi
+    ui_error "Missing downloader (curl or wget required)"
+    exit 1
+}
+
+download_file() {
```
wenjin272, line 68: When I first ran this script, there was no output for a long time; I later discovered it was downloading gum. However, due to my network issues, the download took too long.

I was wondering if we could display a notification when the download time exceeds a certain threshold. For example, showing 'Installing [xxx], please wait,' or even providing a progress bar.

tools/install.sh
```diff
@@ -0,0 +1,838 @@
...
+    if [[ -z "$py_version_output" ]]; then
+        ui_error "Could not parse python3 version"
+        return 1
+    fi
+
+    local py_major py_minor
+    py_major="${py_version_output%%.*}"
+    py_minor="${py_version_output##*.}"
+
+    if [[ "$py_major" -ne 3 ]] || [[ "$py_minor" -lt 10 ]]; then
```
wenjin272: The conditions do not require Python < 3.12.

tools/install.sh
```diff
@@ -0,0 +1,838 @@
...
+
+    local version_lib_dir="${pkg_root}/lib/flink-${FLINK_MAJOR_MINOR}"
+
+    [[ -d "$version_lib_dir" ]] || die "Flink Agents lib directory not found: $version_lib_dir"
+
+    local copied=0
+    local jar
+    for jar in "$version_lib_dir"/flink-agents-dist-*.jar; do
+        [[ -f "$jar" ]] || continue
+        cp "$jar" "$FLINK_HOME/lib/"
```
wenjin272: In #567, we split the dist jar into two parts: common jar and thin jar. Here, only the thin jar was copied.

wenjin272 (commented):
Thanks for your work @JinkunLiu. I tried using this script and found some issues, which I've listed in the comments.

From a user experience perspective, I feel there is a significant issue: users are required to understand too many cli flag and environment variables, such as `--install-flink`, `--enable-pyflink`, `INSTALL_FLINK`, `FLINK_VERSION`, and `FLINK_AGENTS_VERSION`. Even the value of INSTALL_FLINK can be set to ask.

I'm wondering if we could retain only a single CLI flag, --default, to indicate that all operations should follow the default path, while using an interactive mode for user selection in all other cases. Users can interactively choose whether to install Flink, select the Flink version (1.20.3, 2.0.1, 2.1.1, 2.2.0), and specify the flink-agents version (0.1.1, 0.2.1, 0.3). Additionally, users can provide interactive responses regarding the location of an existing Flink installation and the target installation directory.

We can retain these environment variables; when they are detected as set, we will print a log and skip the prompt.

JinkunLiu:
Hello @wenjin272  thanks for you review

> users are required to understand too many cli flag and environment variables. 

CLI flags are reserved for non-interactive installation mode

> I'm wondering if we could retain only a single CLI flag, --default, to indicate that all operations should follow the default path

Actually, the code already does this — you can try running `bash install.sh` directly and the interactive mode for user selection will appear.

> Users can interactively choose whether to install Flink, select the Flink version (1.20.3, 2.0.1, 2.1.1, 2.2.0), and specify the flink-agents version (0.1.1, 0.2.1, 0.3).

Now the install shell provides the latest version as default. Users can use all the features we provide in the latest version.If they need special version about Flink or Flink-agents they can use cli flag.

tools/install.sh
JinkunLiu: Sorry, I may have forgotten the background — do we still need to restrict the Python version in version 0.2.1? In other words, should I modify the code constraints or change the message for users?

tools/install.sh
JinkunLiu: I see that the current directory structure is like this. The shell specifies a specific Flink version, so it will only copy the corresponding thin jar.
```shell
.
├── common
│   └── flink-agents-dist-common-0.2.1.jar
├── flink-1.20
│   └── flink-agents-dist-flink-1.20-0.2.1-thin.jar
├── flink-2.0
│   └── flink-agents-dist-flink-2.0-0.2.1-thin.jar
├── flink-2.1
│   └── flink-agents-dist-flink-2.1-0.2.1-thin.jar
└── flink-2.2
    └── flink-agents-dist-flink-2.2-0.2.1-thin.jar
```

tools/install.sh
JinkunLiu: Done, Fix the help message now

tools/install.sh
JinkunLiu: Done, I have add message and reduced the retry count

tools/install.sh
wenjin272: I think we should modify the code constraints, for flink-agents 0.2.1 only supports python 3.10 and python 3.11.

tools/install.sh
wenjin272: The thin jar contains only the code for flink-agents itself and does not include any Integrations. We need copy both the corresponding thin jar and common jar to flink lib.

wenjin272:
Hi, @JinkunLiu, thanks for your update. 

Since flink-agents provides multi-version support, and considering that some developers' clusters may not yet be compatible with the latest Flink version, I believe it might be too inflexible to always download the latest Flink version.

>  Now the install shell provides the latest version as default. Users can use all the features we provide in the latest version.If they need special version about Flink or Flink-agents they can use cli flag.

This is what I mean when I say that users need to understand many CLI flags. Why can't users be allowed to select a specific Flink version, rather than being required to specify it via a CLI flag? We provide a GUI-like installation wizard, but users are still required to manually specify certain cli flag to achieve specific goals. I believe this creates a fragmented user experience.

I think for some less frequent requirements, such as dry-run and verbose, we can simply only provide CLI flags. However, for some high-frequency requirements, I believe it would be a more user-friendly experience to allow users to select options during the installation guide, rather than requiring them to read and understand the information returned by help and setting CLI flags or environment variables. 

I think these requirements contains:
1. intall dir: we can provide two options: default vs. custom, and if user choose custom, they need input the custom dir.
2. install flink: yes vs. no, and check the FLINK_HOME when user choose no.
3. flink-version: 4 options, 1.20.3, 2.0.1, 2.1.1, 2.2.0(recommend)
4. install pyflink: yes vs. no 

After the user makes the above selections, we will then output the Install Plan.

Additionally, when running the script on my Mac, I encountered the error die: command not found. You may need to replace it with a more universally available command.

tools/install.sh
```diff
@@ -0,0 +1,1117 @@
...
+}
+
+download_file() {
+    local url="$1"
+    local output="$2"
+    if [[ -z "$DOWNLOADER" ]]; then
+        detect_downloader
+    fi
+    if [[ "$DOWNLOADER" == "curl" ]]; then
+        curl -fsSL --proto '=https' --tlsv1.2 --connect-timeout 5 --max-time 30 --retry 2 --retry-delay 1 --retry-connrefused -o "$output" "$url"
```
wenjin272: The `flink-2.2.0-bin-scala_2.12.tgz` file is 544 MB in size, and setting the max-time to 30 seconds imposes high demands on network performance.
Based on my actual tests, it takes approximately 500 seconds in my office network. I recommend setting it to 600 seconds.

wenjin272 (commented):
Hi, @JinkunLiu. I have verified the `install.sh` script locally and successfully submitted the example to the Flink standalone cluster. LGTM. Please take a look at your convenience @xintongsong.

The only issue I encountered was being unable to complete the download of Flink within the timeout period. I'm not sure whether this is a personal network problem or a common occurrence.

Additionally, now that an install script has been provided, we need to refactor the Installation section in the documentation. I think this can be done either in this PR or in a separate one.

tools/install.sh
JinkunLiu: Good catch. This is a mistake which introduced when i fix another issue. Fixed it.

tools/install.sh
```diff
@@ -0,0 +1,1117 @@
...
+}
+
+download_file() {
+    local url="$1"
+    local output="$2"
+    if [[ -z "$DOWNLOADER" ]]; then
+        detect_downloader
+    fi
+    if [[ "$DOWNLOADER" == "curl" ]]; then
+        curl -fsSL --proto '=https' --tlsv1.2 --retry 3 --max-time 600 --retry-delay 1 -- --retry-connrefused -o "$output" "$url"
```
xintongsong: There's an extra `--` that causes the download failure.

tools/install.sh
```diff
@@ -0,0 +1,1117 @@
...
+    if ! validate_python_bin "$input"; then
+        die "Provided Python is invalid or unsupported (need >=3.10 and <3.12): $input"
+    fi
+    PYTHON_BIN="$input"
+    ui_success "Using Python: $PYTHON_BIN"
+    return 0
+}
+
+show_install_plan() {
+    ui_section "Install plan"
```
xintongsong:
```suggestion
    ui_section "Installation plan"
```

xintongsong (changes requested):
Thanks for working on this, @JinkunLiu and @wenjin272 .

I did not closely check all the codes, but have tried the script out instead. Here are a few issues that I run into, which might be improved.

1. An extra `--` in line 68 causes gum downloading failure. However, the script did fail and continued to execute without gum, leading to a poor interaction experience. I think we should make gum required, and fail the script if it's not available.

2. There's no downloading progress. It takes a couple of minutes to download the Flink binary release, and during that time the script acts like dead. Might be better to show some progress, letting users know that it's actually downloading.

3. Setting up the python environment produces too many logs, which I believe is from `pip install`. Can we make the `pip install` output rolling, like it always shows the latest N lines of outputs, taking at most N lines of the screen space?

4. The plan shows `FLINK_HOME          .//flink-2.2.0` on my screen. There's an extra `/`.

5. It's like a one-way road. There's no way to go back.
- There's no way to go back to the previous step and change the settings.
- At the stage of confirming the installation plan, if you choose `No`, the script exits, without giving you a chance to modify the plan.
- When I input a wrong path of exiting flink installation, the script exits, without giving me a second chance to re-input it.

6. I tried `ctrl+c` to interrupt the script. It didn't exit the script, but instead skipped the configuration and directly started the installation.

In addition to the above experience issues, I think for this over 1k loc script, we need test cases to verify it. The `--` bug, introduced by fixing the download time limit, is a strong indicator that the script lacks protection that any simple maintenance update may break things.

wenjin272:
Hi, @JinkunLiu. I see that your updated code has addressed the first two comments from the @xintongsong review. Will the remaining comments and additional tests be updated later? Since I will start my vacation next Thursday, I may need to check whether another developer should be assigned to help track this task.

Additionally, regarding the issue of supplementary testing, I believe we can leverage a coding agent to accomplish this task.

JinkunLiu:
Thanks for the review! @xintongsong 

1. Fixed the extra `--` issue. On whether to make gum required — I referenced openclaw's design, which treats it as an optional dependency rather than a hard requirement. Since gum is just a UI enhancement and not something the script needs to work, I think keeping it optional makes more sense.
2. Added a download progress indicator for the Flink binary. Done.
3. I aware the pip install output is noisy, but haven't found a elegant solution yet. Open to suggestions if you have ideas.
4. I couldn't reproduce the double-slash issue in `FLINK_HOME`. Could you share more details about  how it was triggered?
5. Improved path input validation so users get a second chance on wrong Flink path input. As for going back after reviewing the plan — since the number of configuration fields is still relatively small, I considering leaving that out for now. WDYT?
6. The `Ctrl+C` issue has been fixed.

wenjin272:
Hi, @JinkunLiu, thanks for your update.  These updates look to me.

As you mentioned, there are three remaining issues.
1. make the pip install output rolling
2. second chance to modify install plan
3. add comprehensive tests.

Given that the code freeze (5.31) for Flink-Agents 0.3 is approaching and communication via natural language may be inefficient,  I tried using Claude Code to address the three issues mentioned above and generated some code. I verified this code locally, and it seems to be working fine. Would you mind if I directly push the above commits to this branch? Then, you can accept or reject these commits, or make modifications based on them.

JinkunLiu:
Hi, @wenjin272 
Thanks for you help, please go ahead and push the commits directly to this branch.

wenjin272:
Hi, @JinkunLiu @xintongsong , I pushed 3 commits to address the remaining review comments. PTAL.
  
  - **e8f9c0ea** — Adds an 85-test bats-core suite under `tools/test/` (unit + PATH-shim integration, including an exact-argv pin on `download_file` as a regression guard for the `--` bug) plus a CI job running on Ubuntu and macOS.
  - **525adad5** — `pip_install_quiet` helper replaces noisy `pip install` output with a single rolling status line on tty (full log surfaced on failure), `gum spin` when gum is available, and `--quiet` in CI.
  - **54ccd1b9** — Confirm prompt is now Proceed / Edit a setting / Cancel; the Edit submenu re-prompts any currently-applicable field (Flink Yes/No, version, install dir, FLINK_HOME, PyFlink Yes/No, venv dir) and re-displays the plan.

wenjin272:
Hi, @JinkunLiu @xintongsong, I have made some additional updates to this PR, which address the following issues. PTAL
  
  - **`flink-agents` JAR missing when PyFlink skipped** — download JAR directly from ASF mirror (SHA512-verified); Python no longer required for Java users.
  - **Existing Flink version not detected** — parse `lib/flink-dist-*.jar`, fall back to `bin/flink --version` with a progress hint.
  - **"Install dir" / version shown even when not installing** — plan split into read-only **Environment** + editable **Plan**; install-only rows hidden on `INSTALL_FLINK=No`.
  - **Toggling `Install Flink` didn't re-prompt** — `No → Yes` now cascades into version + dir prompts.
  - **`JAVA_HOME` looked editable** — moved to read-only Environment section.
  - **No Tab-completion on paths** — `read -e -r -p` replaces `gum input`.
  - **`export FLINK_HOME=…` line unclear** — numbered **Next steps** block with rationale.
  - **Double slashes / `~` / trailing `/` in paths** — `normalize_path` helper wired into every path entry.
  - **Ctrl+C swallowed** — global `trap die_cancelled INT`; stderr cancel msg; `confirm` honors subshell exit code.
  - **ESC quit installer / no way back** — edit menu wrapped in `( )` subshell, ESC backs out at any nesting depth.

xintongsong (approved):
Thanks for working on this, @JinkunLiu and @wenjin272 . I tried the wizard locally, and it works well. Much better experience. LGTM.

tools/install.sh
```diff
@@ -0,0 +1,1886 @@
...
+    verify_installation
+
+    echo ""
+    ui_celebrate "Apache Flink Agents installation finished!"
+    echo ""
+    ui_section "Next steps"
+    ui_success "1) Point FLINK_HOME at this install (Flink CLI and clients read it):"
+    ui_info  "       export FLINK_HOME=${FLINK_HOME}"
+    if [[ "$PYFLINK_ACTUALLY_ENABLED" -eq 1 ]]; then
+        ui_success "2) Activate the Python venv (PyFlink + flink-agents are installed there):"
```
JinkunLiu, line 1869: Should we add [configure-pythonpath](https://nightlies.apache.org/flink/flink-agents-docs-main/docs/get-started/installation/#configure-pythonpath) to here?

xintongsong: Make sense to me. Would you like to open a separate PR for that?

PR #600, opened 2026-04-01
[api][runtime][java] Add durable reconcile support with reconciler callables

joeyutong:
Linked issue: #612 

### Purpose of change

This PR adds Java-side durable reconcile support based on a reconciler callable.

With this change, a `DurableCallable` can optionally provide a reconciler via
`reconciler()`. When recovery revisits an in-flight durable call whose original
result has not yet been persisted, the runtime may invoke that reconciler to
recover a successful outcome. If the reconciler returns a result, the runtime
persists and replays that recovered result. If the reconciler throws an
exception, the exception is propagated to the caller and no recovered terminal
outcome is persisted for that durable call.

When no reconciler is provided, durable execution keeps the existing behavior:
recovery replays a previously completed result when one is available; otherwise
it executes the original `call()`.

This PR includes:

- a new `DurableCallable.reconciler()` API for Java durable calls
- runtime changes in `RunnerContextImpl` and `JavaRunnerContextImpl` to route
  sync and async durable execution through the reconcile path when a reconciler
  is provided
- state model updates in `ActionState` and `CallResult` to support `PENDING` /
  terminal transitions for reconcile-based recovery
- review-driven cleanup in the Java runtime path, including shared
  sync/async completion-only handling and clearer base-class async fallback
- unit and operator recovery tests covering recovered success, propagated
  reconcile exceptions, and recovery mismatch restart behavior

The detailed design discussion is available here:
[design discussion](https://github.com/apache/flink-agents/discussions/598)

### Tests

This change is covered by:

- `CallResultTest`: verifies status helpers and legacy null-status compatibility
- `ActionStateSerdeTest`: verifies persisted and restored durable call state,
  including `PENDING`
- `DurableExecutionContextTest`: verifies durable execution context replay and
  persistence behavior
- `RunnerContextImplDurableExecuteTest`: verifies sync durable execution,
  recovered success, propagated reconcile exceptions, and base async fallback
- `JavaRunnerContextImplDurableExecuteAsyncTest`: verifies Java async durable
  execution, recovered success, and propagated reconcile exceptions
- `ActionExecutionOperatorTest`: verifies operator recovery for recovered
  success, propagated reconcile exceptions, and mismatch restart behavior

### API

Yes.

This PR introduces Java public API changes:

- `DurableCallable` adds an optional `@Nullable Callable<T> reconciler()` hook
- `RunnerContext` JavaDoc is updated to describe reconciler-based durable
  execution semantics

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`


api/src/main/java/org/apache/flink/agents/api/context/DurableCallable.java
```diff
@@ -46,4 +50,34 @@ public interface DurableCallable<T> {
...
+     *   <li>Throw {@link ReconcileFallbackException} when recovery cannot determine a terminal
+     *       state. The runtime will continue by executing the original {@link #call()}.
+     * </ul>
+     *
+     * <p>If reconcile logic encounters query failures, temporary external-system unavailability, or
+     * any other condition where it cannot determine a terminal state, it should wrap that
+     * underlying exception in {@link ReconcileFallbackException} rather than surfacing it as a
+     * terminal failure.
+     */
+    default @Nullable Callable<T> reconcile() {
```
xintongsong:
```suggestion
    default @Nullable Callable<T> reconciler() {
```

api/src/main/java/org/apache/flink/agents/api/context/DurableCallable.java
```diff
@@ -46,4 +50,34 @@ public interface DurableCallable<T> {
      * must be JSON-serializable.
      */
     T call() throws Exception;
+
+    /**
+     * Returns an optional callable used to reconcile an in-flight durable call during recovery.
+     *
+     * <p>Returning {@code null} disables reconcile and preserves the existing completion-only
+     * durable execution semantics.
```
xintongsong: "preserves the existing  semantics" sounds more like a changelog than a API description. Users who read this JavaDoc don't need to be aware of the existing/previous semantics.

```suggestion
     * <p>Returning {@code null} disables reconcile and fallback to the default behavior, where all in-flight durable calls will be replayed during recovery.
```

api/src/main/java/org/apache/flink/agents/api/context/DurableCallable.java
```diff
@@ -46,4 +50,34 @@ public interface DurableCallable<T> {
...
+     *
+     * <p>If a reconcile callable is provided, it is invoked only when recovery revisits a matching
+     * {@code PENDING} slot for this durable call.
```
xintongsong: Please note that this is a user-facing API description. `PENDING` slot is an internal concept that should not be exposed to users. An simpler description might be something like "it is invoked during the recovery only if the execution of the callable has started but not completed before the failure".

api/src/main/java/org/apache/flink/agents/api/context/DurableCallable.java
```diff
@@ -46,4 +50,34 @@ public interface DurableCallable<T> {
...
+     *
+     * <ul>
+     *   <li>Return a result when recovery can determine that the durable call already succeeded.
+     *       The runtime will finalize the current slot as succeeded and replay that result.
+     *   <li>Throw the terminal business exception when recovery can determine that the durable call
+     *       already failed. The runtime will finalize the current slot as failed and replay that
+     *       exception.
+     *   <li>Throw {@link ReconcileFallbackException} when recovery cannot determine a terminal
+     *       state. The runtime will continue by executing the original {@link #call()}.
+     * </ul>
```
xintongsong, line 73: The described behaviors are, IMO, a bit too complex and over-constrained. It suggests that, if the callable has not succeeded before the failure, the user has only 2 options: either throws terminal business exception and fails the call, or throws the `ReconcileFallbackException` and re-execute `call()`. What if the call was partially succeeded and the users wants to resume from there rather than completely re-execute the call? Or if the call result can still be delivered but with an approach that is different from `call()`?

I'd suggest to simplify the semantics, providing only 2 options for the users:
- If the proper call result can be delivered, returns it. This will be returned from `durableExecute()` in replying the action.
- If the proper call result cannot be delivered, throws an exception. The exception will also be thrown from `durableExecute()` in replaying the action.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/CallResult.java
```diff
@@ -34,6 +34,13 @@
  */
 public class CallResult {
 
+    /** Persisted status of the durable call. */
+    public enum Status {
+        PENDING,
+        SUCCEEDED,
+        FAILED
+    }
```
xintongsong, line 43: It seems `Status` is never used outside `CallResult` (except for the tests). Shall we make it private?

runtime/src/test/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperatorTest.java
```diff
@@ -1029,6 +1042,225 @@ void testDurableExecuteAsyncExceptionRecovery() throws Exception {
...
+                    (List<StreamRecord<Object>>) testHarness.getRecordOutput();
+            assertThat(recordOutput).hasSize(1);
+            assertThat(recordOutput.get(0).getValue()).isEqualTo(15L);
+        }
+
+        assertThat(TestAgent.RECONCILABLE_CALL_COUNTER.get()).isEqualTo(1);
+        assertThat(TestAgent.RECONCILABLE_RECONCILE_COUNTER.get()).isEqualTo(1);
+    }
+
+    void testDurableExecuteReconcilableRecoveryMismatchStartsNewCall() throws Exception {
```
xintongsong, line 1135: @Test

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -245,71 +247,26 @@ public Object getActionConfigValue(String key) {
...
 
     @Override
     public <T> T durableExecuteAsync(DurableCallable<T> callable) throws Exception {
+        if (durableExecutionContext != null) {
+            Callable<T> reconcileCallable = callable.reconcile();
+            if (reconcileCallable != null) {
+                return durableExecuteWithReconcile(callable, reconcileCallable, callable::call);
+            }
+        }
+
```
xintongsong: It might not be introduced by this PR, but I'm confused here. Does `RunnerContextImpl` supports async execution at all? Seems to me it doesn't. Before this change, it simply create a supplier and calls `get()` on it.

If async execution is only supported in `JavaRunnerContextImpl`, then in this method we can simply call `durableExecute()`, with a log suggesting the fallback.

cc @Sxnan

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -245,71 +247,26 @@ public Object getActionConfigValue(String key) {
...
-        }
-
-        recordDurableCompletion(functionId, argsDigest, result, exception);
-
-        if (exception != null) {
-            throw exception;
+        if (durableExecutionContext != null) {
+            Callable<T> reconcileCallable = callable.reconcile();
+            if (reconcileCallable != null) {
+                return durableExecuteWithReconcileSync(callable, reconcileCallable);
```
xintongsong: Minor: inconsistent names:
- `durableExecuteWithReconcileSync` in `RunnerContextImpl`
- `durableExecuteAsyncWithReconcile` in `JavaRunnerContextImpl`

runtime/src/main/java/org/apache/flink/agents/runtime/context/JavaRunnerContextImpl.java
```diff
@@ -67,6 +77,30 @@ public <T> T durableExecuteAsync(DurableCallable<T> callable) throws Exception {
...
+            throw originalException;
+        }
+        return result;
+    }
+
+    private <T> T durableExecuteAsyncWithReconcile(
+            DurableCallable<T> callable, Callable<T> reconcileCallable) throws Exception {
+        return durableExecuteWithReconcile(
+                callable, reconcileCallable, () -> executeAsyncCallable(callable));
+    }
```
xintongsong, line 76: I like this deduplicating between async and sync execution. Could we do the same for `durableExecuteAsyncCompletionOnly`?

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
Sxnan: You're right, `RunnerContextImpl.durableExecuteAsync` doesn't actually support async — it just synchronously executes the callable. Real async is only in `JavaRunnerContextImpl` via `ContinuationActionExecutor`.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/CallResult.java
joeyutong: Done. `Status` is now private in `CallResult`, and the tests now assert the public state helpers (`isSuccess()`, `isFailure()`, and `isPending()`) instead of referring to the enum directly.

runtime/src/main/java/org/apache/flink/agents/runtime/context/JavaRunnerContextImpl.java
joeyutong: Done. `durableExecuteAsyncCompletionOnly` has been removed, and both sync and Java async completion-only paths now share the same `durableExecuteCompletionOnly(...)` helper with different execution callables.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
joeyutong: Done. `RunnerContextImpl.durableExecuteAsync()` now explicitly falls back to `durableExecute()` with a debug log. Real async durable execution remains only in `JavaRunnerContextImpl` via `ContinuationActionExecutor`.

api/src/main/java/org/apache/flink/agents/api/context/DurableCallable.java
joeyutong: Thanks, we simplified the reconciler contract in this direction. `ReconcileFallbackException` has been removed. The reconciler now has only two user-facing outcomes: returning a result provides a recovered successful outcome that is persisted and replayed; throwing an exception propagates that exception to the caller, and the runtime does not persist a recovered terminal outcome for that durable call.

xintongsong (approved):
Thanks for addressing my comments, @joeyutong . LGTM. Merging.

joeyutong:
Follow-up note: subsequent reconciler work refined the exception semantics introduced in this PR. Reconciler exceptions are now also persisted as terminal failures before being rethrown, and the Java/Python runtimes and docs are being aligned with that contract in follow-up changes.

Issue #541, opened 2026-02-18
[Feature] Support per-event-type configurable log levels for event log


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Summary

  Add per-event-type configurable log levels to the event log system, as identified in the 0.3 planning discussion (#516).

  ## Motivation

  The current event log system only supports binary accept/reject filtering via `EventFilter`. There is no way to:
  - Configure different log levels for different event types (e.g., VERBOSE for ChatRequestEvent, OFF  for ContextRetrievalRequestEvent)
  - Distinguish between standard and verbose logging detail (both behave identically)
  - Filter logged events by level or event type in downstream tools without parsing nested JSON

  ## Changes

  - New `EventLogLevel` enum: `OFF`, `STANDARD`, `VERBOSE`
  - Per-event-type levels configured via `EventLoggerConfig` builder and string-based config options
  - STANDARD truncates long string fields (configurable `eventLogMaxFieldLength`, default 1024); VERBOSE logs full content
  - Top-level `eventType` field in JSON output for easier downstream filtering
  - Startup validation warns about unrecognized event type names in config
  - Backward compatible: existing configs and JSON format continue to work

  ## Config Options

  | Key | Default | Description |
  |-----|---------|-------------|
  | `eventLogLevel` | `STANDARD` | Global default level (OFF, STANDARD, VERBOSE) |
  | `eventLogLevels` | (none) | Per-type overrides. Format: `ChatRequestEvent=VERBOSE,InputEvent=OFF`|
  | `eventLogMaxFieldLength` | `1024` | Max chars for string fields at STANDARD level. 0 disables truncation. |

  ## Related

  - Discussion #516 (Planning Flink Agents 0.3)

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

weiqingy:
Hi @xintongsong, thanks for the thoughtful review! Great points across the board. Let me address each:

- Design doc / discussion: Absolutely, I'll open a discussion thread with the design doc so we can align before moving to implementation.
- Broader truncation semantics for STANDARD: Agreed. Defining STANDARD as "details might be omitted to keep the logs concise" is a better semantic - it gives us room to evolve the truncation strategy (e.g., truncating long lists, capping nested depth) without breaking the contract. I'll update the design to use this framing.
[...]

PR #609, opened 2026-04-07
[runtime] Support per-event-type configurable log levels for event log

weiqingy:
Linked issue: #541 , https://github.com/apache/flink-agents/discussions/552

### Purpose of change

Add per-event-type configurable log levels (`OFF`, `STANDARD`, `VERBOSE`) to the event log system, as designed in https://github.com/apache/flink-agents/discussions/552.

**Key changes:**
- `EventLogLevel` enum (`OFF`, `STANDARD`, `VERBOSE`) with hierarchical config key inheritance (`event-log.type.<EVENT_TYPE>.level`)
- At `STANDARD` level, large content fields are truncated using configurable per-field thresholds (`max-string-length`, `max-array-elements`, `max-depth`) with JSON-parseable wrapper objects (`truncatedString`, `truncatedList`, `truncatedObject`)
- Top-level `logLevel` and `eventType` fields added to JSON log record schema for downstream filtering
- `EventFilter` removed — fully subsumed by the log level system
- `eventLogTruncatedEvents` counter metric for observability
- Python config options and E2E tests

**Config options:**
| Key | Default | Description |
|-----|---------|-------------|
| `event-log.level` | `STANDARD` | Root default log level |
| `event-log.type.<EVENT_TYPE>.level` | _(inherits)_ | Per-type override |
| `event-log.standard.max-string-length` | `2000` | Max string chars at STANDARD |
| `event-log.standard.max-array-elements` | `20` | Max array elements at STANDARD |
| `event-log.standard.max-depth` | `5` | Max nesting depth at STANDARD |

### Tests

- `EventLogLevelResolverTest` — 11 tests: exact match, hierarchy inheritance, root/built-in defaults, case insensitivity, edge cases
- `JsonTruncatorTest` — 8 tests: string/array/depth truncation, wrappers, protected fields, composition, disabled thresholds
- `FileEventLoggerTest` — 7 new tests: STANDARD truncation, VERBOSE preservation, OFF suppression, per-type override, hierarchical inheritance, JSON schema fields, backward-compatible deserialization
- Python E2E tests — 3 new: VERBOSE, OFF, STANDARD truncation

### API

- **New public API**: `EventLogLevel` enum in `api` module
- **Removed public API**: `EventFilter` interface (0.x pre-release, no known users)
- **Modified**: `EventLoggerConfig` — removed `eventFilter` field/builder method

### Documentation

- [x] `doc-needed`


runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogLevelResolver.java
```diff
@@ -0,0 +1,145 @@
...
+     * the explicit per-type level mappings, and {@code event-log.level} for the root default.
+     *
+     * @param confData the flat configuration key-value map (e.g., from {@code
+     *     AgentConfiguration.getConfData()})
+     */
+    public EventLogLevelResolver(Map<String, Object> confData) {
+        Map<String, Object> data = confData != null ? confData : Collections.emptyMap();
+
+        // Parse root default
+        Object rootValue = data.get(ROOT_LEVEL_KEY);
```
wenjin272: Shall we just use `EVENT_LOG_LEVEL.getKey()` here?

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecordJsonSerializer.java
```diff
@@ -67,15 +70,37 @@ public void serialize(EventLogRecord record, JsonGenerator gen, SerializerProvid
...
+                "logLevel", record.getLogLevel() != null ? record.getLogLevel().name() : "VERBOSE");
+        // For PythonEvent, use the Python module path as the top-level eventType
+        // (consistent with the eventType inside the event object)
+        String topLevelEventType = record.getContext().getEventType();
+        if (record.getEvent() instanceof PythonEvent) {
+            String pythonType = ((PythonEvent) record.getEvent()).getEventType();
+            if (pythonType != null) {
+                topLevelEventType = pythonType;
+            }
+        }
```
wenjin272: The logic for extracting the python event type is the same as in `FileEventLogger`. We can extract a common util function.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecord.java
```diff
@@ -31,23 +33,82 @@
...
  */
 @JsonSerialize(using = EventLogRecordJsonSerializer.class)
 @JsonDeserialize(using = EventLogRecordJsonDeserializer.class)
 public class EventLogRecord {
     private final EventContext context;
     private final Event event;
+    private final EventLogLevel logLevel;
+    private final String eventType;
+    private final transient JsonTruncator truncator;
+    private final transient Counter truncatedEventsCounter;
```
wenjin272: Maybe we can mark truncator and truncatedEventsCounter as `@Nullable`.
Additionally, these two members are actually required variables for the serialization of `EventLogRecord`, rather than part of its content. During deserialization, we have to set them to `null`. I wonder if they could be removed from within `EventLogRecord` itself.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
```diff
@@ -54,4 +54,32 @@ public class AgentConfigOptions {
     /** The config parameter specifies the unique identifier of job. */
     public static final ConfigOption<String> JOB_IDENTIFIER =
             new ConfigOption<>("job-identifier", String.class, null);
+
+    /**
+     * The global event log level controlling the default verbosity for all event types. Valid
+     * values are "OFF", "STANDARD", and "VERBOSE". Defaults to "STANDARD".
+     */
+    public static final ConfigOption<String> EVENT_LOG_LEVEL =
+            new ConfigOption<>("event-log.level", String.class, "STANDARD");
```
wenjin272: Maybe we can just use enum here
```
public static final ConfigOption<EventLogLevel> EVENT_LOG_LEVEL =
            new ConfigOption<>("event-log.level", EventLogLevel.class, EventLogLevel.STANDARD);
```

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogLevelResolver.java
```diff
@@ -0,0 +1,145 @@
...
+        this.rootDefault =
+                rootValue != null
+                        ? EventLogLevel.fromString(rootValue.toString())
+                        : BUILT_IN_DEFAULT;
```
wenjin272: Shall we just use `EVENT_LOG_LEVEL.getDefaultValue()` here?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -342,6 +342,10 @@ private void initEventLogger(StreamingRuntimeContext runtimeContext) throws Exce
             return;
         }
         eventLogger.open(new EventLoggerOpenParams(runtimeContext));
+        if (eventLogger instanceof FileEventLogger) {
```
wenjin272: Why is it necessary to specifically check the type of `eventLogger` here?

wenjin272 (commented):
Hi, @weiqingy, thanks for your work. I left some minor comments.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogLevelResolver.java
weiqingy: Good suggestion. Changed to `ConfigOption<EventLogLevel>` with `EventLogLevel.STANDARD` as the default. `AgentConfiguration.get()` already supports enum types via `ConfigurationUtils.convertValue()`, so this works end-to-end.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecordJsonSerializer.java
weiqingy: Good catch. Extracted `EventUtil.resolveEventType(Event, EventContext)` and replaced the duplicated logic in both `FileEventLogger.append()` and `EventLogRecordJsonSerializer.serialize()`.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecord.java
weiqingy: Added `@Nullable` on both fields.

Regarding removing them from `EventLogRecord` — I agree they are serialization plumbing rather than content. However, Jackson's custom serializer (`EventLogRecordJsonSerializer`) receives only the record object via `serialize(EventLogRecord record, ...)`. Since the `ObjectMapper` is shared/static, there's no clean way to pass the truncator and counter to the serializer except through the record itself. The `transient` + `@Nullable` markers signal that these are not part of the logical content. Happy to explore alternatives if you have a preferred pattern in mind.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
weiqingy: Updated the PR.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogLevelResolver.java
weiqingy: Done. Replaced `BUILT_IN_DEFAULT` with `AgentConfigOptions.EVENT_LOG_LEVEL.getDefaultValue()`. Both removed constants are now derived from the single config option definition.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
weiqingy: The `instanceof` check is needed because `setTruncatedEventsCounter()` is specific to `FileEventLogger` — it's not on the `EventLogger` interface. The truncation counter is registered centrally in `BuiltInMetrics` (alongside `numOfEventProcessed`, `numOfActionsExecuted`, etc.) so all operator metrics live in one place under the same `FlinkAgentsMetricGroupImpl` hierarchy. The operator then passes the counter reference to `FileEventLogger` after `open()`.

We considered adding the counter to the `EventLogger` interface or having the logger self-register, but both options would either leak Flink metrics concerns into the public API or break the centralized metrics pattern. The `instanceof` check keeps the interface clean while allowing `FileEventLogger` to participate in the operator's metrics.

Open to suggestions if you'd prefer a different approach here.

weiqingy:
@wenjin272 Thanks for the review! I’ve addressed your comments - could you take another look?

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecord.java
wenjin272: A simple approach might be to move the truncation from the Record serialization process into FileEventLogger.append(), like:
```
@Override
public void append(EventContext context, Event event) throws Exception {
    if (writer == null) {
        throw new IllegalStateException("FileEventLogger not initialized. Call open() first.");
    }

    String eventTypeStr = EventUtil.resolveEventType(event, context);

    // Resolve log level and skip OFF events
    EventLogLevel level =
            levelResolver != null ? levelResolver.resolve(eventTypeStr) : EventLogLevel.VERBOSE;
    if (level == EventLogLevel.OFF) {
        return;
    }

    EventLogRecord record = new EventLogRecord(context, event, level);
    // All events should be JSON serializable, since we check it when sending events to context:
    // RunnerContextImpl.sendEvent
    JsonNode tree = MAPPER.valueToTree(record);
    if (level == EventLogLevel.STANDARD && truncator != null) {
        JsonNode eventNode = tree.get("event");
        if (eventNode instanceof ObjectNode) {
            boolean truncated = truncator.truncate((ObjectNode) eventNode);
            if (truncated && truncatedEventsCounter != null) {
                truncatedEventsCounter.inc();
            }
        }
    }
    String json =
            prettyPrint
                    ? MAPPER.writerWithDefaultPrettyPrinter().writeValueAsString(tree)
                    : MAPPER.writeValueAsString(tree);
    writer.println(json);
}
```

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/JsonTruncator.java
```diff
@@ -0,0 +1,269 @@
...
+            if (child.isTextual()) {
+                JsonNode replacement = truncateString(child.textValue());
+                if (replacement != null) {
+                    node.set(fieldName, replacement);
+                    truncated = true;
+                }
+            } else if (child.isArray()) {
+                // First recurse into retained elements, then truncate the array if needed
+                truncated |= truncateArrayContents((ArrayNode) child, depth + 1);
+                JsonNode replacement = truncateArray((ArrayNode) child);
```
xintongsong: Shall we truncate the array before truncating the array contents? If an element is omitted, we should not bother to truncate its content at all.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecord.java
```diff
@@ -31,23 +35,82 @@
...
  */
 @JsonSerialize(using = EventLogRecordJsonSerializer.class)
 @JsonDeserialize(using = EventLogRecordJsonDeserializer.class)
 public class EventLogRecord {
     private final EventContext context;
     private final Event event;
+    private final EventLogLevel logLevel;
+    private final String eventType;
+    @Nullable private final transient JsonTruncator truncator;
+    @Nullable private final transient Counter truncatedEventsCounter;
```
xintongsong: Why do we need the log-level and truncator in the record?

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecord.java
weiqingy: Thanks for the concrete sketch — followed it in `c4596bc`, and combined it with Xintong's broader concern to drop `logLevel` from `EventLogRecord` as well, since it shares the same "not part of the event content" character as the truncator. `EventLogRecord` is now a pure `(context, event)` carrier, and `FileEventLogger.append()` owns truncation orchestration plus injection of the `logLevel` field on the way out.

The on-disk JSON layout is unchanged — entries still serialize as `timestamp, logLevel, eventType, event` — and the deserializer silently ignores any `logLevel` field present in older log files, so existing logs keep replaying without changes.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/JsonTruncator.java
weiqingy: Agreed — applied in `4356979`. Both call sites now truncate the array first and only recurse into the elements that survive: the `truncateObject` array branch and the symmetric path in `truncateArrayContents` for nested arrays.

Added `testRecursionSkipsDroppedTailElements`, which holds references to the dropped tail elements and asserts their content remains unmutated, so the invariant is locked in against future regressions.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/EventLogRecord.java
weiqingy: Good call. In `c4596bc` `EventLogRecord` is reduced to just `(context, event)`; `logLevel`, `eventType`, `truncator`, and `truncatedEventsCounter` are all gone from the record.

Truncation orchestration and the counter live in `FileEventLogger.append()` now, and the `logLevel` field in the on-disk JSON is emitted by the logger at write time rather than carried on the record.

wenjin272:
LGTM

wenjin272 (approved):

xintongsong (approved):
LGTM

wenjin272:
Hi, @weiqingy. Due to the refactoring of Event in #631, this PR needs to resolve some conflicts. Mainly about the SerDer for Event.

weiqingy:
Thanks @wenjin272 @xintongsong . I’ve updated the PR to resolve the branch conflicts.

Issue #612, opened 2026-04-07
[Feature] Track durable execution reconciler work


joeyutong:
### Summary

This umbrella issue tracks the durable execution reconciler work across the design discussion and follow-up pull requests.

### Design Discussion

- Discussion: #598

### Pull Requests

- Java runtime and operator recovery: #600 #624 
- Python runtime support: #614
- Documentation follow-up: 

### Scope

- Java API and runtime support for `DurableCallable#reconciler()`
- Operator recovery coverage for reconciler-backed durable calls
- Python API and runtime alignment for reconciler-backed durable execution
- Documentation updates for the reconciler design and usage

### Notes

- This issue is intended as a tracking umbrella for the work split across the discussion and PRs above.
- The detailed design rationale lives in discussion #598 and should stay there rather than being duplicated in each PR.

PR #616, opened 2026-04-09
[docs] Document durable reconciler usage

joeyutong:
Linked issue: #612

### Purpose of change

This PR includes:

- documentation for the durable reconciler contract in both Python and Java
- clarification that a reconciler can recover either a successful result or a failed outcome, and that the runtime persists and replays either terminal outcome
- usage examples for `reconciler` / `reconciler()` in the workflow agent guide, including reconciler branches that return recovered success or raise recovered failure
- clarification that reconciler-backed durable execution does not change the existing async support matrix
- documentation that reconciler-based recovery relies on the persisted action state used for exactly-once action consistency
- an additional clarification that durable-call replay and action-level replay are different: code outside durable execution is re-executed only when recovery re-enters an action that has not been recorded as completed

### Tests

Not run. This is a documentation-only change.

### API

No.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`


docs/content/docs/development/workflow_agent.md
```diff
@@ -301,6 +301,11 @@ on how to setup and configure the external action state store.
 - If a failure happens after a function completes but before its result is persisted, the call will be re-executed.
```
xintongsong:
```suggestion
- If a failure happens after a function starts but before it completes and its result is persisted, the call will be re-executed.
```

xintongsong: And we can point to the reconciler from here.

docs/content/docs/development/workflow_agent.md
```diff
@@ -301,6 +301,11 @@ on how to setup and configure the external action state store.
...
 - In Python async actions, if `ctx.durable_execute_async(...)` is not awaited, the result is not recorded and cannot be replayed.
 
+**Recovered terminal outcome with a reconciler:**
+- A durable call may optionally provide a reconciler that is used only during recovery, when the same durable call is revisited and no terminal outcome has been persisted for it yet.
```
xintongsong: It feels a bit obscure. Might be better to start with the two questions that users care most: when to use this feature (call started but not completed/persisted, framework cannot decide should replay or not), and what it can do (support custom reconciling logics).

docs/content/docs/development/workflow_agent.md
```diff
@@ -316,6 +321,28 @@ def process_input(event: InputEvent, ctx: RunnerContext) -> None:
     result = ctx.durable_execute(slow_external_call, event.input)
     ctx.send_event(OutputEvent(output=result))
 ```
+
+You can also pass an optional `reconciler` callable to recover a terminal outcome during recovery.
```
xintongsong: This might be my English issue, but TBH I'm a bit confused about what `terminal outcome` means here.

docs/content/docs/faq/faq.md
```diff
@@ -83,6 +83,8 @@ Async execution can significantly improve performance by allowing multiple opera
 
 > **Cross-language async limitation**: When using cross-language resources (e.g., calling Java integrations from Python or vice versa), async execution is not supported. Cross-language calls always execute synchronously regardless of your JDK version.
 
+Reconciler-backed durable execution is available in both Python and Java, but it only affects how durable calls recover persisted terminal outcomes. It does not change the async execution matrix above.
```
xintongsong: Not sure about the necessity of adding this. Would there be any confusions if we don't add it?

docs/content/docs/operations/deployment.md
```diff
@@ -171,6 +171,8 @@ After recovery from a checkpoint, Flink Agents reprocess events that arrived aft
 
 To ensure exactly-once action consistency, you must configure an external action state store. Flink Agents record action state in this store on a per-action basis. After recovering from a checkpoint, Flink Agents consult the external store and will not re-execute actions that were already completed. This guarantees each action is executed exactly once after recovering from a checkpoint.
 
+The same persisted action state is also used by fine-grained durable execution, including reconciler-based recovery for durable calls that need to recover a terminal outcome after failure.
```
xintongsong:
```suggestion
The same persisted action state is also used by fine-grained durable execution.
```

xintongsong (approved):
LGTM

PR #626, opened 2026-04-13
[test] Avoid Chroma model download in memory test

joeyutong:
### Purpose of change

- replace the test-only mock embedding with a deterministic local embedding
- avoid implicit ChromaDB default embedding downloads in non-Ollama test runs
- keep the unit test focused on long-term memory add/get/search behavior

### Tests

- `test_get_memory_set`
- `test_add_and_get`
- `test_search`
- `test_compact` skipped; depends on an Ollama server

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/runtime/memory/tests/test_vector_store_long_term_memory.py
+++ b/python/flink_agents/runtime/memory/tests/test_vector_store_long_term_memory.py
@@ -16,14 +16,13 @@
 # limitations under the License.
 #################################################################################
 import os
+import re
 import tempfile
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


wenjin272:
Hi, @joeyutong. According to the 0.3 roadmap, we will remove the current implementation of long-term memory and introduce mem0 backend. Consequently, this test will also be removed in the near future.

joeyutong:
@wenjin272 Thanks for your update. I'll close this PR.

Issue #627, opened 2026-04-14
[Feature] Support Fluss as ActionStateStore backend


beryllw:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

#### Summary

Currently only Kafka is supported as the external action state store backend for durable execution. This proposal adds Apache Fluss as an alternative backend.

#### Motivation

The Kafka-based implementation stores action states in an append-only log and rebuilds an in-memory HashMap on recovery by sequentially consuming the entire topic. 
Add Apache Fluss as an alternative backend with efficient recovery via bucket-level offset tracking.

#### Scope

- Add `BackendType.FLUSS` and `FlussActionStateStore` implementation
- Add Fluss connection and authentication config options
- Add integration tests against an embedded Fluss cluster
- Update user documentation (configuration and deployment guides)

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #628, opened 2026-04-14
[Feature] Support Fluss as ActionStateStore backend

beryllw:
Linked issue: #627

### Purpose of change

Uses Fluss log table + in-memory HashMap (similar to Kafka approach):
- Write: Sync append to Fluss log, then update cache
- Read: O(1) lookup from in-memory map
- Recovery: Incremental log scan from bucket offsets (recovery markers), last-write-wins
- Prune: Evict from cache only; physical cleanup via Fluss retention

Table schema: (`state_key STRING, state_payload BYTES, agent_key STRING`), distributed by `agent_key`.

- FlussActionStateStore — core implementation with divergence detection and data-loss validation
- ActionStateSerde — extracted backend-agnostic serializer (shared by Kafka & Fluss)
- ActionStateStore.BackendType.FLUSS — new enum value
- 9 new Fluss config options (bootstrap servers, database, table, buckets, SASL auth)
- Integration tests against embedded Fluss cluster (plain + SASL)
- Docs updated (configuration & deployment)

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`


joeyutong:
Thanks for working on this. I’m also looking into Fluss as an ActionStateStore backend, so I’ve been comparing different designs here.

One question on the PK table approach: with synchronous upsert in put() and no-op rebuildState(), it seems backend upsert/lookup may sit on the hot path. For new seqNums, get() will usually need to hit Fluss, while the Kafka-style design keeps get() local by rebuilding post-checkpoint state into an in-memory cache.

Did you consider using a Log Table instead, plus local cache + recovery marker + changelog rebuild? My main concern is the latency tradeoff of putting PK-table lookup/upsert directly on the action execution path.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStoreIT.java
```diff
@@ -0,0 +1,411 @@
...
+import static org.assertj.core.api.Assertions.assertThatNoException;
+
+/** Integration tests for {@link FlussActionStateStore} against an embedded Fluss cluster. */
+public class FlussActionStateStoreIT {
+
+    private static final String TEST_DATABASE = "test_flink_agents";
+    private static final String TEST_TABLE = "action_state_it";
+    private static final String TEST_KEY = "test-key";
+
+    @RegisterExtension
```
joeyutong, line 50: `FlussActionStateStoreIT` needs the required JVM opens to run reliably. When force-running it on JDK 17+, Fluss/Arrow requires `--add-opens=java.base/java.nio=ALL-UNNAMED`, but the current surefire config does not include it. Could we add this to the test configuration used by the runtime module?

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStoreIT.java
```diff
@@ -0,0 +1,411 @@
...
+
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_ACTION_STATE_DATABASE;
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_ACTION_STATE_TABLE;
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_ACTION_STATE_TABLE_BUCKETS;
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_BOOTSTRAP_SERVERS;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNoException;
+
+/** Integration tests for {@link FlussActionStateStore} against an embedded Fluss cluster. */
+public class FlussActionStateStoreIT {
```
joeyutong, line 44: This class mixes real Fluss IT coverage with cases that are closer to unit tests, such as cache lookup, serialization, and lifecycle behavior. Could we split it into a focused UT and a smaller Fluss IT?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,377 @@
...
+        return state;
+    }
+
+    /**
+     * Rebuilds in-memory state by scanning the Fluss log table from the beginning. Reads until the
+     * log is exhausted ({@link #EMPTY_POLL_THRESHOLD} consecutive empty polls). For the same state
+     * key appearing multiple times in the log, the latest record wins (last-write-wins).
+     */
+    @Override
+    public void rebuildState(List<Object> recoveryMarkers) {
```
joeyutong, line 271: `rebuildState()` currently ignores `recoveryMarkers` and always scans the Fluss log from the beginning. This misses the checkpoint boundary semantics expected by `ActionStateStore` and makes recovery depend on retained history. Could we store and consume Fluss offsets as recovery markers, similar to the Kafka backend?

joeyutong (commented):
Thanks for your work on this.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
```diff
@@ -51,6 +51,44 @@ public class AgentConfigOptions {
...
+    public static final ConfigOption<String> FLUSS_ACTION_STATE_DATABASE =
+            new ConfigOption<>("flussActionStateDatabase", String.class, "flink_agents");
+
+    /** The config parameter specifies the Fluss table name for action state. */
+    public static final ConfigOption<String> FLUSS_ACTION_STATE_TABLE =
+            new ConfigOption<>("flussActionStateTable", String.class, "action_state");
+
+    /** The config parameter specifies the number of buckets for the Fluss action state table. */
+    public static final ConfigOption<Integer> FLUSS_ACTION_STATE_TABLE_BUCKETS =
+            new ConfigOption<>("flussActionStateTableBuckets", Integer.class, 8);
```
joeyutong: Kafka defaults `kafkaActionStateTopicNumPartitions` to 64, while this new Fluss default is 8. If Fluss is intended to be an alternative backend with similar scaling and recovery characteristics, could we align this default or document why the smaller bucket count is intentional? Otherwise switching backend changes the partition/bucket parallelism quite significantly.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,377 @@
...
+                                .column("agent_key", DataTypes.STRING())
+                                .column("completed", DataTypes.BOOLEAN())
+                                .column("updated_at", DataTypes.BIGINT())
+                                .build();
+
+                int buckets = agentConfiguration.get(FLUSS_ACTION_STATE_TABLE_BUCKETS);
+                TableDescriptor descriptor =
+                        TableDescriptor.builder()
+                                .schema(schema)
+                                .distributedBy(buckets, "state_key")
```
joeyutong: Kafka uses `ActionStateKeyPartitioner` to partition by the generated key's first part, i.e. the original input key. Here Fluss is distributed by `state_key`, which includes seqNum/event/action and can spread states for the same key across buckets. Should this use `agent_key` instead to match Kafka's partitioning model? The class comment above also says `agent_key` is the bucket key.

docs/content/docs/operations/configuration.md
```diff
@@ -140,14 +140,35 @@ Here is the list of all built-in core configuration options.
...
-| `actionStateStoreBackend`           | (none)                   | String  | The config parameter specifies the backend for action state store.          |
 | `kafkaBootstrapServers`             | "localhost:9092"         | String  | The config parameter specifies the Kafka bootstrap server.                  |
 | `kafkaActionStateTopic`             | (none)                   | String  | The config parameter specifies the Kafka topic for action state.            |
 | `kafkaActionStateTopicNumPartitions`| 64                       | Integer | The config parameter specifies the number of partitions for the Kafka action state topic. |
 | `kafkaActionStateTopicReplicationFactor` | 1                     | Integer | The config parameter specifies the replication factor for the Kafka action state topic. |
+
+#### Fluss-based Action State Store
+
+Here are the configuration options for Fluss-based Action State Store. Fluss provides O(1) KV point lookup for action state, eliminating the need for in-memory state rebuilding during recovery.
+
```
joeyutong, line 163: This sentence still sounds like the earlier PK/KV-table design. The current implementation uses a log table plus in-memory cache and `rebuildState()`, so the claim about O(1) KV point lookup and no in-memory rebuilding is misleading. Could we keep this as just "Here are the configuration options for Fluss-based Action State Store."?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,377 @@
...
+    private void maybeCreateDatabaseAndTable() {
+        try (Admin admin = connection.getAdmin()) {
+            if (!admin.databaseExists(databaseName).get()) {
+                admin.createDatabase(databaseName, DatabaseDescriptor.EMPTY, true).get();
+                LOG.info("Created Fluss database: {}", databaseName);
+            }
+
+            if (!admin.tableExists(tablePath).get()) {
+                // No primaryKey() call — this creates an append-only log table in Fluss.
+                Schema schema =
```
joeyutong, line 513: Do we need all these metadata columns in the log table? In the current implementation, rebuild only reads `state_key` and `state_payload`; `seq_num`, `completed`, and `updated_at` seem either derivable from `state_key` / payload or unused by the store logic. If `agent_key` is only needed for bucket placement, maybe we can keep `state_key`, `state_payload`, and `agent_key`, and use `agent_key` as the distribution key.

runtime/pom.xml
```diff
@@ -116,6 +116,58 @@ under the License.
...
+            <version>${fluss.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-server</artifactId>
+            <version>${fluss.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
```
joeyutong, line 138: Could we double-check whether all Fluss test dependencies are required here? `fluss-client` is needed by the runtime code, and the embedded cluster test seems to need `fluss-server`, `fluss-server` test-jar, `fluss-test-utils`, and `curator-test`. But I don't see direct usage of the `fluss-common` and `fluss-rpc` test-jars from this test. If they are not needed for `FlussClusterExtension`, it would be better to keep the runtime module dependency surface smaller.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
```diff
@@ -51,6 +51,44 @@ public class AgentConfigOptions {
...
+    public static final ConfigOption<String> FLUSS_BOOTSTRAP_SERVERS =
+            new ConfigOption<>("flussBootstrapServers", String.class, "localhost:9123");
+
+    /** The config parameter specifies the Fluss database for action state. */
+    public static final ConfigOption<String> FLUSS_ACTION_STATE_DATABASE =
+            new ConfigOption<>("flussActionStateDatabase", String.class, "flink_agents");
+
+    /** The config parameter specifies the Fluss table name for action state. */
+    public static final ConfigOption<String> FLUSS_ACTION_STATE_TABLE =
+            new ConfigOption<>("flussActionStateTable", String.class, "action_state");
```
joeyutong:
```suggestion
            new ConfigOption<>("flussActionStateTable", String.class, "action_states");
```

docs/content/docs/operations/configuration.md
```diff
@@ -140,14 +140,35 @@ Here is the list of all built-in core configuration options.
...
+#### Fluss-based Action State Store
+
+Here are the configuration options for Fluss-based Action State Store.
+
+| Key                          | Default          | Type    | Description                                                                              |
+|------------------------------|------------------|---------|------------------------------------------------------------------------------------------|
+| `flussBootstrapServers`      | "localhost:9123" | String  | The Fluss bootstrap servers address.                                                     |
+| `flussActionStateDatabase`   | "flink_agents"   | String  | The Fluss database name for storing action state.                                        |
+| `flussActionStateTable`      | "action_state"   | String  | The Fluss table name for storing action state.                                           |
```
joeyutong:
```suggestion
| `flussActionStateTable`      | "action_states"   | String  | The Fluss table name for storing action state.                                           |
```

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_BOOTSTRAP_SERVERS;
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_SASL_JAAS_CONFIG;
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_SASL_MECHANISM;
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_SASL_PASSWORD;
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_SASL_USERNAME;
+import static org.apache.flink.agents.api.configuration.AgentConfigOptions.FLUSS_SECURITY_PROTOCOL;
+import static org.apache.flink.agents.runtime.actionstate.ActionStateUtil.generateKey;
+
+/**
+ * An implementation of {@link ActionStateStore} that uses an Apache Fluss log table as the backend.
```
joeyutong, line 73: Could we simplify this class-level comment to focus on the current contract instead of the implementation history? The `HashMap` sentence and the “Compared to the previous KV-table implementation” section feel a bit redundant. Also, `agent_key` is the original input key, so “same agent” should probably be “same input key”.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+        try (Admin admin = connection.getAdmin()) {
+            if (!admin.databaseExists(databaseName).get()) {
+                admin.createDatabase(databaseName, DatabaseDescriptor.EMPTY, true).get();
+                LOG.info("Created Fluss database: {}", databaseName);
+            }
+
+            if (!admin.tableExists(tablePath).get()) {
+                // No primaryKey() call — this creates an append-only log table in Fluss.
+                Schema schema =
+                        Schema.newBuilder()
```
joeyutong, line 514: Could we also define constants for the column names?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+                tableName);
+    }
+
+    @Override
+    public void put(Object key, long seqNum, Action action, Event event, ActionState state)
+            throws Exception {
+        String stateKey = generateKey(key, seqNum, action, event);
+        byte[] payload = OBJECT_MAPPER.writeValueAsBytes(state);
+
+        GenericRow row = new GenericRow(NUM_COLUMNS);
```
joeyutong: Nit: this could be simplified with `GenericRow.of(...)`. The row has only three fixed fields, so constructing it in one expression may be easier to read than allocating by column count and setting each field separately.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+        // Synchronous write ensures the record is durable before returning.
+        writer.append(row).get();
+        actionStates.put(stateKey, state);
+
+        LOG.debug("Stored action state: key={}, isCompleted={}", stateKey, state.isCompleted());
+    }
+
+    @Override
+    public ActionState get(Object key, long seqNum, Action action, Event event) throws Exception {
+        String stateKey = generateKey(key, seqNum, action, event);
```
joeyutong, line 217: Could we also handle divergence in `get()`? `KafkaActionStateStore#get()` calls `checkDivergence(key, seqNum)` and removes cached states beyond the requested sequence when divergence is detected. The Fluss backend currently only does a direct map lookup, which makes the two backends behave differently during divergent replay.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    /**
+     * Rebuilds in-memory state by scanning the Fluss log table. If recovery markers are provided,
+     * computes the minimum offset per bucket across all markers and subscribes from those offsets.
+     * Otherwise, subscribes from the beginning. Reads until the log is exhausted ({@link
+     * #EMPTY_POLL_THRESHOLD} consecutive empty polls). For the same state key appearing multiple
+     * times in the log, the latest record wins (last-write-wins).
+     *
+     * <p>Note: Unlike the Kafka implementation which skips rebuild on empty markers, this
```
joeyutong: Should we keep this behavior aligned with the Kafka backend? When `recoveryMarkers` or the derived bucket offsets are empty, Kafka skips rebuild because there is no checkpointed external-store position to recover from. Scanning the Fluss log from the beginning can restore state outside the Flink checkpoint boundary, for example on a fresh start or when the table contains retained data from a previous run. It also hides missing-marker issues and makes recovery cost depend on the full retained log.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+                    scanner.subscribeFromBeginning(b);
+                }
+            } else {
+                for (Map.Entry<Integer, Long> entry : bucketOffsets.entrySet()) {
+                    scanner.subscribe(entry.getKey(), entry.getValue());
+                }
+            }
+
+            int emptyPollCount = 0;
+            while (emptyPollCount < EMPTY_POLL_THRESHOLD) {
```
joeyutong: Could we use explicit stopping offsets instead of consecutive empty polls to terminate rebuild? AFAIK, rebuild can define a precise replay window: read from the marker offset up to the **latest offset** captured at rebuild start. 
`admin.listOffsets(tablePath, buckets, new OffsetSpec.LatestSpec()).all().get()`

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+ * state_key     STRING  -- Generated by ActionStateUtil.generateKey()
+ * state_payload BYTES   -- ActionState JSON serialized bytes
+ * agent_key     STRING  -- Original key; used as bucket key so all records for the same
+ *                          agent are co-located in the same bucket
+ * </pre>
+ */
+public class FlussActionStateStore implements ActionStateStore {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FlussActionStateStore.class);
+    private static final ObjectMapper OBJECT_MAPPER = createObjectMapper();
```
joeyutong: Could we reuse the existing `ActionStateKafkaSeder` serde logic here?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+    @Override
+    public Object getRecoveryMarker() {
+        int numBuckets = table.getTableInfo().getNumBuckets();
+
+        try (Admin admin = connection.getAdmin()) {
+            List<Integer> buckets = new ArrayList<>();
+            for (int b = 0; b < numBuckets; b++) {
+                buckets.add(b);
+            }
+            return new HashMap<>(
```
joeyutong: Nit: do we need the extra `new HashMap<>(...)` here? Since the method returns `Object`, the map from `listOffsets(...).all().get()` can be returned directly unless a defensive copy is intentional.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+                                        return stateSeqNum <= seqNum;
+                                    }
+                                } catch (NumberFormatException e) {
+                                    LOG.warn("Failed to parse seqNum from state key: {}", stateKey);
+                                }
+                            }
+                            return false;
+                        });
+
+        LOG.debug("Pruned in-memory state for key: {} up to seqNum: {}", key, seqNum);
```
joeyutong: Nit: the debug logs at the beginning and end of this method are almost identical

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    @Override
+    public void close() throws Exception {
+        Exception firstException = null;
```
joeyutong: Could we simplify this close logic? Given that each `put()` waits for `writer.append(row).get()`, I am not sure an extra `writer.flush()` is needed here. The `firstException` handling also feels more complex than necessary. Maybe we can just close the table and ensure the connection is closed in `finally`, letting any close exception propagate normally.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStoreTest.java
```diff
@@ -0,0 +1,136 @@
...
+import java.nio.charset.StandardCharsets;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Unit tests for {@link FlussActionStateStore} serialization and in-memory behavior. These tests
+ * verify the ObjectMapper configuration and ActionState serialization round-trips without requiring
+ * a real Fluss cluster.
+ */
+public class FlussActionStateStoreTest {
```
joeyutong, line 45: This test is currently closer to an ActionState serde test than a `FlussActionStateStore` test which shoud focus on store-level behavior, such as `put`, `get`, `rebuildState`, `pruneState` etc.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,412 @@
...
+        byte[] payload = OBJECT_MAPPER.writeValueAsBytes(state);
+
+        GenericRow row = new GenericRow(NUM_COLUMNS);
+        row.setField(COL_STATE_KEY, BinaryString.fromString(stateKey));
+        row.setField(COL_STATE_PAYLOAD, payload);
+        row.setField(COL_AGENT_KEY, BinaryString.fromString(key.toString()));
+
+        // Append to Fluss log for durability, then update in-memory cache.
+        // Synchronous write ensures the record is durable before returning.
+        writer.append(row).get();
```
joeyutong, line 209: For the current synchronous `put()` semantics, should we set `WRITER_BATCH_TIMEOUT` to `0`? Since `put()` calls `writer.append(row).get()`, each put waits for the append future to complete. With a non-zero batch timeout, the batching delay becomes part of the user-visible put latency.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,489 @@
...
+
+    /**
+     * Rebuilds in-memory state by scanning the Fluss log table. If recovery markers are provided,
+     * computes the minimum offset per bucket across all markers and subscribes from those offsets.
+     * Otherwise, skips rebuild since there is no checkpointed position to recover from. Reads from
+     * the start offset up to the latest offset captured at rebuild start. For the same state key
+     * appearing multiple times in the log, the latest record wins (last-write-wins).
+     */
+    @Override
+    public void rebuildState(List<Object> recoveryMarkers) {
```
joeyutong, line 271: Could we split `rebuildState()` into a few smaller helpers? It currently handles marker merging, effective offset calculation, scanner subscription, bounded polling, and row deserialization in one method. Extracting pieces such as `mergeRecoveryMarkerOffsets`, `subscribeEffectiveOffsets`, and the scan/replay loop would make the recovery boundary easier to review and reduce the risk of subtle offset bugs.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,489 @@
...
+                    if (endOffset == null) {
+                        continue;
+                    }
+                    // Track the highest offset seen in this batch (including skipped records
+                    // beyond endOffset) so that we can detect when the bucket has been fully
+                    // consumed even if the last records are past our target window.
+                    long lastSeenOffset = -1;
+                    for (ScanRecord record : records.records(bucket)) {
+                        lastSeenOffset = record.logOffset();
+                        if (record.logOffset() >= endOffset) {
```
joeyutong: Once we see a `endOffset` record for a bucket, shouldn't we stop scanning that bucket instead of continuing through the rest of the records? Offsets within a bucket should be ordered, and records at or beyond `endOffset` are outside the recovery window. Using `break` here would make the bounded-replay semantics clearer and avoid iterating over records that should never be restored.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,489 @@
...
+                            actionStates.put(stateKey, state);
+                        } catch (Exception e) {
+                            LOG.warn(
+                                    "Failed to deserialize action state for key: {}, skipping",
+                                    stateKey,
+                                    e);
+                        }
+                    }
+                    // Remove bucket if the highest seen offset has reached or passed the end
+                    if (lastSeenOffset + 1 >= endOffset) {
```
joeyutong: When a bucket reaches its `endOffset`, should we also unsubscribe it from the scanner? Removing it from `remainingBuckets` prevents further replay, but the scanner is still subscribed to that bucket and may keep fetching records beyond the recovery window while other buckets are still being rebuilt. Calling `scanner.unsubscribe(bucket.getBucket())` here would make the bounded-replay boundary explicit and avoid unnecessary fetches for completed buckets.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,489 @@
...
+        try (LogScanner scanner = table.newScan().createLogScanner()) {
+            // Track which buckets still need to be consumed
+            Map<Integer, Long> remainingBuckets = new HashMap<>();
+            for (Map.Entry<Integer, Long> entry : bucketStartOffsets.entrySet()) {
+                int bucket = entry.getKey();
+                long startOffset = entry.getValue();
+                Long endOffset = bucketEndOffsets.get(bucket);
+
+                // Bucket referenced in recovery marker does not exist in current table
+                if (endOffset == null) {
```
joeyutong: Could we fail recovery instead of skipping buckets when the recovery window is invalid or incomplete? Cases like 
1. `endOffset == null`, 
2. `endOffset < startOffset`,
3. `earliestOffset > startOffset` 
all indicate that the marker no longer matches the available log range. Treating them as no-op or partially rebuilding from `earliestOffset` can silently drop action state from the checkpoint recovery window. Since this store is on the correctness path, I think these cases should fail fast rather than continue with an incomplete cache.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,489 @@
...
+                            continue;
+                        }
+                        InternalRow row = record.getRow();
+                        String stateKey = row.getString(COL_STATE_KEY).toString();
+                        byte[] payload = row.getBytes(COL_STATE_PAYLOAD);
+                        try {
+                            ActionState state = ActionStateSerde.deserialize(payload);
+                            actionStates.put(stateKey, state);
+                        } catch (Exception e) {
+                            LOG.warn(
```
joeyutong: Should deserialization failures fail the rebuild instead of skipping the record? The Fluss log is the durable source for action state recovery. If a record in the recovery window cannot be restored, continuing with an incomplete cache may change replay behavior and cause actions to run again. I think rebuild should fail loudly on corrupted or incompatible state payloads.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/TestAction.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.flink.agents.api.Event;
+import org.apache.flink.agents.api.InputEvent;
+import org.apache.flink.agents.api.context.RunnerContext;
+import org.apache.flink.agents.plan.JavaFunction;
+import org.apache.flink.agents.plan.actions.Action;
+
+import java.util.List;
+
+/** A no-op {@link Action} for use in action state store tests. */
+public class TestAction extends Action {
```
joeyutong:
```suggestion
public class NoOpAction extends Action {
```

xintongsong (approved):
LGTM, except for the code conflicts.

@Sxnan, as one of the contributor of the original Kafka action state store, could you also help check on this PR?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,496 @@
...
+            for (TableBucket bucket : records.buckets()) {
+                Long endOffset = remainingBuckets.get(bucket.getBucket());
+                long lastSeenOffset = replayRecords(records.records(bucket), endOffset);
+                if (lastSeenOffset + 1 >= endOffset) {
+                    remainingBuckets.remove(bucket.getBucket());
+                    scanner.unsubscribe(bucket.getBucket());
+                }
+            }
+        }
+    }
```
Sxnan, line 427: The recovery loop has no timeout. If `scanner.poll()` keeps returning empty (network blips, broker hiccups, metadata refresh, etc.), `remainingBuckets` never shrinks and we loop forever.

Suggest adding a deadline and throwing on timeout so Flink can restart the job. Avoid a simple break-on-empty — an empty poll doesn't necessarily mean we've reached the end offset, so it would silently drop recovery records.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
```diff
@@ -51,6 +51,48 @@ public class AgentConfigOptions {
...
+    public static final ConfigOption<String> FLUSS_BOOTSTRAP_SERVERS =
+            new ConfigOption<>("flussBootstrapServers", String.class, "localhost:9123");
+
+    /** The config parameter specifies the Fluss database for action state. */
+    public static final ConfigOption<String> FLUSS_ACTION_STATE_DATABASE =
+            new ConfigOption<>("flussActionStateDatabase", String.class, "flink_agents");
+
+    /** The config parameter specifies the Fluss table name for action state. */
+    public static final ConfigOption<String> FLUSS_ACTION_STATE_TABLE =
+            new ConfigOption<>("flussActionStateTable", String.class, "action_states");
```
Sxnan: `FLUSS_ACTION_STATE_TABLE` defaults to `"action_states"` and is not validated at use site (see `FlussActionStateStore.java#L122-L124`). With this combination, multiple jobs sharing a Fluss cluster will silently end up writing to the same state table and corrupting each other.

See `KAFKA_ACTION_STATE_TOPIC` for the established pattern: default `null` + `Preconditions.checkArgumentNotNull` at construction. Suggest the same here.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,496 @@
...
+            LogScanner scanner,
+            Map<Integer, Long> bucketStartOffsets,
+            Map<Integer, Long> bucketEndOffsets,
+            Map<Integer, Long> bucketEarliestOffsets) {
+        Map<Integer, Long> remainingBuckets = new HashMap<>();
+        for (Map.Entry<Integer, Long> entry : bucketStartOffsets.entrySet()) {
+            int bucket = entry.getKey();
+            long startOffset = entry.getValue();
+            long endOffset = bucketEndOffsets.get(bucket);
+            long earliestOffset = bucketEarliestOffsets.get(bucket);
```
Sxnan, line 372:
```java
long endOffset = bucketEndOffsets.get(bucket);
long earliestOffset = bucketEarliestOffsets.get(bucket);
```

`bucketStartOffsets` comes from recovery markers (old checkpoint snapshot), while `bucketEndOffsets` / `bucketEarliestOffsets` reflect the current table's bucket set. If the marker references a bucket that no longer exists (e.g. user changed `flussActionStateTableBuckets` after the table was created, or the table was recreated), `Map.get()` returns `null` and auto-unboxing throws NPE with no context.

Suggest a null check that throws `IllegalStateException` with the bucket id and current bucket set in the message.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,496 @@
...
+            String password = agentConfiguration.get(FLUSS_SASL_PASSWORD);
+            if (password != null) {
+                flussConf.setString(CLIENT_SASL_JAAS_PASSWORD, password);
+            }
+        }
+
+        this.connection = ConnectionFactory.createConnection(flussConf);
+        maybeCreateDatabaseAndTable();
+        this.table = connection.getTable(tablePath);
+        this.writer = table.newAppend().createWriter();
```
Sxnan:
```java
this.connection = ConnectionFactory.createConnection(flussConf);
maybeCreateDatabaseAndTable();              // throws -> connection leaks
this.table = connection.getTable(tablePath);
this.writer = table.newAppend().createWriter();
```

If any of the last three steps throws, the already-opened `connection` (and possibly `table`) will never be released. Suggest wrapping the post-`createConnection` work in try/catch and closing what was acquired before rethrowing.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -0,0 +1,496 @@
...
+                                        return stateSeqNum > seqNum;
+                                    }
+                                } catch (NumberFormatException e) {
+                                    LOG.warn(
+                                            "Failed to parse sequence number from state key: {}",
+                                            stateKey);
+                                }
+                                return false;
+                            });
+        }
```
Sxnan, line 224: Two issues here. Both also exist in `KafkaActionStateStore` (L163-L184) — worth fixing in both places.

1. **Cross-agent-key removal**: `removeIf` only compares `seqNum`, ignoring `parts.get(0)`. When divergence is detected for agent A at `seqNum=N`, every entry for agent B with `seqNum > N` is also wiped. Same file's `pruneState` (L425-L442) gets this right via `stateKey.startsWith(key.toString() + "_")`.

2. **Wrong catch type**: `ActionStateUtil.parseKey` throws `IllegalArgumentException` on bad format (see `ActionStateUtil.java:61`), not `NumberFormatException`. So any agent key containing `_` (which produces a 5-part split) will crash `get()` instead of being silently skipped. Suggest catching `Exception` like `pruneState` does at L437.

Sxnan (commented):
Thanks for the PR! I left a few comments belows.

Sxnan (commented):
(comments deduplicated)

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
joeyutong: @Sxnan Good catch. The NPE is a real issue here, but I think the bigger problem is the bucket layout mismatch behind it.
If the table bucket count changes, the offsets in the recovery marker are no longer safe to apply to the current table. Some buckets may be missing, or existing bucket ids may refer to a different key distribution.
Maybe rebuild should validate the recovery marker bucket set against the current table buckets before applying offsets. If they do not match, it would be safer to fail fast with a clear error.

Sxnan (approved):
Thanks for the update! LGTM

PR #631, opened 2026-04-15
Cross Language Unified Event

addu390:
Linked issue: #424

### Purpose of change

Add unified event support. Users can create events with a `type` string and `attributes` map instead of defining subclasses. Enables string-based event routing and JSON-based cross-language transport. Migrates/removes backward compatible with existing subclassed events.

### Tests

New unit tests in Java (`EventTest`, `EventLogRecordJsonSerdeTest`) and Python (`test_event`, `test_decorators`, `test_agent_plan`, `test_local_execution_environment`) covering unified event creation, serialization, routing, and backward compatibility.

### API

Yes. `Event` class (Java/Python) gains `type`, `attributes`, `getType()`/`get_type()`. `@Action` annotation gains `listenEventTypes()`. 

More details about the design in the issue #424

### Documentation

- [x] `doc-needed`
- [ ] `doc-not-needed`
- [ ] `doc-included`


addu390:
Hi @wenjin272, I closed the PR #561 and opened this.
I'll clean-up the PR a bit more, but up for a review otherwise.

Do take a first pass when you get a chance.

api/src/main/java/org/apache/flink/agents/api/annotation/Action.java
```diff
@@ -31,23 +31,40 @@
...
 public @interface Action {
     /**
-     * List of event types that this action should respond to. At least one event type must be
-     * specified.
+     * List of event classes that this action should respond to.
      *
      * @return Array of Event classes that this action listens to
      */
-    Class<? extends Event>[] listenEvents();
+    Class<? extends Event>[] listenEvents() default {};
```
wenjin272: On the Python side, actions only support listening to event string identifiers; on the Java side, we should align with this behavior.

runtime/src/main/java/org/apache/flink/agents/runtime/python/context/PythonRunnerContextImpl.java
```diff
@@ -48,16 +47,18 @@ public PythonRunnerContextImpl(
...
+    /**
+     * Sends an event from Python to Java.
+     *
+     * <p>All Python events are serialized as JSON. The event is deserialized into a Java {@link
+     * Event} and dispatched normally.
+     *
+     * @param eventJson JSON string with at least a {@code "type"} field
+     * @throws IOException if JSON parsing fails
+     */
+    public void sendUnifiedEvent(String eventJson) throws IOException {
```
wenjin272: I think the name `sendUnifiedEvent` is a bit ambiguous. I suggest renaming it to `sendEventJson` or `sentPythonEvent`.

api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -19,36 +19,76 @@
...
     }
 
     public UUID getId() {
         return id;
     }
 
+    /** Returns the event type string used for routing. */
+    @JsonIgnore
+    public String getType() {
+        return type != null ? type : this.getClass().getName();
```
wenjin272: I think we can directly require that the `type` is not null, which would simplify the implementation of Event.

api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -19,36 +19,76 @@
...
-public abstract class Event {
+public class Event {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
     private final UUID id;
+    private final String type;
     private final Map<String, Object> attributes;
     /** The timestamp of the source record. */
     private Long sourceTimestamp;
```
wenjin272, line 44: I think we can remove this field to keep the align between python `Event` and java `Event`

python/flink_agents/api/events/chat_event.py
```diff
@@ -36,9 +41,57 @@ class ChatRequestEvent(Event):
...
+        return ChatRequestEvent(
+            model=event.attributes["model"],
+            messages=messages,
+            output_schema=output_schema_raw,
+        )
+
+    @property
+    def model(self) -> str:
+        """Return the chat model name."""
+        return self.attributes["model"]
```
wenjin272: Shall we `return self.get_attr("model")`?

python/flink_agents/api/events/chat_event.py
wenjin272, line 1: We need to ensure consistency between the chat event on the Python and Java sides, so we must make the same changes on the Java side. Other built-in events also need to be aligned.

python/flink_agents/api/events/chat_event.py
```diff
@@ -36,9 +41,57 @@ class ChatRequestEvent(Event):
...
+        messages: List[ChatMessage],
+        output_schema: OutputSchema | None = None,
+    ) -> None:
+        """Create a ChatRequestEvent."""
+        super().__init__(
+            type=ChatRequestEvent.EVENT_TYPE,
+            attributes={
+                "model": model,
+                "messages": messages,
+                "output_schema": output_schema,
```
wenjin272, line 58: We need to ensure consistency between the built-in events on the Python and Java sides, so we must make the same changes on the Java side.

python/flink_agents/api/events/event.py
```diff
@@ -30,27 +30,71 @@
...
-        event content. This ensures events with identical content have the same
-        ID, which is critical for ActionStateStore divergence detection.
+        event content.
+    type : str | None
+        Event type string used for routing.
+    attributes : Dict[str, Any]
+        Key-value properties for the event data.
     """
 
+    _type_registry: ClassVar[Dict[str, Type["Event"]]] = {}
```
wenjin272: It appears that this variable is intended to handle the deserialization of subclasses, but it is not currently in use. I think the Event class doesn't need to handle this; deserialization of subclass objects can be managed within the subclass's from_event method. Because subclasses may have certain non-trivial fields that require special handling during deserialization.

python/flink_agents/api/tests/test_decorators.py
```diff
@@ -25,25 +23,25 @@
 
 
 def test_action_decorator() -> None:  # noqa D103
-    @action(InputEvent)
+    @action("_input_event")
```
wenjin272: I think we can use InputEvent.EVENT_TYPE, and the same applies to other places where @action is used.

python/flink_agents/runtime/flink_runner_context.py
```diff
@@ -289,18 +224,23 @@ def set_long_term_memory(self, ltm: InternalBaseLongTermMemory) -> None:
...
+        event_json = event.model_dump_json()
         try:
-            self._j_runner_context.sendEvent(class_path, event_bytes, event_json_str)
+            self._j_runner_context.sendUnifiedEvent(event_json)
         except Exception as e:
-            err_msg = "Failed to send event " + class_path + " to runner context"
+            err_msg = (
+                "Failed to send event '"
+                + event.get_type()
+                + "' to runner context"
```
wenjin272: I think we can also add the event json in `err_msg`.

python/flink_agents/runtime/python_java_utils.py
```diff
@@ -46,36 +45,44 @@
...
+
+    Called from Java via PythonActionExecutor to convert a Java Event
+    (serialized as JSON) into a Python Event for action dispatch.
+    If a subclass is registered for the event's type string, the event
+    is automatically reconstructed as that subclass via ``from_event``.
+    """
+    event = Event.from_json(event_json)
+    event_class = Event._type_registry.get(event.type)
+    if event_class is not None and hasattr(event_class, "from_event"):
+        return event_class.from_event(event)
```
wenjin272: Similar to my comment in `Event`, the framework does not need to be aware of the subclass types of `Event`. This would require the framework to handle many details of serialization and deserialization. 
Fro subclass types of `Event`, user need use `SubClass.from_event()` to reconstruct the specific event object in action, like what you do in built-in tool call action and chat action.

wenjin272 (commented):
Hi, @addu390, thanks for you work. Refactoring the Event will have a significant impact on the framework; therefore, this work requires careful design and attention to numerous details. Thanks Again.

I think though some details still need to be addressed, the current Python event design aligns with my expectations, including
* action listen to event.type
* json based DeSer
* remove `PythonEvent` in java, unified in `Event`
* Subclasses of `Event` provide `from_event` method to validate and reconstruct specific event object.

However, the Java event design does not appear to be aligned with the Python approach. In fact, after completing the Event refactoring, we will continue to support cross-language usage of Actions. This means a Python event sent by a Python action may be delivered to a Java action. Therefore, Python events and Java events must maintain consistency in both design and usage.

By the way, there are some conflicts between the PR and the main branch, likely because the main branch was recently reformatted.

addu390:
@wenjin272 Agreed on consistency between Java and Python, the reasoning was that it was quite a larger change, so, was considering having a follow-up PR. 
However, It's a fair call-out, let me actually just do that too in this PR, doing so covers most of the review comments too.

api/src/main/java/org/apache/flink/agents/api/Event.java
addu390: `sourceTimestamp` seems to be used in runtime for timestamp propagation (`ActionExecutionOperator`, `RunnerContextImpl`, `JavaActionTask`, etc.). Safe to remove?

I could however, specifically not use in the cross-language event contract, without any repercussions.

addu390:
@wenjin272 Thanks again for the review! I’ve addressed the broader concern of consistency between Python and Java implementations, along side other review comments. Please take another look.

api/src/main/java/org/apache/flink/agents/api/event/ChatRequestEvent.java
```diff
@@ -27,31 +28,37 @@
...
     public String getModel() {
-        return model;
+        return (String) getAttr("model");
     }
 
+    @JsonIgnore
+    @SuppressWarnings("unchecked")
     public List<ChatMessage> getMessages() {
-        return messages;
+        return (List<ChatMessage>) getAttr("messages");
```
wenjin272, line 95: In the python `ChatRequestEvent`, we deserialize the `messages` field in `from_event` method. But the java `Event` is missing corresponding method. This may cause `ClassCastException`.

api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -20,35 +20,59 @@
...
 
+import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
 
 /** Base class for all event types in the system. */
-public abstract class Event {
+public class Event {
```
wenjin272, line 32: The python `Event` contains a class method `from_event`. In this method, we will perform field validation and deserialization. However, Java's event lacks the corresponding method.

api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -73,17 +97,34 @@ public void setSourceTimestamp(long timestamp) {
...
+     *
+     * @param json the JSON string to deserialize
+     * @return the deserialized Event
+     * @throws IOException if JSON parsing fails
+     * @throws IllegalArgumentException if the 'type' field is missing or empty
+     */
+    public static Event fromJson(String json) throws IOException {
+        Event event = MAPPER.readValue(json, Event.class);
+        if (event.type == null || event.type.isEmpty()) {
+            throw new IllegalArgumentException("Event JSON must contain a 'type' field.");
```
wenjin272: I think we can move this validation to the construct method annotated `@JsonCreator`. This ensures that all code paths constructing an Event will enforce this validation.

api/src/main/java/org/apache/flink/agents/api/EventContext.java
```diff
@@ -25,27 +25,44 @@
...
 public class EventContext {
-    // Type of the event, the class name of the event
+    /** The routing key for the event, matching the {@code EVENT_TYPE} constant or type string. */
     private final String eventType;
+
+    /**
+     * The fully qualified Java class name used for deserialization. For unified events (no
+     * subclass), this is {@code "org.apache.flink.agents.api.Event"}.
+     */
+    private final String eventClass;
```
wenjin272: I think `EventContext` does not need to know the specific event class. 

This information will enable us to restore the specific subclass type when reconstructing events from the event log. However, I believe there is no requirement to deserialize events from the event log. Furthermore, being aware of specific event classes will make the framework's implementation more complex. 

I think that if users utilize event subclasses instead of the base event directly, they should be responsible for handling event deserialization, specifically by uniformly using `from_event`/`fromEvent`.

wenjin272 (commented):
Hi, @addu390. I left some comments, mainly because Java's Event lacks the fromEvent method.

addu390:
@wenjin272 I have addressed the comments related to `fromEvent`, can you take an other pass when you get chance?

wenjin272:
> @wenjin272 I have addressed the comments related to `fromEvent`, can you take an other pass when you get chance?

Sorry for the delay in reviewing — I'll get to this PR today or tomorrow.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
wenjin272: I think this should be changed to ChatRequestEvent.fromEvent(event). The same applies to other actions, including built-in actions, as well as actions in tests and examples. I've checked, and it seems that everything on the Python side has been handled.

wenjin272 (commented):
Hi, @addu390, thanks for your work.

Personally, I think this PR is nearly complete. I left one remaining comment. However, this PR has far-reaching implications, so I’ll also ask others to review it. 

The current commit history is somewhat disorganized, which may make it difficult for others to review. If possible, please reorganize the commits. For example, organize the changes into two commits: one for event refactoring and another for test and example adaptations. Python and Java can also be separated.

addu390:
@wenjin272 Thanks again for the quick reviews, I have addressed the comments and organized the commits, do have a look again! And agreed on the far-reaching implications, @xintongsong could you do a review as well?

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
```diff
@@ -41,7 +41,7 @@ public static Action getToolCallAction() throws Exception {
                         ToolCallAction.class,
                         "processToolRequest",
                         new Class[] {ToolRequestEvent.class, RunnerContext.class}),
```
wenjin272: Here, we constrain the first parameter of `processToolRequest` to be of type `ToolRequestEvent`. However, in cross-language invocations, only generic Event objects are passed rather than specific subclasses, which results in an `argument type mismatch` error.

I think this can be aligned with `ChatModelAction` by changing the parameter type to `Event` and calling `ToolRequestEvent.fromEvent(event)` within the function to convert the `Event` into a `ToolRequestEvent`.

api/src/main/java/org/apache/flink/agents/api/event/ChatRequestEvent.java
```diff
@@ -18,40 +18,75 @@
...
+        if (rawMessages != null) {
+            for (Object m : rawMessages) {
+                if (m instanceof ChatMessage) {
+                    messages.add((ChatMessage) m);
+                } else if (m instanceof Map) {
+                    messages.add(MAPPER.convertValue(m, ChatMessage.class));
+                }
+            }
+        }
+        return new ChatRequestEvent(model, messages, event.getAttr("output_schema"));
```
wenjin272: Here, we do not copy the event's `id`. Therefore, `ChatRequestEvent` generates a new random `id`, which causes the same `ChatRequestEvent` to have different `id` on the sender and receiver sides.

This issue also exists for other built-in Event subclasses.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
wenjin272: The purpose of ActionState is to enable action-level recovery after a job failover. However, the `HashMap` storing attributes does not guarantee strict consistency in key order before and after a JVM restart, which may lead to inconsistent UUIDs.

After the Event refactoring, it seems that there is no distinction made for `InputEvent` here. I think we can unify it into
```
return String.valueOf(
                    UUID.nameUUIDFromBytes(MAPPER.writeValueAsBytes(event.getAttributes())));
```
And the `MAPPER` needs to have sorting enabled.
```
private static final ObjectMapper MAPPER =
            JsonMapper.builder()
                    .configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true)
                    .configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true)
                    .build();
```

wenjin272 (commented):
Hi, @addu390. Sorry for the delayed reply due to the holiday. I left a few comments regarding the details. 

This PR introduces significant changes to the framework, please take a look at your convenience @xintongsong.

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -79,7 +79,7 @@ The key of the pairs store in `MemoryObject` must be string, and the value can b
 
 {{< tab "Python" >}}
 ```python
-@action(InputEvent)
+@action("_input_event")
```
xintongsong: In the documentation, we should also use `InputEvent.EVENT_TYPE` rather than the string literal. Please also check for other occurrences in the documentation.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
xintongsong: +1

xintongsong (commented):
@addu390 @wenjin272

Thank you both for working on this. I have no more comments except for the following two.

api/src/main/java/org/apache/flink/agents/api/event/ChatRequestEvent.java
xintongsong: +1

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
xintongsong: +1

python/flink_agents/api/events/event.py
```diff
@@ -30,27 +30,62 @@
...
-        ID, which is critical for ActionStateStore divergence detection.
+        event content.
+    type : str | None
+        Event type string used for routing.
+    attributes : Dict[str, Any]
+        Key-value properties for the event data.
     """
 
     id: UUID = Field(default=None)
+    type: str | None = Field(default=None)
```
xintongsong: Why do we allow `type` to be `None`?

addu390: Not anymore, I'll remove the `None` support.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
addu390: Yes, that's the pattern. I thought I fixed this in all places, I'll do another full sweep again.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
addu390: Nice find! Now that all event data lives in attributes, can unify the branches and use a sorted ObjectMapper for key ordering. Updated.

addu390:
@wenjin272 @xintongsong Did a full cross-language consistency sweep, all `@Action` methods use base `Event + fromEvent()`, updated docs and JSON fixtures to match. And I resolved a few older comments, so, it's easier to track the open reviews/discussions.

wenjin272:
Hi, @addu390. Some `@Action` methods in the documentation were not using base `Event` + `fromEvent()`. Since they were scattered across various sections, I fixed them directly. Also fixed the conflicts with the main branch.

Thanks for your contribution, merged.

wenjin272 (approved):

Issue #633, opened 2026-04-18
[Feature] JsonParseException in ChatModelAction when LLM returns JSON wrapped in Markdown code blocks


twosom:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

In `ChatModelAction.generateStructuredOutput`, the agent attempts to parse the LLM's response using Jackson's `ObjectMapper`. However, if the LLM includes Markdown formatting (e.g.,  ```json ... ```) in its response, the parser fails with a JsonParseException because it encounters the backtick character instead of a valid JSON start token ({ or [).

## Stack Trace
```text
Caused by: com.fasterxml.jackson.core.JsonParseException: Unexpected character ('`' (code 96)): expected a valid value (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: REDACTED; line: 1, column: 1]
    at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._handleOddValue(ReaderBasedJsonParser.java:2054)
    at org.apache.flink.agents.plan.actions.ChatModelAction.generateStructuredOutput(ChatModelAction.java:213)
    ...

```

At `ChatModelAction`, the code directly passes the raw string from the LLM to `objectMapper.readValue()`. While some LLMs are instructed to return raw JSON, they often default to Markdown-formatted responses. The current implementation lacks a "cleaning" step to strip these common Markdown delimiters.

### How to reproduce

use `Gemma4` model

### Version and environment

latest dev branch

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wenjin272:
I don't consider this a bug, as the large language model's response is indeed not standard JSON, making it reasonable for the framework to throw an exception. Therefore, I am marking this as a feature.

PR #634, opened 2026-04-18
[fix][plan] Fix parsing exception in ChatModelAction

twosom:
Linked issue: #633 

### Purpose of change

The ChatModelAction currently fails with a JsonParseException when an LLM returns a structured output wrapped in Markdown code blocks (e.g., ```json ... ```). This happens because Jackson's ObjectMapper expects a valid JSON start character ({ or [) but encounters a backtick (`) instead.
so, I added a utility method to extract raw JSON string from LLM responses that include Markdown decorators.

### Tests

Added some unit tests for clean llm response

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
@@ -204,10 +204,19 @@ public class ChatModelAction {
         ctx.sendEvent(toolRequestEvent);
     }
 
+    static String cleanLlmResponse(String rawResponse) {
+        String trimmed = rawResponse.trim();
+        if (trimmed.startsWith("```")) {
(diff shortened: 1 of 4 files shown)


wenjin272:
Hi, @twosom, thanks for you work, LGTM.

Could you also add similar logic for the Python chat model action? We should maintain consistent behavior across built-in actions to avoid unexpected outcomes.

twosom:
> Could you also add similar logic for the Python chat model action? We should maintain consistent behavior across built-in actions to avoid unexpected outcomes.

Thank you for your comment @wenjin272. Yes, I will add similar logic to Python and write test cases as well.

wenjin272:
LGTM

Issue #608, opened 2026-04-02
[Feature] Support dynamic refresh of MCP tools and prompts


yanand0909:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

We need to support dynamically refresh MCP tools and prompts using one of the strategies discussed in the discussion doc.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #639, opened 2026-04-24
[Feature][Java] Move MCP resource discovery to runtime

yanand0909:
Linked issue: [608](https://github.com/apache/flink-agents/issues/608)

### Purpose of change

MCP tools and prompts were previously discovered at compile time inside AgentPlan.extractJavaMCPServer(): an MCPServer connection was opened, listTools()/listPrompts() were called over the network, results were serialized into the plan, and the connection was immediately closed. This had three problems:

  1. Build-time server dependency — AgentPlan construction failed if the MCP server was unreachable, even though the agent wouldn't run until later.
  2. Static capabilities — Tools and prompts were frozen at compile time; any server-side changes required recompilation.
  3. No connection reuse — The connection opened for discovery was closed immediately; a new one was established later for actual tool calls.

This PR implements Part 1: Runtime Discovery from discussion #543. MCP tool and prompt discovery is moved from AgentPlan construction to ActionExecutionOperator.open():

### Tests

  - AgentPlanDeclareMCPServerTest — Updated to assert the new compile-time contract: TOOL and PROMPT providers are absent from AgentPlan providers (deferred to runtime), only MCP_SERVER is registered. Tool/prompt retrieval tests now instantiate the MCPServer directly from the plan's provider, simulating what JavaMCPResourceDiscovery does at operator startup.
  - JavaMCPResourceDiscoveryTest (new) — Six unit tests covering JavaMCPResourceDiscovery with stub objects (no live server required):
    - Happy path: tools and prompts are discovered and placed in ResourceCache
    - supportsPrompts() = false: prompt discovery is skipped
    - Non-JavaResourceProvider entries are ignored

### API

No public API changes

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/plan/src/test/java/org/apache/flink/agents/plan/AgentPlanDeclareMCPServerTest.java
+++ b/plan/src/test/java/org/apache/flink/agents/plan/AgentPlanDeclareMCPServerTest.java
@@ -24,14 +24,12 @@ import org.apache.flink.agents.api.InputEvent;
 import org.apache.flink.agents.api.agents.Agent;
 import org.apache.flink.agents.api.annotation.Action;
 import org.apache.flink.agents.api.context.RunnerContext;
-import org.apache.flink.agents.api.prompt.Prompt;
-import org.apache.flink.agents.api.resource.Resource;
 import org.apache.flink.agents.api.resource.ResourceDescriptor;
 import org.apache.flink.agents.api.resource.ResourceName;
 import org.apache.flink.agents.api.resource.ResourceType;
-import org.apache.flink.agents.api.tools.Tool;
 import org.apache.flink.agents.api.tools.ToolMetadata;
 import org.apache.flink.agents.integrations.mcp.MCPPrompt;
+import org.apache.flink.agents.integrations.mcp.MCPServer;
 import org.apache.flink.agents.integrations.mcp.MCPTool;
 import org.apache.flink.agents.plan.resourceprovider.ResourceProvider;
 import org.junit.jupiter.api.*;
@@ -48,14 +46,17 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 /**
  * Tests for MCP server integration with AgentPlan.
  *
- * <p>This test verifies that MCP servers, tools, and prompts are properly discovered and registered
- * in the agent plan, following the pattern from {@link AgentPlanDeclareToolMethodTest}.
+ * <p>Verifies that MCP servers are registered in the agent plan at compile time, while tool and
+ * prompt discovery is deferred to operator startup (runtime). Tool/prompt retrieval tests
+ * instantiate the MCPServer directly from the plan's provider to simulate what {@code
+ * JavaMCPResourceDiscovery} does at runtime.
  *
  * <p>Uses the Python MCP server from python/flink_agents/api/tests/mcp/mcp_server.py.
  */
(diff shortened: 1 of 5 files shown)


wenjin272:
Thanks for your contribution, @yanand0909. Due to the upcoming traditional Chinese holiday, my review may be slightly delayed.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -271,6 +272,8 @@ public void open() throws Exception {
         shortTermMemState = getRuntimeContext().getMapState(shortTermMemStateDescriptor);
 
         resourceCache = new ResourceCache(agentPlan.getResourceProviders());
+        JavaMCPResourceDiscovery.discoverJavaMCPResources(
```
weiqingy, line 275: `JavaMCPResourceDiscovery.discoverJavaMCPResources` runs synchronously here, which means a slow or briefly-unreachable MCP server stalls (or fails) the entire operator's startup. Discussion #543 anticipates this with `failFastOnStartup(true)` as the default and a graceful-degradation opt-in — this PR ships fail-fast only. Is the intent to defer the opt-in to a follow-up, and does the current shape leave room for it (e.g., a per-server `try/catch` boundary around the discovery loop)?

runtime/src/main/java/org/apache/flink/agents/runtime/JavaMCPResourceDiscovery.java
```diff
@@ -0,0 +1,89 @@
...
+        if (servers == null) {
+            return;
+        }
+
+        for (ResourceProvider rp : servers.values()) {
+            if (!(rp instanceof JavaResourceProvider)) {
+                continue;
+            }
+
+            Object mcpServer = rp.provide(null);
```
weiqingy, line 64: `rp.provide(null)` passes `null` as `ResourceContext`. `PythonMCPResourceDiscovery` at `PythonMCPResourceDiscovery.java:73` passes `cache.getResourceContext()` for the same role. Today `MCPServer`'s constructor just stores the field without dereferencing it, so the null is benign — but if a future Java MCP server resolves a dependent resource through `getResourceContext()`, it would NPE here in a way that doesn't manifest on the Python side. Any reason to keep the null rather than match the Python path?

plan/src/test/java/org/apache/flink/agents/plan/AgentPlanDeclareMCPServerTest.java
```diff
@@ -222,116 +221,148 @@ void discoverMCPServer() {
...
-        assertTrue(json.contains("add"), "JSON should contain add tool");
-        assertTrue(json.contains("ask_sum"), "JSON should contain ask_sum prompt");
+        // Serialized plan contains the MCPServer configuration
         assertTrue(json.contains("mcp_server"), "JSON should contain mcp_server type");
+        assertTrue(json.contains("testMcpServer"), "JSON should contain the server provider name");
+        assertTrue(json.contains(MCP_ENDPOINT), "JSON should contain the endpoint");
+
+        // Tools and prompts are NOT serialized into the plan (they are runtime-discovered)
+        assertFalse(
+                json.contains("\"add\"") && json.contains("java_serializable"),
```
weiqingy, line 317:
```java
assertFalse(
    json.contains("\"add\"") && json.contains("java_serializable"),
    "JSON should not contain a serialized 'add' tool provider");
```

This passes when *either* substring is missing, so a regression that drops `"add"` from the JSON for an unrelated reason would silently make this test green — the very class of bug the test is meant to catch. Since the design contract you're locking down is "no `java_serializable` providers for MCP-discovered resources at all," one alternative in case it helps:

```java
assertFalse(json.contains("java_serializable"),
        "JSON should not contain any java_serializable provider entries (MCP discovery is deferred)");
```

weiqingy (commented):
Thanks for taking this on — moving discovery out of plan construction is a real cleanup. Two cross-cutting questions before I leave inline notes:

1. **MCPServer lifecycle when discovery yields nothing.** `discoverJavaMCPResources` instantiates the server and calls `listTools()` / `supportsPrompts()`, both of which lazily initialize an `McpSyncClient` over HTTP (`MCPServer.getClient()` at `MCPServer.java:304-309`, lazy create at `MCPServer.java:316-343`). The only things put in `cache` are the discovered tools and prompts; the server itself isn't cached. Shutdown then relies on `MCPTool.close()` / `MCPPrompt.close()` calling back into `mcpServer.close()` — which works when at least one tool or prompt was discovered. If `listTools()` returns empty AND prompts are unsupported (or also empty), nothing in the cache references the server, so `mcpServer.close()` never runs and the initialized `McpSyncClient` (with its HTTP transport) leaks. An empty-tools server is unusual but possible — a prompts-only server, a misconfigured server, a server going through gradual rollout. Two shapes that would close this gap, in case either helps: (a) put the server itself in the cache (`cache.put(serverName, MCP_SERVER, mcpServer)` after `provide()`), or (b) track discovered servers in a list owned by `ActionExecutionOperator` and close them in the operator's `close()`. How do you want to handle it?

2. **Shutdown contract isn't pinned by a test.** The 6 new unit tests cover discovery happy/sad paths, but none verify that `cache.close()` releases the server. Combined with #1, that means the lifecycle invariant (whichever form it takes) lives entirely in reviewer memory. The `FakeMCPServer` stub already has the right shape — adding a `closeCalled` flag plus a test that closes the cache and asserts it flipped is a few lines. Worth doing?

wenjin272:
Thanks to @yanand0909 for the contribution and @weiqingy for the review. My comments align with @weiqingy's, so I’ll skip adding them here.

Apologies for the delayed review of this PR, as I was occupied with the 0.3 release and FFA-related tasks. There are currently some merge conflicts between this PR branch and main that need to be resolved. No worries if you’re short on time—I can step in and handle it.

lihaosky:
Close as it's stale. @yanand0909  can reopen if still work on this

Issue #640, opened 2026-04-25
[Feature][Java] Add `EventListener` config


twosom:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

We need to establish a way to configure the `EventListener` interface introduced in the previous PR (#102).
Currently, there is no defined method for its configuration. This task requires the implementation of the relevant code and the creation of supporting documentation.

**NOTE** : As the `EventListener` interface is currently defined only in Java, this proposal is strictly limited to the Java implementation.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #641, opened 2026-04-25
[Feature][Java] Add java event listeners config

twosom:
Linked issue: #640

### Purpose of change
Added event listeners config in java

### Tests
ok

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/listener/EventListener.java
+++ b/api/src/main/java/org/apache/flink/agents/api/listener/EventListener.java
@@ -31,6 +31,9 @@ import org.apache.flink.agents.api.EventContext;
  * <p>Event listeners are executed synchronously after the main event processing is complete but
  * before the next event is processed. Implementations should be lightweight and avoid blocking
  * operations to prevent impacting agent performance.
+ *
+ * <p><strong>Note:</strong> Implementing classes must provide a public no-argument constructor to
+ * allow for dynamic instantiation by the agent.
  */
 public interface EventListener {
     /**
--- a/api/src/main/java/org/apache/flink/agents/api/listener/EventListener.java
+++ b/api/src/main/java/org/apache/flink/agents/api/listener/EventListener.java
@@ -22,33 +22,33 @@ import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.api.EventContext;
 
 /**
- * Interface for event listeners that are notified when events are processed.
+ * Interface for event listeners that are notified when events are received for processing.
  *
- * <p>EventListener provides a callback mechanism triggered after event processing completes. This
- * is useful for monitoring, metrics collection, debugging, or triggering side effects based on
- * event processing.
+ * <p>EventListener provides a callback mechanism triggered at the beginning of event processing.
+ * This is useful for monitoring, metrics collection, debugging, or triggering side effects based on
+ * event reception.
  *
- * <p>Event listeners are executed synchronously after the main event processing is complete but
- * before the next event is processed. Implementations should be lightweight and avoid blocking
- * operations to prevent impacting agent performance.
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -191,7 +194,7 @@ public class ActionExecutionOperator<IN, OUT> extends AbstractStreamOperator<OUT
     private transient ListState<Object> currentProcessingKeysOpState;
 
     private final transient EventLogger eventLogger;
-    private final transient List<EventListener> eventListeners;
```
wenjin272: I think we can keep the `final` to maintain alignment with eventLogger. In `initEventListeners`, We can use `addAll` instead of assigning the entire collection at once.

api/src/main/java/org/apache/flink/agents/api/listener/EventListener.java
```diff
@@ -31,6 +31,9 @@
  * <p>Event listeners are executed synchronously after the main event processing is complete but
  * before the next event is processed. Implementations should be lightweight and avoid blocking
  * operations to prevent impacting agent performance.
```
wenjin272: It appears that the EventListener is actually triggered before the Event is processed in `ActionExecutionOperator`, which is inconsistent with the documentation here. We may need to update either the doc or the invocation location.

wenjin272 (commented):
Thanks for your contribution, @twosom. Overall looks good to me. I left two minor comments.

twosom:
> Thanks for your contribution, @twosom. Overall looks good to me. I left two minor comments.

Thanks for the review! @wenjin272 
I’ve applied both minor suggestions! 
Please take another look when you have a moment. Thanks!

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -411,6 +442,8 @@ private void processEvent(Object key, Event event) throws Exception {
             // If the event is an InputEvent, we submit a new mail to try processing the actions.
             mailboxExecutor.submit(() -> tryProcessActionTaskForKey(key), "process action task");
         }
+
+        notifyEventProcessed(event);
```
wenjin272: Hi, @twosom. I carefully reviewed the logic of `processEvent`. In this function, only an `ActionTask` is created and submitted. It's actual execution occurs in subsequent mailbox iterations. Therefore, merely moving `notifyEventProcessed` to the end of the function still remains inconsistent with the documentation of `EventListener`. After considering all factors, I think it's simpler to place it at the beginning and update the `EventListener` documentation accordingly.

twosom: Thanks for the catch, @wenjin272 

I missed the point that the actual execution happens in subsequent mailbox iterations. I agree that calling it at the beginning and updating the Javadoc is the best way to avoid inconsistency. I'll update the code and the documentation accordingly.

twosom: Hi @wenjin272 
I've updated the PR as suggested. Could you take another look? Thanks!

wenjin272 (commented):
LGTM, but there are some conflicts with main branch need resolve.

twosom:
Thanks for the review, @wenjin272! 
I've resolved the merge conflicts and updated the branch. Please take another look.

wenjin272 (approved):
LGTM

Issue #594, opened 2026-03-26
[Feature] Support loading agent skills form python package.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support loading agent skills form python package.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

Issue #592, opened 2026-03-26
[Feature] Support loading skills from remote urls


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support loading agent skills from remote urls.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #650, opened 2026-05-08
[api][runtime] Support agent skills in java

wenjin272:
Linked issue: #591 

### Purpose of change

This PR provides basic support for using agent skills in the Flink Agents Java API. 

### Tests

ut & e2e

### API

Yes, add abstraction for agent skill, and modify the chat model api.

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

Issue #10, opened 2025-06-12
[Tech Debt] Document how-to-contribute in README


xintongsong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Contents to be added:
- how to build the project
- ide settings
- link to the contribution guidelines

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #655, opened 2026-05-11
[api][runtime] Extend  agent skill loading repository.

wenjin272:
Linked issue: #592, #593, #594

### Purpose of change

Extend the way Flink-Agents loads external agent skills:
* load agent skills from remote urls (only http/https point to a zip)
* load agent skills from python package
* load agent skills from java classpath

### Tests

ut

### API

Yes, add `from_xxx` to `Skills`.

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/skills/Skills.java
+++ b/api/src/main/java/org/apache/flink/agents/api/skills/Skills.java
@@ -28,14 +28,29 @@ import org.apache.flink.agents.api.resource.SerializableResource;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Configuration resource describing where to load agent skills from.
  *
- * <p>Mirrors the Python {@code flink_agents.api.skills.Skills}. Use {@link
- * #fromLocalDir(String...)} to construct.
+ * <p>The single field {@code sources} holds an ordered list of {@link SkillSourceSpec} entries.
+ * Each entry has a {@code scheme} (e.g. {@code "local"}, {@code "url"}, {@code "classpath"}, {@code
+ * "package"}) and a scheme-specific {@code params} map. Use one of the factory methods to construct
+ * a {@link Skills} resource:
  *
- * <p>Multiple {@code @Skills} declarations on the same agent are merged at plan-build time.
+ * <ul>
+ *   <li>{@link #fromLocalDir(String...)} for local directories or {@code .zip} files
+ *   <li>{@link #fromUrl(String...)} for http(s) URLs pointing to a {@code .zip}
+ *   <li>{@link #fromClasspath(String...)} for resources on the classpath
+ * </ul>
+ *
+ * <p>The {@code "package"} scheme exists on the Python side only (Java has no analogous concept). A
+ * plan written by Python with {@code scheme=package} deserializes successfully on Java, but {@code
+ * SkillManager} will fail fast at load time with the registered-scheme list.
+ *
+ * <p>Multiple {@code @Skills} declarations on the same agent are merged at plan-build time;
+ * duplicate {@link SkillSourceSpec} entries (same {@code scheme} and {@code params}) are collapsed.
  */
 @JsonIgnoreProperties(
         ignoreUnknown = true,
@@ -51,31 +66,60 @@ public class Skills extends SerializableResource {
     /** Reserved name of the built-in bash tool used to execute skill scripts. */
     public static final String BASH_TOOL = "bash";
 
-    private List<String> paths;
+    private final List<SkillSourceSpec> sources;
 
     /** Required by Jackson. */
     public Skills() {
-        this.paths = Collections.emptyList();
+        this.sources = Collections.emptyList();
     }
 
     @JsonCreator
-    public Skills(@JsonProperty("paths") List<String> paths) {
-        this.paths = paths == null ? Collections.emptyList() : List.copyOf(paths);
+    public Skills(@JsonProperty("sources") List<SkillSourceSpec> sources) {
+        this.sources = sources == null ? Collections.emptyList() : List.copyOf(sources);
     }
 
     /**
-     * Create a {@link Skills} resource from one or more local filesystem directories.
+     * Create a {@link Skills} resource from one or more local paths.
      *
-     * <p>Each path points to a directory whose immediate subdirectories each contain a {@code
-     * SKILL.md} file.
+     * <p>Each path may be a directory whose immediate subdirectories each contain a {@code
+     * SKILL.md} file, or a {@code .zip} file whose top-level entries are the skill subdirectories.
      */
     public static Skills fromLocalDir(String... paths) {
-        return new Skills(Arrays.asList(paths));
+        return new Skills(
+                Arrays.stream(paths)
+                        .map(p -> new SkillSourceSpec("local", Map.of("path", p)))
+                        .collect(Collectors.toList()));
     }
 
-    @JsonProperty("paths")
-    public List<String> getPaths() {
-        return paths;
+    /**
+     * Create a {@link Skills} resource from one or more http(s) URLs.
+     *
+     * <p>Each URL must point to a {@code .zip} whose top level is the baseDir.
+     */
+    public static Skills fromUrl(String... urls) {
+        return new Skills(
+                Arrays.stream(urls)
+                        .map(u -> new SkillSourceSpec("url", Map.of("url", u)))
+                        .collect(Collectors.toList()));
+    }
+
+    /**
+     * Create a {@link Skills} resource from one or more classpath resource paths.
+     *
+     * <p>Each resource may be a directory (e.g. under {@code src/main/resources/skills}) or a
+     * {@code .zip} file. When packaged into a JAR, the resource is loaded via the thread context
+     * class loader and materialized to a temp directory at runtime.
+     */
+    public static Skills fromClasspath(String... resources) {
+        return new Skills(
+                Arrays.stream(resources)
+                        .map(r -> new SkillSourceSpec("classpath", Map.of("resource", r)))
+                        .collect(Collectors.toList()));
+    }
+
+    @JsonProperty("sources")
+    public List<SkillSourceSpec> getSources() {
+        return sources;
     }
 
     @JsonIgnore
--- a/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
@@ -28,6 +28,7 @@ import org.apache.flink.agents.api.resource.ResourceDescriptor;
 import org.apache.flink.agents.api.resource.ResourceType;
 import org.apache.flink.agents.api.resource.SerializableResource;
 import org.apache.flink.agents.api.resource.python.PythonResourceWrapper;
+import org.apache.flink.agents.api.skills.SkillSourceSpec;
 import org.apache.flink.agents.api.skills.Skills;
 import org.apache.flink.agents.api.tools.ToolMetadata;
 import org.apache.flink.agents.plan.actions.Action;
@@ -57,6 +58,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
@@ -572,11 +574,15 @@ public class AgentPlan implements Serializable {
                         ResourceType.TOOL,
                         new ResourceDescriptor(BashTool.class.getName(), new HashMap<>())));
 
-        LinkedHashSet<String> paths = new LinkedHashSet<>();
-        for (Skills s : skillsObjects.values()) {
-            paths.addAll(s.getPaths());
+        // Sort by key before merging: getDeclaredMethods() makes no order guarantee, so without
+        // this the winner on a duplicate skill name would vary across JDK / class layout.
+        List<String> orderedKeys = new ArrayList<>(skillsObjects.keySet());
+        Collections.sort(orderedKeys);
+        LinkedHashSet<SkillSourceSpec> sources = new LinkedHashSet<>();
+        for (String key : orderedKeys) {
+            sources.addAll(skillsObjects.get(key).getSources());
         }
-        Skills merged = Skills.fromLocalDir(paths.toArray(new String[0]));
+        Skills merged = new Skills(new ArrayList<>(sources));
         addResourceProvider(
                 JavaSerializableResourceProvider.createResourceProvider(
                         Skills.SKILLS_CONFIG, ResourceType.SKILLS, merged));
--- a/plan/src/test/java/org/apache/flink/agents/plan/AgentPlanDeclareSkillsTest.java
+++ b/plan/src/test/java/org/apache/flink/agents/plan/AgentPlanDeclareSkillsTest.java
@@ -20,12 +20,14 @@ package org.apache.flink.agents.plan;
 
 import org.apache.flink.agents.api.agents.Agent;
 import org.apache.flink.agents.api.resource.ResourceType;
+import org.apache.flink.agents.api.skills.SkillSourceSpec;
 import org.apache.flink.agents.api.skills.Skills;
 import org.apache.flink.agents.plan.resourceprovider.JavaResourceProvider;
 import org.apache.flink.agents.plan.resourceprovider.JavaSerializableResourceProvider;
 import org.apache.flink.agents.plan.resourceprovider.ResourceProvider;
 import org.junit.jupiter.api.Test;
 
+import java.util.List;
 import java.util.Map;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -51,6 +53,47 @@ class AgentPlanDeclareSkillsTest {
         @org.apache.flink.agents.api.annotation.Skills
         public static Skills second() {
             return Skills.fromLocalDir("/tmp/skill-b", "/tmp/skill-c");
+        }
+    }
+
+    public static class MultiSkillsUrlAgent extends Agent {
+        @org.apache.flink.agents.api.annotation.Skills
+        public static Skills first() {
+            return Skills.fromUrl("https://a.example/x.zip", "https://b.example/x.zip");
+        }
+
+        @org.apache.flink.agents.api.annotation.Skills
+        public static Skills second() {
+            return Skills.fromUrl("https://b.example/x.zip", "https://c.example/x.zip");
+        }
+    }
+
+    public static class MultiSkillsClasspathAgent extends Agent {
+        @org.apache.flink.agents.api.annotation.Skills
+        public static Skills first() {
+            return Skills.fromClasspath("skills-a", "skills-b");
+        }
+
+        @org.apache.flink.agents.api.annotation.Skills
+        public static Skills second() {
+            return Skills.fromClasspath("skills-b", "skills-c");
+        }
+    }
+
+    public static class MixedSchemeAgent extends Agent {
+        @org.apache.flink.agents.api.annotation.Skills
+        public static Skills locals() {
+            return Skills.fromLocalDir("/tmp/skill-a");
+        }
+
+        @org.apache.flink.agents.api.annotation.Skills
+        public static Skills urls() {
+            return Skills.fromUrl("https://a.example/x.zip");
+        }
+
+        @org.apache.flink.agents.api.annotation.Skills
+        public static Skills classpaths() {
+            return Skills.fromClasspath("skills-a");
         }
     }
 
@@ -83,21 +126,49 @@ class AgentPlanDeclareSkillsTest {
     }
 
     @Test
-    void multipleSkillsMethodsMergePathsWithDeduplication() throws Exception {
-        AgentPlan plan = new AgentPlan(new MultiSkillsAgent());
-        ResourceProvider configProvider =
-                plan.getResourceProviders().get(ResourceType.SKILLS).get(Skills.SKILLS_CONFIG);
-        Skills merged =
-                (Skills)
-                        ((JavaSerializableResourceProvider) configProvider)
-                                .provide(
-                                        org.apache.flink.agents.api.resource.ResourceContext
-                                                .fromGetResource((n, t) -> null));
-        // Order is preserved; "/tmp/skill-b" appears once.
-        assertEquals(3, merged.getPaths().size());
-        assertTrue(merged.getPaths().contains("/tmp/skill-a"));
-        assertTrue(merged.getPaths().contains("/tmp/skill-b"));
-        assertTrue(merged.getPaths().contains("/tmp/skill-c"));
+    void multipleSkillsMethodsMergeLocalPathsWithDeduplication() throws Exception {
+        Skills merged = mergedSkillsOf(new MultiSkillsAgent());
+        assertEquals(
+                java.util.Set.of(
+                        new SkillSourceSpec("local", Map.of("path", "/tmp/skill-a")),
+                        new SkillSourceSpec("local", Map.of("path", "/tmp/skill-b")),
+                        new SkillSourceSpec("local", Map.of("path", "/tmp/skill-c"))),
+                java.util.Set.copyOf(merged.getSources()));
+    }
+
+    @Test
+    void multipleSkillsMethodsMergeUrlsWithDeduplication() throws Exception {
+        Skills merged = mergedSkillsOf(new MultiSkillsUrlAgent());
+        assertEquals(
+                java.util.Set.of(
+                        new SkillSourceSpec("url", Map.of("url", "https://a.example/x.zip")),
+                        new SkillSourceSpec("url", Map.of("url", "https://b.example/x.zip")),
+                        new SkillSourceSpec("url", Map.of("url", "https://c.example/x.zip"))),
+                java.util.Set.copyOf(merged.getSources()));
+    }
+
+    @Test
+    void multipleSkillsMethodsMergeClasspathResourcesWithDeduplication() throws Exception {
+        Skills merged = mergedSkillsOf(new MultiSkillsClasspathAgent());
+        assertEquals(
+                java.util.Set.of(
+                        new SkillSourceSpec("classpath", Map.of("resource", "skills-a")),
+                        new SkillSourceSpec("classpath", Map.of("resource", "skills-b")),
+                        new SkillSourceSpec("classpath", Map.of("resource", "skills-c"))),
+                java.util.Set.copyOf(merged.getSources()));
+    }
+
+    @Test
+    void threeSchemeMixMergesInLexicalKeyOrder() throws Exception {
+        // MixedSchemeAgent's @Skills methods are classpaths/locals/urls — merging by sorted
+        // method name fixes the source order regardless of getDeclaredMethods()'s whims.
+        Skills merged = mergedSkillsOf(new MixedSchemeAgent());
+        assertEquals(
+                java.util.List.of(
+                        new SkillSourceSpec("classpath", Map.of("resource", "skills-a")),
+                        new SkillSourceSpec("local", Map.of("path", "/tmp/skill-a")),
+                        new SkillSourceSpec("url", Map.of("url", "https://a.example/x.zip"))),
+                merged.getSources());
     }
 
     @Test
@@ -116,16 +187,20 @@ class AgentPlanDeclareSkillsTest {
     void programmaticSkillsAddResourceParticipates() throws Exception {
         Agent agent = new NoSkillsAgent();
         agent.addResource("more", ResourceType.SKILLS, Skills.fromLocalDir("/tmp/skill-d"));
+        Skills merged = mergedSkillsOf(agent);
+        assertEquals(
+                List.of(new SkillSourceSpec("local", Map.of("path", "/tmp/skill-d"))),
+                merged.getSources());
+    }
+
+    private static Skills mergedSkillsOf(Agent agent) throws Exception {
         AgentPlan plan = new AgentPlan(agent);
         ResourceProvider configProvider =
                 plan.getResourceProviders().get(ResourceType.SKILLS).get(Skills.SKILLS_CONFIG);
-        assertNotNull(configProvider);
-        Skills merged =
-                (Skills)
-                        ((JavaSerializableResourceProvider) configProvider)
-                                .provide(
-                                        org.apache.flink.agents.api.resource.ResourceContext
-                                                .fromGetResource((n, t) -> null));
-        assertEquals(java.util.List.of("/tmp/skill-d"), merged.getPaths());
+        return (Skills)
+                ((JavaSerializableResourceProvider) configProvider)
+                        .provide(
+                                org.apache.flink.agents.api.resource.ResourceContext
+                                        .fromGetResource((n, t) -> null));
     }
 }
--- a/python/flink_agents/api/skills.py
+++ b/python/flink_agents/api/skills.py
@@ -17,7 +17,14 @@
 #################################################################################
 """Skills configuration resource for agent skills discovery.
 
-Use :meth:`Skills.from_local_dir` to construct a :class:`Skills` resource.
+Each :class:`Skills` resource carries a single ordered list of
+:class:`SkillSourceSpec` entries. Each entry has a ``scheme`` (e.g.
+``"local"``, ``"url"``, ``"package"``) and a scheme-specific ``params`` map.
+Use one of the factory methods to construct a :class:`Skills` resource:
+
+* :meth:`Skills.from_local_dir` for local directories or local ``.zip`` files
+* :meth:`Skills.from_url` for http(s) URLs pointing to a ``.zip``
+* :meth:`Skills.from_package` for resources inside installed packages
 
 Example::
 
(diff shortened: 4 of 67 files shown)


wenjin272:
This pr based on #650

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
```diff
@@ -0,0 +1,201 @@
...
+     */
+    public static Path extractClasspathFromJar(URL jarUrl, String resourcePrefix)
+            throws IOException {
+        // Parse the JAR file URL from the jar: URL. The format is jar:<jar-file-url>!/[entry].
+        // We extract just the inner jar-file URL so we can open the whole JarFile and enumerate
+        // all entries — JarURLConnection.getJarFile() would fail when the entry specifier names a
+        // prefix that has no corresponding stored directory entry.
+        String spec = jarUrl.toString();
+        int sep = spec.indexOf("!/");
+        String innerSpec = sep >= 0 ? spec.substring(4, sep) : spec.substring(4);
```
wzhero1: Magic 4 = "jar:".length(). Extract a private static final int JAR_PREFIX_LEN = "jar:".length(); constant.

wzhero1:
Thanks for the PR. Left some comments as below.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
```diff
@@ -0,0 +1,201 @@
...
+     */
+    public static Path extractZipSafely(Path zipPath) throws IOException {
+        Path extractDir = Files.createTempDirectory(TEMP_DIR_PREFIX);
+        // Register cleanup before validation so the empty tempdir is always reclaimed,
+        // even if validation raises.
+        registerCleanup(extractDir);
+        try (ZipFile zf = new ZipFile(zipPath.toFile())) {
+            // First pass: zip-slip validation.
+            Enumeration<? extends ZipEntry> entries = zf.entries();
+            while (entries.hasMoreElements()) {
```
wzhero1: **[Nit] Two-pass scan can be merged into one.**

Validate-then-extract per entry in a single pass halves the IO touches. Performance is not the concern — the code just gets shorter.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/URLSkillRepository.java
```diff
@@ -0,0 +1,50 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Skill repository backed by an http(s) URL pointing to a zip.
+ *
+ * <p>The zip is downloaded to a temp file and extracted into a process-local temp directory
+ * (cleaned up at JVM exit). The downloaded zip itself is removed once extraction completes.
+ */
+public class URLSkillRepository extends FileSystemSkillRepository {
```
wzhero1: **Inheritance is being used as composition.**

A URL repo is not logically *a* FileSystem repo — it fetches over the network, then delegates. Side effects: the constructor performs network I/O before `super(...)`, the original `url` is lost for diagnostics, there's no `close()` lifecycle, and the fetch step can't be mocked alone.

Suggest a `SkillSource { Path materialize() }` interface composed into one `MaterializingSkillRepository`. Same applies to `ClasspathSkillRepository` and `PackageSkillRepository`.

api/src/main/java/org/apache/flink/agents/api/skills/Skills.java
```diff
@@ -51,33 +57,76 @@ public class Skills extends SerializableResource {
     /** Reserved name of the built-in bash tool used to execute skill scripts. */
     public static final String BASH_TOOL = "bash";
 
-    private List<String> paths;
+    private final List<String> paths;
+    private final List<String> urls;
+    private final List<String> classpathResources;
```
wzhero1: **Adding a new source today forces 5 parallel changes** (`Skills` field + factory + `XxxSkillRepository` + `SkillManager` for-loop + `AgentPlan` merge branch).

Suggest a discriminated union — single `sources: List<SkillSourceSpec(scheme, location)>` field + a `scheme → handler` registry. New source = one registration, no other file touched. Bonus: cross-language ser/de becomes uniform.

Wire-format-affecting; much cheaper to fix before 0.3.0 ships.

python/flink_agents/api/skills.py
```diff
@@ -43,23 +59,56 @@ def my_skills() -> Skills:
...
-    # Internal location list populated through the factory methods; kept
-    # as a public pydantic field so it can be serialized/deserialized.
+    # Filesystem paths. Each entry may be a directory whose immediate
+    # subdirectories each contain a ``SKILL.md`` file, or a ``.zip`` file
+    # whose top-level entries are the skill subdirectories.
     paths: List[str] = Field(default_factory=list)
 
+    # http(s) URLs. Each URL must point to a ``.zip`` whose top level is
+    # the baseDir.
+    urls: List[str] = Field(default_factory=list)
```
wzhero1: **Cross-language deserialization silently drops fields.** Python writes `packages` → Java drops it → 0 skills loaded, no warning. Mirror for Java `classpathResources` → Python.

Short term: emit a WARN on unrecognized source fields during deserialization. Long term: unify under one cross-language scheme (see the `sources` discriminator suggested on `Skills.java`).

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/ClasspathSkillRepository.java
```diff
@@ -0,0 +1,129 @@
...
+            if (Files.isRegularFile(p) && p.toString().toLowerCase().endsWith(".zip")) {
+                return SkillMaterializer.extractZipSafely(p);
+            }
+            throw new IllegalArgumentException(
+                    "Classpath resource must be a directory or a .zip: " + p);
+        }
+        if ("jar".equals(protocol)) {
+            return SkillMaterializer.extractClasspathFromJar(url, resource);
+        }
+        throw new IllegalArgumentException(
```
wzhero1, line 149: **⚠️ Flink fat-jar / nested-jar deployment is not covered.**

Flink user JARs typically use Maven Shade or jar-in-jar; `getResource("skills")` returns protocols like `jar:nested:` that throw right here. `findInUrlClassLoader` also assumes `URLClassLoader`, not Flink's `ChildFirstClassLoader`.

Issue #593 explicitly targets this scenario. Before merge: (1) an end-to-end test on a Flink mini-cluster, (2) use `ClassLoader.getResources()` + `getResourceAsStream` instead of dispatching on URL protocol.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java
```diff
@@ -44,7 +47,7 @@ public class SkillManager {
 
     public SkillManager(Skills config) {
```
wzhero1: **Constructor is synchronous, blocking, and not recoverable.** `loadAll()` fetches sources serially; one slow URL stalls startup, one failure kills the whole `SkillManager`, and `repos` is `final` so there's no hot reload.

Follow-up: parallel materialization (`CompletableFuture`), fail-soft mode with a failure list, lazy load on first `getSkill()`, optional `reloadSource()` API.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java
```diff
@@ -134,15 +137,33 @@ public Path resolveResourcePath(String skillName, String resourcePath) {
...
+        }
+        for (String resource : config.getClasspathResources()) {
+            try {
+                registerRepo(new ClasspathSkillRepository(resource));
+            } catch (IOException e) {
+                throw new IllegalStateException(
+                        "Failed to load skills from classpath: " + resource, e);
             }
         }
     }
```
wzhero1, line 254: **`AgentSkill` doesn't know where it came from.** Origin is captured only inside a lambda; `SkillManager` keeps a parallel `repos` map to compensate. Logs and errors can't name the source, the duplicate-name WARN can't either, and serialization loses origin entirely.

Suggest a declarative `record SkillOrigin(String scheme, String location, Instant loadedAt)` field on `AgentSkill`.

api/src/main/java/org/apache/flink/agents/api/skills/Skills.java
```diff
@@ -51,33 +57,76 @@ public class Skills extends SerializableResource {
     /** Reserved name of the built-in bash tool used to execute skill scripts. */
     public static final String BASH_TOOL = "bash";
 
-    private List<String> paths;
+    private final List<String> paths;
```
wzhero1: **Skill names are flat — no namespace, no version.** With this PR opening multiple loading sources, two teams' `github` skills will collide (last writer silently wins).

Options: (1) `namespace:` in SKILL.md frontmatter → `team-a/github`, (2) per-source namespace `fromUrl(url, namespace="team-a")`, (3) source identity *is* the namespace. Wire-format-affecting — easier now than after 0.3.0.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/URLSkillRepository.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private static final int REQUEST_TIMEOUT_MS = 90_000;
+
+    public URLSkillRepository(String url) throws IOException {
+        super(materialize(url));
+    }
+
+    private static Path materialize(String url) throws IOException {
```
wzhero1: **URL zip is loaded with no integrity verification** — vulnerable to truncated downloads, stale CDN cache, MITM with a private CA, and proxy rewrites.

`Skills.fromUrl(url, "sha256:…")` with verification before extraction; mandatory in production. Same place to add an optional host allowlist for SSRF mitigation.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /**
+     * Extract a zip into a fresh temp directory and return that directory. Validates every entry
+     * against zip-slip (paths must resolve inside the extraction directory). Registers a JVM
+     * shutdown hook to remove the extraction directory at process exit.
+     *
+     * @throws IOException if any zip entry resolves outside the extraction directory.
+     */
+    public static Path extractZipSafely(Path zipPath) throws IOException {
```
wzhero1: **Missing zip-bomb protection.** No cap on entry count, per-entry size, total size, or compression ratio — a small DEFLATE bomb can exhaust TaskManager disk/memory.

Add limits (e.g. ≤ 50 MB per entry, ≤ 200 MB total, ≤ 10 000 entries, reject ratio > 100:1) and tests asserting the rejection.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
wzhero1, line 132: **Two-pass scan can be merged into one** — validate-then-extract per entry in a single pass. Not a perf concern, just shorter.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
```diff
@@ -0,0 +1,201 @@
...
+
+    private static final String TEMP_DIR_PREFIX = "flink-agents-skills-";
+
+    private SkillMaterializer() {}
+
+    /**
+     * Register a JVM shutdown hook that removes {@code path} recursively. Failures during deletion
+     * are silently ignored (best-effort cleanup).
+     */
+    public static void registerCleanup(Path path) {
```
wzhero1: **Shutdown hooks accumulate unboundedly.** Each zip / URL / classpath repo registers one hook; long-running processes that rebuild `SkillManager` (retries, hot reload) leak hooks monotonically.

Fix: cache materializations by source digest (best — eliminates the leak entirely), or expose `close()` that calls `removeShutdownHook` + deletes immediately. At minimum javadoc that `SkillManager` must not be reconstructed in a loop.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
wzhero1: **Magic number.** `4` is `"jar:".length()`. Extract `private static final int JAR_URL_PREFIX_LEN = "jar:".length();`.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java
```diff
@@ -134,15 +137,33 @@ public Path resolveResourcePath(String skillName, String resourcePath) {
...
+
+    private void registerRepo(SkillRepository repo) {
```
wzhero1: **Duplicate skill names overwrite silently with no log.** Multi-source name collisions go through `skills.put()` and the later registration wins — no WARN, no error. Users debugging "why doesn't my `SKILL.md` take effect" get no diagnostic.

Minimum: WARN with both origins. Better: document the `paths → urls → classpath` precedence. (Naming the source concretely needs the `SkillOrigin` change on `AgentSkill`.)

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/ClasspathSkillRepository.java
```diff
@@ -0,0 +1,129 @@
...
+        super(materialize(resource, classLoader));
+    }
+
+    private static Path materialize(String resource, ClassLoader classLoader) throws IOException {
+        URL url = classLoader.getResource(resource);
+        if (url == null) {
+            // Fallback for URLClassLoader: JARs without explicit directory entries will not return
+            // a URL for a prefix via getResource(). Scan the classloader's URLs directly to find
+            // any JAR entry that starts with the resource prefix.
+            url = findInUrlClassLoader(resource, classLoader);
```
wzhero1: **`findInUrlClassLoader` silently uses only the first matching jar.** Multiple plugin jars with the same prefix → every jar after the first is dropped, no log. Minimum: log the choice. Ideally merge across all matches, or fail loudly when more than one matches.

Also assumes `URLClassLoader`, not Flink's `ChildFirstClassLoader` — see the nested-jar thread on `materialize`.

api/src/main/java/org/apache/flink/agents/api/skills/Skills.java
```diff
@@ -51,33 +57,76 @@ public class Skills extends SerializableResource {
...
     /** Required by Jackson. */
     public Skills() {
         this.paths = Collections.emptyList();
+        this.urls = Collections.emptyList();
+        this.classpathResources = Collections.emptyList();
     }
 
     @JsonCreator
-    public Skills(@JsonProperty("paths") List<String> paths) {
+    public Skills(
```
wzhero1: **Source-incompatible constructor change.** Wire compat is fine; but `new Skills(List.of("/data/skills"))` no longer compiles. Either keep a `@Deprecated` 1-arg constructor that forwards, or call it out in the 0.3.0 release notes.

If `Skills` later migrates to a single `sources` field (see the discriminator thread), this signature changes again — better to break once.

plan/src/test/java/org/apache/flink/agents/plan/AgentPlanDeclareSkillsTest.java
```diff
@@ -126,6 +127,6 @@ void programmaticSkillsAddResourceParticipates() throws Exception {
                                 .provide(
                                         org.apache.flink.agents.api.resource.ResourceContext
                                                 .fromGetResource((n, t) -> null));
-        assertEquals(java.util.List.of("/tmp/skill-d"), merged.getPaths());
+        assertEquals(List.of("/tmp/skill-d"), merged.getPaths());
```
wzhero1: **Merge-coverage gap.** `AgentPlan.addSkills` now merges three `LinkedHashSet`s but the test only asserts on `getPaths()`. Add cases for overlapping `urls`, overlapping `classpathResources`, and a mixed three-source case asserting all three end up in the merged result.

python/flink_agents/api/skills.py
```diff
@@ -43,23 +59,56 @@ def my_skills() -> Skills:
...
+    def from_url(cls, *urls: str) -> Skills:
+        """Create a Skills resource from one or more http(s) URLs.
+
+        Each URL must point to a ``.zip`` whose top level is the baseDir
+        (i.e. skill subdirectories sit at the top of the zip).
+        """
+        return cls(urls=list(urls))
+
+    @classmethod
+    def from_package(cls, package: str, resource: str) -> Skills:
```
wzhero1: **API asymmetry.** Java `fromClasspath(String...)` is varargs; Python `from_package(package, resource)` is single-pair, so two packages require two `@skills` functions.

Make it symmetric: `def from_package(cls, *pairs: Tuple[str, str])`. See also the cross-language drop thread on the `packages` field.

api/src/main/java/org/apache/flink/agents/api/skills/Skills.java
wenjin272: In current design, skills with the same name are not supported. If there are future user requirements, we can expand further.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/URLSkillRepository.java
wenjin272: By design, we do not intend to support URL download validation and security protections in this version. However, I believe this could indeed become an issue in production environments, so we should consider adding it in the next release.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
wenjin272: Similar to the question above, I believe we can iterate on and implement this improvement in future versions.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java
wenjin272: This behavior will be optimized in subsequent iterations.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/ClasspathSkillRepository.java
wenjin272: What is the significance of highlighting `URLClassLoader` and `ChildFirstClassLoader` here? I understand that ChildFirstClassLoader is also a subclass of URLClassLoader.

Since we set the contextClassLoader to Flink's UserCodeClassloader before executing a JavaAction, it is used in most scenarios. However, there was an issue with the initialization of the skill-managed path in Python Actions. I have modified the code to explicitly pass the UserCodeClassloader.

Typically, I understand that the Flink user JAR does not use the jar-in-jar approach.

`getResourceAsStream` requires a file path as input, but the `resource` here is typically a directory name, so getResourceAsStream cannot be used.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/ClasspathSkillRepository.java
wenjin272: I think merge across all matches makes sense.

wenjin272:
Hi, @wzhero1, Ty for your thorough review. 

I fixed some design issues and bugs, including

  - Skills rewritten as a single discriminated sources: List<SkillSourceSpec(scheme, params)> + SkillSourceRegistry. Uniform Java/Python wire format.
  - URL / Classpath / Package / FileSystem repos are now sibling SkillRepository impls composing a shared SkillDirectoryReader. No more inheritance-as-composition.
  - SkillOrigin(scheme, location) attached to every AgentSkill at registration.
  - SkillRepository extends AutoCloseable; SkillManager.close() cascades through ResourceContextImpl on operator close (failover-safe).
  - Duplicate skill names now LOG.warn with both origins; last-write-wins kept.
  - All matching classpath JARs merged via extractClasspathFromJars; same-path collisions WARN.
  - Single-pass zip extract; JAR_URL_PREFIX_LEN named constant.

The remaining issues related to robustness and security will be iterated on in subsequent phases.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java
```diff
@@ -111,38 +148,94 @@ public List<String> getSkillDirs(List<String> names) {
...
+                registerRepo(repo, originOf(spec));
+            } catch (IOException | IllegalArgumentException e) {
+                throw new IllegalStateException(
+                        "Failed to load skills from " + spec.getScheme() + ":" + spec.getParams(),
+                        e);
+            }
+        }
+    }
+
+    /** Build a {@link SkillOrigin} from a spec for diagnostics (WARN on duplicates, etc.). */
```
wzhero1: **`originOf` re-introduces the parallel scheme ladder the registry was meant to eliminate.**

```java
switch (spec.getScheme()) {
    case "local": ...; case "url": ...;
    case "classpath": ...; case "package": ...;
}
```
Python's `_origin_of` mirrors this. Adding scheme `s3` now requires 3 file edits (registry ×2 + originOf ×2), not 1. Java's `case "package"` is dead code — no Java `package` handler exists.

Fix: add `default String describeLocation(params)` to `SkillSourceHandler`. Each handler describes its own location at registration.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/AgentSkill.java
```diff
@@ -44,6 +44,7 @@ public final class AgentSkill {
     @Nullable private volatile Map<String, String> resources;
     @Nullable private Supplier<Map<String, String>> resourceLoader;
     private volatile boolean activated;
+    @Nullable private volatile SkillOrigin origin;
```
wzhero1: **Java/Python serialization surface for `origin` is asymmetric.**

Python's `AgentSkill` extends `BaseModel`, so `origin` is in default `model_dump()` output. Java's `AgentSkill` is a POJO with no Jackson annotation — any future debug/checkpoint/Pemja serialization path silently drops it on Java while Python emits it. Same class of cross-language drop A3 fixed at the `Skills` envelope.

Add `@JsonProperty("origin")` + `@JsonInclude(NON_NULL)`. Near-zero cost now; cost grows once `AgentSkill` travels on the wire.

wzhero1 (commented):
Round-2 follow-up review against the new commits (`485573a` → `d6d58d8`). Per-comment scope: lifecycle / cross-language consistency / API hygiene gaps left after the prior round, plus one architectural smell carried into the registry refactor. None of these block merge individually; the constructor-leak (#5) and the non-deterministic merge order (#7) are the most worth addressing this PR.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/URLSkillRepository.java
```diff
@@ -0,0 +1,95 @@
...
+ * close()} is never called.
+ *
+ * <p>Composes a {@link SkillDirectoryReader} for the on-disk parsing; the materialization happens
+ * upfront in the constructor (no abuse of {@code super(...)}).
+ */
+public final class URLSkillRepository implements SkillRepository {
+
+    private static final int REQUEST_TIMEOUT_MS = 90_000;
+
+    private final String url;
```
wzhero1, line 38: **Field name `owned` is an adjective without a subject.** maybe `ownerMaterialization`

Same applies to `ClasspathSkillRepository.java:73` and `FileSystemSkillRepository.java:51`.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillRepository.java
```diff
@@ -20,14 +20,19 @@
...
 
+import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 
 /**
  * Source of skills. Mirrors the Python {@code
  * flink_agents.runtime.skill.skill_repository.SkillRepository}.
+ *
+ * <p>Implementations that own a materialized temp directory (URL / classpath / zip) should override
```
wzhero1, line 31: **Three sibling repos duplicate the same 5 delegating methods + state** (`reader`, `materialization`, plus `getSkill`/`getSkills`/`getResources`/`getSkillDir`/`close`) — ~45 lines of boilerplate, +15 per future source.

Extract `AbstractMaterializedSkillRepository` holding the shared state and implementing the delegators once; concrete repos reduce to source-specific I/O + an abstract `getOrigin()`.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java
wzhero1: **`originOf` re-introduces the parallel scheme ladder the registry was meant to eliminate.**

```java
switch (spec.getScheme()) {
    case "local": ...; case "url": ...;
    case "classpath": ...; case "package": ...;
}
```
Python's `_origin_of` mirrors this. Adding scheme `s3` now requires 3 file edits (registry ×2 + originOf ×2), not 1. Java's `case "package"` is dead code — no Java `package` handler exists.

Fix: add `default String describeLocation(params)` to `SkillSourceHandler`. Each handler describes its own location at registration.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/AgentSkill.java
wzhero1, line 47: **Java/Python serialization surface for `origin` is asymmetric.**

Python's `AgentSkill` extends `BaseModel`, so `origin` is in default `model_dump()` output. Java's `AgentSkill` is a POJO with no Jackson annotation — any future debug/checkpoint/Pemja serialization path silently drops it on Java while Python emits it. Same class of cross-language drop A3 fixed at the `Skills` envelope.

Add `@JsonProperty("origin")` + `@JsonInclude(NON_NULL)`. Near-zero cost now; cost grows once `AgentSkill` travels on the wire.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java
```diff
@@ -18,33 +18,63 @@
...
     private final Map<String, SkillRepository> repos = new HashMap<>();
 
-    public SkillManager(Skills config) {
+    /**
+     * Construct a {@code SkillManager} that resolves {@code classpath:} sources against {@code
+     * classLoader}. Production code passes the Flink user-code class loader (threaded through
+     * {@code ResourceCache} from {@code ActionExecutionOperator}); tests / standalone use may use
+     * {@link #SkillManager(Skills)}.
+     */
+    public SkillManager(Skills config, ClassLoader classLoader) {
```
wzhero1, line 72: **Constructor-failure leak: partial `loadAll()` leaves registered repos with no closeable handle.**

If the 2nd of 3 sources succeeds (repo + hook registered) but the 3rd throws, the constructor exits with `IllegalStateException` → caller never gets a `SkillManager` reference → cannot call `close()` → 2nd repo's hook + temp dir leak until JVM exit.

Same monotonic-leak class as review #10, just triggered by partial-construction instead of hot-reload. 5-line fix: wrap `loadAll`'s loop in try/catch, call `close()` before rethrowing.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java
```diff
@@ -111,38 +148,94 @@ public List<String> getSkillDirs(List<String> names) {
...
+
+    /**
+     * Close every owned {@link SkillRepository}, releasing any temp directories materialized for
+     * URL / classpath-zip / classpath-jar sources. Idempotent.
+     */
+    @Override
+    public void close() {
+        // repos may map multiple skill names to the same repo instance; dedup by identity.
+        Set<SkillRepository> unique = Collections.newSetFromMap(new IdentityHashMap<>());
+        unique.addAll(repos.values());
```
wzhero1: **`SkillManager.close()` swallows exceptions silently; inconsistent with `ResourceCache.close()`.**

```java
for (SkillRepository repo : unique) {
    try { repo.close(); } catch (Exception ignored) {}
}
```
`ResourceCache.close()` (line 135-160) collects first-exception + suppressed and rethrows. Same close cascade, two styles. Real failures (locked file, permission denied, disk full) give callers no signal. `ResourceContextImpl.ensureSkillManager()` line 116 has the same silent catch on config swap.

Minimum: `LOG.warn(..., e)`. Preferable: align with `ResourceCache.close()`'s rethrow pattern.

wzhero1 (commented):
Round-2 follow-up review against the new commits (`485573a` → `d6d58d8`). Per-comment scope: lifecycle / cross-language consistency / API hygiene gaps left after the prior round, plus one architectural smell carried into the registry refactor. None of these block merge individually; the constructor-leak (#5) and the non-deterministic merge order (#7) are the most worth addressing this PR.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -532,11 +533,11 @@ private void addSkills(Map<String, Skills> skillsObjects) throws Exception {
                         ResourceType.TOOL,
                         new ResourceDescriptor(BashTool.class.getName(), new HashMap<>())));
 
-        LinkedHashSet<String> paths = new LinkedHashSet<>();
+        LinkedHashSet<SkillSourceSpec> sources = new LinkedHashSet<>();
         for (Skills s : skillsObjects.values()) {
```
wzhero1: **Multi-`@Skills` merge order is non-deterministic on Java; WARN reports collisions but the "winner" is unspecified.**

`AgentPlan.addSkills` iterates `skillsObjects.values()` in `LinkedHashMap` order, but insertion order comes from `Class.getMethods()` — JLS doesn't guarantee declaration order. Test self-discloses at `AgentPlanDeclareSkillsTest:163`: "Class.getMethods() does not preserve declaration order".

Two `@Skills` methods with the same skill name pick different winners across deployments / JDK versions; WARN fires but is non-reproducible. Python `dict` preserves insertion order — Python is deterministic, Java is the bad case.

Min fix: sort method names lexicographically before merging + javadoc note. Stronger: `@Skills(priority=N)` or fail-fast on collision.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/AgentSkill.java
wenjin272: Took a closer look and decided to skip this for now. Grep'd every `AgentSkill` reference in the repo (Java + Python) — no serialization path uses it today, and Jackson's
  default bean introspection would emit `getOrigin()` as `"origin"` automatically even without the annotation, so the "silently drops" failure mode doesn't actually fire under
  the project's default `ObjectMapper`.

  Also worth noting: I ran Python's `model_dump()` and it emits `"origin": null` by default (pydantic v2 doesn't drop None fields), so `@JsonInclude(NON_NULL)` would have made
  the two sides byte-asymmetric. I'd rather design the schema (NON_NULL vs null, `resources` handling, `@JsonCreator`) when there's an actual wire path with concrete 
  requirements. Happy to revisit if you have a specific path in mind.

wenjin272:
Hi, @wzhero1. Ty for the careful review. 

I have updated and addressed the second round comments:

 - **`owned` rename + 5× delegating boilerplate**: the three repos collapsed into `AbstractMaterializedSkillRepository` (Java) / `MaterializedSkillRepository` (Python); the
  field becomes `materialization` on the base. Python's `PackageSkillRepository` simplified to own its tempdir (no more `as_file` ctx + atexit two-step). — 68dc94bb
  
  - **`originOf` scheme ladder**: moved onto `SkillSourceHandler.describeLocation`; built-ins register both functions together. Dead `case "package"` is gone. — 85caed10
  
  - **Partial-load repo leak + `close()` swallow**: constructor closes already-registered repos before propagating; close cascade aligned with `ResourceCache.close()`
  (first-exception rethrow + suppressed). — 47360d2c
  
  - **`@Skills` merge non-determinism**: `addSkills` sorts keys lexically before merging; mixed-scheme test re-asserts as `List` to lock the order. — dcab8359
  
  - **`AgentSkill` JSON asymmetry**: not taking — see inline reply.

  Six of seven taken; cross-language symmetry preserved on every fix.

python/flink_agents/runtime/skill/repository/skill_directory_reader.py
```diff
@@ -0,0 +1,137 @@
...
+            root_path = Path(root)
+            for file_name in files:
+                if file_name == self.SKILL_MD_FILE:
+                    continue
+                file_path = root_path / file_name
+                relative_path = str(file_path.relative_to(skill_dir))
+                try:
+                    resources[relative_path] = file_path.read_text()
+                except UnicodeDecodeError:
+                    resources[relative_path] = f"base64: {file_path.read_bytes()}"
```
wzhero1: **This branch is not base64-encoding — binary resources are silently corrupted.**

```python
resources[relative_path] = f"base64: {file_path.read_bytes()}"
```
The f-string returns the `bytes` repr (literal `"base64: b'\\x89PNG...'"`), not encoded data. `base64.b64decode(value.removeprefix("base64:"))` will raise on any real payload. Java sibling uses `Base64.getEncoder().encodeToString(...)`, so the cross-language contract diverges silently and every binary resource read via `AgentSkill.get_resource()` on Python is unusable.

Fix: `"base64:" + base64.b64encode(file_path.read_bytes()).decode("ascii")`. Worth adding a PNG round-trip test through both languages — none exists today.

python/flink_agents/runtime/skill/skill_manager.py
```diff
@@ -93,42 +108,91 @@ def get_skill_dirs(self, *names: str) -> List[str]:
...
+                handler = skill_source_registry.get(spec.scheme)
+                repo = handler.open(spec.params)
+            except (OSError, ValueError) as e:
+                msg = (
+                    f"Failed to load skills from {spec.scheme}:{spec.params}"
                 )
-                self._skills[skill.name] = skill
-                self._repos[skill.name] = repo
+                raise RuntimeError(msg) from e
+            self._register_repo(repo, _origin_of(spec))
```
wzhero1, line 170: **Partial-load leak: a failed source leaves earlier-opened repos uncleaned.**

Java's `loadAll()` now wraps each `open()` in try/catch and closes already-opened repos on failure (round-2 follow-up). Python here still just re-raises — if source #3 of 5 fails, sources #1 and #2 keep their extracted temp dirs and `atexit` handlers for the interpreter's lifetime. In a long-lived TaskManager Python worker every failed reload leaks.

Same try/finally + first-exception pattern as the Java side: collect opened repos in a list, close them on the failure path before re-raising.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java
```diff
@@ -111,38 +148,113 @@ public List<String> getSkillDirs(List<String> names) {
...
+            skill.setOrigin(origin);
+            AgentSkill previous = skills.put(skillName, skill);
+            if (previous != null) {
+                LOG.warn(
+                        "Skill '{}' from {} overrides earlier registration from {}",
+                        skillName,
+                        origin,
+                        previous.getOrigin() == null ? "<unknown>" : previous.getOrigin());
             }
+            repos.put(skillName, repo);
```
wzhero1, line 269: **Duplicate skill name overwrites the repo-index entry; the displaced repo is orphaned from `closeRepos()`. Bug exists in both languages.**

```java
repos.put(skillName, repo);   // overwrites previous repo ref
```
`closeRepos()` iterates `repos.values()`, so when two sources contribute skill `"foo"` the displaced repo is no longer reachable — its temp dir survives until JVM exit, only the shutdown hook ever fires. The recently-added close-cascade plumbing can't help because the orphan is never visited. Python mirrors the same shape at `python/.../skill_manager.py:177`.

Fix: track opened repos in a separate `List<SkillRepository>` and iterate *that* in `closeRepos()` (dedup with `IdentityHashMap`).

wenjin272:
Hi, @wzhero1, thanks for round three. All three taken:

  - **Python base64 silently corrupting binary resources**: spot on — f-string was inlining the bytes repr, not real base64. Switched to
  `base64.b64encode(...).decode("ascii")`. The single-line `startsWith("base64: ")` assertion is exactly how this hid for a cycle; both Java and Python tests now b64-decode the
   value and assert the round-trip against a PNG signature + non-UTF-8 byte sequence. — e7956487
  - **Python partial-load leak**: Python `_load_skills` now closes already-opened repos before re-raising, mirroring the Java fix from round two. — aa7c9785
  - **Duplicate-name repo orphan in `close()`**: confirmed on both sides — the skill-name → repo map drops the displaced repo's reference, so iterating it on close orphaned the
   earlier repo. Added an `openedRepos`/`_opened_repos` list (identity / `id()` dedup) and changed close to iterate that. New test on each side: two sources contributing the
  same skill name, both repos must be closed. — 8b10d3ab

wzhero1:
approved

Issue #580, opened 2026-03-19
[Feature] Support TTL on Short Term Memory


avx-wlauer:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support controlling TTL on Short Term memory.

Short term memory is backed by Flink's standard MapState facility. Unlike Sensory memory, Short term memory is supposed to last beyond the current call and be available for future operations on the key. The current implementaiton doesn't specify a TTL on the MapState that underlies the short term memory, but in order to control the size of retained state for a long running pipeline, specifying this is important. A new API should be added that allows specifying a StateTtlConfig for the memory, which would get passed down and used when creating the underlying MapState.

Sensory memory doesn't need this facility, as it isn't supposed to last beyond the current call. And Long term memory is handled differently. It also likely needs a way to specify TTL, but the mechanism should be completely different due to the architecture.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wenjin272:
Thanks for offering the contribution, @avx-wlauer. For short-term memory, TTL is indeed crucial. I've assigned you to the ticket.

da-daken:
> > hi@wenjin272 [@avx-wlauer](https://github.com/avx-wlauer) , If short-term memory can be converted into long-term memory through some mechanism, wouldn't that solve the above problem, without using TTL?
> 
> I think periodically converting short-term memory to long-term memory is certainly a solution. However, we still need to store all the information. Even though long-term memory incorporates compression mechanisms, this may still lead to inflated memory size. If users are only concerned with recent memories, then directly expiring historical data is more reasonable. Therefore, I believe that even if such a conversion is possible, a TTL (Time-To-Live) mechanism is still necessary.

i agree ‘If users are only concerned with recent memories, then directly expiring historical data is more reasonable.’, So should this TTL time be made into a configurable item, allowing users to set how much memory they want to obtain within a certain period.
Additionally, may I ask, I notice that currently there seems to be no connection between short-term memory and long-term memory in the project. Is this included in any future plans?

avx-wlauer:
As much as I'd like to submit a PR for this, I'm already consumed in other work and won't be getting to it for a while. If this is still here when my other crises clear up, I'll try to take a look, but given how things have been going, that might be months.

xintongsong:
No worries. The open-source community is entirely based on the principle of voluntarism. Thanks for proposing this feature anyway.

[...]

PR #657, opened 2026-05-11
[Feature] Support TTL on ShortTermMemory

da-daken:
Linked issue: [#580 ](https://github.com/apache/flink-agents/issues/580)

### Purpose of change

Support controlling TTL on Short Term memory.

### Tests

ut

### API

no api

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/agents/AgentExecutionOptions.java
+++ b/api/src/main/java/org/apache/flink/agents/api/agents/AgentExecutionOptions.java
@@ -19,7 +19,6 @@
 package org.apache.flink.agents.api.agents;
 
 import org.apache.flink.agents.api.configuration.ConfigOption;
-import org.apache.flink.api.common.state.StateTtlConfig;
 
 public class AgentExecutionOptions {
     public static final ConfigOption<Agent.ErrorHandlingStrategy> ERROR_HANDLING_STRATEGY =
@@ -52,17 +51,17 @@ public class AgentExecutionOptions {
     public static final ConfigOption<Long> SHORT_TERM_MEMORY_STATE_TTL_MS =
             new ConfigOption<>("short-term-memory.state-ttl.ms", Long.class, 0L);
 
-    public static final ConfigOption<StateTtlConfig.UpdateType>
+    public static final ConfigOption<ShortTermMemoryTtlUpdate>
             SHORT_TERM_MEMORY_STATE_TTL_UPDATE_TYPE =
                     new ConfigOption<>(
                             "short-term-memory.state-ttl.update-type",
-                            StateTtlConfig.UpdateType.class,
-                            StateTtlConfig.UpdateType.OnReadAndWrite);
+                            ShortTermMemoryTtlUpdate.class,
+                            ShortTermMemoryTtlUpdate.ON_READ_AND_WRITE);
 
-    public static final ConfigOption<StateTtlConfig.StateVisibility>
+    public static final ConfigOption<ShortTermMemoryTtlVisibility>
             SHORT_TERM_MEMORY_STATE_TTL_VISIBILITY =
                     new ConfigOption<>(
                             "short-term-memory.state-ttl.visibility",
-                            StateTtlConfig.StateVisibility.class,
-                            StateTtlConfig.StateVisibility.NeverReturnExpired);
+                            ShortTermMemoryTtlVisibility.class,
+                            ShortTermMemoryTtlVisibility.NEVER_RETURN_EXPIRED);
 }
--- a/api/src/main/java/org/apache/flink/agents/api/agents/AgentExecutionOptions.java
+++ b/api/src/main/java/org/apache/flink/agents/api/agents/AgentExecutionOptions.java
@@ -19,6 +19,7 @@
 package org.apache.flink.agents.api.agents;
 
 import org.apache.flink.agents.api.configuration.ConfigOption;
+import org.apache.flink.api.common.state.StateTtlConfig;
 
 public class AgentExecutionOptions {
     public static final ConfigOption<Agent.ErrorHandlingStrategy> ERROR_HANDLING_STRATEGY =
@@ -47,4 +48,21 @@ public class AgentExecutionOptions {
 
     public static final ConfigOption<Boolean> RAG_ASYNC =
             new ConfigOption<>("rag.async", Boolean.class, true);
+
+    public static final ConfigOption<Long> SHORT_TERM_MEMORY_STATE_TTL_MS =
+            new ConfigOption<>("short-term-memory.state-ttl.ms", Long.class, 0L);
+
+    public static final ConfigOption<StateTtlConfig.UpdateType>
+            SHORT_TERM_MEMORY_STATE_TTL_UPDATE_TYPE =
+                    new ConfigOption<>(
+                            "short-term-memory.state-ttl.update-type",
+                            StateTtlConfig.UpdateType.class,
+                            StateTtlConfig.UpdateType.OnReadAndWrite);
+
+    public static final ConfigOption<StateTtlConfig.StateVisibility>
+            SHORT_TERM_MEMORY_STATE_TTL_VISIBILITY =
+                    new ConfigOption<>(
+                            "short-term-memory.state-ttl.visibility",
+                            StateTtlConfig.StateVisibility.class,
+                            StateTtlConfig.StateVisibility.NeverReturnExpired);
 }
--- a/api/src/main/java/org/apache/flink/agents/api/agents/AgentExecutionOptions.java
+++ b/api/src/main/java/org/apache/flink/agents/api/agents/AgentExecutionOptions.java
@@ -48,9 +48,11 @@ public class AgentExecutionOptions {
     public static final ConfigOption<Boolean> RAG_ASYNC =
             new ConfigOption<>("rag.async", Boolean.class, true);
 
+    /** Set to a positive value in milliseconds to enable short-term memory TTL; 0 disables it. */
     public static final ConfigOption<Long> SHORT_TERM_MEMORY_STATE_TTL_MS =
             new ConfigOption<>("short-term-memory.state-ttl.ms", Long.class, 0L);
 
+    /** Update policy for short-term memory TTL, consulted only when TTL is enabled. */
     public static final ConfigOption<ShortTermMemoryTtlUpdate>
             SHORT_TERM_MEMORY_STATE_TTL_UPDATE_TYPE =
                     new ConfigOption<>(
@@ -58,6 +60,9 @@ public class AgentExecutionOptions {
                             ShortTermMemoryTtlUpdate.class,
                             ShortTermMemoryTtlUpdate.ON_READ_AND_WRITE);
 
+    /**
+     * Visibility policy for expired short-term memory state, consulted only when TTL is enabled.
+     */
     public static final ConfigOption<ShortTermMemoryTtlVisibility>
             SHORT_TERM_MEMORY_STATE_TTL_VISIBILITY =
                     new ConfigOption<>(
--- a/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIResponsesModelConnection.java
+++ b/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIResponsesModelConnection.java
@@ -72,6 +72,7 @@ import java.util.Optional;
  *   public static ResourceDesc openAIResponses() {
  *     return ResourceDescriptor.Builder.newBuilder(OpenAIResponsesModelConnection.class.getName())
  *             .addInitialArgument("api_key", System.getenv("OPENAI_API_KEY"))
+ *             .addInitialArgument("api_base_url", System.getenv("OPENAI_API_URL"))
  *             .addInitialArgument("timeout", 120)
  *             .addInitialArgument("max_retries", 3)
  *             .build();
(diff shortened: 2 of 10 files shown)


wenjin272:
Hi, @da-daken, thanks for your contribution. Due to the imminent code freeze for version 0.3, I will prioritize handling some must-have items for this release. This may delay my review of your two PRs, but I estimate that I will complete the reviews within a week at the latest.

runtime/src/test/java/org/apache/flink/agents/runtime/memory/ShortTermMemoryTTLIntegrationTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Test
+    void testTTLConfigurationNotApplied() throws Exception {
+        List<String> results = runScenario(1000L, 0L);
+
+        assertEquals(List.of("event1|NEW", "event2|NEW", "event1|EXISTING"), results);
+    }
+
+    @Test
+    void testTTLConfigurationApplied() throws Exception {
```
weiqingy: Two gaps in coverage worth closing while we're here:

- **TTL disabled path.** No test exercises the early-return in `maybeEnableShortTermMemoryTTL` (`TTL_MS = 0` or unset). That branch is the default for every existing pipeline, so a regression that accidentally enables TTL with 0ms would be silent. A short scenario with `ttlMs = 0` confirming the third `event1` returns `EXISTING` regardless of sleep would cover it.
- **Default update-type / visibility.** Both tests explicitly override the update-type to `OnCreateAndWrite`. The option defaults (`OnReadAndWrite` / `NeverReturnExpired`) are never actually exercised, so a regression in how the defaults are wired would slip through. One scenario that sets only `TTL_MS` and asserts the same expiry behavior would catch that.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/OperatorStateManager.java
```diff
@@ -121,6 +129,39 @@ void initializeKeyedStates(org.apache.flink.api.common.functions.RuntimeContext
                                 PENDING_INPUT_EVENT_STATE_NAME, TypeInformation.of(Event.class)));
     }
 
+    /**
+     * When {@link AgentExecutionOptions#SHORT_TERM_MEMORY_STATE_TTL_MS} is positive, attaches Flink
+     * {@link StateTtlConfig} to the short-term memory {@link MapStateDescriptor}. Unset, null, or
+     * non-positive values disable TTL (Flink does not allow zero/negative TTL).
```
weiqingy, line 138: Nice — recording *"Flink does not allow zero/negative TTL"* at the workaround site spares future maintainers a trip through Flink source.

weiqingy (commented):
Thanks for picking this up — a few cross-cutting questions before this lands:

1. **Python mirror is missing.** `python/flink_agents/api/core_options.py` mirrors every entry in the Java `AgentExecutionOptions` today (`MAX_RETRIES`, `RETRY_WAIT_INTERVAL`, the async flags, etc.). The three new short-term-memory TTL options are not added on the Python side, so a Python user can't reach them via the typed API — they'd have to fall back to raw string keys. Could we add the mirrors (and a matching test) in this PR? CLAUDE.md is explicit about Java↔Python parity for shared options.

2. **Are we comfortable exposing `StateTtlConfig.UpdateType` and `StateTtlConfig.StateVisibility` directly in the public API?** Two of the new options take Flink internal enums as their value type, in a class under `org.apache.flink.agents.api.agents`. That couples the public surface to Flink's state API forever, and makes the Python mirror painful (we'd need to redefine the enums on the Python side or fall back to strings). One alternative is to wrap them in our own enums (`ShortTermMemoryTtlUpdate` / `ShortTermMemoryTtlVisibility`) with one-to-one mappings — costs a few lines and a translation step, buys API independence. What do you think?

da-daken:
> Thanks for picking this up — a few cross-cutting questions before this lands:
> 
> 1. **Python mirror is missing.** `python/flink_agents/api/core_options.py` mirrors every entry in the Java `AgentExecutionOptions` today (`MAX_RETRIES`, `RETRY_WAIT_INTERVAL`, the async flags, etc.). The three new short-term-memory TTL options are not added on the Python side, so a Python user can't reach them via the typed API — they'd have to fall back to raw string keys. Could we add the mirrors (and a matching test) in this PR? CLAUDE.md is explicit about Java↔Python parity for shared options.
> 2. **Are we comfortable exposing `StateTtlConfig.UpdateType` and `StateTtlConfig.StateVisibility` directly in the public API?** Two of the new options take Flink internal enums as their value type, in a class under `org.apache.flink.agents.api.agents`. That couples the public surface to Flink's state API forever, and makes the Python mirror painful (we'd need to redefine the enums on the Python side or fall back to strings). One alternative is to wrap them in our own enums (`ShortTermMemoryTtlUpdate` / `ShortTermMemoryTtlVisibility`) with one-to-one mappings — costs a few lines and a translation step, buys API independence. What do you think?

@weiqingy Thanks for your review. You're right, I submitted the new code. PTAL

api/src/main/java/org/apache/flink/agents/api/agents/AgentExecutionOptions.java
```diff
@@ -47,4 +47,21 @@ public class AgentExecutionOptions {
 
     public static final ConfigOption<Boolean> RAG_ASYNC =
             new ConfigOption<>("rag.async", Boolean.class, true);
+
+    public static final ConfigOption<Long> SHORT_TERM_MEMORY_STATE_TTL_MS =
+            new ConfigOption<>("short-term-memory.state-ttl.ms", Long.class, 0L);
```
weiqingy, line 53: Nit:  0L doubles as the "TTL disabled" sentinel, but that contract only lives in OperatorStateManager.maybeEnableShortTermMemoryTTL. A one-line javadoc here — e.g. "Set to a positive value in milliseconds to enable TTL; 0 (the default) disables it" — would spare future readers a trip into the runtime. Same applies to the two enum options below: worth noting they're only consulted when TTL_MS > 0.

runtime/src/test/java/org/apache/flink/agents/runtime/memory/ShortTermMemoryTTLIntegrationTest.java
```diff
@@ -0,0 +1,158 @@
...
+            shortTermMemory.set(input.eventKey, currentCount + 1);
+            Thread.sleep(input.sleepMs);
+            ctx.sendEvent(
+                    new OutputEvent(
+                            input.eventKey + "|" + (existingValue == null ? "NEW" : "EXISTING")));
+        }
+    }
+
+    @Test
+    void testTTLConfigurationNotApplied() throws Exception {
```
weiqingy: Nit: testTTLConfigurationNotApplied vs testTTLConfigurationApplied (line 109) is a bit misleading — TTL is configured in both; what differs is whether enough time elapsed for entries to expire (sleep_ms=0 vs 2000). Names like testValueStillVisibleBeforeTTLExpiry / testValueExpiresAfterTTL would express the actual assertion. Same suggestion for the Python mirror in short_term_memory_ttl_test.py:126,159.

weiqingy:
LGTM — both prior concerns are addressed cleanly. Only two small readability nits inline. Thanks for the quick turnaround.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/OperatorStateManager.java
```diff
@@ -121,6 +131,63 @@ void initializeKeyedStates(org.apache.flink.api.common.functions.RuntimeContext
...
+        switch (stateVisibility) {
+            case NEVER_RETURN_EXPIRED:
+                return StateTtlConfig.StateVisibility.NeverReturnExpired;
+            case RETURN_EXPIRED_IF_NOT_CLEANED_UP:
+                return StateTtlConfig.StateVisibility.ReturnExpiredIfNotCleanedUp;
+            default:
+                throw new IllegalArgumentException(
+                        "Unsupported TTL state visibility: " + stateVisibility);
+        }
+    }
```
xintongsong, line 187: I suggest placing these two methods in `UpdateType` and `StateVisibility`, respectively. They are only related to the two enum types themselves, not to where they are used.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIResponsesModelConnection.java
```diff
@@ -72,6 +74,7 @@
  *   public static ResourceDesc openAIResponses() {
  *     return ResourceDescriptor.Builder.newBuilder(OpenAIResponsesModelConnection.class.getName())
  *             .addInitialArgument("api_key", System.getenv("OPENAI_API_KEY"))
+ *             .addInitialArgument("api_base_url", System.getenv("OPENAI_API_URL"))
```
xintongsong: This change is unrelated to TTL. It should be placed in a separate commit. Usually, in such cases, we add `[hotfix]` commits for fixing existing issues before commits of the actual PR changes.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/OperatorStateManager.java
```diff
@@ -121,6 +128,39 @@ void initializeKeyedStates(org.apache.flink.api.common.functions.RuntimeContext
...
     }
 
+    /**
+     * When {@link AgentExecutionOptions#SHORT_TERM_MEMORY_STATE_TTL_MS} is positive, attaches Flink
+     * {@link StateTtlConfig} to the short-term memory {@link MapStateDescriptor}. Unset, null, or
+     * non-positive values disable TTL (Flink does not allow zero/negative TTL).
+     */
+    private void maybeEnableShortTermMemoryTTL(
+            MapStateDescriptor<String, MemoryObjectImpl.MemoryItem> descriptor,
+            AgentPlan agentPlan) {
```
xintongsong: It's unnecessary to take the whole `AgentPlan` as an argument. `AgentConfiguration` should be enough.

runtime/src/test/java/org/apache/flink/agents/runtime/memory/ShortTermMemoryTTLIntegrationTest.java
```diff
@@ -0,0 +1,158 @@
...
+        List<String> results = runScenario(1000L, 0L, true, true);
+
+        assertEquals(List.of("event1|NEW", "event2|NEW", "event1|EXISTING"), results);
+    }
+
+    @Test
+    void testTTLConfigurationDisabledWithZeroTtl() throws Exception {
+        List<String> results = runScenario(0L, 2000L, true, true);
```
xintongsong: Sleeping for 2s is unnecessarily long. If we expect the state still exist after the sleep, sleep shorter (e.g., 50ms). If we expect the state does not exist, we can set a small ttl (e.g., 50ms) and wait a bit longer (e.g., 200ms). Either way, 2s is too long.

xintongsong (commented):
Thanks @da-daken for working on this, and thanks @weiqingy for the review.

I also left a few comments. In addition to the inline comments, I'd suggest to re-organize the PR commits.
- We do not allow merge commits in PRs. Please use rebase for conflict resolving.
- Please separate unrelated hotfixes from the major changes.
- You may check the [community guidelines](https://flink.apache.org/how-to-contribute/code-style-and-quality-pull-requests/#3-separate-refactoring-cleanup-and-independent-changes) for instructions and examples.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/OperatorStateManager.java
da-daken: hi @xintongsong ,I’d prefer not to put the mapping methods on the public enum types.                                 
                                                                                                    

> Are we comfortable exposing StateTtlConfig.UpdateType and StateTtlConfig.StateVisibility directly in the public API? Two of the new options take Flink internal enums as their value type, in a class under org.apache.flink.agents.api.agents. That couples the public surface to Flink's state API forever, and makes the Python mirror painful (we'd need to redefine the enums on the Python side or fall back to strings). 

As mentioned in the earlier review from @weiqingy ,  exposing Flink’s StateTtlConfig.UpdateType /                    
StateTtlConfig.StateVisibility directly would couple the Agents public API to Flink state APIs and
make the Python mirror harder. If ShortTermMemoryTtlUpdate or ShortTermMemoryTtlVisibility exposes  
toFlink...() methods, we effectively reintroduce that coupling through the enum API.

I agree the mapping does not have to stay in OperatorStateManager.A good compromise would be to move it into a runtime-internal helper, for example:
```java
final class ShortTermMemoryTtlConfigMapper {                                                        
      static StateTtlConfig.UpdateType toFlinkUpdateType(ShortTermMemoryTtlUpdate updateType) {       
          ...                                                                                         
      }                                                                                               
                                                                                                      
      static StateTtlConfig.StateVisibility toFlinkStateVisibility(                                   
              ShortTermMemoryTtlVisibility visibility) {                                              
          ...                                                                                         
      }                                                                                               
  }
```

xintongsong: Makes sense to me. Thanks for the clarification.

da-daken:
> Thanks @da-daken for working on this, and thanks @weiqingy for the review.
> 
> I also left a few comments. In addition to the inline comments, I'd suggest to re-organize the PR commits.
> 
> * We do not allow merge commits in PRs. Please use rebase for conflict resolving.
> * Please separate unrelated hotfixes from the major changes.
> * You may check the [community guidelines](https://flink.apache.org/how-to-contribute/code-style-and-quality-pull-requests/#3-separate-refactoring-cleanup-and-independent-changes) for instructions and examples.

hi @xintongsong , I have rebased the branch, adopted your review suggestions, and removed the changes in OpenAIResponsesModelConnection because 'api_base_url' is optional, so I will not submit the hotfix. PTAL~

xintongsong (approved):
@da-daken Thanks for addressing my comments. LGTM.

There's one thing left, which is updating the user documentation w.r.t. this feature. We should at least add the new config options to the configuration page, and potentially also add a paragraph in the short-term memory page. 

I'll merge this PR for now, and leave #580 open until the documentation is completed.

Issue #661, opened 2026-05-12
[Feature][integration][Java] Java Implementation of Vector Store with Milvus Support


yanbinyang:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Following the design proposal in [#143](https://github.com/apache/flink-agents/discussions/143) and infrastructure introduced in [#321](https://github.com/apache/flink-agents/issues/321), we want to add a Milvus-based vector store implementation to enhance RAG capabilities with Milvus in Flink Agents. This issue focuses on implementing the Java Milvus vector store integration.

The implementation should include:

- Add `MilvusVectorStore` under `integrations/vector-stores/milvus`
- Reuse the existing `BaseVectorStore`, `Document`, `VectorStoreQuery`, and `VectorStoreQueryResult` APIs
- Support Milvus connection configuration, such as `uri`, `token`, `username`, `password`, and `database`
- Support collection/search configuration, such as `collection_name`, `vector_field`, `text_field`, `primary_key_field`, `output_fields`, and search params
- Convert Milvus query results into the common `Document` type
- Add unit tests and/or integration tests for query argument parsing and result conversion

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wenjin272:
Thanks @yanbinyang. Assigned.

PR #663, opened 2026-05-12
[Feature][Integration][Java] Add Milvus vector store integration

yanbinyang:
Linked issue: [#661](https://github.com/apache/flink-agents/issues/661)

### Purpose of change

This PR implements the Java version of Milvus-based Vector Store functionality for Flink Agents, following the design proposal in [#143](https://github.com/apache/flink-agents/discussions/143). This implementation enables RAG (Retrieval-Augmented Generation) capabilities by providing vector-based context retrieval.

- Added Milvus vector store integration
- Added Milvus vector store module under `integrations/vector-stores`
- Supported Milvus connection and query configuration
- Converted Milvus search results into the common `Document` format
- Added tests for Milvus vector store behavior

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/vector_stores.md
+++ b/docs/content/docs/development/vector_stores.md
@@ -41,6 +41,7 @@ In Flink Agents, vector stores are essential for:
 
 ### Concepts
 * **Document**: Document is the abstraction that represents a piece of text and associated metadata. A document may also carry a pre-computed `embedding` vector and a `score` populated by query results.
+* **Collection**: Collection is the abstraction that represents a set of documents. It corresponds to different concept for different vector store specification, like index in Elasticsearch/OpenSearch and collection in Chroma/Milvus.
 * **Filter DSL**: A unified, equality-only metadata filter dialect shared by `query`, `get`, and `delete`. The DSL covers only the subset every supported backend can honour (equality matching), so callers don't need to know each store's native operators. See the [Filter DSL](#filter-dsl) section below for details.
 
 ## How to use
@@ -172,7 +173,7 @@ For vector stores that implement `CollectionManageableVectorStore`, you can crea
 * `delete_collection` / `deleteCollection`: Delete a collection by name.
 
 {{< hint info >}}
-Collection-level operations are only supported for vector stores that implement `CollectionManageableVectorStore`. Among the built-in providers, Chroma (Python), Elasticsearch (Java) and OpenSearch (Java) implement this interface.
+Collection-level operations are only supported for vector stores that implement `CollectionManageableVectorStore`. Among the built-in providers, Chroma (Python), Elasticsearch (Java), OpenSearch (Java), and Milvus (Java) implement this interface.
 {{< /hint >}}
 
 {{< tabs "Collection level operations" >}}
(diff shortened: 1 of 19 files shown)


docs/content/docs/development/vector_stores.md
```diff
@@ -41,6 +41,7 @@ In Flink Agents, vector stores are essential for:
 
 ### Concepts
 * **Document**: Document is the abstraction that represents a piece of text and associated metadata. A document may also carry a pre-computed `embedding` vector and a `score` populated by query results.
+* **Collection**: Collection is the abstraction that represents a set of documents. It corresponds to different concept for different vector store specification, like index in Elasticsearch/OpenSearch and collection in Chroma/Milvus.
```
wenjin272: It looks like there were some conflicts during the rebase. In a previous PR, we removed the `Collection` data structure and this line of description.

integrations/vector-stores/milvus/src/main/java/org/apache/flink/agents/integrations/vectorstores/milvus/MilvusVectorStore.java
```diff
@@ -0,0 +1,1185 @@
...
+        kwargs.put("consistency_level", this.consistencyLevel.name());
+        kwargs.put("load_timeout_ms", this.loadTimeoutMs);
+        if (this.databaseName != null) {
+            kwargs.put("db_name", this.databaseName);
+        }
+        return kwargs;
+    }
+
+    /** Returns the lazily-created Milvus client. */
+    private MilvusClientV2 client() {
```
wenjin272, line 324: Since flink-agents supports asynchronous execution of actions, meaning that multiple threads can invoke the same vector store simultaneously, connection leaks caused by multi-threading may occur here.

I suggest adding DCL.

tools/docker/elasticsearch/docker-compose.yml
wenjin272, line 1: It appears that this file is not being used in the CI. I'm wondering if we could also align the current Elasticsearch startup process in the CI with that of Milvus.

wenjin272 (commented):
Thanks for your contribution @yanbinyang.  Overall looks good to me, and I left serveral minor comments.

I think we should also verify Milvus in the test case `Mem0LongTermMemoryTest`. This test is currently disabled due to a bug in Flink itself, which is expected to be fixed in the coming weeks. There are also workarounds available for local execution. You can add the test path for Milvus, and I will verify it locally.

docs/content/docs/development/vector_stores.md
yanbinyang: > It looks like there were some conflicts during the rebase. In a previous PR, we removed the `Collection` data structure and this line of description.

Thanks for catching this. Removed the stale `Collection` concept line in ae1261a

integrations/vector-stores/milvus/src/main/java/org/apache/flink/agents/integrations/vectorstores/milvus/MilvusVectorStore.java
yanbinyang: > Since flink-agents supports asynchronous execution of actions, meaning that multiple threads can invoke the same vector store simultaneously, connection leaks caused by multi-threading may occur here.
> 
> I suggest adding DCL.

Good point. Addressed in 56cc845 by making the Milvus client `volatile` and using double-checked locking for lazy initialization. I also synchronized `close()` so concurrent action execution will not create or leak multiple clients.

tools/docker/elasticsearch/docker-compose.yml
yanbinyang: > It appears that this file is not being used in the CI. I'm wondering if we could also align the current Elasticsearch startup process in the CI with that of Milvus.

Addressed in 37c5436. The cross-language CI job now starts and stops Elasticsearch via `tools/docker/elasticsearch/docker-compose.yml`, aligned with the Milvus startup flow.

yanbinyang:
> Thanks for your contribution @yanbinyang. Overall looks good to me, and I left serveral minor comments.
> 
> I think we should also verify Milvus in the test case `Mem0LongTermMemoryTest`. This test is currently disabled due to a bug in Flink itself, which is expected to be fixed in the coming weeks. There are also workarounds available for local execution. You can add the test path for Milvus, and I will verify it locally.

Thanks for the suggestion. Addressed in fa34054 by adding a Milvus path to `Mem0LongTermMemoryTest`: the test now parameterizes the Mem0 vector store over Elasticsearch and Milvus, and `Mem0LongTermMemoryAgent` declares a `milvusLtmStore` resource.

I couldn't run this e2e test locally because it still depends on the existing Pemja/Flink issue workaround and a local Milvus/Mem0 environment. Could you please help verify the Milvus path locally?

wenjin272 (commented):
The code looks good to me, and I have verified using mem0 with mivlus locally, it works well.

However, there are still some issues with the commit organization in this PR. We require each PR branch to be rebased onto main rather than merged, in order to maintain a linear history on the main branch.

yanbinyang:
Thanks. I have rebased the branch onto the latest `main` and force-pushed the updated linear history. The merge commit has been removed, and the PR is now mergeable.

wenjin272 (approved):
LGTM

Issue #645, opened 2026-05-01
[Bug][runtime] Fix memory leak in DurableExecutionManager.checkpointIdToSeqNums


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Summary

`DurableExecutionManager.recordCheckpointSequenceNumbers()` unconditionally adds entries to `checkpointIdToSeqNums`, but `notifyCheckpointComplete()` only cleans them up when `actionStateStore != null`. When durable execution is disabled (`actionStateStore == null`), checkpoint entries accumulate indefinitely — a memory leak.

## Root Cause

```java
// Always records — no null guard
void recordCheckpointSequenceNumbers(long checkpointId, Map<Object, Long> seqNums) {
    checkpointIdToSeqNums.put(checkpointId, seqNums);
}

// Only cleans up when actionStateStore is non-null
void notifyCheckpointComplete(long checkpointId) {
    if (actionStateStore != null) {
        ...
        checkpointIdToSeqNums.remove(checkpointId);  // never reached when actionStateStore == null
    }
}
```

## Fix

Add a `checkpointIdToSeqNums.remove(checkpointId)` call unconditionally in `notifyCheckpointComplete`, or guard `recordCheckpointSequenceNumbers` with an early return when `actionStateStore == null`.

## Notes

- This bug existed before the `ActionExecutionOperator` refactoring (PR #546); it was surfaced during review of that PR by @wenjin272.
- The fix and a test verifying the behaviour can be added in the same PR.

### How to reproduce

This is a structural code bug — no specific input data is needed to trigger it.

1. Configure an agent job with durable execution **disabled** (i.e., do not set `ActionStateStore` — the default).
2. Enable Flink checkpointing (e.g., `env.enableCheckpointing(60_000)`).
3. Run the job and allow several checkpoints to complete.

**Expected**: `checkpointIdToSeqNums` stays empty (no entries needed when durable execution is off).  
**Actual**: Each completed checkpoint adds an entry via `recordCheckpointSequenceNumbers()` that is never removed, because `notifyCheckpointComplete()` skips cleanup when `actionStateStore == null`. The map grows unbounded for the lifetime of the job.

### Version and environment

- **Flink Agents**: 0.3-SNAPSHOT (identified in PR #546 review)
- **Apache Flink**: 2.2.0
- **Java**: 11
- **Deployment mode**: any (bug is in operator logic, not deployment-specific)

[...]

PR #666, opened 2026-05-13
[runtime] Lock null-store symmetry invariant in DurableExecutionManager

weiqingy:
Issue #645 originally described a memory leak in checkpointIdToSeqNums when actionStateStore == null: an unconditional put on the snapshot side paired with a guarded remove on the cleanup side, causing the map to grow unboundedly when durable execution was disabled.

That asymmetry was structurally fixed during the DurableExecutionManager extraction in #546, which added a symmetric actionStateStore == null early-return at the top of snapshotLastCompletedSequenceNumbers. The symmetry is now held by two independent conditionals with no test or javadoc tying them together — a future refactor that drops either guard would silently reintroduce the leak.

This commit locks the invariant in place:

- Add a @VisibleForTesting accessor for checkpointIdToSeqNums mirroring the existing getActionStateStore() precedent.
- Add a regression test that, with actionStateStore == null, runs two snapshot + notifyCheckpointComplete cycles and asserts the map stays empty. verifyNoInteractions on the mock KeyedStateBackend additionally proves the snapshot-side early-return fires before any backend access.
- Strengthen javadoc on both methods to call out the symmetric guard invariant and cross-link to issue #645.

No production behavior change.

### Documentation

- [ ] `doc-needed` 
- [X] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
@@ -333,6 +333,12 @@ class DurableExecutionManager implements ActionStatePersister, AutoCloseable {
      * via {@link #snapshotLastCompletedSequenceNumbers}. After pruning, the entry for that
      * checkpoint is removed. No-op when durable execution is disabled.
      *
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):
LGTM

Issue #646, opened 2026-05-01
[Tech Debt] [runtime] Add comprehensive tests for ActionExecutionOperator manager classes


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Summary

PR #546 introduced 5 focused manager classes by decomposing `ActionExecutionOperator`. Minimal contract tests (4 files, 7 tests) were added to keep that PR reviewable. Heavier coverage was explicitly deferred and is tracked here.

## Deferred Coverage

| Manager | Missing Coverage |
|---------|-----------------|
| `OperatorStateManager` | Flink-harness-based tests (real `StateInitializationContext`, keyed state, operator state backend) |
| `DurableExecutionManager` | Full snapshot + recovery cycle; `notifyCheckpointComplete` pruning behaviour |
| `EventRouter` | Listener notification on event processed; watermark draining order under concurrent keys |
| `ActionTaskContextManager` | Full memory-context lifecycle (create → transfer → complete → cleanup) |

## Notes

- The contract tests added in PR #546 (`e2e4e03`) cover the core promise of each manager but use mocks and simple stubs.
- Flink-harness tests require `AbstractStreamOperatorTestHarness` and real Flink state backends — heavier setup, better suited to a focused follow-up (keep PR #546 reviewable).

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #659, opened 2026-05-12
[runtime] Add comprehensive tests for ActionExecutionOperator manager classes 

weiqingy:
Linked issue: #646

### Purpose of change

Closes the heavier test coverage explicitly deferred at PR #546's review for the manager classes that decomposed `ActionExecutionOperator`. PR #546 added 4 contract test files (7 tests, 295 lines) using mocks and stubs; heavier coverage — snapshot/recovery cycle, listener notification, watermark draining order, memory-context lifecycle — was deferred and tracked in #646. This PR closes that gap.

**Scope** (matches issue #646's deferred-coverage table):

| Manager | New coverage |
|---------|--------------|
| `DurableExecutionManager` | Full snapshot + recovery cycle; `notifyCheckpointComplete` pruning behaviour (3 new tests) |
| `EventRouter` | Listener notification on event processed; watermark draining order under concurrent keys (5 new tests) |
| `ActionTaskContextManager` | Full memory-context lifecycle: create → transfer → complete → cleanup (5 new tests) |

`OperatorStateManager` is dropped from this PR — its observable contracts are already covered by `ActionExecutionOperatorTest.java:176–231` (rescale + recovery) and `:254–306` (durable state recovery). Adding a manager-scoped harness test would duplicate without adding coverage.

`PythonBridgeManager` is out of scope per the issue body's deferred-coverage table.

### Commit structure (4 commits)

1. **[runtime] Add @VisibleForTesting seams for manager tests** — three narrow seams in production code, all following the existing `DurableExecutionManager.getActionStateStore()` precedent:
   - `DurableExecutionManager.getCheckpointIdToSeqNums()` accessor
   - `EventRouter.addEventListener(EventListener)` accessor
   - `EventRouter` 3-arg `@VisibleForTesting` constructor for `EventLogger` injection (2-arg constructor delegates; production signature unchanged)
2. **[runtime] Add comprehensive tests for DurableExecutionManager** — 3 new tests
3. **[runtime] Add comprehensive tests for EventRouter** — 5 new tests
4. **[runtime] Add comprehensive tests for ActionTaskContextManager** — 5 new tests

### Tests

- **13 new tests** added; 6 existing manager tests untouched. After this PR: 5 + 7 + 7 + 1 = **20 tests** across the four manager test classes (was 7).
- Full `runtime` module: **262 tests, 0 failures, 0 errors, 0 skipped**.
- Spotless format check: `### FORMAT CHECK PASSED ###`.
- License check: clean (no tracked files missing headers).

[...]

PR #667, opened 2026-05-13
[runtime] Handle notifyCheckpointAborted to stop leaking checkpoint entries

### LINKED ISSUE #665: [Bug][runtime] DurableExecutionManager leaks checkpointIdToSeqNums entries on aborted checkpoints
### Search before asking

## Summary

`DurableExecutionManager` records per-key sequence numbers under a checkpoint id in `snapshotLastCompletedSequenceNumbers(...)`, and `notifyCheckpointComplete(...)` later removes that entry while pruning durable state. However, Flink can also *abort* a checkpoint — in which case the framework calls `notifyCheckpointAborted(...)` instead of `notifyCheckpointComplete(...)`. The `DurableExecutionManager` does not currently override `notifyCheckpointAborted`, so the entry recorded for an aborted checkpoint is never removed from `checkpointIdToSeqNums`. When durable execution is enabled, this causes the map to grow over time as checkpoints abort (e.g., due to timeouts, alignment failures, or backend pressure).

This is structurally distinct from the null-store case tracked in #645 — that one is already structurally fixed by the symmetric `actionStateStore == null` guard. The aborted-checkpoint leak only manifests when `actionStateStore != null`.

## Root Cause

```java
// Records entry on snapshot — works fine for completed checkpoints
void snapshotLastCompletedSequenceNumbers(KeyedStateBackend<?> backend, long checkpointId) {
    if (actionStateStore == null) return;
    ...
    checkpointIdToSeqNums.put(checkpointId, keyToSeqNum);
}

// Removes entry only on successful completion
void notifyCheckpointComplete(long checkpointId) {
    if (actionStateStore != null) {
        ...
        checkpointIdToSeqNums.remove(checkpointId);
    }
}

// MISSING: no notifyCheckpointAborted override → aborted checkpoints leak their entry
```

In `ActionExecutionOperator`, only `notifyCheckpointComplete` is wired through to the manager; there is no corresponding `notifyCheckpointAborted` handler that forwards the call to the manager.

## Fix (proposed)

1. Add `notifyCheckpointAborted(long checkpointId)` to `DurableExecutionManager` that removes the corresponding entry from `checkpointIdToSeqNums` (no pruning needed — the durable state for an aborted checkpoint was never committed).
2. Override `notifyCheckpointAborted` on `ActionExecutionOperator` and forward to the manager.
3. Add a regression test in `DurableExecutionManagerTest` covering both:
   - Aborted checkpoint removes its entry.
   - Completed and aborted checkpoints interleaved leave only the in-flight ones in the map.

### How to reproduce

This is a structural code bug — no specific input data is needed.

1. Configure an agent job with durable execution **enabled** (set `ActionStateStore`).
2. Enable Flink checkpointing with conditions that cause some checkpoints to abort (e.g., short timeout, backend pressure).
3. Run the job and observe `checkpointIdToSeqNums` size.

**Expected**: only entries for in-flight (not yet completed or aborted) checkpoints remain.
**Actual**: entries for aborted checkpoints persist for the lifetime of the job, growing unboundedly under sustained abort conditions.

### Version and environment

- **Flink Agents**: 0.3-SNAPSHOT
- **Apache Flink**: 2.2.0
- **Java**: 11
- **Deployment mode**: any (bug is in operator logic)

weiqingy:
Closes #665.

## What this PR does

`DurableExecutionManager.checkpointIdToSeqNums` leaked entries on aborted checkpoints. Flink calls `notifyCheckpointAborted(...)` when a checkpoint is aborted (timeout, alignment failure, backend pressure); the existing code only handled the complete path. Under sustained abort pressure the map grew unboundedly.

### Production changes

1. **`DurableExecutionManager.notifyCheckpointAborted(long)`** (new, package-private) — removes the entry from `checkpointIdToSeqNums`. **No `pruneState` call** — durable writes for an aborted checkpoint were never committed, so the prior committed checkpoint's recovery state is still load-bearing and must not be pruned. Guarded by `actionStateStore != null` to mirror the symmetric guard from #645.
2. **`ActionExecutionOperator.notifyCheckpointAborted(long)`** (new `@Override`) — thin delegate to the manager, then `super.notifyCheckpointAborted(...)`. Mirrors the existing `notifyCheckpointComplete` override exactly.
3. **Javadoc invariant statement** on `snapshotLastCompletedSequenceNumbers` and `notifyCheckpointComplete` strengthened to name BOTH release paths (complete OR abort). The `actionStateStore != null` guard now lives on three methods; the javadoc makes the three-way symmetry explicit and cross-links #645 + #665.
4. **`@VisibleForTesting getCheckpointIdToSeqNums()`** accessor — mirrors `getActionStateStore()` precedent. (Same addition as in #666; the second PR to land drops the duplicate on rebase.)

### Tests (DEM-level, three new)

- `notifyAbortedRemovesEntryWithoutPruning` — entry released, durable state untouched. Uses a real `InMemoryActionStateStore` (not a mock) so wrongful pruning would be observable.
- `completedAndAbortedInterleavedKeepsInFlightEntries` — three in-flight checkpoints; one completes (state pruned), one aborts (state preserved), one remains.
- `noStoreModeNotifyCheckpointAbortedIsNoOp` — symmetric null-store no-op coverage.

Sanity-mutation verified locally:
- Emptying the new method's body → 2 of 3 new tests fail.
- Adding a wrongful `actionStateStore.pruneState(...)` call → 2 of 3 new tests fail (state was incorrectly pruned).

Operator-level harness test deferred to #646 — the new operator override is a one-line delegate; the logic is in the manager.

## Test plan

- [x] `mvn test -Dtest=DurableExecutionManagerTest -pl runtime` — 5/5 pass (2 existing + 3 new)
- [x] `mvn test -Dtest=ActionExecutionOperatorTest -pl runtime` — 28/28 pass
- [x] `mvn test -pl runtime` — 307/307 pass (no regressions)
- [x] `./tools/lint.sh -c` — 0 violations
- [x] `./tools/check-license.sh` — clean (no new tracked files)
- [x] Sanity mutation: empty new method body → expected tests fail
- [x] Sanity mutation: wrongful `pruneState` call on abort → expected tests fail

### Documentation

- [ ] `doc-needed` 
- [X] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
@@ -333,11 +333,15 @@ class DurableExecutionManager implements ActionStatePersister, AutoCloseable {
      * via {@link #snapshotLastCompletedSequenceNumbers}. After pruning, the entry for that
      * checkpoint is removed. No-op when durable execution is disabled.
      *
-     * <p><b>Invariant:</b> the {@code checkpointIdToSeqNums.remove} below and the {@code put} in
-     * {@link #snapshotLastCompletedSequenceNumbers} MUST share the same {@code actionStateStore !=
-     * null} guard. Dropping the guard on either side breaks the symmetry and reintroduces the
+     * <p><b>Invariant:</b> the {@code checkpointIdToSeqNums.remove} below, the {@code put} in
+     * {@link #snapshotLastCompletedSequenceNumbers}, and the {@code remove} in {@link
+     * #notifyCheckpointAborted} MUST all share the same {@code actionStateStore != null} guard.
+     * Every snapshotted entry is released by exactly one of the two paths — Flink notifies either
+     * {@code notifyCheckpointComplete} OR {@code notifyCheckpointAborted} for each checkpoint,
+     * never both. Dropping the guard on any side breaks the symmetry and reintroduces the
      * unbounded-map leak tracked by <a href="https://github.com/apache/flink-agents/issues/645">
-     * issue #645</a>.
+     * issue #645</a> (complete path) or <a
+     * href="https://github.com/apache/flink-agents/issues/665">issue #665</a> (abort path).
      *
      * @param checkpointId the id of the completed checkpoint.
      */
@@ -348,6 +352,35 @@ class DurableExecutionManager implements ActionStatePersister, AutoCloseable {
             for (Map.Entry<Object, Long> entry : keyToSeqNum.entrySet()) {
                 actionStateStore.pruneState(entry.getKey(), entry.getValue());
             }
+            checkpointIdToSeqNums.remove(checkpointId);
+        }
+    }
+
+    /**
+     * Releases the per-checkpoint sequence-number snapshot recorded by {@link
+     * #snapshotLastCompletedSequenceNumbers} when Flink aborts the checkpoint instead of completing
+     * it. Unlike {@link #notifyCheckpointComplete}, this method does NOT prune durable action
+     * state: the aborted checkpoint's writes were never committed, so the previously-pruned-up-to
+     * point is still the {@code lastCompletedSequenceNumber} from the last successful checkpoint,
+     * and any state recorded since is still load-bearing for recovery from that prior checkpoint.
+     * We only release the in-memory tracking entry to prevent unbounded growth of {@code
+     * checkpointIdToSeqNums} when checkpoints abort under sustained pressure (issue #665).
+     *
+     * <p>Safe when no entry exists for {@code checkpointId} (e.g., abort fires for a checkpoint
+     * this task never snapshotted): {@link Map#remove} returns {@code null}. No-op when durable
+     * execution is disabled.
+     *
+     * <p><b>Invariant:</b> see {@link #snapshotLastCompletedSequenceNumbers} — together with {@link
+     * #notifyCheckpointComplete}, this method shares the same {@code actionStateStore != null}
+     * guard that releases entries recorded by the snapshot side. Dropping the guard on any side
+     * breaks the symmetry and reintroduces the unbounded-map leak tracked by <a
+     * href="https://github.com/apache/flink-agents/issues/645">issue #645</a> (complete path) or <a
+     * href="https://github.com/apache/flink-agents/issues/665">issue #665</a> (abort path).
+     *
+     * @param checkpointId the id of the aborted checkpoint.
+     */
+    void notifyCheckpointAborted(long checkpointId) {
+        if (actionStateStore != null) {
             checkpointIdToSeqNums.remove(checkpointId);
         }
     }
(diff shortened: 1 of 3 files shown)


wenjin272:
Hi, @weiqingy. It appears that after the merge of #659, both #666  and #667  have some conflicts.

weiqingy:
Hi @wenjin272, the PR has been updated to resolve the conflicts.

weiqingy:
The one CI failure (`it-python [java-17] [python-3.12] [flink-2.1]`) is the known `test_react_agent_on_local_runner` LLM flake against Ollama `qwen3:1.7b`, not caused by this PR:

```
FAILED flink_agents/e2e_tests/e2e_tests_integration/react_agent_test.py::test_react_agent_on_local_runner
  - assert 432596736 == 1386528
```

The test expects `4444 × 312 = 1386528`, but the LLM made an extra unnecessary `multiply(1386528, 312)` call and returned `432596736`. The test source has a comment right next to the assertion: *"This may be caused by the LLM response does not match the output schema, you can rerun this case."*

This same failure (same exact numbers, `432596736 == 1386528`) is currently failing on `main` at `b38ae21` — the commit this PR is rebased onto — and on several other recent main-branch runs. Failure runs through the Python `local_runner`, which logs `"Local runner does not support durable execution; recovery is not available."` — the Java `DurableExecutionManager` / `ActionExecutionOperator` paths changed by this PR are never exercised.

Will re-run CI.

wenjin272:
> The one CI failure (`it-python [java-17] [python-3.12] [flink-2.1]`) is the known `test_react_agent_on_local_runner` LLM flake against Ollama `qwen3:1.7b`, not caused by this PR:
> 
> ```
> FAILED flink_agents/e2e_tests/e2e_tests_integration/react_agent_test.py::test_react_agent_on_local_runner
>   - assert 432596736 == 1386528
> ```
> 
> The test expects `4444 × 312 = 1386528`, but the LLM made an extra unnecessary `multiply(1386528, 312)` call and returned `432596736`. The test source has a comment right next to the assertion: _"This may be caused by the LLM response does not match the output schema, you can rerun this case."_
> 
> This same failure (same exact numbers, `432596736 == 1386528`) is currently failing on `main` at `b38ae21` — the commit this PR is rebased onto — and on several other recent main-branch runs. Failure runs through the Python `local_runner`, which logs `"Local runner does not support durable execution; recovery is not available."` — the Java `DurableExecutionManager` / `ActionExecutionOperator` paths changed by this PR are never exercised.
> 
> Will re-run CI.

I believe we need to polish the stability and observability of CI in version 0.4. If you encounter any unstable cases, please contact me to rerun them. I now have the permission to rerun failed CI jobs.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
wenjin272, line 322: It appears that only tests are calling this method. We may need to verify whether this is an unnecessary interface or if the call was accidentally removed during a previous conflict resolution.

wenjin272 (commented):
Hi, @weiqingy, LGTM. I left a comment that may need to be confirmed.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/DurableExecutionManager.java
weiqingy: Thanks for catching this, @wenjin272. Confirmed it's dead code on `main` today, but it predates this PR — not a conflict-resolution loss.

Timeline from `git log -S "maybePruneState"`:

1. **#138** introduced `private void maybePruneState(...)` on `ActionExecutionOperator`, called from the per-key completion path: `maybePruneState(key, sequenceNumber);`.
2. **#603 "Fix prune state corner cases around checkpoints"** *deliberately* removed both the call site and the helper. The fix moved pruning from the per-key path to the checkpoint-driven path inside `notifyCheckpointComplete`, where it now prunes using `lastCompletedSequenceNumber` captured at the checkpoint barrier (via `snapshotLastCompletedSequenceNumbers`). That's the current/correct architecture.
3. **#546 "Refactor ActionExecutionOperator into focused manager classes"** extracted `DurableExecutionManager` and re-surfaced `maybePruneState(...)` as a package-private method on the new class — but no caller was re-added (correctly, since #603 eliminated that path). This is the orphan.

This seems to be leftover from the #546 refactor. This PR didn’t touch this method. Since the CI for this PR has already passed, and considering that the CI checks are not fully stable, I created a separate PR to fix it: https://github.com/apache/flink-agents/pull/682

wenjin272 (approved):
LGTM

Issue #625, opened 2026-04-13
[Feature] Support YAML api


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, Flink-Agents supports Python and Java APIs, both of which are programmatic. We aim to provide a declarative API, similar to the YAML-based API offered by CrewAI. We believe this will also help reduce the difficulty for Coding Agents when developing jobs based on Flink-Agents.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

qingfei1994:
@wenjin272 I'm pretty interested in making Flink-Agents more user friendly, but I think it may be better if it also support SQL?  Data engineers and data analyst are more familiar with SQL instead of yaml. What do you think?

wenjin272:
Hi, @qingfei1994, the YAML API mentioned here is equivalent to the Python and Java APIs we already support. Users can declare an Agent using the Python API, Java API, or YAML API.

I agree with your point that big data engineers are more proficient in SQL. However, supporting SQL is a far more ambitious topic than providing a YAML API for Flink Agents. Flink Agents is built on top of Flink, you can regard an Agent constructed with Flink Agents as a custom operator in Flink. If SQL is to be used for describing Flink jobs that include flink-agents, modifications to Flink's table layer would be required. However, as Flink is a mature framework, ensuring API stability and compatibility is paramount. Therefore, until flink-agents reach sufficient maturity, we will not consider making changes to Flink at this time.

PR #670, opened 2026-05-14
[api][python] Introduce YAML API for declaring agents

wenjin272:
Linked issue: #625

### Purpose of change

1. Add data structure to represent a java function in python
2. Support execute java function tool in python.
3. Introduce YAML api in python. 

### Tests

ut & e2e

### API

Yes, add yaml api.

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/api/yaml/loader.py
+++ b/python/flink_agents/api/yaml/loader.py
@@ -1,0 +1,387 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#################################################################################
+"""YAML loader: parse a YAML document and register agents on an execution
+environment.
+"""
+
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Dict, List, Tuple
+
+if TYPE_CHECKING:
+    from flink_agents.api.execution_environment import AgentsExecutionEnvironment
+
+import yaml
+
+from flink_agents.api.agents.agent import Agent
+from flink_agents.api.chat_message import ChatMessage, MessageRole
+from flink_agents.api.function import Function, JavaFunction, PythonFunction
+from flink_agents.api.prompts.prompt import Prompt
+from flink_agents.api.resource import ResourceDescriptor, ResourceType
+from flink_agents.api.skills import Skills
+from flink_agents.api.tools.function_tool import FunctionTool
+from flink_agents.api.yaml.aliases import (
+    JAVA_WRAPPER_CLAZZ,
+    resolve_clazz,
+    resolve_event_type,
+)
+from flink_agents.api.yaml.specs import (
+    ActionSpec,
+    AgentSpec,
+    DescriptorSpec,
+    Language,
+    PromptSpec,
+    SkillsSpec,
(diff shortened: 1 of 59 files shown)


python/flink_agents/api/yaml/specs.py
```diff
@@ -0,0 +1,208 @@
...
+
+    Holds the agent's own resources and actions. Resources/actions declared
+    at the file level (siblings of ``agents:``) are merged in by the loader.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    name: str
+    description: str | None = None
+    namespace: str | None = None
```
xintongsong: The name `namespace` is confusing. Maybe something like `default_func_path`. Anyway, the semantic of this needs well description in documents.

xintongsong: Do we need separate defaults for python and java?

xintongsong: I think this default needs more careful design, otherwise it may introduce more confusion than convenience. I'd suggest to make this a follow-up and exclude from this PR.

python/flink_agents/api/yaml/specs.py
```diff
@@ -0,0 +1,208 @@
...
+    ``parameter_types`` is required when ``type: java`` (Java method
+    signatures vary across tools).
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    name: str
+    function: str | None = None
+    type: Language | None = None
+    parameter_types: List[str] | None = None
```
xintongsong, line 123: This needs a more detailed description. What are the required formats/values of the parameter type strings.

python/flink_agents/api/yaml/loader.py
```diff
@@ -0,0 +1,496 @@
...
+    ToolSpec,
+    YamlAgentsDocument,
+)
+
+
+class _FlinkAgentsYamlLoader(yaml.SafeLoader):
+    """SafeLoader that recognizes only ``true``/``false`` as booleans.
+
+    PyYAML follows YAML 1.1 by default, where ``on``/``off``/``yes``/``no``
+    also coerce to booleans. ``on:`` is the natural way to declare an
```
xintongsong: Can we change to another keyword?

python/flink_agents/api/yaml/loader.py
```diff
@@ -0,0 +1,496 @@
...
+    - ``function`` given and contains ``.``: use verbatim.
+    - ``function`` given without ``.`` and ``namespace`` given:
+      ``{namespace}.{function}``.
+    - ``function`` given without ``.`` and no ``namespace``: error.
+    - ``function`` missing and ``namespace`` given: ``{namespace}.{name}``.
+    - ``function`` missing and no ``namespace``: error.
+    """
+    if function is not None:
+        if "." in function:
+            qualified = function
```
xintongsong: What if `namespace` is `foo` and the function is `foo.bar.Func`?

python/flink_agents/api/yaml/loader.py
```diff
@@ -0,0 +1,496 @@
...
+    for action_spec in doc.actions:
+        if action_spec.name in shared_actions:
+            msg = f"Duplicate shared action name {action_spec.name!r} in {path}"
+            raise ValueError(msg)
+        shared_actions[action_spec.name] = action_spec
+
+    return agents, shared_resources, shared_actions, agent_specs, doc
+
+
+def _assert_has_python_implementation(
```
xintongsong: Is this necessary? What if I implemented an agent in yaml + java, but need to use it in a pyflink job?

python/flink_agents/api/yaml/tests/test_loader.py
```diff
@@ -0,0 +1,648 @@
...
+
+def test_load_yaml_registers_shared_resources_on_env() -> None:
+    env = AgentsExecutionEnvironment.get_execution_environment()
+    load_yaml(env, _FIXTURES / "with_shared.yaml")
+    assert "shared_conn" in env.resources[ResourceType.CHAT_MODEL_CONNECTION]
+
+
+def test_load_yaml_string_ref_to_missing_shared_action_errors() -> None:
+    env = AgentsExecutionEnvironment.get_execution_environment()
+    bad = _FIXTURES / "bad_missing_shared_action.yaml"
```
xintongsong: Should not write temporal contents in the source directory.

xintongsong (approved):
LGTM

PR #677, opened 2026-05-14
[FLINK-AGENTS][api] Fix Kryo deserialization failure for Event classes with immutable lists

### LINKED ISSUE #673: [Bug] Kryo serialization fails for Event classes with immutable lists on JDK 17+
### Search before asking

### Description

ChatRequestEvent, ContextRetrievalResponseEvent, and ToolRequestEvent store their List<> constructor arguments by reference. When callers pass an immutable list (List.of(...), Arrays.asList(...), Collections.unmodifiableList(...)), Flink's Kryo serializer fails at runtime with an InaccessibleObjectException because JDK 17+ restricts reflective access to internal fields of java.util.Arrays$ArrayList and similar immutable list implementations.

The failure is non-obvious: the event is constructed successfully, but the job crashes when the first record is serialized/cloned between operators.

### How to reproduce

// In any agent action:
ChatMessage msg = new ChatMessage(MessageRole.USER, "hello");
ctx.sendEvent(new ChatRequestEvent("myModel", List.of(msg)));
// ^^^ List.of() produces an immutable list

Run the pipeline on JDK 17+ (the minimum for Flink 1.20). The job fails with:

Caused by: java.lang.reflect.InaccessibleObjectException:
  Unable to make field private final java.lang.Object[]
  java.util.Arrays$ArrayList.a accessible: module java.base
  does not "opens java.util" to unnamed module @...

**Expected Behaviour**

Events should accept any valid List<T> regardless of mutability. The user should not need to know that Kryo requires mutable lists internally.
Affected Classes

    org.apache.flink.agents.api.event.ChatRequestEvent
    org.apache.flink.agents.api.event.ContextRetrievalResponseEvent
    org.apache.flink.agents.api.event.ToolRequestEvent

### Version and environment

Flink Agents: 0.3-SNAPSHOT (current main)
Flink: 1.20.0
Java: JDK 17+

avichaym:
fixes #673

Wraps list arguments in new ArrayList<>() before passing to setAttr() in ChatRequestEvent, ContextRetrievalResponseEvent, and ToolRequestEvent. Without this, passing List.of(...) crashes Kryo on JDK 17+ with InaccessibleObjectException.

### DIFF
--- a/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/EventKryoSerializationTest.java
+++ b/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/EventKryoSerializationTest.java
@@ -1,0 +1,83 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.integration.test;
+
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.apache.flink.agents.api.chat.messages.MessageRole;
+import org.apache.flink.agents.api.event.ChatRequestEvent;
+import org.apache.flink.agents.api.event.ContextRetrievalResponseEvent;
+import org.apache.flink.agents.api.vectorstores.Document;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+/**
+ * Verifies that the user-constructed Event classes which carry collection-typed payloads remain
+ * Kryo-friendly when constructed with immutable collections (e.g. List.of(...)).
+ *
+ * <p>On JDK 17+, Flink's Kryo serializer cannot reflectively access internal fields of {@code
+ * java.util.ImmutableCollections} subclasses without {@code --add-opens
+ * java.base/java.util=ALL-UNNAMED}. The fix is to defensive-copy the input collection at event
+ * construction time so the stored collection is always a plain {@link java.util.ArrayList}, which
+ * Kryo handles natively.
+ *
+ * <p>Scope: this test only covers {@link ChatRequestEvent} and {@link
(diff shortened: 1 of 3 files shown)


wenjin272 (commented):
Hi, @avichaym, Ty for this fix.

The current fix looks good to me, but I think two additional points need to be added:
1. If user use `Map.of` to consturct `ToolResponseEvent`, there will likely be issues as well.
2. Supplement with relevant tests. Since in the CI pipeline, the test cases in `flink-agents-end-to-end-tests-integration` module will be executed in jdk 17 with flink 1.20, 2.0, 2.1 and 2.2. I think we can add a `EventKryoSerializationTest` in this module like:

```
@Test
void chatRequestEventWithImmutableListSurvivesKryoRoundTrip() throws Exception {
    ChatRequestEvent event =
            new ChatRequestEvent(
                    "myModel", List.of(new ChatMessage(MessageRole.USER, "hello")));

    ChatRequestEvent roundTripped = roundTrip(event, ChatRequestEvent.class);

    assertEquals("myModel", roundTripped.getModel());
    assertNotNull(roundTripped.getMessages());
    assertEquals(1, roundTripped.getMessages().size());
    assertEquals("hello", roundTripped.getMessages().get(0).getContent());
}
private static <T extends Event> T roundTrip(T event, Class<T> cls) throws Exception {
    KryoSerializer<T> serializer = new KryoSerializer<>(cls, new SerializerConfigImpl());
    DataOutputSerializer out = new DataOutputSerializer(64);
    serializer.serialize(event, out);
    DataInputDeserializer in = new DataInputDeserializer(out.getCopyOfBuffer());
    return serializer.deserialize(in);
}
```

avichaym:
Thanks for the careful review @wenjin272 - both points addressed in the new commit : 

1. ToolResponseEvent now defensive-copies all four maps (responses, success, error, externalIds) so Map.of(...) callers are safe - same shape as the list fix.

2. Added EventKryoSerializationTest under flink-agents-end-to-end-tests-integration covering all four affected events.

e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/EventKryoSerializationTest.java
```diff
@@ -0,0 +1,163 @@
...
+    @Test
+    void toolRequestEventDefensiveCopiesImmutableList() throws Exception {
+        ToolRequestEvent event =
+                new ToolRequestEvent(
+                        "myModel",
+                        List.of(
+                                Map.of(
+                                        "id", "call_1",
+                                        "name", "myTool",
+                                        "arguments", Map.of("x", 1))));
```
wenjin272: Currently, we only copy the outer ImmutableList/ImmutableMap without handling nested ImmutableList/ImmutableMap cases. Therefore, this test will fail under Flink 1.20.

I think there are probably two solutions to this problem:
1. Supports copying arbitrarily nested ImmutableList/ImmutableMap structures into nested MutableList/MutableMap, but this might be too heavy for constructing Events.
2. Since both ToolRequestEvent and ToolResponseEvent are constructed by built-in actions, users typically do not create these events themselves. Therefore, we forgo defensive checks for Tool Events.

I personally prefer option 2, WDYT？

avichaym: Agreed, option 2 makes more sense. Updated to revert the Tool event changes and keeping defensive copies only on ChatRequestEvent and ContextRetrievalResponseEvent. Test trimmed to those two cases.

wenjin272 (approved):
LGTM

PR #678, opened 2026-05-14
[FLINK-AGENTS][integrations] Fix OpenSearchVectorStore for Amazon OpenSearch Serverless

### LINKED ISSUE #674: [Bug] OpenSearchVectorStore fails against Amazon OpenSearch Serverless (AOSS)
### Search before asking

### Description

The OpenSearchVectorStore integration (introduced in #533) supports both Amazon OpenSearch Service (provisioned domains) and Amazon OpenSearch Serverless (AOSS) via a service_type parameter that defaults to "serverless". However, for serverless , some operations fail because AOSS has some differences specifically around SigV4 signing requirements, supported index APIs, bulk ingestion constraints, and KNN engine capabilities.

### How to reproduce

1) Create an AOSS collection of type VECTORSEARCH in any region.
Create a data-access policy granting aoss:* to your IAM principal for both collection/<name> and index/<name>/*.
2)Wait for the collection to reach ACTIVE status.
3)Run any pipeline that uses OpenSearchVectorStore with the default service_type=serverless (or explicitly set it), pointing at the AOSS endpoint.

Some operations will fail with issues - 

POST /<index>/_refresh after add/delete - API surface issue

filtered KNN ( createKnnIndex omits method.engine, AOSS defaults to NMSLIB which cannot serve filtered KNN)

### Version and environment

Flink Agents: 0.3-SNAPSHOT (current main)
Flink: 1.20.0
Java: JDK 17+
OS: macOS (arm64) / Linux

avichaym:
Fixes #674

The integration defaults to service_type=serverless but doesn't actually work against AOSS. Five fixes:

    Add x-amz-content-sha256 + Content-Length headers for SigV4 signing (AOSS returns 403 without them)
    Skip _refresh calls on serverless (AOSS returns 404 — API not exposed)
    Omit custom _id in _bulk actions on serverless (AOSS rejects them)
    Validate _bulk responses for partial failures (was silently losing data)
    Use FAISS engine instead of default NMSLIB for index creation (NMSLIB doesn't support filtered KNN)

Also adds a 15s settle after index creation on serverless for AOSS propagation.

Tested end-to-end against a live AOSS VECTORSEARCH collection.

### DIFF
--- a/integrations/vector-stores/opensearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/opensearch/OpenSearchVectorStore.java
+++ b/integrations/vector-stores/opensearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/opensearch/OpenSearchVectorStore.java
@@ -46,6 +46,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
@@ -245,11 +246,18 @@ public class OpenSearchVectorStore extends BaseVectorStore
     }
 
     private void createKnnIndex(String idx) {
+        // Use the FAISS engine with HNSW: it supports both pre-filtered and post-filtered KNN
+        // queries (the default NMSLIB engine on AOSS does NOT support filters and rejects
+        // queries with "Engine [NMSLIB] does not support filters"), and is the recommended
+        // engine for both AOSS VECTORSEARCH collections and OpenSearch Service domains 2.x+.
         String body =
                 String.format(
                         "{\"settings\":{\"index\":{\"knn\":true}},"
                                 + "\"mappings\":{\"properties\":{\"%s\":{\"type\":\"knn_vector\","
-                                + "\"dimension\":%d},\"%s\":{\"type\":\"text\"},"
+                                + "\"dimension\":%d,"
+                                + "\"method\":{\"engine\":\"faiss\",\"name\":\"hnsw\","
+                                + "\"space_type\":\"l2\"}},"
+                                + "\"%s\":{\"type\":\"text\"},"
                                 + "\"metadata\":{\"type\":\"object\"}}}}",
                         vectorField, dims, contentField);
         try {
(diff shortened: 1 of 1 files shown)


integrations/vector-stores/opensearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/opensearch/OpenSearchVectorStore.java
wenjin272, line 413: Currently, updateEmbedding is implemented by inserting a document with the same ID. In the AOSS scenario, since the client cannot specify an ID, this approach fails to achieve the update purpose. Therefore, we should throw an "unsupported operation" exception.

integrations/vector-stores/opensearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/opensearch/OpenSearchVectorStore.java
```diff
@@ -400,7 +418,14 @@ public List<String> addEmbedding(
             allIds.add(id);
```
wenjin272: Since AOSS does not allow clients to specify IDs, returning this ID in the AOSS scenario is meaningless, as users cannot use it for get or delete operations. I believe that in the AOSS scenario, we should return the ID generated by AOSS.

wenjin272 (commented):
Hi, @avichaym, thanks for your work. I left two comments.

integrations/vector-stores/opensearch/src/main/java/org/apache/flink/agents/integrations/vectorstores/opensearch/OpenSearchVectorStore.java
avichaym: Both points addressed. updateEmbedding on AOSS now throws UnsupportedOperationException (with a message pointing to the provisioned domain as an alternative), and add() on serverless parses items[].index._id out of each _bulk response and returns those AOSS-generated ids instead of the client-side UUIDs. Provisioned-domain behaviour is unchanged. Verified end-to-end against the live AOSS test collection.

wenjin272 (approved):
LGTM

Issue #159, opened 2025-09-11
[Tech Debt] Specify model to be used in ChatModelSetup.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, different implementation of chat model has different behavior about specifying which model to use. We'd better specify it in ChatModelSetup abstraction, which is aligned with embedding model, so that users don't get confused.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #685, opened 2026-05-17
[api] Specify model in ChatModelSetup (Java + Python)

weiqingy:
Linked issue: #159

### Purpose of change

Different `ChatModelSetup` implementations diverged in how they exposed the `model` parameter. On the Java side, every chat-model subclass except `OllamaChatModelSetup` already used the inherited `BaseChatModelSetup.model` field — Ollama re-declared its own `private final String model` shadow. On the Python side, `BaseChatModelSetup` had no `model` field at all, so every chat-model subclass redeclared `model: str = Field(...)` locally. This diverged from `BaseEmbeddingModelSetup`, which has owned the field at the base on both languages from the start.

This PR aligns `ChatModelSetup` with `EmbeddingModelSetup`:

- **Commit 1 (Java)** — removes the redundant `model` field shadow in `OllamaChatModelSetup`. `getParameters()` now reads from the inherited `BaseChatModelSetup.model` field that was already populated by the same descriptor argument. Behavior is bit-for-bit identical; the redundant surface is gone.
- **Commit 2 (Python)** — lifts `model: str = Field(description="Name of the chat model to use.")` to `BaseChatModelSetup` (required, no default — matching `BaseEmbeddingModelSetup`). Removes the now-redundant per-subclass `Field` redeclarations from Ollama, Anthropic, OpenAI, Tongyi, and AzureOpenAI. `__init__` signatures, per-integration defaults, and `super().__init__(model=model, ...)` call sites are all preserved.

The cross-language bridge (`runtime/java/java_chat_model.py`) pops `model` from kwargs and forwards it to `super().__init__()`, symmetric with the existing `connection` handling. Test mocks and the cross-language compatibility fixture (`agent_plan.json`) are updated to satisfy the new required field.

### Tests

- New: `OllamaChatModelSetupTest` (Java) locks the invariant that `getModel()` and `getParameters().get("model")` both return the descriptor-supplied value after the shadow removal.
- New: `test_chat_model_base.py` (Python) — a minimal subclass that omits the local `model` field inherits it from the base; missing `model` raises `ValidationError`.
- Each Python integration test file gains `test_model_field_roundtrip` (5 files); Anthropic / OpenAI / Tongyi also gain `test_default_model_when_omitted` to confirm `__init__`-level defaults still apply post-refactor.
- All existing tests stay green: `./tools/ut.sh -j` (Java), `cd python && uv run pytest flink_agents -k "not e2e_tests"` (Python).
- `./tools/lint.sh -c` and `./tools/check-license.sh` pass.

### API

`BaseChatModelSetup` (Python) now declares `model: str` as a required base-class field, matching `BaseEmbeddingModelSetup`. Public `__init__` signatures of every chat-model subclass are unchanged. Descriptor argument key remains `"model"`.

One behavior tightening, intentional and aligned with the embedding contract: calling `<Subclass>.model_validate({"connection": "x"})` without a `"model"` key now raises `pydantic.ValidationError` instead of falling back to a subclass field-level default. Construction through `__init__` (the normal path) is unaffected — per-integration defaults still apply.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/integrations/chat-models/ollama/src/main/java/org/apache/flink/agents/integrations/chatmodels/ollama/OllamaChatModelSetup.java
+++ b/integrations/chat-models/ollama/src/main/java/org/apache/flink/agents/integrations/chatmodels/ollama/OllamaChatModelSetup.java
@@ -53,13 +53,11 @@ import java.util.Map;
  */
 public class OllamaChatModelSetup extends BaseChatModelSetup {
 
-    private final String model;
     private final Object think;
     private final boolean extractReasoning;
 
     public OllamaChatModelSetup(ResourceDescriptor descriptor, ResourceContext resourceContext) {
         super(descriptor, resourceContext);
-        this.model = descriptor.getArgument("model");
         this.think = descriptor.getArgument("think", true);
         this.extractReasoning = descriptor.getArgument("extract_reasoning", true);
     }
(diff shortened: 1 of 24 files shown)


wenjin272 (approved):
LGTM

Issue #687, opened 2026-05-18
[Feature][Python] Implement EventListener to align with Java SDK


twosom:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Related PR
- #641
- #102 

### Description

This feature introduces the `EventListener` interface to enable event-driven monitoring and custom logic extensions within the Flink Agent Python SDK, aligning with the existing Java SDK implementation.

#### Design Considerations

During the initial design phase, I explored two different approaches for registering event listeners:

**Option 1: Decorator-based Registration (Considered)**
I initially evaluated a Pythonic approach where a decorator registers a standalone function as a listener by storing its metadata in an internal registry.
```python
# Information stored in an internal registry
REGISTRY = []

def event_listener(func):
    REGISTRY.append({
        "module": func.__module__,
        "qualname": func.__qualname__
    })
    return func

@event_listener
def my_handler(context, event):
    print(f"Received: {event}")
```

*Pros:* Highly idiomatic and concise for Python developers.
*Cons:* This approach deviates significantly from the class/interface-based structure of the Java SDK, breaking API consistency across different language SDKs.

**Option 2: Class-based Interface (Final Decision)**
To maintain a unified developer experience and ensure structural alignment with the Java SDK, I decided to adopt a **class-based interface design**.

#### Proposed API Usage

The user implements the `EventListener` interface, and the framework dynamically instantiates and triggers it during the event lifecycle.

```python
from flink_agents.api.listener import EventListener
from flink_agents.api.event_context import EventContext
from flink_agents.api.events import Event

class MyCustomListener(EventListener):
    def on_event_processed(self, context: EventContext, event: Event) -> None:
        print(f"Event processed: {event.type} at {context.timestamp}")

# Registration via environment configuration
agents_env.get_config().set(
    AgentConfigOptions.EVENT_LISTENERS, 
    [str(MyCustomListener)]
)
```

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #688, opened 2026-05-18
[Feature][Python] Implement EventListener in python sdk

twosom:
Linked issue: #687 

### Purpose of change
This PR implements the `EventListener` mechanism for both Java and Python runtimes, ensuring API consistency across different language SDKs. It enables developers to monitor and react to events synchronously during the agent's event lifecycle.

### Tests
ok

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`


wenjin272:
Hi, @twosom. Thanks for supporting EventListener in python sdk.

Overall, I think the current design has become somewhat overly complex. In my view, we should have a single `PythonEventListenerWrapper` that, at runtime, uses pemja to instantiate the corresponding Python `EventListener` object and simply invokes its `on_event` method.

I think some of the current complexity stems from supporting `str(MyPythonListener)`. Could we use a standard Python fully-qualified class name (`"my_module.MyPythonListener"`) as the listener identifier, instead of the `str(MyPythonListener)` → `"module:qualname.method_name"` protocol?

That would let us drop `EventListenerMeta`, `_resolve_module`, the `<locals>` / `__main__` / `sys.modules` fallbacks, and the matching `split(":")` on the Java side — Python-side instantiation collapses to `importlib.import_module` + `getattr` + `rpartition(".")`.

The cost is that listeners must live at module top level (which the PR already enforces against `<locals>`) and can't be defined in `__main__`.

Additionally, I believe there is no need to support event listeners in the `local_runner`. In fact, we are also considering whether the `local_runner` still needs to exist

Additionally, I believe the current testing is insufficient. We have some unit tests, but no end-to-end tests. For this cross-language scenario, we need to add real tests that submit jobs to the remote environment in `e2e_tests_resource_cross_language`.

python/flink_agents/runtime/python_java_utils.py
```diff
@@ -321,6 +322,73 @@ def get_java_tool_metadata_from_tool(tool: Tool) -> typing.Dict[str, str]:
...
+    searching in sys.modules if standard import fails.
+    """
+    import importlib
+    import sys
+
+    if module_name == "__main__":
+        module = sys.modules["__main__"]
+        first_part = class_qualname.split(".")[0]
+        if not hasattr(module, first_part):
+            for m in list(sys.modules.values()):
```
weiqingy, line 343: Both the `__main__` and `ImportError` fallbacks walk `sys.modules.values()` and return the first module that has `class_qualname.split(".")[0]` as an attribute. If two modules top-level-export a class with the same name — not an unusual user mistake — the result is non-deterministic and silent: wrong listener class instantiated and notified. What invariant are we relying on here? If it's "module top-level uniqueness across the whole interpreter," can we raise on collision instead of letting first-match win?

python/flink_agents/api/tests/test_event_listener.py
```diff
@@ -0,0 +1,101 @@
...
+        """Test string representation when the module is '__main__' but lacks a '__file__' attribute."""
+        from unittest.mock import MagicMock, patch
+
+        mock_module = MagicMock()
+        mock_module.__name__ = "__main__"
+        # __file__ attribute is missing
+
+        with patch("inspect.getmodule", return_value=mock_module):
+            # Should fallback to "__main__" if __file__ is missing
+            assert str(MainListenerMock) == "__main__:MainListenerMock.on_event_processed"
```
weiqingy, line 91: This test claims to cover the missing-`__file__` branch, but `MagicMock()` auto-creates `__file__` on access — `hasattr(module_obj, "__file__")` is always `True` and the fallback branch never runs. The assertion happens to pass because `inspect.getmodule` is patched, but for the wrong reason. Switching to `spec=ModuleType` and explicitly deleting `__file__` would make the test actually exercise the fallback path.

python/flink_agents/api/event_context.py
```diff
@@ -0,0 +1,35 @@
...
+    Attributes:
+    ----------
+    eventType : str
+        The routing key for the event, matching the ``EVENT_TYPE`` constant or
+        type string.
+    timestamp : str
+        Timestamp of when the event occurred.
+    """
+
+    eventType: str
```
weiqingy, line 33: camelCase here clashes with the rest of `flink_agents/api`, which is snake_case throughout. User code reading `context.eventType` will stand out. Would `event_type` with `alias="eventType"` satisfy the Java JSON contract while keeping the Python attribute idiomatic?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/PythonBridgeManager.java
```diff
@@ -301,4 +290,76 @@ public void close() throws Exception {
...
+                // Convert java event context to python event context
+                final Object pythonEventContext =
+                        pythonResourceAdapter.toPythonEventContext(context);
+                for (Map.Entry<Object, PythonFunction> entry : this.listenerEntries) {
+                    final PythonFunction listenerFunction = entry.getValue();
+                    final Object listenerObject = entry.getKey();
+                    listenerFunction.call(listenerObject, pythonEventContext, pythonEvent);
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(e);
```
weiqingy, line 334: `throw new RuntimeException(e)` strips the descriptor of the listener that failed. The wrapper already holds `listenerEntries` — including the current entry's `module:qualname` in the message would turn "which of N listeners threw?" from a multi-minute debug into a one-line answer.

python/flink_agents/runtime/local_runner.py
```diff
@@ -358,6 +381,15 @@ def run(self, **data: Dict[str, Any]) -> Any:
...
         return key
 
+    def _trigger_listener(
+        self, listener: EventListener, event_context: EventContext, event: Event
+    ) -> None:
+        """Trigger a single listener with error handling."""
+        try:
+            listener.on_event_processed(event_context, event)
+        except Exception:
+            logger.exception("Error in EventListener execution")
```
weiqingy, line 391: This swallows and logs; `EventRouter#notifyEventProcessed` on the Java side lets listener exceptions propagate. Same interface, two failure semantics depending on the runner. What's the rationale for diverging here? If it's deliberate, worth documenting the asymmetry in `event_listener.md` so users of both runners aren't surprised.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonResourceAdapterImpl.java
```diff
@@ -77,6 +78,11 @@ public class PythonResourceAdapterImpl implements PythonResourceAdapter {
 
     static final String INVOKE_PYTHON_TOOL = PYTHON_MODULE_PREFIX + "invoke_python_tool";
 
+    static final String FROM_JAVA_EVENT_CONTEXT = PYTHON_MODULE_PREFIX + "from_java_event_context";
+
+    static final String INSTANTIATE_PYTHON_EVENT_LISTER =
```
weiqingy, line 83: Typo — `INSTANTIATE_PYTHON_EVENT_LISTER` is missing an `N`. Should be `INSTANTIATE_PYTHON_EVENT_LISTENER`.

weiqingy (commented):
Thanks for the PR! 

A few thoughts after going through the diff:

**1. Java and Python listeners end up mutually exclusive at the operator level.** The `if (pythonBridge.isInitialized())` branch in `ActionExecutionOperator#open` (lines 197–208) calls either `initForPythonEventListeners` or `eventRouter.initEventListeners`, never both. `pythonBridge.isInitialized()` is true whenever the plan contains a Python action, a Python resource, or Mem0 config — so a user who registers a Java listener FQN under any of those conditions sees their listener silently never fire. On top of that, `PythonBridgeManager#initForPythonEventListeners` will `ArrayIndexOutOfBounds` on `parts[1]` for a Java FQN with no `:` in it. The `event-listeners` key is being asked to mean two different things based on unrelated runtime state. What do you think about splitting it — by descriptor shape, a `python:` prefix on Python entries, or two separate config keys?

**2. No e2e coverage for the Java→Python listener path.** `initForPythonEventListeners`, `PythonEventListenerWrapper`, `toPythonEventContext`, and the new branch in `ActionExecutionOperator#open` are only exercised by Python unit tests and a `LocalRunner` smoke test. None of it touches a real `ActionExecutionOperator` before merge, and that's the whole reason this feature exists. Could you add a scenario under `python/flink_agents/e2e_tests/e2e_tests_resource_cross_language` that submits a job with a Python listener and asserts `on_event_processed` fires per event?

**3. The `str(MyListener)` identifier protocol is doing the work of a function.** `EventListenerMeta.__str__` emits `module:Class.on_event_processed`, and `_resolve_module` reverses it with `__main__`-by-file-stem plus an attribute-search across all loaded modules. A standard Python FQN (`"my_module.MyClass"`) plus `importlib.import_module` + `getattr` + `rpartition(".")` would collapse the whole protocol — the metaclass, the resolver, and all the fallbacks. The current cost (listeners must live at module top level) is already enforced by the `<locals>` check. Hiding the descriptor inside `str()` also bites anyone who later logs `str(MyListener)` expecting `<class …>`. Would a classmethod (`MyListener.descriptor()`) or a free `listener_id(cls)` helper work better?

**4. `PythonResourceAdapter` is picking up non-resource concerns.** It gains `toPythonEventContext()` and `initPythonEventListener()`, neither of which is a resource. They happen to need the same interpreter, but that's the only thing they share with the rest of this interface. Once a second non-resource feature reaches in for its own bootstrapping, the name stops describing what the type does. Could we either extract a sibling (e.g. `PythonEventListenerAdapter`) or just call `pythonInterpreter.invoke(...)` directly from `PythonBridgeManager` for these two?

PR #695, opened 2026-05-20
[Feature][Integration][Java] Add built-in support for Azure OpenAI Chat Model

alnzng:
Linked issue: #694

### Purpose of change

Add Azure OpenAI as a first-class chat model provider for Java agents.

### Tests

- Run end-to-end test on top of a real Azure OpenAI model and verified the tests passed
- Added new unit tests: `AzureOpenAIChatModelConnectionTest` and `AzureOpenAIChatModelSetupTest`.

### API

- Adds two new public resource classes (`AzureOpenAIChatModelConnection`, `AzureOpenAIChatModelSetup`) and two new constants on ResourceName.ChatModel`.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
+++ b/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
@@ -64,8 +64,10 @@ import java.util.Map;
  * <p>Optional connection arguments:
  *
  * <ul>
- *   <li><b>timeout</b> (Number): seconds before an API call times out (default 60)
- *   <li><b>max_retries</b> (Number): retry attempts on failure (default 3)
+ *   <li><b>timeout</b> (Number): seconds before an API call times out; must be greater than 0,
+ *       otherwise ignored (SDK default applies)
+ *   <li><b>max_retries</b> (Number): retry attempts on failure; must be non-negative, otherwise
+ *       ignored (SDK default applies)
  *   <li><b>azure_url_path_mode</b> (String): one of {@code "AUTO"}, {@code "LEGACY"}, or {@code
  *       "UNIFIED"} (default {@code "AUTO"}). Controls how the SDK constructs Azure OpenAI request
  *       URLs. In {@code AUTO} mode the SDK only treats the endpoint as Azure when its hostname
@@ -112,23 +114,21 @@ public class AzureOpenAIChatModelConnection extends BaseChatModelConnection {
             throw new IllegalArgumentException("azure_endpoint should not be null or empty.");
         }
 
-        Integer timeoutSeconds = descriptor.getArgument("timeout");
-        if (timeoutSeconds == null) {
-            timeoutSeconds = 60;
-        }
-
-        Integer maxRetries = descriptor.getArgument("max_retries");
-        if (maxRetries == null) {
-            maxRetries = 3;
-        }
-
         OpenAIOkHttpClient.Builder clientBuilder =
                 OpenAIOkHttpClient.builder()
                         .baseUrl(azureEndpoint)
                         .credential(AzureApiKeyCredential.create(apiKey))
-                        .azureServiceVersion(AzureOpenAIServiceVersion.fromString(apiVersion))
-                        .timeout(Duration.ofSeconds(timeoutSeconds))
-                        .maxRetries(maxRetries);
+                        .azureServiceVersion(AzureOpenAIServiceVersion.fromString(apiVersion));
+
+        Integer timeoutSeconds = descriptor.getArgument("timeout");
+        if (timeoutSeconds != null && timeoutSeconds > 0) {
+            clientBuilder.timeout(Duration.ofSeconds(timeoutSeconds));
+        }
+
+        Integer maxRetries = descriptor.getArgument("max_retries");
+        if (maxRetries != null && maxRetries >= 0) {
+            clientBuilder.maxRetries(maxRetries);
+        }
 
         String azureUrlPathMode = descriptor.getArgument("azure_url_path_mode");
         if (azureUrlPathMode != null && !azureUrlPathMode.isBlank()) {
--- a/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
+++ b/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
@@ -1,0 +1,279 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.integrations.chatmodels.openai;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.openai.azure.AzureOpenAIServiceVersion;
+import com.openai.azure.AzureUrlPathMode;
+import com.openai.azure.credential.AzureApiKeyCredential;
+import com.openai.client.OpenAIClient;
+import com.openai.client.okhttp.OpenAIOkHttpClient;
+import com.openai.core.JsonValue;
+import com.openai.models.ChatModel;
+import com.openai.models.FunctionDefinition;
+import com.openai.models.FunctionParameters;
+import com.openai.models.chat.completions.ChatCompletion;
+import com.openai.models.chat.completions.ChatCompletionCreateParams;
+import com.openai.models.chat.completions.ChatCompletionFunctionTool;
+import com.openai.models.chat.completions.ChatCompletionTool;
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.apache.flink.agents.api.chat.model.BaseChatModelConnection;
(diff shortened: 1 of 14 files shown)


alnzng:
@wenjin272 @xintongsong @weiqingy please help take a look this PR, thanks.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
```diff
@@ -0,0 +1,276 @@
...
+        if (apiVersion == null || apiVersion.isBlank()) {
+            throw new IllegalArgumentException("api_version should not be null or empty.");
+        }
+
+        String azureEndpoint = descriptor.getArgument("azure_endpoint");
+        if (azureEndpoint == null || azureEndpoint.isBlank()) {
+            throw new IllegalArgumentException("azure_endpoint should not be null or empty.");
+        }
+
+        Integer timeoutSeconds = descriptor.getArgument("timeout");
```
wenjin272, line 128: The `OpenAiCompletionsConnection` validates the boundaries for `timeout` and `maxRetries`, so we can align with that here.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelSetup.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+        params.put("logprobs", logprobs);
+        if (temperature != null) {
+            params.put("temperature", temperature);
+        }
+        if (maxTokens != null) {
+            params.put("max_tokens", maxTokens);
+        }
+        if (additionalKwargs != null && !additionalKwargs.isEmpty()) {
+            params.putAll(additionalKwargs);
```
wenjin272: `putAll(additionalKwargs)` runs **after** the first-class fields, so a colliding key (`temperature`, `model`, `max_tokens`, ...) silently overwrites the validated value. For example, `additional_kwargs {"temperature": 5.0}` bypasses the constructor's `0.0–2.0` range check.

It appears that the Python side has the same issue.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
```diff
@@ -0,0 +1,276 @@
...
+            ChatCompletion completion = client.chat().completions().create(builder.build());
+
+            ChatMessage response =
+                    OpenAIChatCompletionsUtils.convertFromOpenAIMessage(
+                            completion.choices().get(0).message(), Map.of());
+
+            if (modelOfAzureDeployment != null
+                    && !modelOfAzureDeployment.isBlank()
+                    && completion.usage().isPresent()) {
+                recordTokenMetrics(
```
wenjin272, line 219: If the user has not configured `modelOfAzureDeployment`, we will not record token metrics at all, which may result in a poor user experience. I am wondering whether we can use the `model` as a fallback.

The same issue exists on the Python side as well.

wenjin272 (commented):
Hi, @alnzng, Ty for contributing this integration. I left server comments.

In addition, I believe we need to update the documentation by removing references to Python-only implementations and adding Java parameters and examples.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
alnzng: Azure OpenAI APIs have different setup on the `model`, the `model` means a Azure deployment name, not a real LLM name. We have discussed this in the past: https://github.com/apache/flink-agents/pull/478#discussion_r2726510844

Fallback to `model` will confuse the user, I would prefer to keep current behavior. One thing we can improve is that we can call out the in the doc that `modelOfAzureDeployment` is required if the users want to see token metrics. 

WDYT?

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelSetup.java
alnzng: Updated and use a consistent logic from OpenAI / Anthropic model to handle `additionalKwargs `.

alnzng:
@wenjin272 PR is updated based on your comments.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
```diff
@@ -0,0 +1,279 @@
...
+            if (Boolean.TRUE.equals(logprobs)) {
+                builder.logprobs(true);
+            }
+
+            @SuppressWarnings("unchecked")
+            Map<String, Object> additionalKwargs =
+                    (Map<String, Object>) mutableArgs.remove("additional_kwargs");
+            if (additionalKwargs != null) {
+                for (Map.Entry<String, Object> entry : additionalKwargs.entrySet()) {
+                    builder.putAdditionalBodyProperty(
```
weiqingy, line 205: The Setup-side fix (nesting `additional_kwargs` under its own key) closes the obvious case, but here in the connection we then forward those kwargs via `builder.putAdditionalBodyProperty(...)`. The openai-java SDK serializes typed fields and additional body properties into the same JSON request body, so a caller can still pass e.g. `additional_kwargs={"temperature": 5.0}` alongside the typed `temperature=0.3` and end up with both keys in the body (or one silently winning depending on SDK serialization order), bypassing the `[0.0, 2.0]` range check.

Should we detect collisions against the typed fields we already pop here (`model`, `temperature`, `max_tokens`, `logprobs`) and throw an `IllegalArgumentException`? That would keep the validation contract intact regardless of how the SDK merges the body.

The Python side has the same shape with `**kwargs, **additional_kwargs` (last one wins via dict expansion order) — should the fix apply on both sides?

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIChatCompletionsUtils.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /**
+     * Convert an OpenAI {@link ChatCompletionMessage} to a Flink Agents {@link ChatMessage}.
+     * Caller-provided {@code extraArgs} are copied into the returned ChatMessage's own extraArgs
+     * map (the caller's input is treated as read-only; {@code Map.of()} is safe). Additionally,
+     * {@code message.refusal()} is written as {@code extraArgs["refusal"]} when present, preserving
+     * prior Java behavior.
+     */
+    public static ChatMessage convertFromOpenAIMessage(
```
weiqingy: The `extraArgs` parameter is dead today — both call sites (this one and `AzureOpenAIChatModelConnection.chat`) pass `Map.of()`. Also, the body copies caller args first (`response.getExtraArgs().putAll(extraArgs)`) and then writes `refusal` after, so a caller-supplied `"refusal"` key would be silently overwritten by `message.refusal()`. The precedence isn't obvious from the signature.

Should we drop the parameter until there's a real caller that needs it? That would keep the API surface minimal and avoid locking in surprising precedence. Or, if we want to keep it as a forward-looking hook, should we document the override behavior in the Javadoc?

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
wenjin272: Sorry, I forgot our previous discussion. I think it’s reasonable to clarify this matter in the documentation.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
xintongsong: +1

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIChatCompletionsUtils.java
xintongsong: +1

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
xintongsong: +1 for the proposed solution, and for also fixing this in the python side.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
xintongsong: In addition to the documentation, I think we can also print a warning log if we detect that `modelOfAzureDeployment` is not set during initializing the Setup.

alnzng: sg - we can add a warning log for this.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIChatCompletionsUtils.java
alnzng: Good catch! Let's remove it for now.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/AzureOpenAIChatModelConnection.java
alnzng: >Should we detect collisions against the typed fields we already pop here (model, temperature, max_tokens, logprobs) and throw an IllegalArgumentException?

Sounds good. We can fail this fast if any collision detected.

alnzng:
@weiqingy @xintongsong @weiqingy Thanks for your review. PR is updated based on your comments, please take a look again.

xintongsong (approved):
LGTM

Issue #220, opened 2025-09-26
[Feature] Separate messages and arguments for prompt of chat method in BaseChatModelSetup


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, the `chat` method in `BaseChatModelSetup` will get arguments from messages `extra_args` to fill the prompt template, which is not clear.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

weiqingy:
@wenjin272 I came up with a change proposal on this https://github.com/weiqingy/flink-agents/blob/220-design-review/docs/PR-220/220-DESIGN.md Could you please take a look when you get a chance? If it looks good to you, I can help work on this.

weiqingy:
@wenjin272 The PR has been created: https://github.com/apache/flink-agents/pull/698 . Could you take a look when you get a chance? Thanks!

PR #698, opened 2026-05-22
[api][plan][runtime] Separate prompt arguments from message extra_args in BaseChatModelSetup.chat()

weiqingy:
Linked issue: #220

### Purpose of change

`BaseChatModelSetup.chat()` (Java + Python) previously filled prompt templates by flattening every input message's `extra_args` into a single map. This conflated chat-message metadata with prompt-template variables and forced callers to stuff template values into a generic metadata bag.

This PR adds an explicit `arguments` parameter on `chat()` and carries the same field on `ChatRequestEvent`. `ChatModelAction` extracts it from the event and forwards to the setup on both round 1 and tool-response continuations, so multi-turn flows keep re-filling the template correctly.

`ChatMessage.extra_args` is unchanged and still carries `externalId`, `STRUCTURED_OUTPUT`, OpenAI `refusal`, Ollama `reasoning`, and other provider-specific metadata used by chat-model connections.

### Tests

- New positive test (Java + Python): `chat(messages, arguments, parameters)` formats the `Prompt` template using values from `arguments`.
- New negative test (Java + Python): a `ChatMessage` with `extra_args` set no longer feeds the prompt template — proves the cutover.
- New multi-turn test (Java + Python): two consecutive `chat()` invocations with the same `arguments` re-fill the template each time.
- New action-layer regression test (Java + Python): `processChatRequestOrToolResponse` with a `ToolResponseEvent` extracts the persisted `arguments` from the saved tool-request context and forwards to `chat_model.chat(...)` on round 2 — locks the multi-turn contract.
- Bridge test updated: `PythonChatModelSetupTest.testChat` asserts `"arguments"` flows into Pemja kwargs marshalled to Python.
- Existing migrated tests in `test_built_in_actions.py`, `built_in_action_async_execution_test.py`, and `e2e_tests_mcp/mcp_test.py` preserve their outcome assertions post-cutover — proves behavior parity.
- Full Java + Python test sweeps green (`./tools/ut.sh -j`; `uv run pytest`: 509 passed / 13 skipped).
- `./tools/lint.sh -c` clean.

### API

Yes — this is a breaking API change:

- **Java** `BaseChatModelSetup`: the existing 2-arg `chat(List<ChatMessage>, Map<String, Object> parameters)` overload is removed (it would erase to the same signature as a hypothetical `chat(messages, arguments)`). The new primary form is `chat(List<ChatMessage>, Map<String, Object> arguments, Map<String, Object> parameters)`. The 1-arg `chat(List<ChatMessage>)` convenience overload is retained.
- **Python** `BaseChatModelSetup.chat`: `arguments: Mapping[str, Any] | None = None` is added between `messages` and `**kwargs`.
- **Java** `ChatRequestEvent`: new 4-arg constructor `(String model, List<ChatMessage> messages, @Nullable Map<String, Object> arguments, @Nullable Object outputSchema)`. Existing 2-arg and 3-arg legacy constructors continue to work (delegating with empty arguments).
- **Python** `ChatRequestEvent.__init__`: `arguments: Dict[str, Any] | None = None` added between `messages` and `output_schema`.

Migration: callers that previously set template variables via `ChatMessage.extra_args` should move them to `ChatRequestEvent.arguments`. All in-repo callers (3 Java examples, 3 Python examples, 2 e2e tests, 1 runtime test) are migrated in this PR.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
@@ -113,8 +113,8 @@ public abstract class BaseChatModelSetup extends Resource {
 
     public ChatMessage chat(
             List<ChatMessage> messages,
-            Map<String, Object> arguments,
-            Map<String, Object> parameters) {
+            Map<String, Object> promptArgs,
+            Map<String, Object> modelParams) {
         Preconditions.checkNotNull(
                 connection,
                 "Connection is not initialized. Ensure open() is called before chat().");
@@ -128,8 +128,8 @@ public abstract class BaseChatModelSetup extends Resource {
                     "Prompt is not initialized. Ensure open() is called before chat().");
             Prompt prompt = (Prompt) this.prompt;
             Map<String, String> stringified = new HashMap<>();
-            if (arguments != null) {
-                for (Map.Entry<String, Object> entry : arguments.entrySet()) {
+            if (promptArgs != null) {
+                for (Map.Entry<String, Object> entry : promptArgs.entrySet()) {
                     stringified.put(
                             entry.getKey(),
                             entry.getValue() != null ? entry.getValue().toString() : "");
@@ -155,8 +155,8 @@ public abstract class BaseChatModelSetup extends Resource {
         }
 
         Map<String, Object> params = this.getParameters();
-        if (parameters != null) {
-            params.putAll(parameters);
+        if (modelParams != null) {
+            params.putAll(modelParams);
         }
         return connection.chat(messages, tools, params);
     }
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
@@ -108,10 +108,13 @@ public abstract class BaseChatModelSetup extends Resource {
     public abstract Map<String, Object> getParameters();
 
     public ChatMessage chat(List<ChatMessage> messages) {
-        return this.chat(messages, Collections.emptyMap());
+        return this.chat(messages, Collections.emptyMap(), Collections.emptyMap());
     }
 
-    public ChatMessage chat(List<ChatMessage> messages, Map<String, Object> parameters) {
+    public ChatMessage chat(
+            List<ChatMessage> messages,
+            Map<String, Object> arguments,
+            Map<String, Object> parameters) {
         Preconditions.checkNotNull(
                 connection,
                 "Connection is not initialized. Ensure open() is called before chat().");
@@ -124,15 +127,17 @@ public abstract class BaseChatModelSetup extends Resource {
                     prompt instanceof Prompt,
                     "Prompt is not initialized. Ensure open() is called before chat().");
             Prompt prompt = (Prompt) this.prompt;
-            Map<String, String> arguments = new HashMap<>();
-            for (ChatMessage message : messages) {
-                for (Map.Entry<String, Object> entry : message.getExtraArgs().entrySet()) {
-                    arguments.put(entry.getKey(), entry.getValue().toString());
+            Map<String, String> stringified = new HashMap<>();
+            if (arguments != null) {
+                for (Map.Entry<String, Object> entry : arguments.entrySet()) {
+                    stringified.put(
+                            entry.getKey(),
+                            entry.getValue() != null ? entry.getValue().toString() : "");
                 }
             }
 
             // append meaningful messages
-            List<ChatMessage> promptMessages = prompt.formatMessages(MessageRole.USER, arguments);
+            List<ChatMessage> promptMessages = prompt.formatMessages(MessageRole.USER, stringified);
             for (ChatMessage message : messages) {
                 if ((message.getContent() != null && !message.getContent().isEmpty())
                         || message.getRole() == MessageRole.ASSISTANT) {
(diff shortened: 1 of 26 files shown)


weiqingy:
Looks like the 1 failing check is unrelated to this PR:

- **it-python [ubuntu-latest] [java-17] [python-3.12] [flink-2.1]**: `LLM-output nondeterminism in real-Ollama e2e test` — `test_react_agent_on_local_runner` fails at `_generate_structured_output` (`chat_model_action.py:236`) because the local `qwen3:1.7b` returned a stray tool-call JSON instead of `{"result": N}`; all 3 retries produced different malformed responses. The same test passes on 3 sibling matrix slots in the same CI run (Python 3.11 + Flink 1.20, Python 3.11 + Flink 2.0, Python 3.12 + Flink 2.2), and this PR doesn't modify `_generate_structured_output` or the schema-validation path. Evidence: [failing job log](https://github.com/apache/flink-agents/actions/runs/26266532101/job/77310915528).

weiqingy:
After the retrigger, the prior `it-python` flake passed. A different job is now failing, and the evidence points to an environment flake again:

- **cross-language [ubuntu-latest] [python-3.12] [java-21]**: `Ollama HTTP timeout in multi-turn tool-calling test` — `test_yaml_cross_language_agent` hung ~5 minutes before `java.net.http.HttpTimeoutException: request timed out` against `localhost:11434`. The runner was CPU-only (`No NVIDIA/AMD GPU detected. Ollama will run in CPU-only mode.`), and the test requires tool-calling with continuations on `qwen3:1.7b` — the heaviest scenario in the cross-language matrix. The sibling `test_java_chat_model_integration` passed on the same commit, exercising the same new `chat(messages, arguments, kwargs)` Pemja plumbing — so the new code path is working; the failure surface is the downstream HTTP layer (`java.net.http.HttpClient`), not PR-touched marshalling code. A signature/marshalling regression would surface as `TypeError`/`ClassCastException`/serialization error, not an HTTP timeout. Evidence: [failing job log](https://github.com/apache/flink-agents/actions/runs/26272211932/job/77328109146).

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
```diff
@@ -108,10 +108,13 @@ public void open() throws Exception {
...
     public ChatMessage chat(List<ChatMessage> messages) {
-        return this.chat(messages, Collections.emptyMap());
+        return this.chat(messages, Collections.emptyMap(), Collections.emptyMap());
     }
 
-    public ChatMessage chat(List<ChatMessage> messages, Map<String, Object> parameters) {
+    public ChatMessage chat(
+            List<ChatMessage> messages,
+            Map<String, Object> arguments,
+            Map<String, Object> parameters) {
```
xintongsong: It could be confusing what are the differences between arguments and parameters. I'd suggest to user more explicit names, e.g., `modelParams`, `promptArgs`.

api/src/main/java/org/apache/flink/agents/api/event/ChatRequestEvent.java
```diff
@@ -39,17 +40,26 @@ public class ChatRequestEvent extends Event {
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
     public ChatRequestEvent(
-            String model, List<ChatMessage> messages, @Nullable Object outputSchema) {
+            String model,
+            List<ChatMessage> messages,
+            @Nullable Map<String, Object> arguments,
```
xintongsong: Same here (and other occurrences). It's a bit hard to connect this `arguments` with the prompt of model setup.

xintongsong (commented):
Thanks for working on this, @weiqingy. LGTM in general. I just have some minor comments. In addition to the inline comments, I think we also need to update the documentation w.r.t. this change.

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
weiqingy: Good call — renamed for clarity in `1aba525`. The two maps are now `promptArgs` (the prompt-template variables) and `modelParams` (the values merged into the model request), which also lines up with the model-params meaning already carried one layer down in `BaseChatModelConnection.chat(...)`. The same split is threaded through `ChatRequestEvent`, `ChatModelAction`, and the Pemja bridge.

api/src/main/java/org/apache/flink/agents/api/event/ChatRequestEvent.java
weiqingy: Renamed here and at the other occurrences in `1aba525`: the constructor parameter and getter are now `promptArgs` / `getPromptArgs()` (Python `prompt_args`), and the event attribute, Pemja kwargs key, and persisted tool-request context key are all `prompt_args`. I left the connection-layer `arguments` (which means model params) and the LLM tool-call function `arguments` untouched, since those are separate concepts and out of this change's scope. I filed #715 to rename the connection-layer `arguments` → `modelParams` as a follow-up, so the naming is consistent end-to-end.

weiqingy:
@xintongsong Thanks for the review! Pushed the changes:

- **Naming** (`1aba525`): `arguments` → `promptArgs`/`prompt_args` and the setup-level `parameters` → `modelParams`, across the Java and Python `chat()` surface, `ChatRequestEvent`, `ChatModelAction`, and the Pemja bridge.
- **Docs** (`f698fb4`): `prompts.md` and the workflow-agent docs now pass prompt-template variables via `ChatRequestEvent(prompt_args=...)` / the `promptArgs` constructor argument, and I fixed the prose that still referenced `ChatMessage.extra_args`. Updated the documentation checkbox to `doc-included`.

I also merged the latest `main` and regenerated the cross-language `ChatRequestEvent` snapshot for the new `prompt_args` attribute (`9411c49`), so CI is green again.

xintongsong (approved):
Thanks, @weiqingy. LGTM

Issue #693, opened 2026-05-20
[doc] Add doc and example for introducing YAML API


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

After introducing yaml api in flink-agents, we need to supplement the relevant documentation and examples.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Issue #696, opened 2026-05-20
[Doc] Add doc and example for using skills


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Add documentation and example for agent skills integration.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #705, opened 2026-05-25
[doc] Add YAML API documentation and quickstart example.

wenjin272:
Linked issue: #693

### Purpose of change

Add YAML API documentation and quickstart example.

### Tests

No

### API

No

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/vector_stores.md
+++ b/docs/content/docs/development/vector_stores.md
@@ -1,6 +1,6 @@
 ---
 title: Vector Stores
-weight: 6
+weight: 7
 type: docs
 ---
 <!--
--- a/docs/content/docs/development/yaml.md
+++ b/docs/content/docs/development/yaml.md
@@ -153,7 +153,7 @@ tools:
     parameter_types: [java.lang.Integer, java.lang.Integer]
 ```
 
-**Skills** — bundles of agent skill assets loaded from one or more sources. At least one of `paths` / `urls` / `classpath` / `package` must be non-empty; multiple sources can coexist. See [Agent Skills]({{< ref "docs/development/workflow_agent" >}}) for the skill loading model.
+**Skills** — bundles of agent skill assets loaded from one or more sources. At least one of `paths` / `urls` / `classpath` / `package` must be non-empty; multiple sources can coexist.
 
 | Field | Required | Description |
 |-------|----------|-------------|
--- a/docs/content/docs/development/yaml.md
+++ b/docs/content/docs/development/yaml.md
@@ -1,0 +1,555 @@
+---
+title: YAML API
+weight: 3
+type: docs
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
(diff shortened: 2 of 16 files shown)


docs/content/docs/development/vector_stores.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Vector Stores
-weight: 6
+weight: 7
```
weiqingy, line 3: `weight: 7` here ties with `docs/content/docs/development/memory/_index.md`, which is also pinned to `weight: 7`. At a tie, Hugo's sidebar order falls back to alphabetic path, so Memory and Vector Stores can swap depending on Hugo version. How do you feel about bumping Memory to 8 in the same PR so each entry in the development tree has a unique weight? Open to other ideas if there's a reason to keep Memory at 7.

docs/content/docs/development/yaml.md
```diff
@@ -0,0 +1,555 @@
...
+
+```yaml
+tools:
+  - name: add
+    type: java
+    function: com.example.MyTools:add
+    parameter_types: [java.lang.Integer, java.lang.Integer]
+```
+
+**Skills** — bundles of agent skill assets loaded from one or more sources. At least one of `paths` / `urls` / `classpath` / `package` must be non-empty; multiple sources can coexist. See [Agent Skills]({{< ref "docs/development/workflow_agent" >}}) for the skill loading model.
```
weiqingy: The Agent Skills cross-link here — `[Agent Skills]({{< ref "docs/development/workflow_agent" >}})` — currently lands on `workflow_agent.md`, which doesn't have any Skills content yet (`grep -i skills` on that file is empty). Is there a different destination doc that's planned for the skill loading model — maybe one tracked by #696? If a Skills doc isn't on the way yet, would it work to drop the link and revisit once one lands, or do you have a different angle in mind?

python/flink_agents/examples/quickstart/yaml_review_analysis_agent.yaml
```diff
@@ -0,0 +1,63 @@
...
+              If the review mentions shipping dissatisfaction, first call
+              notify_shipping_manager, then return the JSON above with no
+              mention of the tool call.
+          - role: user
+            content: |
+              "input":
+              {input}
+
+    chat_model_setups:
+      - name: review_analysis_model
```
weiqingy, line 55: Noticed the Java sibling YAML (`examples/src/main/resources/yaml/yaml_review_analysis_agent.yaml`) has an inline comment right above this line explaining that `processInput` sends `ChatRequestEvent("reviewAnalysisModel", ...)` and the chat-model setup name has to match. The Python side has the same coupling — `ReviewAnalysisAgent.process_input` sends `ChatRequestEvent(model="review_analysis_model", ...)` — but the Python YAML doesn't mention it. Worth mirroring the same comment on this side so a reader who renames the setup sees the constraint upfront? Suggested wording, in case it helps:

```yaml
# Action `process_input` sends `ChatRequestEvent(model="review_analysis_model", ...)`,
# so the chat-model setup MUST be named `review_analysis_model`.
chat_model_setups:
  - name: review_analysis_model
```

weiqingy (commented):
Thanks for putting this together — the YAML doc is dense in the right places (cross-language section + JSON Schema contract), and the quickstart mirrors the workflow-agent quickstart cleanly. Left a few comments inline on specific lines below.

wenjin272:
Thanks for the careful review @weiqingy! All three addressed in the latest commit:

  1. **weight collision** — Good catch. Bumping Memory to 8 alone would have collided with Tool Use, so I shifted the tail of the development tree: Memory → 8, Tool Use → 9, MCP → 10, Integrate with Flink → 11. Every sidebar entry now has a unique weight (1–11).
  2. **dead Agent Skills link** — You're right, `workflow_agent.md` has no Skills content. We plan to add a dedicated Agent Skills chapter later, so I've dropped the link for now and will link it once that doc lands.
  3. **missing comment on the Python YAML** — Mirrored the naming-constraint comment from the Java sibling onto the Python quickstart YAML. Thanks for the suggested wording.

weiqingy (approved):
LGTM

xintongsong (approved):
LGTM

Issue #622, opened 2026-04-10
[Feature] Support using cross-language actions.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, flink-agents has already supported using cross-language resources. The next step is support using cross-language actions. This depend on the event refactor #424 

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

addu390:
@wenjin272 I can take this on, given it's related work. But only after #424 is complete.

wenjin272:
@addu390, thanks for your interest in contributing. I've assigned this issue to you. As you mentioned, this needs to wait until #424 is completed.

wenjin272:
Hi, @addu390, since #424 is completed, we can start this task.

Before implementing, I think you could prepare a design document, and post it to github discussions. This way, we can discuss the design document first to avoid rework caused by differing opinions.

addu390:
@wenjin272 Agreed! I'll come with a design document and a few snippets soon.

wenjin272:
Hi @addu390, the code freeze for flink-agents 0.3 is on May 31. However, I'll be on vacation for two weeks starting May 22. So if you could send out the design for discussion before I leave, I'd be happy to help review it. That said, please go at your own pace — after May 22, we have other developers who can help review your design as well.

Additionally, in #670, I added support for representing JavaFunction in Python. I believe this is also necessary for cross-language action execution, so you may refer to it.

addu390:
Hi @wenjin272, I'm away as well, until May 19th, but I'll post the design document after I'm back on 20th. And agreed! Would to nice to get a review from you before the vacation.

addu390:
Hi @wenjin272 
While going over the current state of where things stand after #631 and #670, the cross-language action mechanism seems to be almost there? A Python agent can do `add_action(JavaFunction(...))` and a Java agent can do `addAction(new PythonFunction(...))`, and the operator dispatches to the right side.

Haven't tested, but something like the below should work:
```
class MyAgent(Agent):
    def __init__(self):
        super().__init__()
        self.add_action(
            name="handle",
            events=["my_event"],
            func=JavaFunction(
                qualname="com.example.Handlers",
                method_name="handle",
                parameter_types=["org.apache.flink.agents.api.Event",
                                 "org.apache.flink.agents.api.context.RunnerContext"],
            ),
        )
```
[...]

PR #709, opened 2026-05-27
[Feature] Cross Language Actions

addu390:
Linked issue: #622

### Purpose of change

Add symmetrical cross-language action support so a Java agent can declare a Python-bodied action via `@Action(target=@PythonFunction(...))` and a Python agent can declare a Java-bodied action via `@action(target=JavaFunction(...))`. Aligns built-in event SerDe (`Event`, `ChatMessage`, `MessageRole`, `ToolResponse`, `Document`) so events round-trip across the Java<->Python wire.

### Tests

- Plan-compile and JSON-shape tests on both sides (`AgentPlanCrossLanguageTest`, `test_agent_plan_cross_language`).
- Event snapshot round-trip tests against shared fixtures under `e2e-test/cross-language-event-snapshots/{java,python}/`.
- Runtime dispatch tests (`CrossLanguageActionRuntimeTest`, `test_local_runner_cross_language`).
- Flink mini-cluster e2e: `JavaAgentWithPythonActionTest` and `python_agent_with_java_action_test`.

### API

Yes:
- Java: `@Action.target()` field, new `@PythonFunction` annotation.
- Python: `@action(target=...)` kwarg.

### Call-outs

Output extraction dispatches on pipeline wire format:
- `EventRouter.getOutputFromOutputEvent` now branches on `inputIsJava` instead of
`event instanceof OutputEvent`. Fixes a `ClassCastException` on Java pipelines whose
action body is Python (the action's `_output_event` deserialized as base `Event`,
not typed `OutputEvent`). Same-language pipelines
are unchanged. Pinned in `EventRouterTest` by three new tests.
- `RowTypeInfo` `output_schema` known gap (no behavior change):
Java emits `{"fieldNames", types:[<Class>]}`, Python emits `{"names",
types:[<BasicTypeInfo ordinal>]}`, doesn't round-trip. Has snapshot tests
with `*_knownGap` with docstrings.

### Documentation

- [x] `doc-not-needed` (Docs covered in a follow-up)

### DIFF
--- a/api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
+++ b/api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
@@ -91,11 +91,13 @@ class CrossLanguageEventSnapshotTest {
         JsonNode actual = MAPPER.readTree(actualJson);
 
         Path committed = snapshotDir.resolve("java/" + fileName);
-        assumeTrue(
+        assertTrue(
                 Files.exists(committed),
                 "Java snapshot "
                         + fileName
-                        + " not committed; run with -Dregenerate.snapshots=true first.");
+                        + " missing from "
+                        + committed
+                        + ". If you added a new event, regenerate with -Dregenerate.snapshots=true and commit alongside the test.");
         JsonNode expected = MAPPER.readTree(Files.readString(committed));
 
         assertEquals(
@@ -108,9 +110,13 @@ class CrossLanguageEventSnapshotTest {
 
     private static Event readPythonSnapshot(String fileName) throws Exception {
         Path pythonSnapshot = snapshotDir.resolve("python/" + fileName);
-        assumeTrue(
+        assertTrue(
                 Files.exists(pythonSnapshot),
-                "Python snapshot " + fileName + " not present; run Python generator first.");
+                "Python snapshot "
+                        + fileName
+                        + " missing from "
+                        + pythonSnapshot
+                        + ". Regenerate the Python side with REGENERATE_SNAPSHOTS=1 and commit alongside this test.");
         return Event.fromJson(Files.readString(pythonSnapshot));
     }
 
--- a/api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
+++ b/api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
@@ -63,6 +63,8 @@ class CrossLanguageEventSnapshotTest {
     private static final UUID FIXED_REQUEST_ID =
             UUID.fromString("00000000-0000-0000-0000-000000000002");
     private static final String FIXED_TOOL_CALL_ID = "call_aaaa";
+    private static final String FIXED_TOOL_CALL_ID_NUMERIC = "call_bbbb";
+    private static final String FIXED_TOOL_CALL_ID_BOOL = "call_cccc";
     private static final long FIXED_TIMESTAMP = 1_700_000_000_000L;
 
     private static Path snapshotDir;
@@ -215,8 +217,18 @@ class CrossLanguageEventSnapshotTest {
         assertEquals("hello world", msg.getContent());
     }
 
+    /**
+     * Known 0.3 gap — {@link RowTypeInfo}-typed {@code output_schema} does not round-trip across
+     * the language boundary. Java emits {@code {"fieldNames": [...], "types": [<Class>]}} while
+     * Python emits {@code {"names": [...], "types": [<BasicTypeInfo ordinal>]}}, so a {@link
+     * ChatRequestEvent} carrying a {@code RowTypeInfo} schema cannot be deserialized on the other
+     * side. The {@code BaseModel} (Pydantic class) branch is symmetric and works. Reconciling the
+     * {@code RowTypeInfo} wire format requires a canonical shape + bilateral {@code OutputSchema}
+     * serdes shims; tracked as a follow-up.
+     */
     @Test
-    void chatRequestOutputSchemaWireFormatIsJavaShaped() throws Exception {
+    void chatRequestRowTypeInfoOutputSchemaIsNotPortableAcrossLanguages_knownGap()
+            throws Exception {
         OutputSchema schema =
                 new OutputSchema(
                         new RowTypeInfo(
@@ -227,6 +239,8 @@ class CrossLanguageEventSnapshotTest {
                         "test-model", List.of(new ChatMessage(MessageRole.USER, "hi")), schema);
         String json = MAPPER.writeValueAsString(event);
 
+        // Pin Java's local shape so a future regression can't silently change it. The gap with
+        // Python's `{"names": ...}` shape is the documented limitation, not the assertion.
         assertTrue(json.contains("\"fieldNames\""), "Java wire format uses `fieldNames`.");
         assertFalse(json.contains("\"names\""), "Java wire format does not use Python's `names`.");
     }
@@ -331,13 +345,30 @@ class CrossLanguageEventSnapshotTest {
     }
 
     @Test
-    void pythonToolResponseEventLosesDataWhenConsumedByJava() throws Exception {
+    void pythonToolResponseEventRoundTripsScalarResponses() throws Exception {
         Event base = readPythonSnapshot("tool_response_event.json");
         ToolResponseEvent typed = ToolResponseEvent.fromEvent(base);
 
         assertEquals(FIXED_REQUEST_ID, typed.getRequestId());
-        assertTrue(typed.getResponses().isEmpty());
 
+        ToolResponse stringResp = typed.getResponses().get(FIXED_TOOL_CALL_ID);
+        assertNotNull(stringResp, "String response should be wrapped as ToolResponse.success.");
+        assertEquals("pong", stringResp.getResult());
+        assertTrue(stringResp.isSuccess());
+
+        ToolResponse numericResp = typed.getResponses().get(FIXED_TOOL_CALL_ID_NUMERIC);
+        assertNotNull(numericResp, "Number response should be wrapped as ToolResponse.success.");
+        assertEquals(42, ((Number) numericResp.getResult()).intValue());
+        assertTrue(numericResp.isSuccess());
+
+        ToolResponse boolResp = typed.getResponses().get(FIXED_TOOL_CALL_ID_BOOL);
+        assertNotNull(boolResp, "Boolean response should be wrapped as ToolResponse.success.");
+        assertEquals(Boolean.TRUE, boolResp.getResult());
+        assertTrue(boolResp.isSuccess());
+
+        // Remaining shape gap: Python's ToolResponseEvent model has no top-level success/error/
+        // timestamp maps (those live inside each ToolResponse on the Java side). Pin it so the
+        // divergence stays visible.
         Map<String, Object> attrs = typed.getAttributes();
         assertFalse(attrs.containsKey("success"));
         assertFalse(attrs.containsKey("error"));
--- a/api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
+++ b/api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
@@ -58,8 +58,10 @@ class CrossLanguageEventSnapshotTest {
 
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
-    private static final UUID FIXED_EVENT_ID = UUID.fromString("00000000-0000-0000-0000-000000000001");
-    private static final UUID FIXED_REQUEST_ID = UUID.fromString("00000000-0000-0000-0000-000000000002");
+    private static final UUID FIXED_EVENT_ID =
+            UUID.fromString("00000000-0000-0000-0000-000000000001");
+    private static final UUID FIXED_REQUEST_ID =
+            UUID.fromString("00000000-0000-0000-0000-000000000002");
     private static final String FIXED_TOOL_CALL_ID = "call_aaaa";
     private static final long FIXED_TIMESTAMP = 1_700_000_000_000L;
 
@@ -91,13 +93,17 @@ class CrossLanguageEventSnapshotTest {
         Path committed = snapshotDir.resolve("java/" + fileName);
         assumeTrue(
                 Files.exists(committed),
-                "Java snapshot " + fileName + " not committed; run with -Dregenerate.snapshots=true first.");
+                "Java snapshot "
+                        + fileName
+                        + " not committed; run with -Dregenerate.snapshots=true first.");
         JsonNode expected = MAPPER.readTree(Files.readString(committed));
 
         assertEquals(
                 expected,
                 actual,
-                "Java serialization of " + fileName + " drifted from committed snapshot; if intentional, regenerate.");
+                "Java serialization of "
+                        + fileName
+                        + " drifted from committed snapshot; if intentional, regenerate.");
     }
 
     private static Event readPythonSnapshot(String fileName) throws Exception {
@@ -132,7 +138,8 @@ class CrossLanguageEventSnapshotTest {
         Event base = readPythonSnapshot("input_event.json");
         InputEvent typed = InputEvent.fromEvent(base);
 
-        assertEquals(FIXED_EVENT_ID, typed.getId(), "ID lost when deserializing Python InputEvent.");
+        assertEquals(
+                FIXED_EVENT_ID, typed.getId(), "ID lost when deserializing Python InputEvent.");
         assertEquals(InputEvent.EVENT_TYPE, typed.getType());
         assertEquals("hello", typed.getInput(), "InputEvent.input mismatch.");
     }
@@ -161,7 +168,8 @@ class CrossLanguageEventSnapshotTest {
         Event base = readPythonSnapshot("output_event.json");
         OutputEvent typed = OutputEvent.fromEvent(base);
 
-        assertEquals(FIXED_EVENT_ID, typed.getId(), "ID lost when deserializing Python OutputEvent.");
+        assertEquals(
+                FIXED_EVENT_ID, typed.getId(), "ID lost when deserializing Python OutputEvent.");
         assertEquals(OutputEvent.EVENT_TYPE, typed.getType());
         assertEquals("world", typed.getOutput(), "OutputEvent.output mismatch.");
     }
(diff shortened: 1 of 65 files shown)


addu390:
Hi @wzhero1, could you please review when you get a chance. Noticed that cross-language actions is planned to be part of `0.3` with code-freeze on May 31st. Although close, I think it's possible to target.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/EventRouter.java
```diff
@@ -186,31 +186,33 @@ Event wrapToInputEvent(IN input, PythonActionExecutor pythonActionExecutor) thro
...
-                        pythonActionExecutor.getOutputFromOutputEvent(eventJson);
-                return (OUT) outputFromOutputEvent;
-            } catch (Exception e) {
-                throw new IllegalStateException(
-                        "Failed to extract output from event: " + event.getType(), e);
-            }
+        OutputEvent typedEvent =
+                (event instanceof OutputEvent) ? (OutputEvent) event : OutputEvent.fromEvent(event);
+        if (inputIsJava) {
+            return (OUT) typedEvent.getOutput();
```
weiqingy, line 208: For the `inputIsJava == true && event is unified Event` case (e.g. a Python action emitting `_output_event` in a Java pipeline), this now calls `OutputEvent.fromEvent(event).getOutput()` and bypasses `PythonActionExecutor` entirely. The old branch routed everything that wasn't `instanceof OutputEvent` through Python — so this is a deliberate behavior change to a hot path, not just a refactor, and it affects every output event in every pipeline, not just cross-language ones.

Two asks:
- Could this be called out in the PR description so reviewers (and the 0.3 release notes) don't miss it?
- Is there a test under `runtime/` that specifically pins the Java-pipeline + Python-action output path so a future Pemja change doesn't silently regress payload reconstruction? `CrossLanguageActionRuntimeTest` exercises dispatch  — point me at the one covering the recovered output type/value if I missed it.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -228,20 +228,30 @@ private void extractActionsFromAgent(Agent agent) throws Exception {
...
+            org.apache.flink.agents.api.annotation.Action actionAnnotation =
+                    Objects.requireNonNull(
+                            method.getAnnotation(
+                                    org.apache.flink.agents.api.annotation.Action.class));
+            String[] listenEventTypeStrings = actionAnnotation.listenEventTypes();
+            org.apache.flink.agents.api.annotation.PythonFunction target =
+                    actionAnnotation.target();
+
+            org.apache.flink.agents.plan.Function execFunction;
+            if (target.module().isEmpty()) {
```
weiqingy: The `module().isEmpty()` branch picks Java vs Python, but `qualname()` is read unconditionally on the Python path without a non-empty check. `@Action(target = @PythonFunction(module = "pkg"))` (no qualname) would build a `PythonFunction("pkg", "")` and fail much later inside the interpreter with an obscure error.

Would it make sense to validate here — e.g. both empty → Java; exactly one set → reject with a message naming the action; both set → Python? Catching it at extraction time would surface the typo at the layer where the user can actually correct it. (Same question on the Python side — does `@action(target=...)` reject a `JavaFunction` with an empty `method_name` or `parameter_types`?)

api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
```diff
@@ -0,0 +1,483 @@
...
+        Files.createDirectories(target.getParent());
+        Files.writeString(target, json + "\n");
+    }
+
+    private static void assertJavaSnapshotStable(String fileName, Event event) throws Exception {
+        String actualJson = MAPPER.writeValueAsString(event);
+        JsonNode actual = MAPPER.readTree(actualJson);
+
+        Path committed = snapshotDir.resolve("java/" + fileName);
+        assumeTrue(
```
weiqingy: `assumeTrue` here means the test reports as "skipped, not failed" if `chat_request_event.json` (or any other committed snapshot) gets renamed or deleted. The whole point of the snapshot suite is to catch silent wire-format drift, so silently-skipped "stable" tests undercut that guarantee.

The `regenerate*` tests genuinely need conditional skipping (the system property is the gate). For `*JavaSnapshotIsStable` and `javaCanDeserialize*FromPythonSnapshot`, was there a reason for the symmetric `assumeTrue` — any case where a missing snapshot should soft-pass? If not, would a hard `assertTrue(Files.exists(...))` be better? Same question for `readPythonSnapshot:111`.

python/flink_agents/api/tests/test_cross_language_event_snapshots.py
```diff
@@ -0,0 +1,455 @@
...
+def _write_python_snapshot(name: str, event: Event) -> None:
+    target = _SNAPSHOT_DIR / "python" / name
+    target.parent.mkdir(parents=True, exist_ok=True)
+    target.write_text(event.model_dump_json(indent=2) + "\n")
+
+
+def _assert_python_snapshot_stable(name: str, event: Event) -> None:
+    actual = json.loads(event.model_dump_json())
+    committed_path = _SNAPSHOT_DIR / "python" / name
+    if not committed_path.exists():
```
weiqingy: Mirror of the same concern on the Java side: `pytest.skip` when the committed snapshot is missing turns a real wire-format regression into a silent pass-as-skipped. The `REGENERATE_SNAPSHOTS=1` skips on the regenerate tests are legitimate (the env var is the gate).

For the stability / round-trip helpers at lines 64 and 76, was the symmetric `pytest.skip` intentional — any case where a missing snapshot should soft-pass? If not, would `pytest.fail(...)` or a plain `assert committed_path.exists()` be a tighter fit?

weiqingy (commented):
Thanks for taking this on — the Java↔Python snapshot pairs are a clean way to pin the wire format. A few questions inline.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
addu390: Agreed. Validation in `AgentPlan` now checks both fields: both empty is Java, both set is Python, anything in between throws with the action name. Same check on the Python decorator for both `PythonFunction` and `JavaFunction` targets. Also added tests for the reject cases on both sides.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/EventRouter.java
addu390: Missed a few tests, added 3 new unit tests in `EventRouterTest`, `JavaAgentWithPythonActionTest` would be the Pemja regression guard and also added a `call-outs` section in the PR description.

api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
```diff
@@ -0,0 +1,489 @@
...
+        writeJavaSnapshot("tool_response_event.json", buildToolResponseEvent());
+    }
+
+    @Test
+    void toolResponseEventJavaSnapshotIsStable() throws Exception {
+        assertJavaSnapshotStable("tool_response_event.json", buildToolResponseEvent());
+    }
+
+    @Test
+    void pythonToolResponseEventLosesDataWhenConsumedByJava() throws Exception {
```
wzhero1: This test pins the data loss as the expected contract: `ToolResponseEvent.fromEvent` only handles `ToolResponse`/`Map` values (no `String` branch), so a Python action emitting `responses={"call_id": "pong"}` (string-valued, which Python does emit) yields an empty `responses` map on the Java side with no exception — the core Python-action → Java-action path. Because this asserts the loss, the suite stays green while the data is gone and a fix has to break this test first.

Could we add a `String` branch in `fromEvent` so string responses survive, or — if they're genuinely out of scope for 0.3 — rename this to make the gap explicit (e.g. `..._knownGap_NotYetSupported`) and note it in the PR description rather than encoding silent loss as the expected behavior?

api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
```diff
@@ -0,0 +1,489 @@
...
+        assertEquals("test-model", typed.getModel());
+        assertNotNull(typed.getMessages());
+        assertEquals(1, typed.getMessages().size(), "Expected one message.");
+        ChatMessage msg = typed.getMessages().get(0);
+        assertEquals(MessageRole.USER, msg.getRole(), "Role mismatch on Python-produced message.");
+        assertEquals("hello world", msg.getContent());
+    }
+
+    @Test
+    void chatRequestOutputSchemaWireFormatIsJavaShaped() throws Exception {
```
wzhero1: This (and the Python mirror `test_chat_request_output_schema_wire_format_is_python_shaped`) pins a wire incompatibility rather than reconciling it: for a `RowTypeInfo`-typed `output_schema`, Java emits `{"fieldNames": [...], "types": [<Class>]}` while Python emits `{"names": [...], "types": [<BasicType int>]}` — both the keys and the type encodings differ, so a `ChatRequestEvent` carrying a non-null `RowTypeInfo` output_schema can't be deserialized on the other side. (The `BaseModel` `module`/`class` path does look symmetric; this is specifically the `RowTypeInfo` branch.)

Is structured `RowTypeInfo` output across the boundary in scope for 0.3? If yes the two sides need a shared key/format; if no, same ask — make it an explicit documented limitation instead of a green test asserting the mismatch.

api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
```diff
@@ -0,0 +1,489 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+/** Cross-language event SerDe snapshot tests. */
+class CrossLanguageEventSnapshotTest {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    private static final UUID FIXED_EVENT_ID =
```
wzhero1, line 61: Forcing a fixed id here masks a real cross-language divergence: Java assigns event ids with `UUID.randomUUID()` (`Event.java`) while Python derives a deterministic content-based id (`event.py` `_generate_content_based_id`). Because the two sides use different identity models, the committed snapshots only line up if the id is forced — so they don't reflect what either side naturally emits, and the same logical event gets a different id depending on its producer, which breaks id-based dedup/replay/correlation across languages.

Is cross-language event-id stability a requirement? If so the two sides should agree on one scheme (content-derived on both, most likely); if not, worth documenting that ids are language-local and must not be used for cross-boundary correlation.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/JavaResourceAdapter.java
```diff
@@ -180,6 +180,18 @@ public Object invokeJavaTool(
...
+    /** Invoke a Java static action method with positional arguments from Python. */
+    public Object invokeJavaAction(
+            String className,
+            String methodName,
+            List<String> parameterTypes,
+            List<Object> arguments)
+            throws Exception {
+        Method method = resolveMethod(className, methodName, parameterTypes);
+        Object[] args = arguments == null ? new Object[0] : arguments.toArray();
+        return method.invoke(null, args);
```
wzhero1, line 200: `invoke(null, args)` assumes the resolved target is `static`, but `Class.getMethod(...)` doesn't enforce that. A `JavaFunction` whose `method_name` points at an instance method resolves fine here and then throws `NullPointerException` from `Method.invoke` with a message that says nothing about the static requirement — hard to diagnose from the user side.

Could we check `Modifier.isStatic(method.getModifiers())` at resolution time and throw a message naming the method (mirroring `FunctionTool`'s static check), so the misconfiguration surfaces where the user can act on it?

python/flink_agents/plan/agent_plan.py
```diff
@@ -242,21 +242,31 @@ def _get_actions(agent: Agent) -> List[Action]:
     actions = []
     for name, value in agent.__class__.__dict__.items():
         if isinstance(value, staticmethod) and hasattr(value, "_listen_events"):
+            exec_ = (
```
wzhero1: `hasattr(value, "_listen_events")` is checked on the `staticmethod` wrapper, which makes this order-sensitive. With `@action` outermost over `@staticmethod` (the order the examples use), the attribute lands on the staticmethod object and this branch hits. With the conventional `@staticmethod` outermost over `@action`, the attribute is on `__func__`, `hasattr(wrapper, ...)` is False (3.9), and the `elif callable(value)` below also misses (staticmethod isn't callable pre-3.10) — so the action is silently dropped with no error.

Unwrapping first (`inner = value.__func__ if isinstance(value, staticmethod) else value`, then check `inner`) would make both orders work. No example hits the broken order today, so it's preventive — but it's an easy silent-drop trap.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -228,20 +228,39 @@ private void extractActionsFromAgent(Agent agent) throws Exception {
...
         for (Method method : agentClass.getDeclaredMethods()) {
-            if (method.isAnnotationPresent(org.apache.flink.agents.api.annotation.Action.class)) {
-                org.apache.flink.agents.api.annotation.Action actionAnnotation =
-                        method.getAnnotation(org.apache.flink.agents.api.annotation.Action.class);
-
-                String[] listenEventTypeStrings =
-                        Objects.requireNonNull(actionAnnotation).listenEventTypes();
-
-                org.apache.flink.agents.plan.JavaFunction javaFunction =
+            if (!method.isAnnotationPresent(org.apache.flink.agents.api.annotation.Action.class)) {
```
wzhero1, line 247: `getDeclaredMethods()` returns only methods declared on the concrete class, so `@Action` methods inherited from a parent agent class aren't registered. Every agent today is a single-level `extends Agent` with `@Action` on the final class, so nothing triggers this — but factoring shared actions into an intermediate base class (a natural refactor) would make them silently stop registering.

Note `getMethods()` only half-fixes it (catches `public` `@Action` by convention but misses `protected`/package-private); walking the superclass chain calling `getDeclaredMethods()` is the visibility-insensitive fix. Low urgency — worth a guard or a documented constraint.

python/flink_agents/plan/function.py
```diff
@@ -272,17 +272,23 @@ def set_java_resource_adapter(self, adapter: Any) -> None:
...
         """
         if self._j_resource_adapter is None:
             msg = (
                 "JavaFunction requires the JVM resource adapter; not set "
-                "on this descriptor. The runtime should inject it via "
+                "on this descriptor. The runtime injects it via "
                 "set_java_resource_adapter before invocation."
             )
             raise RuntimeError(msg)
+        if args:
```
wzhero1, line 291: Routing on `if args:` (positional → `invokeJavaAction`, else → `invokeJavaTool`) leaves the "positional = action, keyword = tool" contract enforced only by the docstring. A mixed call `func(event, x=1)` would go to `invokeJavaAction` and silently drop `kwargs`; a no-arg call falls to the tool path. Both real call sites are clean today (the action path passes exactly `(event, context)` positionally, the tool path passes pure kwargs), so this isn't a live bug — but it's an unenforced convention that a future caller or refactor could misroute silently.

A cheap assertion (reject simultaneous `args` and `kwargs`, or dispatch on an explicit flag) would make the contract enforced rather than assumed.

wzhero1 (commented):
Thanks for pushing this through before the freeze — the api/plan descriptor split and the Java↔Python snapshot pairs are a solid foundation. Most of my notes are inline: a couple of the snapshot tests pin a cross-language gap as the expected contract (silent data loss / wire incompatibility on the core path), there's an event-identity question, and a few latent dispatch/registration traps.

api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
addu390: 1. **ID survives the language hop via `from_event`.** There's a fix in this PR. `InputEvent.from_event` and `OutputEvent.from_event` in `event.py` now does `result.id = event.id` instead of letting `Event.__setattr__` regenerate a content-based ID. 
The `assertEquals(FIXED_EVENT_ID, typed.getId(), ...)` lines in the round-trip tests would fail without that fix, so `FIXED_EVENT_ID` isn't masking. It's just the fixture that makes the snapshot JSON byte-stable so the preservation assertion can exist.

2. **Independent Java and Python producers don't agree on the ID for logically equivalent events.** Java uses `UUID.randomUUID()` so two identical-payload events get different IDs anyway. Python uses content-derived UUIDv3 so two identical-payload events get the same ID, plus the ID regenerates on every `__setattr__` write (maybe a bug?). So cross-producer dedup/replay across the boundary doesn't work today. Aligning that is a real design call, out of scope for 0.3, unless you'd rather tackle it here.

api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
addu390: Fixed. `ToolResponseEvent.fromEvent` now wraps non-`ToolResponse`/non-Map values via `ToolResponse.success(v)`, so covers primitives `string/number/bool..` round-trip. Renamed the test to infer that and extended the Python snapshot with numeric and boolean entries to pin all three.

Root cause: `responses` is `Dict[UUID, Any]` on Python but `Map<String, ToolResponse>` on Java, so raw scalars satisfied Python's schema with nowhere to land on Java. Tightening Python's type to `Dict[UUID, ToolResponse]` would be the cleaner long-term fix/follow-up.

One caveat: when Python encodes an error case as a `string` in responses (e.g. `tool_call_action.py:48`), Java now wraps it as `success(error_string)` because the Python wire shape doesn't distinguish success from error. The cleaner fix would be tightening Python's responses in a follow-up.

api/src/test/java/org/apache/flink/agents/api/CrossLanguageEventSnapshotTest.java
addu390: Renamed the test (and Python mirror) to `...RowTypeInfoOutputSchemaIsNotPortableAcrossLanguages_knownGap` with Javadoc spelling out the divergence: Java's `{"fieldNames", types:[<Class>]}` vs Python's `{"names", types:[<BasicTypeInfo ordinal>]}`. The BaseModel branch is symmetric and works; only `RowTypeInfo` is affected.

addu390:
@weiqingy @wzhero1 Thanks for the review, replied/Addressed the comments. Ready for an other review.

wzhero1:
Thanks for the update! LGTM. @xintongsong could you take a final look before merge?

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -227,21 +227,56 @@ private void extractActionsFromAgent(Agent agent) throws Exception {
...
+                        org.apache.flink.agents.api.annotation.Action.class)) {
+                    throw new IllegalStateException(
+                            "Inherited @Action '"
+                                    + parent.getName()
+                                    + "#"
+                                    + inherited.getName()
+                                    + "' is not supported; declare on the concrete agent.");
+                }
+            }
+        }
```
xintongsong, line 245: Ideally, inherited actions should also be included. Given that this issue is not introduced by this PR, I'm okay with not fixing it in this PR. We'd better create an issue to track and fix this in future.

xintongsong (approved):
@addu390, thanks for working on this feature. And Thanks @weiqingy @wzhero1 for the review. LGTM. Merging.

Issue #706, opened 2026-05-26
[Bug] Java chat token metrics are recorded inside async call boundary


joeyutong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The Java built-in chat action can record token metrics from inside the async chat callable.

`ChatModelAction` can execute `chatModel.chat(...)` through `ctx.durableExecuteAsync(...)` when chat async execution is enabled. On the Java chat-model path, token metrics are recorded inside `BaseChatModelConnection.recordTokenMetrics(...)`, which is called from the chat-model connection while the chat call is still running.

This is different from the Python built-in chat path. The Python chat action collects token usage from the chat response and calls `chat_model._record_token_metrics(...)` only after `await ctx.durable_execute_async(...)` resumes. That keeps metric recording outside the async callable.

The metric group is exposed through `RunnerContext#getAgentMetricGroup()` / `getActionMetricGroup()` and Python `ctx.agent_metric_group` / `ctx.action_metric_group`. If the runtime metric group is intended to be used only from the operator/mailbox execution path, the Java built-in chat token recording should follow the Python pattern and avoid touching the metric group inside the async call boundary.

Expected behavior:

- Java and Python built-in chat actions should record token metrics at the same execution boundary.
- Token metrics for Java chat models should be recorded after the async chat call returns, not inside the `durableExecuteAsync` callable.
- The context metric-group getter contract should make it clear whether the returned metric group may be used from async callables, or whether callers must only use it from the operator/mailbox path.

Possible fix directions:

- Move Java chat token metric recording out of `BaseChatModelConnection.recordTokenMetrics(...)` / the connection call path and into `ChatModelAction` after `ctx.durableExecuteAsync(...)` returns.
- Mirror the Python approach by carrying token usage in the `ChatMessage` metadata/extra args and recording it after the chat response is available.
- Clarify the context metric-group getter contract so future code does not accidentally record metrics inside async callables.

### How to reproduce

Use a Java chat model with async chat execution enabled.

1. `ChatModelAction` creates a `DurableCallable` whose `call()` invokes `chatModel.chat(messages, Map.of())`.
2. With chat async enabled, `ChatModelAction` executes the callable through `ctx.durableExecuteAsync(callable)`.
3. The Java chat-model connection records token metrics from inside the chat call via `BaseChatModelConnection.recordTokenMetrics(...)`.

[...]

PR #712, opened 2026-05-29
[api][plan][integrations] Record built-in chat token metrics outside the async call boundary

weiqingy:
Linked issue: #706

### Purpose of change

The Java built-in chat action recorded token metrics from *inside* the durable async callable: each chat-model connection called `BaseChatModelConnection.recordTokenMetrics(...)` within its `chat()` method, which runs on a durable-execution pool thread. The runtime metric group is meant to be used from the operator/mailbox (action) thread, so touching it from the callable crosses that boundary.

This brings Java to the same execution boundary as Python. Mirroring `chat_model_action.py`:

- Each of the 7 Java connections (Ollama, Anthropic, Bedrock, AzureAI, OpenAI Completions/Responses/AzureOpenAI) now stashes `model_name` / `promptTokens` / `completionTokens` into the response `ChatMessage.extraArgs` instead of recording inside `chat()`.
- `ChatModelAction` records **after** the durable call returns (before structured-output reassignment, which would drop the keys) via a new `BaseChatModelSetup.recordTokenMetrics(...)` — the Python-parity record site (`chat_model._record_token_metrics`). The setup's bound metric group is the action metric group, so the **emitted metric path and counter names are unchanged**.
- The old `BaseChatModelConnection.recordTokenMetrics(...)` and its now-dead `connection.setMetricGroup(...)` forwarding are removed.
- The `RunnerContext` metric-group getter Javadoc now documents that the returned group must only be accessed from the operator/mailbox thread, not inside a durable callable.

Recording matches Python's guard exactly: the model name must be non-empty and both token counts greater than zero; values are read as `Number` and converted with `longValue()` to tolerate `Integer`/`Long` across the Pemja bridge and durable recovery.

This also fixes a latent gap: Python-backed chat models invoked from the Java action previously recorded **no** token metrics (the path bypasses the Java connection's recording); they are now captured once.

### Tests

- Relocated the base token-metrics test to the setup (`BaseChatModelSetupTokenMetricsTest`), mirroring Python's `test_token_metrics.py`: records to the per-model sub-group counters, no-ops without a metric group, separate sub-groups per model, `getResourceType() == CHAT_MODEL`.
- Extended `ChatModelActionTest` with cases for the new `recordChatTokenMetrics` helper: records once when all keys are present and positive; `Integer`-typed token values still recorded via `Number.longValue()`; skips when a key is missing or non-numeric; skips when a token is `0` or the model name is empty (Python parity).
- `./tools/build.sh -j`, `./tools/ut.sh -j`, and `./tools/lint.sh -c` all pass.

### API

Adds `public BaseChatModelSetup.recordTokenMetrics(String, long, long)` (the Python-parity record site) and removes the previously `protected` `BaseChatModelConnection.recordTokenMetrics(...)`. No change to public configuration, event, or resource APIs. Emitted metric names and paths are unchanged.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelConnection.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelConnection.java
@@ -19,7 +19,6 @@
 package org.apache.flink.agents.api.chat.model;
 
 import org.apache.flink.agents.api.chat.messages.ChatMessage;
-import org.apache.flink.agents.api.metrics.FlinkAgentsMetricGroup;
 import org.apache.flink.agents.api.resource.Resource;
 import org.apache.flink.agents.api.resource.ResourceContext;
 import org.apache.flink.agents.api.resource.ResourceDescriptor;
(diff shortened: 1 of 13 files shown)


xintongsong (approved):
LGTM

joeyutong:
@xintongsong Should this PR be backported to the release-0.2 branch?

xintongsong:
Exactly. That's what I was wondering by asking for the `affectVersion` label in #706.

PR #714, opened 2026-05-29
[Docs] Cross-Language Actions

addu390:
Linked issue: follow-up to #709

### Purpose of change

Doc-only follow-up to PR #709 (Cross Language Actions), which shipped without doc updates.

### Tests

Docs-only, no tests.

### API

No API changes.

### Documentation

- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/workflow_agent.md
+++ b/docs/content/docs/development/workflow_agent.md
@@ -535,6 +535,13 @@ public class MyAgent extends Agent {
 
 {{< /tabs >}}
 
+{{< hint warning >}}
+**Limitations:**
+
+- Cross-language actions are currently supported only when [running in Flink]({{< ref "docs/operations/deployment#run-in-flink" >}}), not in local development mode
+- Complex object serialization between languages may have limitations
+{{< /hint >}}
+
 ## Event
 
 Events are JSON-serializable messages passed between actions. Every event has a `type` string used for routing and an `attributes` map that carries the payload. A single event may trigger multiple actions if they are all listening to its type.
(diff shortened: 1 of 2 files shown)


addu390:
@xintongsong Today, cross-language is mentioned in a few docs plus the FAQs (Q2/Q3), but there's no concept landing page as such. 

Worth adding a dedicated `docs/content/docs/development/cross_language.md` to consolidate the what / how / limitations, with per-area sections linking to it? Happy to cover that in this PR if useful.

weiqingy (commented):
Thanks for filling the doc gap from #709. One question inline.

docs/content/docs/development/workflow_agent.md
```diff
@@ -490,6 +490,51 @@ To use async execution on JDK 21+, user should append jvm option `--add-exports=
 {{< /tab >}}
 {{< /tabs >}}
 
+### Cross-language Actions
+
+An action declared in one language can dispatch its body to the other language by setting a `target` on the decorator/annotation. The decorated function or annotated method then acts as a stub — it should raise so direct calls outside the framework fail loud.
```
weiqingy, line 495: The three other cross-language sections — chat models (`chat_models.md`), embedding models, and vector stores — each carry the same caveat: *"Cross-language resources are currently supported only when running in Flink, not in local development mode."* The Java-from-Python action path documented here goes through the same Pemja bridge inside `ActionExecutionOperator` (`CrossLanguageActionRuntimeTest` exercises it via the Flink operator harness, not local mode). Does that same local-dev limitation apply to cross-language actions? If so, a matching one-line note here would save users a confusing local-mode failure — and there's no general caveat elsewhere in this page that already covers it.

weiqingy (approved):
Thanks for the quick turnaround. The limitations block matches the wording in `chat_models.md`, `embedding_models.md`, and `vector_stores.md` line-for-line — including the `run-in-flink` ref, which resolves to the `## Run in Flink` heading in `deployment.md`. That covers the local-mode caveat I raised. LGTM.

xintongsong (approved):
LGTM

Issue #716, opened 2026-05-31
[Tech Debt] Flaky live-LLM e2e/cross-language CI tests cause frequent red CI


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Summary

Several end-to-end / cross-language CI tests that depend on a **live Ollama (`qwen3:1.7b`) server** fail intermittently, turning CI red on unrelated PRs and occasionally on `main` itself. The failures are not caused by the code under test — the same tests pass on re-run with no change. This issue proposes mitigating the flakiness so CI signal becomes trustworthy.

### Evidence (last ~35 `Flink Agents CI` runs)

Across the most recent ~35 workflow runs, **~18 ended in failure**. Separating *real* failures from *flaky* ones:

| Category | Jobs | Nature |
|---|---|---|
| **Flaky** (live-LLM e2e) | `cross-language`, `it-python` (e2e), `it-java` (e2e) | Non-deterministic — pass on re-run |
| Real (excluded here) | `ut-python` snapshot/unit, `Code Style Check` | Deterministic, branch-specific — *not* part of this issue |

The flaky failures reproduce **across many different branches, including `main`**, which confirms they come from shared test infrastructure rather than any single PR's code:

```
main                          → httpx.ReadTimeout (×2)
feature/cross_language_actions → ReadTimeout / assert ≠ 1386528
fix/cross-language-test-order  → ReadTimeout / assert ≠ 1386528 (failed 6 of 7 runs)
docs/cross_language_actions    → httpx.ReadTimeout
new_short_memory_ttl           → httpx.ReadTimeout
220-impl                       → assert ≠ 1386528
```

### Two distinct flaky signatures

**1. LLM non-determinism — wrong tool-calling result**
`flink_agents/e2e_tests/e2e_tests_integration/react_agent_test.py::test_react_agent_on_local_runner`

The ReAct agent should compute `(2123 + 2321) * 312 = 1386528` via the `add`/`multiply` tools, but the 1.7B model returns a wrong value. Observed across runs (all on the *same* assertion, all different garbage values):

```
assert 45750    == 1386528
assert 4927483  == 1386528
assert 432596736 == 1386528   (×2)
```

The test already carries an in-line admission of the flakiness:
> `"This may be caused by the LLM response does not match the output schema, you can rerun this case."`

**2. Ollama request timeout**
[...]

PR #717, opened 2026-05-31
[test] Mitigate flaky live-LLM e2e/cross-language CI tests

weiqingy:
Linked issue: #716

### Purpose of change

The live-LLM e2e and cross-language CI tests run a small Ollama model (`qwen3:1.7b`) and fail intermittently — either the model returns a wrong tool-call result (e.g. `assert <varies> == 1386528`) or an Ollama call exceeds its read timeout (`httpx.ReadTimeout`). These flakes reproduce across many branches including `main`, turning CI red on unrelated PRs. See #716 for the failure statistics and evidence.

This PR mitigates the flakiness without masking real, deterministic failures:

1. **Per-test retry, scoped to the live-LLM e2e/cross-language suites only.** Python uses `pytest-rerunfailures` (`--reruns 2 --reruns-delay 5`); Java uses Surefire `-Dsurefire.rerunFailingTestsCount=2`. Both are applied only at the e2e/cross-language test invocations — the unit and style invocations are untouched, so a genuine regression still fails immediately. A test that passes on retry produces a green build but is reported as a flake (pytest `R` markers / Surefire "Flakes"), so the signal is preserved, not hidden.

2. **Close the one remaining 30 s timeout gap.** The cross-language test's Python-wrapped Ollama connection fell back to the Python default `request_timeout` of 30 s; this sets it to 240 s, matching the Java-native connection already configured in the same test agent.

Loosening exact-equality assertions on LLM output (asserting tool-call shape rather than an exact value) is noted in #716 as longer-term hardening and is intentionally out of scope here.

### Tests

This is a CI/test-configuration change, so no new product tests are added (a test asserting "retry is configured" would be tautological). Verified:
- The retry flags are scoped correctly — a deliberately-failing test is retried 3× under an e2e selector but runs once under the unit selector; the unit pytest (`-k "not e2e_tests"`) and unit mvn (`-pl "${exclude_list}"`) invocations are unchanged.
- `pytest-rerunfailures==16.3` resolves against `pytest==9.0.3`, and the e2e jobs install it via `tools/build.sh`'s `uv sync --extra dev` (the `dev` extra composes `test`) before the `--no-sync` pytest runs.
- `mvn ... -Dsurefire.rerunFailingTestsCount=2 test-compile` is accepted on Surefire 3.5.2; the cross-language module compiles with the timeout change.

### API

No public API change. All changes are test-configuration (CI scripts, the Python test extra, and one e2e test agent).

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/e2e-test/flink-agents-end-to-end-tests-resource-cross-language/src/test/java/org/apache/flink/agents/resource/test/ChatModelCrossLanguageAgent.java
+++ b/e2e-test/flink-agents-end-to-end-tests-resource-cross-language/src/test/java/org/apache/flink/agents/resource/test/ChatModelCrossLanguageAgent.java
@@ -78,6 +78,7 @@ public class ChatModelCrossLanguageAgent extends Agent {
         return ResourceDescriptor.Builder.newBuilder(
                         ResourceName.ChatModel.PYTHON_WRAPPER_CONNECTION)
                 .addInitialArgument("pythonClazz", ResourceName.ChatModel.Python.OLLAMA_CONNECTION)
+                .addInitialArgument("request_timeout", 240)
                 .build();
     }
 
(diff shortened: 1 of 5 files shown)


xintongsong (approved):
LGTM

Issue #715, opened 2026-05-30
[Tech Debt] Rename connection-layer chat() argument from "arguments" to "modelParams" for namin


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

#698 renamed the setup-layer `BaseChatModelSetup.chat()` parameters to `promptArgs` (prompt-template variables) and `modelParams` (model parameters), to remove the confusion between two near-synonymous names ([reviewer feedback](https://github.com/apache/flink-agents/pull/698#discussion_r3322887198)).

One overloaded use was deliberately scoped out of that PR: the connection layer still names its model-parameters argument `arguments`:

```java
// BaseChatModelConnection.java
public abstract ChatMessage chat(
        List<ChatMessage> messages, List<Tool> tools, Map<String, Object> arguments);
```

So the setup's `modelParams` is passed straight into the connection's `arguments` — the same concept under two names across a single call hop — and `arguments` here also collides with the (now-renamed) prompt-template `arguments`.

**Proposed:** rename the connection-layer `arguments` → `modelParams` (Java) so the name is consistent end-to-end.

**Scope (Java):**
- `BaseChatModelConnection.chat(...)` (interface + Javadoc)
- `PythonChatModelConnection` (Pemja bridge)
- Provider connections: Anthropic, Ollama, Bedrock, OpenAI (Responses + Completions + AzureOpenAI), AzureAI
- Affected tests

**Out of scope:**
- Python connection layer — already uses `**kwargs` for model params, so it is already consistent.
- MCP `arguments` (`MCPTool` / `MCPPrompt` / `MCPServer`) — an unrelated concept (MCP tool-call parameters).

**Compatibility:** this changes a parameter name on a public abstract method, so custom `BaseChatModelConnection` implementations would update their override signature. It is erasure-compatible (positional), so call sites are unaffected.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #720, opened 2026-05-31
[integrations][api] Rename connection-layer chat() argument to modelParams

weiqingy:
Linked issue: #715

### Purpose of change

After #698 renamed the setup-layer `BaseChatModelSetup.chat()` model-parameters argument to `modelParams`, the connection layer still named the same value `arguments` — the setup's `modelParams` was passed straight into `BaseChatModelConnection.chat(... Map<String, Object> arguments)`, i.e. one concept under two names across a single call hop, and `arguments` also collided with the (now-renamed) prompt-template and the tool-call `arguments`.

This renames the connection-layer model-parameters identifier `arguments` → `modelParams` so the concept has one consistent name end-to-end. It covers the `BaseChatModelConnection.chat(...)` abstract method (signature + Javadoc), the Pemja bridge override, the seven provider connections (Anthropic, Ollama, Bedrock, OpenAI Responses/Completions/AzureOpenAI, AzureAI), and their in-file model-params locals/helper params (including `rawArguments` → `rawModelParams` in the `buildRequest` helpers).

Unrelated `arguments` are deliberately left untouched: tool-call `"arguments"` JSON keys, the `parseArguments`/`serializeArguments` tool-call helpers, MCP `arguments`, and `JavaResourceAdapter`'s tool-call `arguments`.

**Compatibility:** `BaseChatModelConnection.chat(...)` is a public abstract method, so custom connection implementations update their override's parameter name. The change is erasure-compatible (positional), so call sites are unaffected, and it is invisible across the Python boundary — the Pemja bridge forwards the map contents, never a key named `"arguments"`, so the Python connection layer (which already uses `**kwargs`) needs no change.

### Tests

No new tests — this is a behavior-preserving, name-only positional rename; existing tests recompiling and passing against the new signature is the correctness guarantee. Verified locally:
- `mvn -pl api -am test-compile` and `test-compile` of the five chat-model modules: pass
- `mvn -pl api test -Dtest='BaseChatModelSetupSkillsTest,BaseChatModelTest,PythonChatModelConnectionTest'`: 20/20 pass
- `spotless:check` on all touched modules: clean

### API

Yes — renames the parameter of the public abstract method `BaseChatModelConnection.chat(...)` (and all provider overrides) from `arguments` to `modelParams`. Erasure-compatible/positional: call sites unaffected; only custom subclass overrides update the parameter name.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelConnection.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelConnection.java
@@ -50,9 +50,9 @@ public abstract class BaseChatModelConnection extends Resource {
      *
      * @param messages the input chat messages
      * @param tools the tools can be called by the model
-     * @param arguments the additional arguments passed to the model
+     * @param modelParams the additional arguments passed to the model
      * @return the chat response containing model outputs
      */
     public abstract ChatMessage chat(
-            List<ChatMessage> messages, List<Tool> tools, Map<String, Object> arguments);
+            List<ChatMessage> messages, List<Tool> tools, Map<String, Object> modelParams);
 }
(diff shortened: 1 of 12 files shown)


xintongsong (approved):
LGTM

Issue #711, opened 2026-05-28
[Feature] Expose short-term memory state TTL cleanup strategy as a config option


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

#### Background

PR #657 added TTL support for short-term memory state via three config options under `AgentExecutionOptions`:
- `SHORT_TERM_MEMORY_STATE_TTL_MS`
- `SHORT_TERM_MEMORY_STATE_TTL_UPDATE_TYPE`
- `SHORT_TERM_MEMORY_STATE_TTL_VISIBILITY`

Inside `OperatorStateManager.maybeEnableShortTermMemoryTTL`, the cleanup strategy is hardcoded to `cleanupFullSnapshot()`:

```java
StateTtlConfig ttlConfig =
        StateTtlConfig.newBuilder(Duration.ofMillis(ttlMs))
                .setUpdateType(toFlinkUpdateType(updateType))
                .setStateVisibility(toFlinkStateVisibility(stateVisibility))
                .cleanupFullSnapshot()   // hardcoded
                .build();
```

#### Problem

Flink's `StateTtlConfig.CleanupStrategies` offers four cleanup paths, each with different operational trade-offs:

| Strategy | When state is reclaimed | Best for |
|---|---|---|
| Lazy (default, no explicit setter) | On read access only | Read-heavy workloads |
| `cleanupFullSnapshot()` (current hardcoded choice) | During full checkpoint snapshots | Pipelines with frequent full checkpoints |
| `cleanupIncrementally(...)` | Incrementally on every state access | Bounded latency overhead per access |
| `cleanupInRocksdbCompactFilter(...)` | During RocksDB background compaction | Long-running RocksDB-backed pipelines |

For long-running RocksDB pipelines using incremental checkpoints, full-snapshot cleanup means expired entries can persist on disk for extended periods, increasing on-disk state size unnecessarily. `cleanupInRocksdbCompactFilter` is the recommended strategy in that scenario.

Today users cannot pick the right strategy for their deployment without forking the runtime.

#### Proposal

[...]

PR #721, opened 2026-05-31
[api][runtime][python] Expose short-term memory state TTL cleanup strategy as a config option

weiqingy:
Linked issue: #711

### Purpose of change

PR #657 added short-term memory state TTL but hardcoded the cleanup strategy to `cleanupFullSnapshot()` in `OperatorStateManager`. This exposes the cleanup strategy as a public config option `SHORT_TERM_MEMORY_STATE_TTL_CLEANUP_STRATEGY`, backed by a new public `ShortTermMemoryTtlCleanupStrategy` enum (same pattern as `ShortTermMemoryTtlUpdate` / `ShortTermMemoryTtlVisibility`), so users can pick the strategy that fits their deployment — notably `ROCKSDB_COMPACTION_FILTER` for long-running RocksDB pipelines on incremental checkpoints. Default `FULL_SNAPSHOT` preserves prior behavior exactly.

    FULL_SNAPSHOT             -> cleanupFullSnapshot()
    INCREMENTAL               -> cleanupIncrementally(5, false)
    ROCKSDB_COMPACTION_FILTER -> cleanupInRocksdbCompactFilter(1000L)
    LAZY                      -> disableCleanupInBackground()

The incremental (`5, false`) and RocksDB (`1000L`) parameters reproduce Flink's internal defaults (which are package-private and not referenceable). `ROCKSDB_COMPACTION_FILTER` is only effective with the RocksDB state backend; `LAZY` disables background cleanup so expired entries are reclaimed on read access only.

### Tests

- Java unit test on the strategy translator, asserting each enum value produces the correct `StateTtlConfig` cleanup strategy via `getCleanupStrategies()`.
- Existing `ShortTermMemoryTTLIntegrationTest` still passes (the `FULL_SNAPSHOT` default is unchanged).
- Python enum round-trip test, plus the cross-language compatibility-test assertions (key/type/default parity with the Java option).

### API

Yes. Adds one public config option `AgentExecutionOptions.SHORT_TERM_MEMORY_STATE_TTL_CLEANUP_STRATEGY` and the public enum `ShortTermMemoryTtlCleanupStrategy` (Java + Python). No Flink types are exposed in the public surface.

### Documentation

The configuration reference (`docs/operations/configuration.md`) now documents the full short-term-memory TTL option family in the Core Options table — `state-ttl.ms` / `.update-type` / `.visibility` (introduced in #657 but previously undocumented) plus the new `.cleanup-strategy`.

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`


weiqingy:
@xintongsong @da-daken Could you please help review this PR when you get a chance? Thanks!

xintongsong:
Hi @weiqingy ,

It seems to me there are some misunderstanding of the TTL clean-up strategies. I checked the implementation of [StateTtlConfig](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java) and find that:
1. The cleanup strategies are additive rather than exclusive. This means a config option that only allows one value may not be expressive enough.
2. `INCREMENTAL` and `ROCKSDB_COMPACTION_FILTER` is enabled by default. The only strategy what won't be enabled by default is `FULL_SNAPSHOT`.

Based on these observations, I'm leaning towards not to introduce this new option. Because the whole cleanup strategy mechanism is too complex for the users to understand and tune, and the default behaviors should be good enough for most of the use cases. Unless we see a concrete need for this fine-tuning, these knobs might be unnecessary.

WDYT?

weiqingy:
> Hi @weiqingy ,
> 
> It seems to me there are some misunderstanding of the TTL clean-up strategies. I checked the implementation of [StateTtlConfig](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java) and find that:
> 
> 1. The cleanup strategies are additive rather than exclusive. This means a config option that only allows one value may not be expressive enough.
> 2. `INCREMENTAL` and `ROCKSDB_COMPACTION_FILTER` is enabled by default. The only strategy what won't be enabled by default is `FULL_SNAPSHOT`.
> 
> Based on these observations, I'm leaning towards not to introduce this new option. Because the whole cleanup strategy mechanism is too complex for the users to understand and tune, and the default behaviors should be good enough for most of the use cases. Unless we see a concrete need for this fine-tuning, these knobs might be unnecessary.
> 
> WDYT?

Thanks for digging into `StateTtlConfig` — you're right on both counts. I went through the Flink 2.2 bytecode to confirm:

1. **Additive, not exclusive.** `CleanupStrategies` keeps an `EnumMap<Strategies, CleanupStrategy>` alongside the `isCleanupInBackground` flag, so the strategies accumulate. A single-choice enum is the wrong model — selecting `INCREMENTAL` wouldn't turn off the RocksDB compaction filter, and vice versa.

2. **Background cleanup is on by default.** `Builder`'s constructor sets `isCleanupInBackground = true`. With that flag set, `getIncrementalCleanupStrategy()` and `inRocksdbCompactFilter()` are already active with default parameters — no explicit call required. `FULL_SNAPSHOT` is the only strategy that needs opting in.

The key consequence: the scenario that motivated the issue — long-running RocksDB pipelines wanting `cleanupInRocksdbCompactFilter` — **is already covered by Flink's default**. The option wouldn't unlock anything new there, and the only knob that genuinely deviates from the default is disabling background cleanup, which is niche.

So I agree there's no concrete need that justifies the added public surface and tuning complexity. I'll close this PR and the issue. Thanks for catching the misunderstanding early.

weiqingy:
Closing per review discussion — Flink's default TTL cleanup (incremental + RocksDB compaction filter, on by default) already covers the motivating use case, so the new option isn't needed.

Issue #719, opened 2026-05-31
[Tech Debt] Strengthen live-LLM e2e assertions via structured tool-invocation events


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Follow-up to #716 (item 3, the optional hardening direction). #716's load-bearing flakiness fixes — per-test retry and the Python-connection timeout — shipped in #717 and restored trustworthy CI signal. This issue tracks the remaining improvement: making the live-LLM e2e tests assert on **what the agent did** rather than an exact numeric result.

#### Problem

Several live-LLM e2e tests assert an **exact** final value produced by a small, non-deterministic model (`qwen3:1.7b`). The clearest case is `react_agent_test.py:135` (latest `main`):

```python
input_list.append({"key": "0001", "value": InputData(a=2123, b=2321, c=312)})  # :128
...
assert output_list[0]["0001"].result == 1386528                                 # :135  (2123 + 2321) * 312
```

This single assertion conflates three independent dimensions that the model gets right at very different rates. From the #713 review discussion, the 1.7B model has been observed to: answer directly without tools, call the right tool with wrong arguments, emit a tool call as plain text instead of a real tool call, miss a later calculation step, get a correct tool result but still produce a wrong final answer, or return a response that doesn't match the output schema. So **"the right tool was invoked" and "the final answer is correct" are separate concerns**, and an exact-value check tests the most model-capability-dependent dimension while saying nothing about the rest.

#### Direction — layered assertions

Decompose the single exact-value check into separable assertions, ordered by how deterministic each is:

1. **The correct tool was invoked** (`add` / `multiply`) — order- and model-capability-independent; the most stable signal for the tool-calling path.
2. **The tool was invoked with the expected arguments** (e.g. `add(2123, 2321)`, `multiply(4444, 312)`) — validates input parsing without depending on the model's final reasoning. (Per the #713 thread: invoking the right tool isn't meaningful unless the arguments are also right.)
3. **Final-output correctness** — kept as a *separate* concern. Where layers 1–2 already pin the tool inputs, the final value is deterministic arithmetic the tools computed, so the exact-value check becomes redundant and can relax to a schema/shape check; retries (#717) already cover the residual model-capability flakiness.

[...]

PR #722, opened 2026-06-01
[test][runtime] Strengthen live-LLM e2e tests with structured tool-invocation assertions

weiqingy:
Linked issue: #719

### Purpose of change

Follow-up to #716 (item 3). The live-LLM e2e tests assert on the agent's final output value (`react_agent_test` asserted `result == 1386528`) or weak substrings of it (`"3" in output`, `"22" in output`). A single check conflates three things a small non-deterministic model gets right at different rates: which tool was invoked, with what arguments, and the final synthesized answer — so a failure cannot be localized, and the substring checks barely test anything.

This PR adds layered assertions on the structured `ToolRequestEvent`s the runtime already produces, so CI checks *what the agent did* (which tool, what arguments) rather than only an exact number the model often gets wrong.

Sourced two ways, by execution path:
- **Flink path** (cross-language tests + react remote): a shared helper `collect_tool_invocations(log_dir)` reads the `events-*.log` the `FileEventLogger` already writes.
- **Local runner** (react local, `from_list`/`to_list`): the pure-Python `LocalRunner` has no event log, so a small in-memory capture hook exposes the `ToolRequestEvent`s that already flow through its event deque.

Both yield the same `{name, arguments}` shape, so assertions read identically.

Notes on two deliberate choices:
- Final-value checks are kept (not relaxed). The agent's `.result` is a separate model synthesis step, not a tool output, so it can be wrong even when the tool calls are correct — this was observed live (the model called `multiply(4444, 312)` correctly but emitted a wrong final number). The value check catches a failure the tool assertions cannot, so it remains; its residual flakiness is covered by the agent's retries and the per-test retry from #716.
- For the react local test we assert the `multiply` invocation, not `add`. The small model frequently computes the addition itself and only calls the multiply tool, so an `add` assertion would be an unreliable signal; `multiply`'s first argument is the threaded sum, so asserting it proves the addition was computed correctly and the tool was used.

### Tests

- New fixture-based unit tests for the helpers (`collect_tool_invocations`, `assert_tool_invoked`, `tool_invocations_from_events`) — no live model required.
- New unit test for the local-runner capture hook, asserting both that the event is captured and that it still dispatches to its action (so capture cannot silently break tool execution).
- The strengthened e2e tests: chat_model / yaml / react (remote + local) cross-language tests.
- The react local-runner test was exercised live (Ollama, qwen3:1.7b) across many runs to confirm the captured tool calls and arguments end-to-end. The Flink-path e2e tests run in CI.

### API

Adds a test-facing accessor `get_tool_request_events()` on `LocalRunner` and `LocalExecutionEnvironment` (Python runtime), returning the `ToolRequestEvent`s captured during execution. No change to the Java side; the events were already emitted there.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`


wenjin272 (approved):
LGTM.

I believe it is highly meaningful to incorporate structured validation rather than merely comparing the final results. Ty @weiqingy.

At the moment, I feel that `local_runner` and `local_execution_environment` are of limited significance; however, we must adapt to them when making certain changes. In the next version, we can consider deprecating `local_runner` and `local_execution_environment`.

weiqingy:
Thanks for the review, and agreed on the structured-validation direction.

On `local_runner` / `local_execution_environment` — good point. They exist as a pure-Python, in-process path (selected when no Flink `StreamExecutionEnvironment` is passed), so they're not the production path, yet every execution-layer change has to be carried through them too — as this PR's tool-invocation assertions were. I've opened #829 to scope what deprecating them would involve (migration path for the e2e tests and quickstart/example scripts that depend on the local path, and a deprecation timeline) so we can weigh it for a future version.

Issue #724, opened 2026-06-01
[Bug] AgentPlan resource getter is not thread-safe on release-0.2


joeyutong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

On the `release-0.2` branch, resource resolution is cached in `AgentPlan`, but `AgentPlan#getResource(String, ResourceType)` is not thread-safe.

The method currently performs a multi-step check/create/cache sequence without a shared lock:

1. check `resourceCache` for an existing resource
2. call `ResourceProvider#provide(...)` if missing
3. put the new resource into `resourceCache`

Because the check and put are not atomic, two concurrent first accesses to the same resource can both pass the cache check, both create a resource instance, and then race when storing the result. That violates the expected cached-resource semantics. `AgentPlanTest#testGetResourceFromResourceProvider` already asserts that repeated `getResource(...)` calls should return the same resource instance, but the current implementation only guarantees that for sequential calls.

This can be reached from the public runtime path on JDK 21:

- `JavaActionTask` executes Java actions through `ContinuationActionExecutor`.
- `JavaRunnerContextImpl#durableExecuteAsync(...)` delegates to `ContinuationActionExecutor#executeAsync(...)` when a continuation context is present.
- The JDK 21 `ContinuationActionExecutor` implementation submits the callable to a fixed async thread pool.
- If the callable resolves a resource through `RunnerContext#getResource(...)`, `RunnerContextImpl#getResource(...)` delegates to `agentPlan.getResource(...)` on that async thread.

Potential symptoms include duplicate model/tool/connection initialization and duplicate registration of resource-side state such as metrics when two instances of the same logical resource are created concurrently.

The issue appears to be incidentally covered on `main` by later refactoring commits rather than by a dedicated bug-fix commit:

- https://github.com/apache/flink-agents/commit/f8eac10b870ad62ab1799fe9d7d35cc25d40185c (`[plan] Extract ResourceCache and PythonResourceBridge from AgentPlan (#548)`) moved runtime resource caching out of `AgentPlan` into `ResourceCache` and introduced `public synchronized Resource getResource(...)`.
- https://github.com/apache/flink-agents/commit/487bd5844229f9b6712b56fac58a21a8026db62a (`[runtime] Own ResourceContext inside ResourceCache for shared lifecycle.`) later adjusted the same cache/lifecycle area while keeping the synchronized resource resolution path.

### How to reproduce

[...]

PR #727, opened 2026-06-02
[python][plan] Make AgentPlan.getResource thread-safe (backport #548 to release-0.2)

weiqingy:
Linked issue: #724

### Purpose of change

Backport the resource-cache thread-safety fix to `release-0.2`.

`AgentPlan#getResource` performs a non-atomic check-create-cache sequence. When two actions concurrently trigger the first access to the same uncached resource, both can invoke the provider and create duplicate instances (duplicate model/tool init, duplicate metric registration) before the last writer wins the cache.

On `main` this was addressed as a side effect of #548 (f8eac10), which moved resource caching out of `AgentPlan` into a dedicated `ResourceCache` whose `getResource` is synchronized. That refactor never landed on `release-0.2`, so the race is still live there. Backporting the full `ResourceCache` extraction to a maintenance branch is heavier than warranted; the minimal equivalent is to guard the lazy resolution with a reentrant lock — the same mechanism `main` relies on.

This PR applies that minimal guard on both language sides:

- Java: make `AgentPlan#getResource` `synchronized`. The intrinsic lock is reentrant, so the nested resolution callback can still call back into `getResource` on the same thread.
- Python: guard `AgentPlan.get_resource` with a per-instance reentrant lock. The async action thread pool runs actions on multiple threads, and `provider.provide` can release the GIL when it calls into Java via Pemja, so the same race exists. The lock is a `cached_property` (not a Pydantic private attribute) so it stays out of model equality.

### Tests

- Java: `AgentPlanTest#getResourceShouldCreateOnlyOneInstanceUnderConcurrentAccess` — two threads race on the first access to a shared resource whose provider sleeps; asserts a single instance is created and both callers see it.
- Python: `test_get_resource_single_instance_under_concurrent_access` — same contract via two threads and a `threading.Barrier`; asserts the provider runs once.

Both tests fail without the corresponding guard and pass with it.

### API

No public API changes.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
@@ -223,12 +223,17 @@ public class AgentPlan implements Serializable {
     /**
      * Get resource from agent plan.
      *
+     * <p>Synchronized so the check-create-cache sequence is atomic: concurrent first accesses to
+     * the same uncached resource invoke the provider once and share a single instance. The
+     * intrinsic lock is reentrant, so the nested resolution callback below can call back into this
+     * method on the same thread.
+     *
      * @param name the resource name
      * @param type the resource type
      * @return the resource instance
      * @throws Exception if the resource cannot be found or created
      */
-    public Resource getResource(String name, ResourceType type) throws Exception {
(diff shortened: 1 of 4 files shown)


xintongsong (approved):
LGTM

Issue #754, opened 2026-06-04
[api][plan][runtime] Support triggering action by Event field values


rosemarYuan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Summary

This umbrella issue tracks the CEL (Common Expression Language) action condition filtering work across the design discussion and follow-up pull requests. The change introduces an `EventType` constants/registry, unifies the `@action` / `@Action` trigger entry, and adds an `ActionRouter` that combines typed routing with CEL evaluation in both Java and Python.

## Design Discussion

- Discussion: [Discussion] Trigger action by Event Field Values #726

## Pull Requests

The work ships as a four-PR stack and must merge in order (PR1 → PR2 → PR3 → PR4).

- EventType constants and unified trigger entry: #756 
- Java / Python CEL implementation: #821 
- Documentation:  #977
## Current Status

- All four PRs are pending (will be opened sequentially as each parent merges).

## Scope

- Public API rename: `@Action(listenEventTypes = …)` → `@Action(…)`; `@action(*listen_events)` → `@action(*trigger_conditions)`
- New `EventType` constants and runtime registry for built-in and user-defined event types (Java + Python)
- CEL plan-time classifier (`PlanCelClassifier`): bare-ident → `TypeMatch`, otherwise → `CelExpression`
- CEL whitelist: only the `has` macro is allowed; `exists` / `exists_one` / `all` / `filter` / `map` are rejected
- AST rewriter that folds `EventType.X`, normalises type-equality aliases, qualifies bare identifiers under `attributes.`, and rebuilds `has` calls
- `CelExpressionFacade` (Java) / `cel_facade` (Python) compile pipeline with a process-wide LRU program cache
- `ActionRouter` two-phase routing: typed-index fast path + CEL slow path, deduped across the union
- Wiring into `ActionExecutionOperator` (Java) and `LocalRunner` (Python)
- JSON serialization: `listen_event_types` → `trigger_conditions`, with backward-compat fallback in both deserializers
- Cross-language conformance fixture (`cel_conformance_cases.json`, `disallowed-macros.txt`) plus a CI workflow that diffs them and runs both evaluators
- User documentation in `docs/content/docs/development/workflow_agent.md`

## Notes

- This issue is intended as a tracking umbrella for the work split across discussion and the four PRs above.
[...]

PR #756, opened 2026-06-04
[api][java][python] Introduce EventType constants and unify Action trigger entry

rosemarYuan:
Linked issue: #754

### Purpose of change

This PR introduces `EventType` constants and a registry for built-in and user-defined event types, and unifies the Action trigger entry point across Java and Python:

- Java: `@Action(value = ...)`
- Python: `@action(*trigger_conditions)`

This is a preparatory API and plan change for the CEL Action Condition filtering feature tracked in #754. Follow-up PRs will add Java runtime CEL evaluation, Python runtime CEL evaluation, and documentation.

### Scope

Since #709, `@Action` / `@action` carry two orthogonal concerns:

| concern | Java element | Python API | decides |
| --- | --- | --- | --- |
| trigger | `value()` | `*trigger_conditions` | when the action fires |
| dispatch | `target()` | `target=` | where it runs |

This PR only changes the trigger side:

- Java: `listenEventTypes` is renamed to `value`.
- Python: `*listen_events` is renamed to `*trigger_conditions`.

The dispatch side is unchanged. `target`, `PythonFunction`, and cross-language dispatch semantics are reused as-is.

### Tests

- Java: `EventTypeTest`, `AgentPlan*Test`, `Action*SerializerTest` — all green
- Python: `api/tests` + `plan/tests` — 218 passed, 11 skipped
- Verified that cross-language `target` dispatch still works after the trigger rename

### API

This is a source-level API rename on the trigger side.

Java:

- `@Action(listenEventTypes = {X.EVENT_TYPE})` → `@Action(EventType.X)`
- `@Action(listenEventTypes =...,  target = @PythonFunction(...))` → `@Action(value = ..., target = @PythonFunction(...))` 

Python:

- `@action(*listen_events, target=None)` → `@action(*trigger_conditions, target=None)`
- `target=` remains unchanged.

All in-tree Java and Python callers have been migrated. 

- [x] `doc-included`

Updated 10 markdown files under `docs/content/docs/` to use the new `EventType.X` / `@Action(EventType.X)` form while preserving the existing cross-language `target` examples.

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/EventType.java
+++ b/api/src/main/java/org/apache/flink/agents/api/EventType.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api;
+
+/**
+ * Compile-time constants for built-in event types, sourced from each {@code XxxEvent.EVENT_TYPE}.
+ *
+ * <p>Usage: {@code @Action(EventType.InputEvent)}.
+ */
+public final class EventType {
+
+    public static final String InputEvent = org.apache.flink.agents.api.InputEvent.EVENT_TYPE;
+    public static final String OutputEvent = org.apache.flink.agents.api.OutputEvent.EVENT_TYPE;
+    public static final String ChatRequestEvent =
+            org.apache.flink.agents.api.event.ChatRequestEvent.EVENT_TYPE;
+    public static final String ChatResponseEvent =
+            org.apache.flink.agents.api.event.ChatResponseEvent.EVENT_TYPE;
+    public static final String ToolRequestEvent =
+            org.apache.flink.agents.api.event.ToolRequestEvent.EVENT_TYPE;
+    public static final String ToolResponseEvent =
+            org.apache.flink.agents.api.event.ToolResponseEvent.EVENT_TYPE;
+    public static final String ContextRetrievalRequestEvent =
+            org.apache.flink.agents.api.event.ContextRetrievalRequestEvent.EVENT_TYPE;
+    public static final String ContextRetrievalResponseEvent =
+            org.apache.flink.agents.api.event.ContextRetrievalResponseEvent.EVENT_TYPE;
+
+    private EventType() {}
+}
--- a/api/src/main/java/org/apache/flink/agents/api/annotation/Action.java
+++ b/api/src/main/java/org/apache/flink/agents/api/annotation/Action.java
@@ -24,23 +24,24 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Annotation for marking a method as an agent action.
+ * Marks a method as an agent action triggered by matching events.
  *
- * <p>This annotation specifies which event types the action should respond to. The annotated method
- * will be triggered when any of the specified event types occur.
- *
- * <p>Events are specified as type strings via {@link #listenEventTypes()}. Use the {@code
- * EVENT_TYPE} constants on built-in event classes for standard events, or plain strings for custom
- * events.
- *
- * <p>Example usage:
+ * <p>Each {@link #value()} entry is an event type name string. Use the {@code EVENT_TYPE} constants
+ * on built-in event classes, the {@link org.apache.flink.agents.api.EventType} constants, or plain
+ * strings for custom events. Multiple entries combine with OR.
  *
  * <pre>{@code
- * @Action(listenEventTypes = {InputEvent.EVENT_TYPE})
- * public void handleInput(Event event, RunnerContext ctx) { ... }
+ * // Built-in event type via the EventType constant
+ * @Action(EventType.InputEvent)
  *
- * @Action(listenEventTypes = {InputEvent.EVENT_TYPE, "MyCustomEvent"})
- * public void handleMultiple(Event event, RunnerContext ctx) { ... }
+ * // Equivalent via the legacy class constant
+ * @Action(InputEvent.EVENT_TYPE)
+ *
+ * // User-defined event type
+ * @Action("MyCustomEvent")
+ *
+ * // Multiple types (OR semantics)
+ * @Action({EventType.InputEvent, "MyCustomEvent"})
  * }</pre>
  *
  * <p>For a cross-language action, set {@link #target()} to a {@link PythonFunction} with a
@@ -49,7 +50,7 @@ import java.lang.annotation.Target;
  *
  * <pre>{@code
  * @Action(
- *     listenEventTypes = {InputEvent.EVENT_TYPE},
+ *     value = EventType.InputEvent,
  *     target = @PythonFunction(module = "my_pkg.handlers", qualname = "handle_input"))
  * public void handleInput(Event event, RunnerContext ctx) {
  *     throw new UnsupportedOperationException("cross-language stub");
@@ -60,11 +61,11 @@ import java.lang.annotation.Target;
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Action {
     /**
-     * List of event type strings that this action should respond to.
-     *
-     * @return Array of event type strings
+     * Event type name strings; multiple entries have OR semantics. Named {@code value} (not {@code
+     * triggerConditions}) to enable the {@code @Action({...})} shorthand (JLS §9.7.3); corresponds
+     * to Python's {@code *trigger_conditions}.
      */
-    String[] listenEventTypes();
+    String[] value();
 
     /**
      * Cross-language target. When {@link PythonFunction#module()} is non-empty, dispatch routes to
--- a/api/src/test/java/org/apache/flink/agents/api/EventTypeTest.java
+++ b/api/src/test/java/org/apache/flink/agents/api/EventTypeTest.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api;
+
+import org.apache.flink.agents.api.event.ChatRequestEvent;
+import org.apache.flink.agents.api.event.ChatResponseEvent;
+import org.apache.flink.agents.api.event.ContextRetrievalRequestEvent;
+import org.apache.flink.agents.api.event.ContextRetrievalResponseEvent;
+import org.apache.flink.agents.api.event.ToolRequestEvent;
+import org.apache.flink.agents.api.event.ToolResponseEvent;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/** Tests for {@link EventType}. */
+class EventTypeTest {
+
+    @Test
+    void builtInConstantsMatchEventClassConstants() {
+        assertEquals(InputEvent.EVENT_TYPE, EventType.InputEvent);
+        assertEquals(OutputEvent.EVENT_TYPE, EventType.OutputEvent);
+        assertEquals(ChatRequestEvent.EVENT_TYPE, EventType.ChatRequestEvent);
+        assertEquals(ChatResponseEvent.EVENT_TYPE, EventType.ChatResponseEvent);
+        assertEquals(ToolRequestEvent.EVENT_TYPE, EventType.ToolRequestEvent);
+        assertEquals(ToolResponseEvent.EVENT_TYPE, EventType.ToolResponseEvent);
+        assertEquals(
+                ContextRetrievalRequestEvent.EVENT_TYPE, EventType.ContextRetrievalRequestEvent);
+        assertEquals(
+                ContextRetrievalResponseEvent.EVENT_TYPE, EventType.ContextRetrievalResponseEvent);
+    }
+}
--- a/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
@@ -184,27 +184,27 @@ public class AgentPlan implements Serializable {
 
     private void extractActions(
             String actionName,
-            String[] listenEventTypeStrings,
+            String[] triggerEntries,
             org.apache.flink.agents.plan.Function function,
             Map<String, Object> config)
             throws Exception {
-        List<String> eventTypeNames = new ArrayList<>(Arrays.asList(listenEventTypeStrings));
+        List<String> triggerConditions = new ArrayList<>(Arrays.asList(triggerEntries));
 
-        if (eventTypeNames.isEmpty()) {
+        if (triggerConditions.isEmpty()) {
             throw new IllegalArgumentException(
                     "Action "
                             + actionName
-                            + " must specify at least one event type via listenEventTypes.");
+                            + " must specify at least one trigger entry via @Action(EventType.x).");
         }
 
         // Create an Action
-        Action action = new Action(actionName, function, eventTypeNames, config);
+        Action action = new Action(actionName, function, triggerConditions, config);
 
         // Add to actions map
         actions.put(action.getName(), action);
 
         // Add to actionsByEvent map
-        for (String eventTypeName : eventTypeNames) {
+        for (String eventTypeName : triggerConditions) {
             actionsByEvent.computeIfAbsent(eventTypeName, k -> new ArrayList<>()).add(action);
         }
     }
@@ -251,7 +251,7 @@ public class AgentPlan implements Serializable {
                     Objects.requireNonNull(
                             method.getAnnotation(
                                     org.apache.flink.agents.api.annotation.Action.class));
-            String[] listenEventTypeStrings = actionAnnotation.listenEventTypes();
+            String[] triggerEntries = actionAnnotation.value();
             org.apache.flink.agents.api.annotation.PythonFunction target =
                     actionAnnotation.target();
             String targetModule = target.module();
@@ -276,7 +276,7 @@ public class AgentPlan implements Serializable {
                                 + method.getName()
                                 + "' must set both module and qualname");
             }
-            extractActions(method.getName(), listenEventTypeStrings, execFunction, null);
+            extractActions(method.getName(), triggerEntries, execFunction, null);
         }
 
         for (Map.Entry<
--- a/plan/src/main/java/org/apache/flink/agents/plan/actions/Action.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/actions/Action.java
@@ -33,17 +33,17 @@ import java.util.Map;
 import java.util.Objects;
 
 /**
- * Representation of an agent action with event listening and function execution.
+ * Representation of an agent action with unified trigger conditions.
  *
- * <p>This class encapsulates a named agent action that listens for specific event types and
- * executes an associated function when those events occur.
+ * <p>Each entry of {@code triggerConditions} is an event type name string. Multiple entries combine
+ * with OR.
  */
 @JsonSerialize(using = ActionJsonSerializer.class)
 @JsonDeserialize(using = ActionJsonDeserializer.class)
 public class Action {
     private final String name;
     private final Function exec;
-    private final List<String> listenEventTypes;
+    private final List<String> triggerConditions;
 
     // TODO: support nested map/list with non primitive type value.
     @Nullable private final Map<String, Object> config;
@@ -51,18 +51,18 @@ public class Action {
     public Action(
             String name,
             Function exec,
-            List<String> listenEventTypes,
+            List<String> triggerConditions,
             @Nullable Map<String, Object> config)
             throws Exception {
         this.name = name;
         this.exec = exec;
-        this.listenEventTypes = listenEventTypes;
+        this.triggerConditions = triggerConditions;
         this.config = config;
         exec.checkSignature(new Class[] {Event.class, RunnerContext.class});
     }
 
-    public Action(String name, Function exec, List<String> listenEventTypes) throws Exception {
-        this(name, exec, listenEventTypes, null);
+    public Action(String name, Function exec, List<String> triggerConditions) throws Exception {
+        this(name, exec, triggerConditions, null);
     }
 
     public String getName() {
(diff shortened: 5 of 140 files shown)


python/flink_agents/plan/actions/action.py
```diff
@@ -29,29 +29,39 @@
...
+        entries combine with OR semantics.
     """
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     name: str
     # TODO: Raise a warning when the action has a return value, as it will be ignored.
     exec: PythonFunction | JavaFunction
-    listen_event_types: List[str]
+    trigger_conditions: List[str]
```
wzhero1, line 53: **Python Action model lacks backward compat for old JSON key**

Java `ActionJsonDeserializer` (L71-73) falls back to `listen_event_types` when `trigger_conditions` is absent. The Python pydantic `Action` model only declares `trigger_conditions: List[str]` with no alias or model_validator fallback — deserializing old-format plan JSON from persisted Flink state will raise `ValidationError`.

plan/src/main/java/org/apache/flink/agents/plan/actions/Action.java
```diff
@@ -33,36 +33,36 @@
...
+ * <p>Each entry of {@code triggerConditions} is an event type name string. Multiple entries combine
+ * with OR.
  */
 @JsonSerialize(using = ActionJsonSerializer.class)
 @JsonDeserialize(using = ActionJsonDeserializer.class)
 public class Action {
     private final String name;
     private final Function exec;
-    private final List<String> listenEventTypes;
+    private final List<String> triggerConditions;
```
wzhero1, line 46: Discuss：**Keep `listen_event_types`, add `trigger_conditions` as a new field**

Rather than renaming `listen_event_types` → `trigger_conditions`, I'd suggest keeping both:

- `listen_event_types`: which events route to this action (static dispatch)
- `trigger_conditions`: CEL expression filter for whether to process a matched event (dynamic filtering)

These are orthogonal concerns — routing vs. filtering. Keeping `listen_event_types` also avoids JSON backward-compat issues entirely (the Python pydantic model currently lacks the fallback that the Java deserializer has for the old key).

api/src/main/java/org/apache/flink/agents/api/EventType.java
```diff
@@ -0,0 +1,167 @@
...
+ * <p>Usage in {@code @Action}:
+ *
+ * <ul>
+ *   <li>Built-in events: {@code @Action(EventType.InputEvent)}
+ *   <li>User-defined events: {@code @Action("MyCustomEvent")}
+ * </ul>
+ *
+ * <p>Resolution via {@link #lookupOrSelf}: built-in &rarr; user-registered &rarr; passthrough.
+ */
+public final class EventType {
```
wzhero1, line 26: **Reconsider the `EventType` aggregation class**

`EventType.InputEvent` duplicates `InputEvent.EVENT_TYPE` — two sources of truth for the same value. The `register()` / `lookup()` / `lookupOrSelf()` registry has zero callers in production code today.

`InputEvent.EVENT_TYPE` is already self-documenting and lives in its natural namespace. Suggest deferring the registry until the CEL PR actually needs it, and just using the event class constants directly.

plan/src/main/java/org/apache/flink/agents/plan/actions/Action.java
rosemarYuan: Thanks for the comment — this was actually my first instinct too: a clean routing-vs-filtering separation. The more I worked through the problem though, the more I came to think it doesn't quite map onto what users actually want to express. 

Event filtering is just event routing at a finer granularity, not a separate concern. Each entry in `trigger_conditions` answers exactly one question — "should this action fire for this event?" — and a bare event type is simply the most common (and most degenerate) form of that condition.

The case that really pins this down: for a single action, users often want per-event-type conditions, with different conditions for different types — and sometimes multiple conditions on the same type. With a unified list this falls out naturally. [#726](https://github.com/apache/flink-agents/issues/726) shows the canonical shape:

```
// Mixed triggers. Entries are OR'ed.
@Action({
    EventType.InputEvent,                                                           // direct type
    "type == EventType.ChatResponseEvent && retryCount > 0",                        // metadata condition
    "type == EventType.ToolResponseEvent && response.success == false",             // payload condition
    "type == EventType.ChatResponseEvent && response.plantype.contains('Day')"      // metadata + payload
})
```

Split into` listen_event_types + trigger_conditions`, there's no clean way to express any of this. The two fields are parallel lists with no shared index, so "this condition belongs to that type" has no representation — and "ChatResponseEvent has two distinct conditions" is even less expressible. You'd end up either forcing one CEL expression to gate all listened types (loses per-type granularity), or inventing a multi-map like `{ChatResponseEvent: [cond1, cond2], ToolResponseEvent: [cond3]}` — which is just the unified list with extra ceremony, plus an awkward way to represent the "no condition" case for direct-type entries. The unified field is what naturally fits the actual shape of user intent.

Curious if you have a concrete scenario in mind where the two-field split would be cleaner — happy to revisit if there's a use case I'm missing.

python/flink_agents/plan/actions/action.py
rosemarYuan: Fixed in lines 84-88. Added a model_validator fallback that mirrors the Java ActionJsonDeserializer behavior:
```
 # Legacy fallback: listen_event_types → trigger_conditions
        if "trigger_conditions" not in self or self.get("trigger_conditions") is None:
            if self.get("listen_event_types"):
                self["trigger_conditions"] = list(self["listen_event_types"])
            self.pop("listen_event_types", None)
```
Old-format plan JSON with listen_event_types will now deserialize without ValidationError.

api/src/main/java/org/apache/flink/agents/api/EventType.java
rosemarYuan: Agreed. I've refactored EventType to be a pure namespace of constants — each one simply re-exports the value from the corresponding event class (e.g., EventType.InputEvent = _InputEvent.EVENT_TYPE), so there's a single source of truth. The register() / lookup() / lookupOrSelf() registry has been removed from this PR and will be introduced later when the CEL PR actually requires dynamic resolution.

Meanwhile, I've also slimmed down the EventType class by removing the internal utility functions — their concrete implementations are deferred to a follow-up PR.

plan/src/main/java/org/apache/flink/agents/plan/serializer/ActionJsonDeserializer.java
```diff
@@ -66,10 +66,15 @@ public Action deserialize(JsonParser jsonParser, DeserializationContext deserial
...
 
-        // Deserialize listenEventTypes
-        List<String> listenEventTypes = new ArrayList<>();
-        node.get("listen_event_types")
-                .forEach(eventTypeNode -> listenEventTypes.add(eventTypeNode.asText()));
+        // Deserialize trigger_conditions (fall back to legacy listen_event_types for older JSONs)
+        List<String> triggerConditions = new ArrayList<>();
+        JsonNode triggerNode = node.get("trigger_conditions");
+        if (triggerNode == null) {
+            triggerNode = node.get("listen_event_types");
```
weiqingy: This fallback to the legacy `listen_event_types` key is the load-bearing guarantee that plan JSON persisted in older Flink state still deserializes after the rename — but nothing exercises it. The cross-version snapshot resources were all regenerated to `trigger_conditions` in this PR, so the compat tests now compare new-format against new-format and this branch never fires under test. A future refactor could delete the fallback and every test would stay green.

Worth a small deserialize test that feeds a JSON blob carrying `listen_event_types` and asserts it lands in `getTriggerConditions()` / `getListenEventTypes()`? That pins the old-key path so it can't be silently dropped.

python/flink_agents/plan/actions/action.py
```diff
@@ -71,7 +81,13 @@ def __serialize_config(self, config: Dict[str, Any]) -> Dict[str, Any] | None:
 
     @model_validator(mode="before")
     def __custom_deserialize(self) -> "Action":
-        config = self["config"]
+        # Legacy fallback: listen_event_types → trigger_conditions
+        if "trigger_conditions" not in self or self.get("trigger_conditions") is None:
+            if self.get("listen_event_types"):
+                self["trigger_conditions"] = list(self["listen_event_types"])
+            self.pop("listen_event_types", None)
```
weiqingy: Same thought as the Java side: this `model_validator` is the only thing keeping older persisted plans (with `listen_event_types`) deserializable, but no test feeds it the old key — the regenerated snapshot fixtures all use `trigger_conditions` now, so this branch is untested.

Would a small round-trip test help here — construct an `Action` from a dict with `listen_event_types` and assert it surfaces as `trigger_conditions`? That guards the fallback against a future cleanup.

api/src/test/java/org/apache/flink/agents/api/EventTypeTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+/** Tests for {@link EventType}. */
+class EventTypeTest {
+
+    @Test
+    void builtInConstantsAreNonNull() {
```
weiqingy: `assertNotNull` on inlined `String` constants can't really fail — the constants are non-null by construction — so this test passes regardless of what each constant holds. A copy-paste slip like `OutputEvent = InputEvent.EVENT_TYPE` would sail through.

Since the point of `EventType` is to be a single source of truth, would asserting value-equality against each source (e.g. `assertEquals(InputEvent.EVENT_TYPE, EventType.InputEvent)`) be a stronger check? That actually guards the mapping. Same applies to the Python `test_event_type.py` smoke test.

python/flink_agents/api/tests/test_event_type.py
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+#################################################################################
+"""Smoke tests for :mod:`flink_agents.api.events.event_type`."""
+
+from __future__ import annotations
+
+from flink_agents.api.events.event_type import EventType
+
+
+def test_builtin_constants_are_non_empty_strings() -> None:
```
weiqingy: This asserts each constant is a non-empty string, which holds for any string value and so won't catch a wrong mapping (e.g. two constants pointing at the same source `EVENT_TYPE`). Would asserting each constant against its source event's `EVENT_TYPE` be a better guard on the single-source-of-truth contract? Mirrors the suggestion on the Java `EventTypeTest`, so whatever shape you land on there can carry over here.

weiqingy (commented):
Thanks for taking this on — the rename is clean, and I appreciate that you kept a backward-compat fallback for the old JSON key on *both* sides (the `listen_event_types` branch in `ActionJsonDeserializer` and the `model_validator` in `action.py`), so plan JSON persisted in older Flink state still deserializes. That's an easy thing to miss in a rename PR. A few questions inline.

plan/src/main/java/org/apache/flink/agents/plan/serializer/ActionJsonDeserializer.java
rosemarYuan: Good catch — you're right that the regenerated snapshot fixtures mean the legacy branch is never exercised under test. I'll add a focused deserialization test that feeds a JSON blob with the old listen_event_types key and asserts it maps correctly to getTriggerConditions().

python/flink_agents/plan/actions/action.py
rosemarYuan: Same idea on the Python side. While adding the test I found that the custom __init__ was rejecting the old key before model_validator(mode="before") had a chance to rename it — so the fallback was silently broken. Fixed __init__ to accept **kwargs and handle the legacy key, and added a matching deserialization test.

api/src/test/java/org/apache/flink/agents/api/EventTypeTest.java
rosemarYuan: You're right — assertNotNull on inlined constants is vacuous. Originally addressed in a follow-up PR; now pulled forward into this commit with assertEquals(InputEvent.EVENT_TYPE, EventType.InputEvent) and similar for all 8 built-ins.

python/flink_agents/api/tests/test_event_type.py
rosemarYuan: Mirrors the Java fix — the Python test now asserts EventType.InputEvent == InputEvent.EVENT_TYPE directly for all 8 constants. Consistent shape across both languages.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -184,27 +184,27 @@ private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundE
...
         // Create an Action
-        Action action = new Action(actionName, function, eventTypeNames, config);
+        Action action = new Action(actionName, function, triggerConditions, config);
 
         // Add to actions map
         actions.put(action.getName(), action);
 
         // Add to actionsByEvent map
-        for (String eventTypeName : eventTypeNames) {
+        for (String eventTypeName : triggerConditions) {
```
weiqingy, line 207: This site builds the routing map from the raw `triggerConditions`, but the built-in path just below (`addBuiltAction`, L217) builds it through `action.getListenEventTypes()` — and that accessor's javadoc says a follow-up overrides it to "filter out non-type entries" once CEL lands. Python's `agent_plan.py:148` also reads the raw field. So of the three routing sites, only the built-in one — which never carries CEL — goes through the filtering seam; the two that handle user-supplied conditions bypass it.

Today they're equivalent, since every entry is a plain event-type name. But once CEL expressions enter `trigger_conditions`, these two sites would register strings like `"type == EventType.ChatResponseEvent && retryCount > 0"` as literal keys in `actionsByEvent` — exactly what the accessor exists to strip. Would it make sense to route all three through the accessor now, or to defer the accessor until the CEL PR and read the raw field everywhere in the meantime? Either way the three sites agreeing seems worth locking in. Not a blocker for this PR — flagging while the context is fresh.

weiqingy (commented):
Thanks for the quick turnaround — all four threads are addressed and CI is green. The legacy-key fallbacks now have focused tests on both sides, the `EventType` tests assert value-equality against each source, and the `__init__` fix closes a real gap, not just a coverage one. One forward-looking question inline.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
rosemarYuan: Thanks for flagging this. The inconsistency wasn't intentional — when writing `extractActions`, the `triggerConditions` local variable was right at hand so I used it directly; same on the Python. Since all three return identical values in this PR, the divergence slipped through unnoticed. A simple oversight on my part.

The follow-up CEL PR consolidates all three sites into a single `registerAction() `method that routes exclusively via `getListenEventTypes()`, which by then performs real filtering. So the three sites end up in lockstep once CEL lands.

weiqingy: Agreed, fine to leave as-is — the three sites are equivalent today and nothing can carry a CEL string yet. Thanks for the context, resolved on my end.

python/flink_agents/api/events/event_type.py
```diff
@@ -0,0 +1,66 @@
...
+    InputEvent: str = _InputEvent.EVENT_TYPE
+    OutputEvent: str = _OutputEvent.EVENT_TYPE
+    ChatRequestEvent: str = _ChatRequestEvent.EVENT_TYPE
+    ChatResponseEvent: str = _ChatResponseEvent.EVENT_TYPE
+    ToolRequestEvent: str = _ToolRequestEvent.EVENT_TYPE
+    ToolResponseEvent: str = _ToolResponseEvent.EVENT_TYPE
+    ContextRetrievalRequestEvent: str = _ContextRetrievalRequestEvent.EVENT_TYPE
+    ContextRetrievalResponseEvent: str = _ContextRetrievalResponseEvent.EVENT_TYPE
+
+    def __init__(self) -> None:
```
wenjin272: I think the __init__ method is unnecessary. We don’t have this design in `ResourceName` either.

plan/src/main/java/org/apache/flink/agents/plan/serializer/ActionJsonDeserializer.java
wenjin272: Although both @weiqingy  and @wzhero1  agree on handling backward compatibility, I believe it is unnecessary at this stage. Flink-Agents is still in the 0.x series, where API compatibility is not guaranteed. In fact, we have already introduced some breaking API changes in version 0.3 (e.g., #631), so maintaining backward compatibility specifically for listen_event_types offers limited value.

We will begin preparing for API stability and compatibility commitments starting with the next release (0.4), and formally commit to backward compatibility from version 1.0 onwards. Until then, I consider keeping the codebase clean and simple to be a higher priority.

api/src/main/java/org/apache/flink/agents/api/annotation/Action.java
```diff
@@ -59,12 +60,8 @@
...
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Action {
-    /**
-     * List of event type strings that this action should respond to.
-     *
-     * @return Array of event type strings
-     */
-    String[] listenEventTypes();
+    /** Event type name strings; multiple entries have OR semantics. */
+    String[] value();
```
wenjin272, line 68: Here, we use `value` as the parameter name, whereas Python’s `Action` decorator uses `trigger_conditions`. 

I understand this discrepancy arises because in Java annotations, only a member named `value` can be used without explicitly specifying the parameter name during declaration. 

Therefore, to improve usability for users, I believe this inconsistency is acceptable. However, we should probably explain this in the comments.

wenjin272 (commented):
Ty for taking this on @rosemarYuan. I think overall the work is clear. but I have two main questions:

* It appears that the YAML API and the `addAction/add_action` paths are not covered.
* I don't see a need to keep the backward-compatibility fallback. WDYT @xintongsong? 

Additionally, there are some conflicts with the main branch that can be resolved after cutting the release-0.3 branch.

python/flink_agents/api/events/event_type.py
rosemarYuan: Thanks for your suggestions regarding the review. Dropped the `__init__` guard. Agreed it adds little value here, so removing it aligns with that idiom.

api/src/main/java/org/apache/flink/agents/api/annotation/Action.java
rosemarYuan: Good point — added a one-sentence note on explaining:
- (1) The JLS §9.7.3 reason for the name.
- (2) Its mapping to Python's `*trigger_conditions`. 
Kept it tight to avoid bloating the annotation javadoc.

rosemarYuan:
Hi @xintongsong, need a policy call from you.

This PR renames `listen_event_types` -> `trigger_conditions`. I added a fallback on both Java (`ActionJsonDeserializer`) and Python (`Action.model_validator`), so old plan JSON still deserializes.

@wenjin272  pointed out we're still in 0.x and don't owe API/state compatibility yet (#631 already broke API in 0.3), so the fallback may be debt we shouldn't take on until the 0.4 / 1.0 commitment.

That policy angle wasn't on the table in the earlier review — we only discussed whether the fallback was correctly implemented. Could you decide:

  (A) keep the fallback + tests, or
  (B) drop them under the 0.x "no compat" stance?

Thanks!

rosemarYuan:
## YAML `trigger_conditions`: Should CEL Also Accept YAML Event Aliases?

I have a local implementation for the YAML API path, but there is one syntax question I would like to align on before pushing it.

When condition-expression support was originally designed, the YAML API had not yet been introduced. As a result, the interaction between YAML event aliases and CEL expressions was not considered.

### Current Behavior

The YAML loader already supports short aliases for built-in event types. For example:

```yaml
trigger_conditions:
  - input
```

The YAML layer resolves `input` to the corresponding built-in event type string before constructing `plan.Action`.

CEL conditions, however, use the `EventType.X` namespace, consistent with the existing Java and Python APIs:

```yaml
trigger_conditions:
  - >-
    type == EventType.ChatResponseEvent
    && priority == 'high'
```

A mixed condition list can therefore contain two syntactic forms for built-in event types:

```yaml
actions:
  - name: handle_priority_input
    type: java
    function: com.example.MyActions:handlePriorityInput
    trigger_conditions:
      - input
      - >-
        type == EventType.ChatResponseEvent
        && priority == 'high'
```

For example, `input` and `EventType.InputEvent` refer to the same conceptual event type, but they look unrelated in the YAML configuration.

### Concern: Overloading the Alias Name

I do not think short aliases should be introduced into CEL.

Consider a natural-looking condition:

```cel
type == chat_request && chat_request.id > 10
```

Here, `chat_request` appears to play two different roles:

* on the left, it is an event-type constant;
* on the right, it appears to be an event or payload object.

These roles are conceptually distinct. Mainstream languages normally make the distinction explicit—for example, a type check and an object reference are represented using different syntax.

Using the `EventType.X` namespace keeps the roles unambiguous:

```cel
type == EventType.ChatRequestEvent
&& chat_request > 10
```

The `EventType.` prefix makes the identifier's role visible directly from the syntax. Readers do not need to infer from context whether `chat_request` represents a type constant, an event object, or an attribute.

Binding aliases as top-level CEL variables would also introduce a silent name-collision risk. User events may legitimately expose fields or variables named `input`, `output`, or `chat_request`, especially in LLM and RAG workloads. Reserving those common names as CEL variables could make expressions harder to understand and evolve.

### Proposed Boundary

My preference is to keep the current boundary explicit:

* YAML short aliases are supported only for standalone event-type conditions;
* CEL expressions continue to use `EventType.X` for built-in event types.

For example:

```yaml
trigger_conditions:
  - input
  - >-
    type == EventType.ChatResponseEvent
    && priority == 'high'
```

This does mean that YAML may contain two notations for event types, but each notation has a clear scope:

* `input` is YAML shorthand for a standalone type match;
* `EventType.InputEvent` is a typed constant inside a CEL expression.


### Question

Does the proposed boundary look acceptable for the YAML API?

In particular:

* should aliases remain YAML-only shorthand for standalone type matches;
* or should CEL also expose them as top-level variables?

Feedback from anyone who has written non-trivial YAML agents would be especially useful, particularly regarding which form feels more natural in practice.

/cc @wenjin272

wenjin272 (commented):
LGTM. The point about backward compatibility may need to be confirmed. PLTA @xintongsong

xintongsong:
I think there's no need to stay compatible for agent plans from older version. First of all, we are still in 0.x versions and should avoid carry any burdens due to trying to stay compatible with immature legacy apis. And secondly, the json agent plan should not be considered as an public API. It's internal. We should not support executing a json agent plan generated with an older version.

plan/src/main/java/org/apache/flink/agents/plan/serializer/ActionJsonDeserializer.java
```diff
@@ -66,10 +66,12 @@ public Action deserialize(JsonParser jsonParser, DeserializationContext deserial
...
         }
 
-        // Deserialize listenEventTypes
-        List<String> listenEventTypes = new ArrayList<>();
-        node.get("listen_event_types")
-                .forEach(eventTypeNode -> listenEventTypes.add(eventTypeNode.asText()));
+        // Deserialize trigger_conditions
+        List<String> triggerConditions = new ArrayList<>();
+        JsonNode triggerNode = node.get("trigger_conditions");
+        if (triggerNode != null) {
```
wenjin272: The `trigger_conditions` should not be null.

plan/src/test/resources/actions/action_java_function.json
```diff
@@ -6,5 +6,5 @@
     "method_name": "legal",
     "parameter_types": ["org.apache.flink.agents.api.Event", "org.apache.flink.agents.api.context.RunnerContext"]
   },
-  "listen_event_types": ["_input_event"]
+  "trigger_conditions": ["_input_event"]
```
wenjin272, line 9: The field name in `action_invalid_event_type.json`, `action_missing_fields.json` and `action_invalid_function_type.json` were not adapted.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -184,27 +184,27 @@ private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundE
...
-        List<String> eventTypeNames = new ArrayList<>(Arrays.asList(listenEventTypeStrings));
+        List<String> triggerConditions = new ArrayList<>(Arrays.asList(triggerEntries));
 
-        if (eventTypeNames.isEmpty()) {
+        if (triggerConditions.isEmpty()) {
             throw new IllegalArgumentException(
                     "Action "
                             + actionName
-                            + " must specify at least one event type via listenEventTypes.");
+                            + " must specify at least one trigger entry via @Action(value = ...).");
```
wenjin272: Consider using `@Action(EventType.x)` in the error message.

wenjin272 (commented):
Several minor comments.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
rosemarYuan: **Thanks.** Addressed all three in the latest revision: dropped the null guard (trigger_conditions is always present), renamed the field in the remaining three fixtures, and switched the error message to @Action(EventType.x). PTAL.

wenjin272 (approved):
LGTM

Issue #762, opened 2026-06-04
[Release] Manual verification mcp section of user documentation


wzhero1:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Verification Checklist — MCP

Doc page: https://nightlies.apache.org/flink/flink-agents-docs-main/docs/development/mcp/
Source: `docs/content/docs/development/mcp.md`

#### A. Documentation Review

- [x] **Overview** — Verify MCP protocol description is accurate
- [x] **Declare MCP Server** — Python `@mcp_server` / Java `@MCPServer` usage is clear and complete
- [x] **Authentication** — Bearer / Basic / ApiKey examples are correct
- [x] **Python authentication** — headers dict approach vs Java auth object approach difference is clearly explained
- [x] **Use MCP prompts and tools** — "auto-discovery" description is accurate, prompt/tool reference approach is clear
- [x] **Appendix: MCP SDK** — Selection table (Python/Java SDK based on Agent language and JDK version) is accurate
- [x] **lang="python" override** — Explanation is clear, when to use it
- [x] **Link validity** — Links to prompts and tool_use pages are correct
- [ ] **Code snippets** — Java snippets use `addInitialArgument("timeout", 30)` but implementation expects `"timeoutSeconds"` — see [comment](https://github.com/apache/flink-agents/issues/762#issuecomment-4627779459)

#### B. Code–Documentation Consistency

- [x] Python `@mcp_server` decorator exists (`flink_agents.api.decorators`)
- [x] Java `@MCPServer` annotation exists and fields match documentation
- [x] `ResourceName.MCP_SERVER` constant exists
- [x] Java `BearerTokenAuth` class exists (`integrations/mcp/.../auth/`)
- [x] Java `BasicAuth` class exists
- [x] Java `ApiKeyAuth` class exists
- [x] `@MCPServer` `lang` parameter exists with default value "auto"
- [x] Python `endpoint` / `headers` parameter passing matches ResourceDescriptor usage
- [ ] Java Builder `addInitialArgument("auth", ...)` approach matches actual implementation — `"auth"` key is correct, but `"timeout"` key should be `"timeoutSeconds"` — see [comment](https://github.com/apache/flink-agents/issues/762#issuecomment-4627779459)

#### C. Run Examples

- [x] Start test MCP server (`mcp_server.py`)
- [x] Python MCP agent e2e test connects to MCP server successfully
[...]

PR #771, opened 2026-06-05
[integrations][mcp] Read MCP server timeout from the "timeout" descriptor argument

### LINKED ISSUE #770: [Bug] Java MCPServer ignores the "timeout" ResourceDescriptor argument, always uses 30s default
### Search before asking

### Description

Found during MCP doc verification (#762).

The Java `MCPServer` reads its timeout from the `ResourceDescriptor` with key `"timeoutSeconds"`,
but the docs, the test, the Python implementation, and all other connections use `"timeout"`.
So `addInitialArgument("timeout", N)` is silently ignored and the 30s default always applies.

- Bug: `integrations/mcp/.../MCPServer.java` — `FIELD_TIMEOUT_SECONDS = "timeoutSeconds"`, read in the
  `MCPServer(ResourceDescriptor, ResourceContext)` constructor.
- Docs (already correct): `docs/content/docs/development/mcp.md` lines 56, 102, 163 use `"timeout"`.

Fix: make the Java implementation read/serialize `"timeout"` (Python uses `timeout`; every other
connection reads `getArgument("timeout")`). No doc change needed.

### How to reproduce

```java
ResourceDescriptor.Builder.newBuilder(ResourceName.MCP_SERVER)
        .addInitialArgument("endpoint", "http://127.0.0.1:8000/mcp")
        .addInitialArgument("timeout", 60)   // expected 60s
        .build();
```
The resolved `MCPServer.getTimeoutSeconds()` returns `30`, not `60`.

### Version and environment

Flink Agents 0.3.0 (`main`), all Flink versions. Java MCP only; Python unaffected.

wzhero1:
Linked issue: Closes #770, refs #762

### Purpose of change

Java `MCPServer` read its connection timeout from the `ResourceDescriptor` using the key
`"timeoutSeconds"`, while the docs, the test, the Python implementation and all other
connections use `"timeout"`. As a result `addInitialArgument("timeout", N)` was silently
ignored and the hard-coded 30s default always applied.

This renames the descriptor/JSON key to `"timeout"` — a single constant in `MCPServer.java`,
which also aligns the Jackson wire key with the Python side. The documentation already uses
`"timeout"` and needs no change.

### Tests

Added regression test `MCPServerTest#testTimeoutFromResourceDescriptor`: builds an `MCPServer`
from a `ResourceDescriptor` with `timeout=60` and asserts `getTimeoutSeconds() == 60` (returns
`30` before the fix). Verified locally:
- `mvn -pl integrations/mcp test`: 38/38 pass
- `spotless:check` on the module: clean

### API

No public API changes. Only the internal descriptor-argument / JSON key string changes; the
`MCPServer` Java field, the `getTimeoutSeconds()` getter and the default value are unchanged.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPServerTest.java
+++ b/integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPServerTest.java
@@ -19,6 +19,9 @@
 package org.apache.flink.agents.integrations.mcp;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.flink.agents.api.resource.ResourceContext;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
+import org.apache.flink.agents.api.resource.ResourceName;
 import org.apache.flink.agents.api.resource.ResourceType;
 import org.apache.flink.agents.integrations.mcp.auth.ApiKeyAuth;
 import org.apache.flink.agents.integrations.mcp.auth.BasicAuth;
@@ -69,6 +72,28 @@ class MCPServerTest {
         assertThat(server.getHeaders()).isEmpty();
         assertThat(server.getTimeoutSeconds()).isEqualTo(30);
         assertThat(server.getAuth()).isNull();
+    }
+
+    @Test
+    @DisabledOnJre(JRE.JAVA_11)
+    @DisplayName("Read timeout from ResourceDescriptor")
+    void testTimeoutFromResourceDescriptor() {
+        ResourceDescriptor descriptor =
+                ResourceDescriptor.Builder.newBuilder(ResourceName.MCP_SERVER)
+                        .addInitialArgument("endpoint", DEFAULT_ENDPOINT)
+                        .addInitialArgument("timeout", 60)
+                        .build();
+
+        MCPServer server =
+                new MCPServer(
+                        descriptor,
+                        ResourceContext.fromGetResource(
+                                (name, type) -> {
+                                    throw new UnsupportedOperationException(
+                                            "No dependencies expected");
+                                }));
+
+        assertThat(server.getTimeoutSeconds()).isEqualTo(60);
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


integrations/mcp/src/test/java/org/apache/flink/agents/integrations/mcp/MCPServerTest.java
```diff
@@ -71,6 +74,28 @@ void testSimpleConstructor() {
...
+        MCPServer server =
+                new MCPServer(
+                        descriptor,
+                        ResourceContext.fromGetResource(
+                                (name, type) -> {
+                                    throw new UnsupportedOperationException(
+                                            "No dependencies expected");
+                                }));
+
+        assertThat(server.getTimeoutSeconds()).isEqualTo(60);
```
weiqingy, line 96: The bug here was a cross-language wire-key mismatch — Java emitted and read `"timeoutSeconds"` while Python and the descriptor use `"timeout"`. This test covers the descriptor→field path, and the existing `testJsonSerialization` round-trips through Jackson. But since that round-trip's write and read sides both go through `FIELD_TIMEOUT`, it stays green even if the constant were renamed back to `"timeoutSeconds"` — the one thing none of the tests pin is the literal wire key that Python depends on.

Would it be worth asserting on the serialized string so a future rename can't silently reintroduce the same Java/Python mismatch? Something like, in `testJsonSerialization`:

```java
assertThat(json).contains("\"timeout\"");
```

xintongsong: I think this is a bigger question: How do we systematically guarantee apis / option-keys consistency across docs, java / python / yaml apis, and the json plan? Not only for `timeout`, but potentially for all args passed via resource descriptors.

I don't have any concrete plan. Just one preliminary idea: we may introduce some option key constants (like `ResourceName`), so that we can verify the constants are consistent between java and python with test cases, and also reduce the chance of typo and silent ignore on user side.

And that should be out of the scope of this PR.

xintongsong (approved):

Issue #776, opened 2026-06-05
[Bug] workflow_single_agent_example.py fails — FileSource reads SKILL.md as ProductReview JSON


wzhero1:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

`python -m flink_agents.examples.quickstart.workflow_single_agent_example` fails because `FileSource` scans `resources/` recursively and picks up `skills/math-calculator/SKILL.md`, which is not valid `ProductReview` JSON.

**Root cause:** The FileSource is configured with `f"file:///{current_dir}/resources"` (line 66 of `workflow_single_agent_example.py`), but the `resources/` directory contains both `product_review.txt` and the `skills/` subdirectory added for the skills example. FileSource reads all files recursively, so `SKILL.md` gets fed into the `ProductReview.model_validate_json()` deserializer and fails.

**Suggested fix:** Either:
1. Point FileSource to the specific file `resources/product_review.txt` instead of the directory, or
2. Move skills resources to a separate directory outside `resources/`.

**Found during:** Manual verification of tool_use documentation (#761, checklist item C3).

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #707, opened 2026-05-26
[docs][examples] Add Agent Skills documentation and quickstart example

wenjin272:
Linked issue: #696 

### Purpose of change

Add Agent Skills documentation and quickstart example

### Tests

Manually test quickstart example locally.

### API

No

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

PR #782, opened 2026-06-05
[bugfix][examples] Point quickstart FileSource at product_review.txt instead of the resources dir

wzhero1:
Linked issue: #776

### Purpose of change

The three product-review quickstart examples build a `FileSource` over the whole
`resources/` directory. Flink's default file enumerator recurses into
subdirectories, so after #707 added `skills/math-calculator/SKILL.md` under
`resources/`, the source reads that markdown file and feeds each line to
`ProductReview.model_validate_json()`, which raises a `ValidationError` and fails
the job.

This affects all three Python quickstart examples that read the directory, not
just the one in the issue:
- `workflow_single_agent_example.py`
- `yaml_workflow_agent_example.py`
- `react_agent_example.py`

The fix points each `FileSource` at the specific `resources/product_review.txt`
file — the same single-file targeting the Java examples already use via
`copyResource`. The `skills/` resources and packaging are left untouched, so the
skills example keeps working.

For reference, the following are **not** affected and need no change:
- The 3 Java examples (`WorkflowSingleAgentExample` / `ReActAgentExample` /
  `YamlWorkflowAgentExample`) — their `resources/` is polluted the same way, but
  they copy a single `input_data.txt` to a temp dir and target that one file.
- The e2e tests — each `FileSource` targets a dedicated single-record input
  subdir (`execute_test_input/`, `input/`, `java_chat_module_input/`, …).

### Tests

- `ruff check` and `ruff format --check` pass on the changed examples.
- `python -m py_compile` passes on the three examples.
- Root cause verified against the bundled Flink 2.2.1 `flink-connector-files`
  jar: the default `NonSplittingRecursiveEnumerator` + `DefaultFileFilter`
  recurse into `skills/` and emit `SKILL.md` (only names starting with `.`/`_`
  are skipped). Narrowing the source to the single file avoids it.

### API

No public API changes.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

### DIFF
--- a/python/flink_agents/examples/quickstart/react_agent_example.py
+++ b/python/flink_agents/examples/quickstart/react_agent_example.py
@@ -76,7 +76,7 @@ def main() -> None:
     product_review_stream = env.from_source(
         source=FileSource.for_record_stream_format(
             StreamFormat.text_line_format(),
-            f"file:///{current_dir}/resources/",
+            f"file:///{current_dir}/resources/product_review.txt",
         )
         .monitor_continuously(Duration.of_minutes(1))
         .build(),
--- a/python/flink_agents/examples/quickstart/react_agent_example.py
+++ b/python/flink_agents/examples/quickstart/react_agent_example.py
@@ -74,6 +74,8 @@ def main() -> None:
     # Read product reviews from a text file as a streaming source.
     # Each line in the file should be a JSON string representing a ProductReview.
     product_review_stream = env.from_source(
+        # Target the single file, not the resources/ dir: Flink's enumerator
+        # recurses, so files like skills/SKILL.md would be parsed as reviews.
         source=FileSource.for_record_stream_format(
             StreamFormat.text_line_format(),
             f"file:///{current_dir}/resources/product_review.txt",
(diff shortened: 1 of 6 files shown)


python/flink_agents/examples/quickstart/react_agent_example.py
```diff
@@ -76,7 +76,7 @@ def main() -> None:
     product_review_stream = env.from_source(
         source=FileSource.for_record_stream_format(
             StreamFormat.text_line_format(),
-            f"file:///{current_dir}/resources/",
+            f"file:///{current_dir}/resources/product_review.txt",
```
weiqingy, line 81: One optional note inline, LGTM

Optional: a one-line note on *why* this targets the file rather than `resources/` would stop someone from later broadening it back to the dir and rehitting #776 — e.g. `# Target the specific file so non-record resources (e.g. skills/) aren't read as reviews`. Same applies to the other two examples.

xintongsong (approved):
LGTM. Let's merge the PR after resolving @weiqingy 's comments.

wzhero1:
Good call — added the rationale inline at all three examples (and the matching doc snippets) so the single-file targeting isn't accidentally broadened back to `resources/`. Thanks!

wenjin272:
LGTM

PR #760, opened 2026-06-04
[metric] Add key-value metric group for model and action dimensions

yunfengzhou-hub:
<!-- Please link the PR to the relevant issue(s). Hotfix doesn't need this. -->
  Linked issue: #744
                                                                                                                                                                                                                                                      
  ### Purpose of change                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                      
Use key-value metric groups for `action` and `model` dimensions instead of plain named sub-groups. This makes these dimensions available as labels in reporters that support them (e.g., Prometheus), and produces more structured metric identifiers:         
                                                                                                                                                                                                                                                      
  - Action metrics: `<action_name>.numOfActionsExecuted` → `action.<action_name>.numOfActionsExecuted`                                                                                                                                                
  - Token metrics: `<action_name>.<model_name>.promptTokens` → `action.<action_name>.model.<model_name>.promptTokens`
                                                                                                                                                                                                                                                      
  ### Tests            
                                                                                                                                                                                                                                                      
  - Unit tests (value-based and mock-based) verify the key-value group hierarchy and counter accumulation.                                                                                                                                            
  - An end-to-end integration test using `InMemoryReporter` + `MiniClusterExtension` validates the complete metric identifier through the real Flink metric system.
                                                                                                                                                                                                                                                      
  ### API              
                                                                                                                                                                                                                                                      
  - **Breaking**: The metric identifiers for `promptTokens`, `completionTokens`, and per-action execution counters now include the key prefix (`action.`, `model.`). Metric reporter queries or dashboards referencing the old names need updating.   
  - A new `getSubGroup(String key, String value)` method is added to the `FlinkAgentsMetricGroup` interface (Java) and `get_sub_group(name, value)` (Python).
  - `getSubGroup(String name)` now rejects names containing `=` to prevent cache-key collisions with the key-value variant.                                                                                                                           
                                                                                                                                                                                                                                                      
  ### Documentation                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                      
  <!-- Do not remove this section. Check the proper box only. -->                                                                                                                                                                                     
   
  - [ ] `doc-needed` <!-- Your PR changes impact docs -->                                                                                                                                                                                             
  - [ ] `doc-not-needed` <!-- Your PR changes do not impact docs -->
  - [x] `doc-included` <!-- Your PR already contains the necessary documentation updates -->

Issue #744, opened 2026-06-03
[Bug] Name of metrics dynamically change


yunfengzhou-hub:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, when I start a Flink Agent job and check the metrics on Prometheus, I can see that the name of the metric is as follows:

flink_taskmanager_job_task_operator_chat_model_action_serviceType_qwen3_6_plus_totalTokens

In this metric group, "chat_model_action" and "qwen3_6_plus" are the name of the action and the model. It means this metric name would alter according to different models and actions, which makes it hard to track Flink Agent metrics

### How to reproduce

Configure a Prometheus metric server for a Flink Agent job, and check the reported metrics.

### Version and environment

Flink 1.20, currently latest Flink Agent code (1ca367d7).

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #783, opened 2026-06-05
[metric] Add key-value metric group for model and action dimensions (#760)

yunfengzhou-hub:
Linked issue: #744

Backport of #760 to `release-0.2`.                                                                                                                                                               
                                                                                                                                                                                                                                                      
  ### Purpose of change                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                      
Use key-value metric groups for `action` and `model` dimensions instead of plain named sub-groups. This makes these dimensions available as labels in reporters that support them (e.g., Prometheus), and produces more structured metric identifiers:         
                                                                                                                                                                                                                                                      
  - Action metrics: `<action_name>.numOfActionsExecuted` → `action.<action_name>.numOfActionsExecuted`                                                                                                                                                
  - Token metrics: `<action_name>.<model_name>.promptTokens` → `action.<action_name>.model.<model_name>.promptTokens`
                                                                                                                                                                                                                                                      
  ### Tests            
                                                                                                                                                                                                                                                      
  - Unit tests (value-based and mock-based) verify the key-value group hierarchy and counter accumulation.                                                                                                                                            
  - An end-to-end integration test using `InMemoryReporter` + `MiniClusterExtension` validates the complete metric identifier through the real Flink metric system.
                                                                                                                                                                                                                                                      
  ### API              
                                                                                                                                                                                                                                                      
  - **Breaking**: The metric identifiers for `promptTokens`, `completionTokens`, and per-action execution counters now include the key prefix (`action.`, `model.`). Metric reporter queries or dashboards referencing the old names need updating.   
  - A new `getSubGroup(String key, String value)` method is added to the `FlinkAgentsMetricGroup` interface (Java) and `get_sub_group(name, value)` (Python).
  - `getSubGroup(String name)` now rejects names containing `=` to prevent cache-key collisions with the key-value variant.                                                                                                                           
                                                                                                                                                                                                                                                      
  ### Documentation                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                      
  <!-- Do not remove this section. Check the proper box only. -->                                                                                                                                                                                     
   
  - [ ] `doc-needed` <!-- Your PR changes impact docs -->                                                                                                                                                                                             
  - [ ] `doc-not-needed` <!-- Your PR changes do not impact docs -->
  - [x] `doc-included` <!-- Your PR already contains the necessary documentation updates -->


python/flink_agents/api/chat_models/chat_model.py
```diff
@@ -226,6 +226,6 @@ def _record_token_metrics(
         if metric_group is None:
             return
 
-        model_group = metric_group.get_sub_group(model_name)
+        model_group = metric_group.get_sub_group("model", model_name)
```
weiqingy, line 229: This same migration — `get_sub_group(model_name)` → `get_sub_group("model", model_name)` — also lives in `vector_store_long_term_memory.py`'s `_report_token_metrics` (around line 267), where long-term-memory token usage is still recorded under the flat `get_sub_group(metric["model_name"])`. On `main` that path was the mem0 file #760 migrated, but mem0 doesn't exist on `release-0.2`, so the vector-store LTM is where it lives here — and it sits outside this PR's diff. Left as-is, chat-model token metrics pick up the `model.` prefix while LTM metrics keep `long-term-memory.<model>.promptTokens`, and since the rename is breaking it'd take a second breaking change to bring LTM in line later. Was leaving it out a deliberate scope call, or would it be worth folding the same one-line change in here? (`test_vector_store_long_term_memory.py` mocks `get_sub_group` with an arg-agnostic autospec, so it shouldn't need a test rewrite.)

yunfengzhou-hub: Thanks for pointing out this. It is worth folding the same one line change here. I've made the change.

wenjin272:
Thanks @yunfengzhou-hub and @weiqingy. Since this issue is a backport of #760 and the comments from @weiqing have been addressed, I will merge this PR.

Issue #753, opened 2026-06-04
[Tech Debt] Add a document for Support TTL on Short Term Memory


GreatEugenius:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Add a document for Support TTL on Short Term Memory

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

GreatEugenius:
Hi @da-daken, as the implementer of the relevant PR, would you like to add documentation for this feature?

da-daken:
> Hi [@da-daken](https://github.com/da-daken), as the implementer of the relevant PR, would you like to add documentation for this feature?

@GreatEugenius Yes, I will submit it when I'm available, within these few days

PR #784, opened 2026-06-06
[doc] Support TTL on ShortTermMemory in User Documentation

da-daken:
Linked issue: [#753](https://github.com/apache/flink-agents/issues/753) https://github.com/apache/flink-agents/issues/580

### Purpose of change

Add a document for Support TTL on Short Term Memory

### Tests

No need to test

### API

no api

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/memory/sensory_and_short_term_memory.md
+++ b/docs/content/docs/development/memory/sensory_and_short_term_memory.md
@@ -252,7 +252,10 @@ public static void secondAction(Event event, RunnerContext ctx) throws Exception
 {{< /tab >}}
 
 {{< /tabs >}}
+
 ## Auto-Cleanup Behavior
+
+### Sensory Memory
 
 Sensory Memory is automatically cleared by the framework after each agent run completes. This cleanup happens:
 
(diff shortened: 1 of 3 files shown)


da-daken:
hi @xintongsong @GreatEugenius , I have added user documentation and configuration documentation about short-term memory TTL, but I'm not sure if it meets the requirements. PTAL～

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -263,4 +266,56 @@ Sensory Memory is automatically cleared by the framework after each agent run co
...
\ No newline at end of file
+{{< /hint >}}
+
+### Short-Term Memory
+
+Short-term memory can be configured with a time-to-live (TTL) so that older state expires automatically. This is useful for agents that may run for a long time: if the agent only needs recent memories, expiring historical data directly keeps the stored state focused on the latest context and avoids retaining stale information.
+
+Set `short-term-memory.state-ttl.ms` to a value greater than 0 in milliseconds to enable TTL. You can also configure how the TTL is refreshed and whether expired state can be returned before Flink cleans it up:
+
+- `short-term-memory.state-ttl.update-type`: controls whether TTL is refreshed on create/write or on read/write.
```
weiqingy, line 277: Both the prose ("older state expires automatically… expiring historical data") and the two examples use `ON_READ_AND_WRITE`, the default — under which a *read* also resets the TTL clock. So an old entry that's still being read periodically won't expire. A reader skimming the intro might take "older" to mean by creation time. Would it be worth one clarifying sentence here — that with `ON_READ_AND_WRITE` "age" is measured from last access, and `ON_CREATE_AND_WRITE` is the one that expires by write/creation time? It's okay with me if you'd rather keep it brief.

weiqingy (commented):
Thanks for adding this — nice that the Python and Java snippets mirror the real e2e/integration tests exactly, and every config key, default, and enum value matches `AgentExecutionOptions` / `core_options.py`. One optional question inline.

docs/content/docs/development/memory/sensory_and_short_term_memory.md
da-daken: I think 'controls whether TTL is refreshed on create/write or on read/write.' already sufficiently explains the above situation, so I don't intend to provide any additional supplements. Thank you for such a prompt reply.

wenjin272: I believe the literals themselves, along with a single line of additional explanation, are sufficient; therefore, I agree with keeping it brief.

wenjin272 (approved):
LGTM

Issue #772, opened 2026-06-05
[Bug] Java toTable signature description in integrate_with_flink doc is wrong


Sxnan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Found during integrate_with_flink doc verification (#743).

The general-text paragraph after the `from_table`/`to_table` tabs in
`docs/content/docs/development/integrate_with_flink.md` (lines 175-178) reads:

> User should provide `KeySelector` in `from_table()` to tell how to convert
> the input `Table` to `KeyedStream` internally. And provide `Schema` and
> `TypeInformation` in `to_table()` to tell the output `Table` schema.
>
> {{< hint info >}}
> Currently, user should provide both `Schema` and `TypeInformation` when call
> `to_table()`, we will support only provide one of them in the future.
> {{< /hint >}}

This text sits outside the `{{< /tabs >}}` block, so readers take it as
applying to both languages. It is true for Python only; the Java API only
accepts `Schema`.

- Python — `python/flink_agents/api/execution_environment.py:78`:
  ```python
  def to_table(self, schema: Schema, output_type: TypeInformation) -> Table:
  ```
- Java — `api/src/main/java/org/apache/flink/agents/api/AgentBuilder.java:90`:
  ```java
  Table toTable(Schema schema);
  ```

The Java integration test confirms the single-arg signature:
`e2e-test/.../FlinkIntegrationTest.java:140` calls `.toTable(outputSchema)`.

Java users following the general text will look for a two-arg overload that
does not exist.

Fix: scope the "Schema + TypeInformation" claim to Python, and document that
Java's `toTable(Schema)` does not need TypeInformation.

### How to reproduce

Read the doc as a Java user; the trailing paragraph after the
from_table/to_table tabs describes API behavior that does not match the Java
side.

### Version and environment

Flink Agents 0.3.0 (`main`).

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #790, opened 2026-06-08
[docs] Clarify Java toTable takes only Schema in integrate_with_flink

wzhero1:
Linked issue: #772

### Purpose of change

The general-text paragraph after the `from_table`/`to_table` tabs in
`docs/content/docs/development/integrate_with_flink.md` sits outside the
`{{< /tabs >}}` block, so readers apply it to both languages. It claimed
`to_table()` requires both `Schema` and `TypeInformation`, which is true for
Python only — the Java API accepts only `Schema`:

- Java: `Table toTable(Schema schema)` (`api/src/main/java/org/apache/flink/agents/api/AgentBuilder.java:90`)
- Python: `def to_table(self, schema, output_type)` (`python/flink_agents/api/execution_environment.py:78`)
- e2e tests call single-arg `.toTable(outputSchema)` in `FlinkIntegrationTest`, `ReActAgentTest`, `SkillsIntegrationTest`.

This change scopes the "Schema + TypeInformation" requirement to Python, adds an
explicit note that Java's `toTable(Schema)` does not need `TypeInformation`, and
narrows the info hint to Python.

### Tests

Docs-only change. Verified locally by building the site with Hugo (`hugo` in
`docs/`); the page renders the per-language list and the info hint correctly.

### API

No. Documentation-only; no public API changed.

### Documentation

- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/integrate_with_flink.md
+++ b/docs/content/docs/development/integrate_with_flink.md
@@ -172,7 +172,13 @@ Table outputTable =
 {{< /tabs >}}
 
 
-User should provide `KeySelector` in `from_table()` to tell how to convert the input `Table` to `KeyedStream` internally. And provide `Schema` and `TypeInformation` in `to_table()` to tell the output `Table` schema.
+User should provide `KeySelector` in `from_table()` to tell how to convert the input `Table` to `KeyedStream` internally.
+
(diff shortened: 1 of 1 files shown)


wenjin272 (approved):
LGTM

PR #813, opened 2026-06-08
[docs] Clarify Workflow Agent event emission and event metadata

### LINKED ISSUE #793: [Tech Debt] Clarify Workflow Agent documentation around event emission and event metadata
### Search before asking

### Description

There are two documentation/API-contract follow-ups around the Workflow Agent page.

1. The `Send Event` snippet shows one action sending both a `ChatRequestEvent` and an `OutputEvent`.

   This is valid API usage, but it can be misleading as a workflow example. In the runtime, an `OutputEvent` is emitted downstream immediately, while the `ChatRequestEvent` continues through the chat-model path. If a `ChatResponseEvent` handler also emits an output, one input may produce both an immediate output and a later model-based output.

   Source / evidence:
   - The snippet is in `docs/content/docs/development/workflow_agent.md`, in the `Send Event` section. It shows `ctx.send_event(ChatRequestEvent(...))` followed by `ctx.send_event(OutputEvent(...))` in the same action. The Java tab shows the same pattern with `ctx.sendEvent(...)`.
   - Runtime behavior is in `runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java`: `OutputEvent` is collected downstream immediately, while non-output events are routed to triggered actions.

   Suggested fix: split the snippet into separate examples for "trigger another action" and "emit downstream output", or explicitly state that sending an `OutputEvent` in the same action is only for immediate downstream output and is not the normal chat request/response workflow pattern.

2. The typed event reconstruction contract should be explicit about base `Event` metadata such as `id` and `sourceTimestamp`.

   `fromEvent` / `from_event` helpers reconstruct typed events from a base `Event`. If metadata such as `id` or `sourceTimestamp` is dropped in any typed reconstruction path, event logs, listeners, correlation, deduplication, and downstream timestamp propagation become harder to reason about.

   Source / evidence:
   - The custom event subclass guidance is in `docs/content/docs/development/workflow_agent.md`, in the `Custom Event Subclasses` section. The examples require users to implement `from_event` / `fromEvent`, but do not state whether base metadata should be copied.
   - Built-in Java event reconstruction paths such as `InputEvent.fromEvent(...)` and `ChatRequestEvent.fromEvent(...)` preserve the base event `id`, and Java paths also preserve `sourceTimestamp` where applicable. This should either become part of the documented contract or be explicitly described as implementation-specific behavior.
   - Python built-in event reconstruction paths preserve the base `id`, for example `InputEvent.from_event(...)`, but the workflow page does not tell users writing custom events what metadata they are responsible for preserving.

   Suggested fix: audit typed event reconstruction implementations and either preserve base metadata consistently or document the intended contract and limitations.

wzhero1:
Linked issue: #793

### Purpose of change

Two documentation/API-contract clarifications on `docs/content/docs/development/workflow_agent.md`, per #793 (docs-only, no runtime/API change):

- **Send Event snippet (Part 1)**: the previous snippet sent both a `ChatRequestEvent` and an `OutputEvent` from a single action, which is valid but misleading — at runtime an `OutputEvent` is emitted downstream immediately while the `ChatRequestEvent` continues through the chat-model path, so one input can produce both an immediate output and a later model-based output. Split it into two separate examples — **"Trigger another action"** (sends only `ChatRequestEvent`) and **"Emit downstream output"** (sends `OutputEvent` from the `ChatResponseEvent` handler) — and add a hint stating that `OutputEvent` is collected/emitted downstream immediately, bypassing action routing, while other events are routed to their listening actions.
- **Custom event metadata contract (Part 2)**: the `from_event` / `fromEvent` examples did not say whether base `Event` metadata should be preserved. Document the contract — reconstruction should preserve the base `id` (both languages) and `sourceTimestamp` (Java only, runtime-internal; the Python `Event` has no equivalent) — matching all built-in events, and update the `MyEvent` example to follow it (Python assigns `result.id = event.id` last because the content-based id regenerates on field changes; Java gains a `@JsonCreator` id-accepting constructor and carries `sourceTimestamp` when present).

### Tests

Documentation-only change; no runtime impact. Verified manually:

- Both claims audited against source: all 8 built-in Java events and 8 Python events preserve `id` in their `fromEvent` / `from_event`; Java additionally preserves `sourceTimestamp`; `ActionExecutionOperator.processEvent` collects `OutputEvent` downstream immediately and routes all other events to triggered actions.
- The updated `MyEvent` Java snippet compiles (JDK 11 + Flink 2.2.0 classpath) and runs — `fromEvent` preserves both `id` and `sourceTimestamp`. The Python snippet runs in the venv — `from_event` preserves a custom `id`.
- Hugo shortcodes are balanced (tabs/tab/hint), all tab-group ids are unique, and code fences are even.

### API

No public API changes.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

Closes #793

### DIFF
--- a/docs/content/docs/development/workflow_agent.md
+++ b/docs/content/docs/development/workflow_agent.md
@@ -266,34 +266,66 @@ public class ReviewAnalysisAgent extends Agent {
 
 In the function, user can also send new events, to trigger other actions, or output the data.
 
-{{< tabs "Send Event" >}}
+**Trigger another action** — send a built-in or custom event that another action listens to:
+
+{{< tabs "Trigger Another Action" >}}
 
 {{< tab "Python" >}}
 ```python
 @action(InputEvent.EVENT_TYPE)
 @staticmethod
 def process_input(event: Event, ctx: RunnerContext) -> None:
-    # send ChatRequestEvent
-    ctx.send_event(ChatRequestEvent(model=xxx, messages=xxx))
-    # output data to downstream
-    ctx.send_event(OutputEvent(output=xxx))
+    # send a ChatRequestEvent to trigger the built-in chat-model action
+    ctx.send_event(ChatRequestEvent(model="my_model", messages=messages))
 ```
 {{< /tab >}}
 
 {{< tab "Java" >}}
 ```java
 @Action(listenEventTypes = {InputEvent.EVENT_TYPE})
-public static void processInput(Event event, RunnerContext ctx) throws Exception {
-    InputEvent inputEvent = InputEvent.fromEvent(event);
-    // send ChatRequestEvent
+public static void processInput(Event event, RunnerContext ctx) {
+    // send a ChatRequestEvent to trigger the built-in chat-model action
     ctx.sendEvent(new ChatRequestEvent("my_model", messages));
-    // output data to downstream
-    ctx.sendEvent(new OutputEvent(xxx));
-}   
+}
 ```
 {{< /tab >}}
 
 {{< /tabs >}}
+
+**Emit downstream output** — send an `OutputEvent` to produce an output of the agent:
+
+{{< tabs "Emit Output" >}}
+
+{{< tab "Python" >}}
+```python
+@action(ChatResponseEvent.EVENT_TYPE)
+@staticmethod
+def emit_output(event: Event, ctx: RunnerContext) -> None:
+    # output data to downstream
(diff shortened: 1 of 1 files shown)


wenjin272 (approved):
LGTM

Issue #751, opened 2026-06-04
[Tech Debt] Fix JAR copy instruction in installation doc after common/thin JAR split


yunfengzhou-hub:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The "Install Flink Agents Java Library — From Official Release" section in `docs/content/docs/get-started/installation.md` only copies the thin JAR from the Python package, missing the common JAR introduced by #557. This affects users who install via `pip install flink-agents` and manually copy JARs to `$FLINK_HOME/lib/`.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

Sxnan:
Took a look while reviewing. The observation that only the thin JAR is copied is correct, but I think the right fix may be to drop this copy step for the Python path rather than add the common JAR alongside it.

`AgentsExecutionEnvironment.get_execution_environment(env=...)` already calls `env.add_jars(...)` for both the common and thin JARs ([execution_environment.py:139–164](https://github.com/apache/flink-agents/blob/main/python/flink_agents/api/execution_environment.py#L139-L164)). PyFlink's `add_jars` injects them into the client-side py4j JVM context classloader AND into `pipeline.jars` for JM/TM user-code classloader (child-first). Verified against a clean PyFlink install (no flink-agents JARs in `pyflink/lib/`): both a thin-JAR probe class and a common-JAR probe class are unreachable before the call and reachable after. So nothing in `$FLINK_HOME/lib/` is needed for Python users.

Copying the common JAR into `$FLINK_HOME/lib/` would also push all integration deps (AWS / OpenAI / Anthropic / Milvus / ES / OpenSearch / Azure / MCP …) — no relocation — into the system classloader, enlarging the dependency conflict surface for every user job, including ones that don't touch those integrations.

Will send a PR shortly that scopes this step to the Java-only path.

PR #818, opened 2026-06-09
[doc] Scope Flink Agents Java Library install step to Java jobs

Sxnan:
Linked issue: #751

### Purpose of change

The current `Install Flink Agents Java Library` section in `installation.md` instructs every user to copy a JAR into `$FLINK_HOME/lib/`. This is unnecessary for Python users — `AgentsExecutionEnvironment.get_execution_environment(env=...)` already calls `env.add_jars(...)` for both the common and version-specific (thin) JARs, which PyFlink injects into the client-side py4j JVM context classloader and into `pipeline.jars` for the JM/TM user-code classloader. Copying integration JARs into `$FLINK_HOME/lib/` also pushes them into the system classloader, which enlarges the dependency conflict surface for unrelated user jobs.

The original `cp` glob also matched both the fat and thin JARs, leading to duplicate classes on the classpath.

This PR:

- Scopes the section to Java jobs (or other deployments that don't go through the Python entry point) and adds a hint telling Python users to skip it.
- Switches the official-release path from copying out of `$PYTHONPATH/flink_agents/lib/` to a direct Maven Central download of the self-contained fat JAR, so Java-only users no longer need a Python install. Mirrors what `tools/install.sh` already does.
- Narrows the From Source `cp` to the exact fat JAR filename so the `-thin` variant is no longer accidentally included.

### Tests

Docs-only change. Verified that:

- The pip-installed `flink_agents/lib/{common, flink-<ver>}/` layout is reached automatically by `env.add_jars` via `add_jars_to_context_class_loader`: starting from a clean PyFlink install (no flink-agents JARs in `pyflink/lib/`), a thin-JAR probe class (`org.apache.flink.agents.integrations.chatmodels.ollama.OllamaChatModelConnection`) and a common-JAR probe class (`io.github.ollama4j.Ollama`) are both `ClassNotFoundException` before the call and resolve after — confirming nothing in `$FLINK_HOME/lib/` is required for the Python path.
- Maven Central publishes the self-contained fat JAR under `flink-agents-dist-flink-<minor>-<version>.jar` for each released version (matches the URL in the new official-release snippet).

### API

No public API changes.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/get-started/installation.md
+++ b/docs/content/docs/get-started/installation.md
@@ -200,28 +200,46 @@ export PYTHONPATH=$(python -c 'import sysconfig; print(sysconfig.get_paths()["pu
 **Note:** You can add the `export PYTHONPATH=...` line to your shell profile (`~/.bash_profile`, `~/.bashrc`, `~/.zprofile`, or `~/.zshrc`) to set it permanently. This way, it will be automatically configured in all future terminal sessions.
 {{< /hint >}}
 
-### Install Flink Agents Java Library
+### Install Flink Agents Java Library (Java jobs only)
 
-Copy the Flink Agents distribution JAR to your Flink installation's `lib` directory:
+{{< hint info >}}
+**Skip this section if you only use the Python API.** When you call
+`AgentsExecutionEnvironment.get_execution_environment(env=...)`, the framework
+automatically registers the required JARs via `pipeline.jars` so Flink loads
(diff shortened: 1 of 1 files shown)


wenjin272 (approved):
LGTM

Issue #824, opened 2026-06-09
[Bug] Stale processing keys in union list state can be snapshotted by non-owner subtasks


joeyutong:
### Search before asking

I searched in the issues and found nothing similar.

### Description

`currentProcessingKeysOpState` is registered as operator **union list state**. After a rescale or restore, every new subtask restores the union of all previous subtasks' list state.

`ActionExecutionOperator#tryResumeProcessActionTasks` already filters restored keys by key ownership before scheduling recovery work. This means a non-owner subtask will not resume processing a key that belongs to another subtask. However, the non-owner subtask still keeps the non-owned key in its local `currentProcessingKeysOpState`.

This stale local state can be written back into the next checkpoint:

1. A job runs with parallelism 1 and starts processing key `K`.
2. `K` is added to `currentProcessingKeysOpState`.
3. A checkpoint is taken while `K` is still in flight, so the checkpoint contains `[K]`.
4. The job is restored or rescaled to parallelism 2.
5. Because this is union list state, both new subtasks restore `[K]`.
6. The owner subtask schedules recovery for `K`; the non-owner subtask skips execution because `K` is not in its key-group range.
7. If the non-owner subtask does not prune `K` from `currentProcessingKeysOpState`, it snapshots `[K]` again in the next checkpoint.
8. The owner subtask also snapshots `[K]` while `K` is still in flight.
9. On the next restore, union list state restores `[K, K]`.

This violates the invariant that one in-flight key should appear only once in `currentProcessingKeysOpState`. It can lead to duplicate recovery mailbox submissions for the same key. When the key finishes, `removeProcessingKey(K)` removes all matching entries and returns a count greater than 1, causing the operator check to fail:

```text
Current processing key count for key K should be 1, but got 2
```

Expected behavior: after restore, each subtask should keep only the processing keys it owns in its local operator state. Non-owned restored keys should be pruned so they are not checkpointed again by non-owner subtasks. Duplicate restored keys should also not result in duplicate recovery submissions.

Actual behavior: non-owned keys are skipped for execution but remain in the non-owner subtask's union list state, so stale keys can be re-published into later checkpoints.

### How to reproduce

This can be reproduced with a unit test around `ActionExecutionOperator`:

1. Start an `ActionExecutionOperator` test harness with `maxParallelism = 4`, `parallelism = 1`, and key selector `value -> value`.
2. Process one element with key `K`.
[...]

PR #825, opened 2026-06-10
[runtime] Prune restored processing keys before snapshot

joeyutong:
Linked issue: #824

### Purpose of change

This PR fixes duplicate processing-key recovery after rescale or restore.

`currentProcessingKeysOpState` is union list state, so every restored subtask sees the union of all previous subtasks' processing keys. The existing restore logic skips non-owned keys when scheduling recovery work, but non-owner subtasks still kept those keys in their local operator state. If a checkpoint is taken before the owner finishes the key, non-owner subtasks can snapshot stale processing keys again. A later restore can then see duplicate entries for the same key, causing duplicate recovery mailbox submissions and eventually making `removeProcessingKey` return a count greater than one.

The fix keeps only distinct keys owned by the current subtask in the local operator state after restore. This prevents non-owner subtasks from re-publishing stale keys into later checkpoints.

### Tests

- `mvn -pl runtime -am -Dtest=ActionExecutionOperatorTest#testRestoreOnlyResumesKeysOwnedByCurrentSubtask -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -pl runtime -am -Dtest=ActionExecutionOperatorTest -Dsurefire.failIfNoSpecifiedTests=false test`

### API

No public API changes. The change only adds a package-private helper in `OperatorStateManager`.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
@@ -55,8 +55,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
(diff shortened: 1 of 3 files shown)


joeyutong:
cc @wenjin272

wenjin272 (approved):
LGTM

Issue #723, opened 2026-06-01
[Bug] Python memory can persist Pemja-incompatible values and fail after state backend restore


joeyutong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Python-side memory values can currently accept objects that are not stable across the Pemja / Flink state backend boundary. When such values are written to sensory memory or short-term memory and then checkpointed/restored, the restored Java state may contain `pemja.core.object.PyObject` / `PyJObject` wrappers rather than a materialized, process-independent payload. After a TaskManager / Python process restart, converting those restored objects back to Python can fail or crash in Pemja native conversion.

There are two related user-facing problems:

1. The memory documentation does not clearly define the Python-side value contract.
   The docs mention primitive types, collections, Java POJOs, general class types, and `MemoryObject`, but do not say whether Python values must be recursively JSON-like / checkpoint-stable. For example, it is unclear whether `uuid.UUID`, Pydantic models, custom Python classes, or a `dict` containing those objects are valid memory values.

2. The built-in tool context path writes `UUID` objects into sensory memory.
   In `python/flink_agents/plan/actions/chat_model_action.py`, tool call context uses `initial_request_id: UUID` / `tool_request_event_id: UUID` in dictionaries stored in `ctx.sensory_memory`. For example:
   - `_update_tool_call_context(...)` stores a dict keyed by `initial_request_id`.
   - `_save_tool_request_event_context(...)` stores `"initial_request_id": initial_request_id` and stores the context under `tool_request_event_id`.

With Pemja, `uuid.UUID` does not have a dedicated Python-to-Java conversion branch, so it falls back to a `PyObject` wrapper. That makes the built-in tool-call flow capable of storing a non-checkpoint-stable payload in Flink state.

Expected behavior:

- The Python memory value contract should be explicit and enforceable.
- Built-in runtime memory payloads should not store raw Python objects such as `uuid.UUID`.
- Values stored in Flink-backed memory should be safe to checkpoint and restore across TaskManager / Python process restarts.

A reasonable Python-side contract could be to accept only recursively checkpoint-stable values such as:

```python
None | bool | int | float | str | bytes | list[MemoryValue] | dict[str, MemoryValue]
```

Pydantic values should be materialized before storing, for example:

```python
ctx.sensory_memory.set("x", model.model_dump(mode="json"))
```

or:

```python
[...]

PR #828, opened 2026-06-10
[python] Normalize built-in tool-call context for checkpoint safety

weiqingy:
Linked issue: #723

### Purpose of change

The built-in chat-model action stored non-primitive Python objects in sensory memory: `UUID` values, an `OutputSchema`, and `ChatMessage` lists. Pemja wraps such objects as `PyObject` holders whose JNI pointers go stale after a TaskManager/Python restart, so restoring the checkpointed tool context crashes in `JcpPyObject_FromJObject`.

This normalizes those values to a primitive-only form before they reach memory and reconstructs the rich types on read, entirely inside the three tool-context helpers in `chat_model_action.py` (no caller or signature changes):

- `ChatMessage` lists are stored via `model_dump(mode="json")` and reconstructed via `ChatMessage.model_validate`. `mode="json"` is required because `MessageRole` is a `str, Enum` that a bare `model_dump()` would leave as an enum member.
- `initial_request_id` is stored as `str` and reconstructed to `UUID`.
- `output_schema` is stored via `OutputSchema.model_dump()` and reconstructed via `OutputSchema.model_validate` (None-guarded).

Dict keys were already strings and the retry-stats context already holds only `int`s, so both are unchanged. `prompt_args` is user-supplied and already round-trips as a `ChatRequestEvent` attribute, so it is left as-is.

This is the first of the agreed changes on #723. A follow-up PR will add a `set()`-time validator for user-supplied memory values and document the Python memory value contract.

### Tests

`plan/tests/actions/test_chat_model_action.py`: 9 new unit tests plus a recursive `_assert_primitive` helper that asserts the stored form is primitive-only (a checkpoint-safety proxy, since no Python checkpoint/restore harness exists), and that round-trips reconstruct `UUID` / `OutputSchema` / `ChatMessage`, preserve the `UUID`→`str` dict-key match and `RowTypeInfo`, handle a `None` output_schema, and keep `model` / `prompt_args`.

`plan/tests/actions/test_chat_model_action_retry.py`: updated the hand-seeded context in `test_forwards_saved_prompt_args_to_chat` to the new primitive-only stored form.

```
cd python
uv run --no-sync pytest flink_agents/plan/tests/actions/ -v   # 22 passed
uv run --no-sync pytest flink_agents -k "not e2e_tests" -q    # all green
```

### API

No public API change. The normalization is fully encapsulated in the existing tool-context helpers; callers and method signatures are unchanged.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/plan/actions/chat_model_action.py
+++ b/python/flink_agents/plan/actions/chat_model_action.py
@@ -15,7 +15,6 @@
 #  See the License for the specific language governing permissions and
 # limitations under the License.
 #################################################################################
-import copy
 import json
 import logging
 import re
@@ -59,6 +58,16 @@ _logger = logging.getLogger(__name__)
 # ============================================================================
 # Helper Functions for Tool Call Context Management
 # ============================================================================
+def _serialize_messages(messages: List[ChatMessage]) -> List[Dict]:
+    """Materialize chat messages into JSON-safe dicts for checkpoint-stable memory."""
+    return [message.model_dump(mode="json") for message in messages]
+
+
+def _deserialize_messages(messages: List[Dict]) -> List[ChatMessage]:
+    """Reconstruct chat messages from their stored JSON-safe dict form."""
+    return [ChatMessage.model_validate(message) for message in messages]
+
+
 def _update_tool_call_context(
     sensory_memory: MemoryObject,
     initial_request_id: UUID,
@@ -81,12 +90,12 @@ def _update_tool_call_context(
     key = str(initial_request_id)
     tool_call_context = sensory_memory.get(_TOOL_CALL_CONTEXT) or {}
     if key not in tool_call_context and initial_messages is not None:
-        tool_call_context[key] = copy.deepcopy(initial_messages)
+        tool_call_context[key] = _serialize_messages(initial_messages)
 
-    tool_call_context[key].extend(added_messages)
+    tool_call_context[key].extend(_serialize_messages(added_messages))
 
     sensory_memory.set(_TOOL_CALL_CONTEXT, tool_call_context)
-    return tool_call_context[key]
+    return _deserialize_messages(tool_call_context[key])
 
 
 def _save_tool_request_event_context(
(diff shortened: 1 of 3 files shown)


joeyutong:
LGTM. One thing to note: this fixes assume the job is restarted from a clean state.

wenjin272 (approved):
LGTM.

I tried to add an e2e case to truly verify that recovery from a checkpoint fails before the fix and is resolved after it. However, in the MiniCluster, triggering an in-place recovery by throwing an exception does not recreate the JVM, so the problematic code path is never exercised. I think we can, after #708 merge, add a case that manually kills the TM process in a standalone cluster to trigger recovery, and use that to verify this fix.

weiqingy:
> LGTM. One thing to note: this fixes assume the job is restarted from a clean state.

Thanks for the review @joeyutong .
Agreed — that's the right way to read it. The change makes the context written to memory primitive-only from here on, and the read path now expects that primitive layout, so it doesn't migrate state already checkpointed by the old code: a savepoint taken before this fix still holds the non-primitive form and would hit the original problem on restore. So it's forward-looking — correct for fresh runs and for checkpoints written after it lands, not for restoring a pre-fix checkpoint. That fits where the project is today (no state-migration guarantees on the built-in tool context yet); if we ever need pre-fix-checkpoint compatibility we can add a migration shim on the read path.

weiqingy:
> LGTM.
> 
> I tried to add an e2e case to truly verify that recovery from a checkpoint fails before the fix and is resolved after it. However, in the MiniCluster, triggering an in-place recovery by throwing an exception does not recreate the JVM, so the problematic code path is never exercised. I think we can, after #708 merge, add a case that manually kills the TM process in a standalone cluster to trigger recovery, and use that to verify this fix.

@wenjin272 Thanks for digging into the e2e angle. That matches what we found while scoping this — local/MiniCluster mode never crosses Pemja, so the SIGSEGV path can't be reproduced there, which is why the unit tests assert the stored form is recursively primitive as a checkpoint-safety proxy instead of driving a real restore. Killing the TM process in a standalone cluster after #708 is exactly the right way to get true before/after verification. I filed #836 to track it so it doesn't get lost.

Issue #822, opened 2026-06-09
[Feature] Add parallel LLM quickstart and example for multi-action fan-out


Ryan-Nightwish:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Flink Agents already supports parallel LLM invocations via multi-action fan-out — an agent action
can emit multiple `ChatRequestEvent` events, and the framework executes the corresponding LLM calls
concurrently (on JDK 21+ for Java; always for Python). However, none of the existing quickstart
examples or documentation covers this capability.

**Proposed changes:**

1. **New quickstart doc** (`docs/content/docs/get-started/quickstart/parallel_llm.md`) — walks
   through a "Parallel Sentiment Analysis" example end-to-end: environment setup, agent design,
   Flink integration, and deployment instructions. Includes both Python and Java tabs.

2. **New Java example**
   - `examples/.../ParallelChatRequestExample.java` — entry point
   - `examples/.../agents/ParallelChatAgent.java` — agent that fans out three sentiment-dimension
     `ChatRequestEvent` events in parallel, collects responses, and aggregates with a final LLM call

3. **New Python example**
   - `python/.../parallel_chat_request_example.py` — entry point
   - `python/.../agents/parallel_chat_agent.py` — Python equivalent of the Java agent

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #830, opened 2026-06-10
[docs][examples] Add parallel LLM quickstart and examples for multi-action fan-out

Ryan-Nightwish:
Linked issue: #822 

### Purpose of change

Flink Agents supports parallel LLM invocations via multi-action fan-out, but no existing quickstart or example demonstrates this capability. This PR adds a complete "Parallel Sentiment Analysis" example (Java + Python) and a quickstart doc page covering the end-to-end workflow.

**New files:**

- `docs/content/docs/get-started/quickstart/parallel_llm.md` — quickstart page with code walkthrough (both Python and Java tabs), deployment instructions, and constraints for multi-action fan-out
- `examples/.../ParallelChatRequestExample.java` — Java entry point
- `examples/.../agents/ParallelChatAgent.java` — Java agent that fans out three `ChatRequestEvent` events (one per sentiment dimension) in parallel, collects responses via sensory memory, and aggregates with a final LLM call
- `python/.../parallel_chat_request_example.py` — Python entry point
- `python/.../agents/parallel_chat_agent.py` — Python equivalent of the Java agent

**Modified files:**

- `examples/.../agents/CustomTypesAndResources.java` — add shared types (`SentimentRequest`, `SentimentKeySelector`, `AspectResponse`, `SummaryResponse`) following project convention
- `python/.../agents/custom_types_and_resources.py` — add shared types (`AspectResponse`, `SummaryResponse`)

### Tests

- Verified Java example runs successfully (JDK 21, parallel execution)
- Verified Python example runs successfully (conda + Python 3.11)
- Both produce correct output: `taste:positive, service:negative, price:not_mentioned`

### API

No public API changes.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/get-started/quickstart/parallel_llm.md
+++ b/docs/content/docs/get-started/quickstart/parallel_llm.md
@@ -84,13 +84,13 @@ agentsEnv.addResource(
 
 ### Create the Agent
 
-Below is the example code for the `ParallelChatAgent`. Two system prompts — `PARALLEL_SYSTEM_PROMPT` for the parallel aspect judgments and `AGGREGATE_SYSTEM_PROMPT` for the aggregation call — are packaged into `ChatRequestEvent`s by `_build_aspect_request` / `buildAspectRequest` and `_build_summarize_request` / `buildSummarizeRequest`. The agent defines a chat model and four actions: `request_aspect_judgments` initializes the row state and emits a single `SentimentInputEvent`; `handle_taste_input` and `handle_service_input` each listen on `SentimentInputEvent` and independently dispatch one `ChatRequestEvent`, recording a `{request_id → aspect}` entry in the shared `aspect_map`; and `handle_response` looks up the dispatched aspect by `request_id`, accumulates the results, and emits the final output. For more details, please refer to the [Workflow Agent]({{< ref "docs/development/workflow_agent" >}}) documentation.
+Below is the example code for the `ParallelChatAgent`. Two system prompts — `PARALLEL_SYSTEM_PROMPT` for the parallel aspect judgments and `AGGREGATE_SYSTEM_PROMPT` for the aggregation call — are packaged into `ChatRequestEvent`s by `_build_aspect_request` / `buildAspectRequest` and `_build_summarize_request` / `buildSummarizeRequest`. The agent defines a chat model and four actions: `request_aspect_judgments` stores the review `id` and `text` in sensory memory and emits a single `SentimentInputEvent`; `handle_taste_input` and `handle_service_input` each listen on `SentimentInputEvent` and independently dispatch one `ChatRequestEvent`, recording a `{request_id → aspect}` entry using path-based memory access (`aspect_map.<request_id>`); and `handle_response` looks up the dispatched aspect, stores the result under `sentiments.<aspect>`, and triggers aggregation once all aspects are collected. For more details, please refer to the [Workflow Agent]({{< ref "docs/development/workflow_agent" >}}) documentation.
 
 {{< tabs "Create the Agent" >}}
 
 {{< tab "Python" >}}
 ```python
-ASPECTS: Tuple[str, ...] = ("taste", "service")
+ASPECTS: tuple = ("taste", "service")
 N_ASPECTS = len(ASPECTS)
 
 PARALLEL_SYSTEM_PROMPT = (
@@ -107,9 +107,12 @@ AGGREGATE_SYSTEM_PROMPT = (
 
 
 class SentimentInputEvent(Event):
+    """Intermediate event that broadcasts the review input to all aspect handlers."""
+
     EVENT_TYPE: ClassVar[str] = "SentimentInputEvent"
 
     def __init__(self, input_id: int, text: str) -> None:
+        """Initialize with the review id and text."""
         super().__init__(
             type=SentimentInputEvent.EVENT_TYPE,
             attributes={"input_id": input_id, "text": text},
@@ -131,11 +134,10 @@ def _build_aspect_request(text: str, aspect: str) -> ChatRequestEvent:
     )
 
 
-def _build_summarize_request(row: Dict[str, Any]) -> ChatRequestEvent:
+def _build_summarize_request(text: str, sentiments: Dict[str, str]) -> ChatRequestEvent:
     """Build a ChatRequestEvent for the aggregation phase."""
-    sentiments = row["sentiments"]
     body = (
-        f"Original: {row['text']}\n"
+        f"Original: {text}\n"
         + "Judgments: "
         + " ".join(f"{a}:{sentiments[a]}" for a in ASPECTS)
     )
@@ -181,32 +183,27 @@ class ParallelChatAgent(Agent):
     @action(InputEvent.EVENT_TYPE)
     @staticmethod
     def request_aspect_judgments(event: Event, ctx: RunnerContext) -> None:
-        """Process input event and dispatch a SentimentInputEvent for each aspect handler."""
-        row = _init_row(event)
-        # Sensory memory requires JSON serialization across the Pemja JVM boundary.
-        ctx.sensory_memory.set("res", json.dumps(row, ensure_ascii=False))
-        ctx.send_event(SentimentInputEvent(input_id=row["id"], text=row["text"]))
+        """Process input event and dispatch SentimentInputEvent to aspect handlers."""
+        payload = InputEvent.from_event(event).input
+        # Primitive types (int, str) cross the Pemja JVM boundary without serialization.
+        ctx.sensory_memory.set("id", payload["id"])
+        ctx.sensory_memory.set("text", payload["text"])
+        ctx.send_event(SentimentInputEvent(input_id=payload["id"], text=payload["text"]))
 
     @action(SentimentInputEvent.EVENT_TYPE)
     @staticmethod
     def handle_taste_input(event: Event, ctx: RunnerContext) -> None:
         """Handle taste aspect: build and send ChatRequestEvent for taste judgment."""
-        row = json.loads(ctx.sensory_memory.get("res"))
         req = _build_aspect_request(event.get_attr("text"), "taste")
-        row["aspect_map"][str(req.id)] = "taste"
-        # Sensory memory requires JSON serialization across the Pemja JVM boundary.
-        ctx.sensory_memory.set("res", json.dumps(row, ensure_ascii=False))
+        ctx.sensory_memory.set(f"aspect_map.{req.id}", "taste")
         ctx.send_event(req)
 
     @action(SentimentInputEvent.EVENT_TYPE)
     @staticmethod
     def handle_service_input(event: Event, ctx: RunnerContext) -> None:
-        """Handle service aspect: build and send ChatRequestEvent for service judgment."""
-        row = json.loads(ctx.sensory_memory.get("res"))
+        """Handle service aspect: build and send ChatRequestEvent for service."""
         req = _build_aspect_request(event.get_attr("text"), "service")
-        row["aspect_map"][str(req.id)] = "service"
-        # Sensory memory requires JSON serialization across the Pemja JVM boundary.
-        ctx.sensory_memory.set("res", json.dumps(row, ensure_ascii=False))
+        ctx.sensory_memory.set(f"aspect_map.{req.id}", "service")
         ctx.send_event(req)
 
     @action(ChatResponseEvent.EVENT_TYPE)
@@ -216,7 +213,7 @@ class ParallelChatAgent(Agent):
         ...
 ```
 
-The complete source including `_init_row`, `_build_output_event`, `_all_aspects_received`, and other supporting functions can be found in [`parallel_chat_agent.py`](https://github.com/apache/flink-agents/blob/main/python/flink_agents/examples/quickstart/agents/parallel_chat_agent.py).
+The complete source including `_build_output_event` and other supporting functions can be found in [`parallel_chat_agent.py`](https://github.com/apache/flink-agents/blob/main/python/flink_agents/examples/quickstart/agents/parallel_chat_agent.py).
 {{< /tab >}}
 
 {{< tab "Java" >}}
@@ -229,7 +226,7 @@ The complete source including `_init_row`, `_build_output_event`, `_all_aspects_
  *   <li>InputEvent → requestAspectJudgments → emits SentimentInputEvent
  *   <li>SentimentInputEvent triggers handlers in parallel:
  *       <ul>
- *         <li>handleTasteInput   → ChatRequestEvent (taste LLM call)
+ *         <li>handleTasteInput → ChatRequestEvent (taste LLM call)
  *         <li>handleServiceInput → ChatRequestEvent (service LLM call)
  *       </ul>
  *   <li>Each ChatResponseEvent → handleResponse (accumulates aspect results)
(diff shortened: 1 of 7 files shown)


Ryan-Nightwish:
Hi @wenjin272 @GreatEugenius , I have added examples and docs about parallel LLM invocations via multi-action fan-out, but I am not sure if it meets the requirements. PTAL orz.

docs/content/docs/get-started/quickstart/parallel_llm.md
```diff
@@ -0,0 +1,349 @@
...
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+{{< hint warning >}}
+**Constraints when using multi-action fan-out:**
+- **Fan-out action must exit immediately after sending events.** The action that emits multiple `ChatRequestEvent` events (e.g. `request_aspect_judgments`) should return right after calling `ctx.send_event(...)`. Do not `await`, block, or perform further business logic in the same action — the framework needs control back to schedule the parallel chat actions.
+- **Response handlers on the same key execute sequentially.** Multiple `ChatResponseEvent` triggers for the same key are processed one at a time, not concurrently. This means the read-modify-write pattern on sensory memory (e.g. `load_row → update → save_row`) is safe and will not suffer from concurrent overwrites.
+  {{< /hint >}}
```
GreatEugenius:
```suggestion
{{< /hint >}}
```

docs/content/docs/get-started/quickstart/parallel_llm.md
```diff
@@ -0,0 +1,349 @@
+---
+title: 'Parallel LLM Calls'
+weight: 3
```
GreatEugenius: frontmatter weight: 3 conflicts with existing pages

docs/content/docs/get-started/quickstart/parallel_llm.md
```diff
@@ -0,0 +1,349 @@
...
+        ResourceDescriptor.Builder.newBuilder(ResourceName.ChatModel.OLLAMA_CONNECTION)
+                .addInitialArgument("endpoint", "http://localhost:11434")
+                .addInitialArgument("requestTimeout", 240)
+                .build());
+```
+{{< /tab >}}
+
+{{< /tabs >}}
+
+### Create the Agent
```
GreatEugenius, line 85: The walkthrough snippets reference a number of helper functions that are never shown in the doc — _init_row, _save_row, _load_row, _build_aspect_request, _build_summarize_request, _parse_response, _is_final, _all_aspects_received (and their Java counterparts initRow, buildAspectRequest, etc.). The other quickstarts (e.g. workflow_agent.md) show self-contained action bodies that a reader can follow end-to-end, so this doc reads as incomplete by comparison: a reader can't tell what the prompts look like or how the response is parsed without leaving the page, and there's no pointer telling them where to go.

Suggest either inlining the key helpers (at least the prompt-building ones), or adding a sentence linking to the full sources, e.g.:
`The full implementation including helper functions can be found in parallel_chat_agent.py and ParallelChatAgent.java.`

GreatEugenius (commented):
Thanks for the PR. Overall, it looks good to me; I only left a few minor comments.

python/flink_agents/examples/quickstart/agents/parallel_chat_agent.py
```diff
@@ -0,0 +1,174 @@
...
+    @action(ChatResponseEvent.EVENT_TYPE)
+    @staticmethod
+    def handle_response(event: Event, ctx: RunnerContext) -> None:
+        """Process chat response event and send output event."""
+        parsed = _parse_response(event)
+        row = _load_row(ctx)
+        if _is_final(parsed):
+            ctx.send_event(_build_output_event(row, parsed))
+            return
+        row["sentiments"][parsed.aspect] = parsed.result
```
weiqingy: The accumulator is keyed on `parsed.aspect` — the string the LLM put in its response — while the completion check is `len(row["sentiments"]) == N_ASPECTS`. Since the key is whatever the model echoes back rather than the aspect each request was dispatched for, two failure modes open up with a small/non-deterministic model like the documented `qwen3:1.7b`: if the model returns the same `aspect` value twice, the map never reaches size 3 and the summarize request is never sent (the input silently produces no output); if it returns a label outside `ASPECTS`, the later `_build_summarize_request` lookup over the fixed `ASPECTS` list misses. The two languages even diverge on that miss — Python raises `KeyError` (line 92), while Java yields a silent `"taste:null"` (`ParallelChatAgent.java:123`).

Was correlating by the dispatched aspect considered here — e.g. carrying it through a `prompt_args` round-trip or a `request_id` — rather than trusting the model to echo it back? If the "trust the echoed aspect" simplification is intentional for the demo, would it be worth a doc note that the design assumes the model returns exactly one of the three labels, and that a non-conforming response breaks collection? That assumption is currently the load-bearing part of the example and it's invisible to a reader.

python/flink_agents/examples/quickstart/agents/parallel_chat_agent.py
```diff
@@ -0,0 +1,174 @@
...
+INPUT_TEXT = "The food here is great, but the service is too slow"
+ASPECTS: Tuple[str, ...] = ("taste", "service", "price")
+N_ASPECTS = len(ASPECTS)
+
+PARALLEL_SYSTEM_PROMPT = (
+    "You are a sentiment analysis assistant. Return JSON: "
+    '{"aspect":"<dimension>", "result":"<positive|negative|not_mentioned>"}'
+    " — no explanation, no extra fields."
+)
+AGGREGATE_SYSTEM_PROMPT = (
```
weiqingy, line 46: The two `AGGREGATE_SYSTEM_PROMPT`s ask the model for materially different summary formats. The Java side (`ParallelChatAgent.java:71-76`) spells out the value vocabulary inline — `{"summary":"taste:<positive/negative/not_mentioned>, service:<...>, price:<...>"}` — while this one gives an empty template, `{"summary":"taste: service: price:"}`. From the same input the two examples will produce visibly different summaries. The `PARALLEL_SYSTEM_PROMPT` is identical across both languages, which is what makes this read as an oversight rather than a deliberate per-language choice — and CLAUDE.md asks for Java/Python parity on cross-language features. Would it make sense to bring these two in line?

examples/src/main/java/org/apache/flink/agents/examples/agents/ParallelChatAgent.java
```diff
@@ -0,0 +1,188 @@
...
+                (CustomTypesAndResources.SentimentRequest) inputEvent.getInput();
+        Map<String, Object> row = new HashMap<>();
+        row.put("id", request.getId());
+        row.put("text", request.getText());
+        row.put("sentiments", new HashMap<String, String>());
+        return row;
+    }
+
+    private static void saveRow(RunnerContext ctx, Map<String, Object> row) throws Exception {
+        ctx.getSensoryMemory().set("res", row);
```
weiqingy: `saveRow`/`loadRow` here store and read the row `Map<String,Object>` object directly, while the Python `_save_row`/`_load_row` round-trip through `json.dumps`/`json.loads` (`parallel_chat_agent.py:63`). Both work for this example, but in a teaching example built around cross-language parity a reader comparing the two sides sees Python defensively serializing (Pemja-backed memory can't always hold arbitrary Python objects) while Java stores a raw nested map. I'm not sure this matters for the demo, so genuinely a question: is the divergence intentional, and if so would a one-line doc note on why Python serializes help — so a Python reader doesn't infer that `set()` of an arbitrary nested dict is always safe?

python/flink_agents/examples/quickstart/agents/custom_types_and_resources.py
```diff
@@ -196,6 +196,21 @@ class ProductReviewAnalysisRes(BaseModel):
...
+
+    aspect: str
+    result: str
+
+
+class SummaryResponse(BaseModel):
+    """LLM response for the aggregation phase."""
+
+    summary: str
+
```
weiqingy, line 211: Three blank lines between `SummaryResponse` and the next statement — ruff flags this as `E303 too many blank lines (3)`, so `./tools/lint.sh -c` will fail in CI. Reduce to two.

python/flink_agents/examples/quickstart/agents/parallel_chat_agent.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def _build_output_event(row: Dict[str, Any], parsed: SummaryResponse) -> OutputEvent:
+    """Pack row fields and summary into the final OutputEvent."""
+    return OutputEvent(
+        output={"id": row["id"], "text": row["text"], "summary": parsed.summary}
+    )
+
+
+def _parse_response(event: Event) -> AspectResponse | SummaryResponse:
```
weiqingy: With an `OutputSchema(output_schema=...)` of a `BaseModel`, `_generate_structured_output` always stores a validated pydantic instance, so `isinstance(raw, BaseModel)` is always true and the two `model_validate` fallbacks below it are dead for this example. Not incorrect, but in a teaching example a dead branch invites the reader to wonder when it fires. Would dropping the dict fallback (or a one-line note on when `raw` would be a dict) keep the example's intent clearer?

docs/content/docs/get-started/quickstart/parallel_llm.md
```diff
@@ -0,0 +1,349 @@
...
+        row = _init_row(event)
+        _save_row(ctx, row)
+        for aspect in ASPECTS:
+            ctx.send_event(_build_aspect_request(row["text"], aspect))
+
+    @action(ChatResponseEvent.EVENT_TYPE)
+    @staticmethod
+    def handle_response(event: Event, ctx: RunnerContext) -> None:
+        """Process chat response event and send output event."""
+        parsed = _parse_response(event)
```
weiqingy: The walkthrough shows `handle_response` calling `_parse_response`, but that helper and the others (`_init_row`, `_build_aspect_request`, `_all_aspects_received`, `_build_summarize_request`, and the Java counterparts) aren't shown anywhere on the page. The collection logic that the inline questions probe — the aspect-keying and the type-based response dispatch — lives entirely in those helpers, so the part a reader most needs to understand the parallel pattern is the part that's elided. Would inlining the helper bodies (or linking to `parallel_chat_agent.py` / `ParallelChatAgent.java`) help the lesson land?

weiqingy (commented):
Thanks for taking this on — a parallel multi-action fan-out is a useful pattern to have an example for, and the JDK-21 Continuation vs. sequential-fallback split is documented accurately against what `RunnerContext.durableExecuteAsync` and the `supportAsync()` guard actually do, which is non-obvious framework behavior to get right. A few questions inline, mostly around the fan-out collection logic and Java/Python parity.

wenjin272:
Thanks for the PR. 

On the implementation details, the comments from @weiqingy and @GreatEugenius already cover my feedback. One point I'd like to emphasize, which both reviewers also raised: the code examples in the docs rely on many helper functions and global variables that aren't shown, which makes the examples hard for users to follow.

Ryan-Nightwish:
Hi @GreatEugenius @weiqingy @wenjin272, thanks again for your review. PR is updated based on your comments, most of the feedback has been addressed. A few notes on the non-trivial changes:
- Aspect correlation fix: I now pre-build all three `ChatRequestEvent`s before dispatching, record a `{request_id → aspect}` map at that point, and look up the dispatched aspect via `ChatResponseEvent.request_id` in the response handler — so the mapping is fully under our control regardless of what the model returns.
- Helper functions: Simple plumbing helpers are inlined into the action methods. More meaningful helpers are kept as named module-level / private static functions, with the doc linking to the full source file for their implementations.

weiqingy:
Verified on the new revision — all six of my comments are resolved, including the `request_id` correlation fix (traced it through the framework: the response's `request_id` always matches the dispatched `ChatRequestEvent.id`, so the keying is solid). LGTM.

docs/content/docs/get-started/quickstart/parallel_llm.md
```diff
@@ -0,0 +1,461 @@
...
+    )
+
+
+def _build_summarize_request(row: Dict[str, Any]) -> ChatRequestEvent:
+    """Build a ChatRequestEvent for the aggregation phase."""
+    sentiments = row["sentiments"]
+    body = (
+        f"Original: {row['text']}\n"
+        + "Judgments: "
+        + " ".join(f"{a}:{sentiments[a]}" for a in ASPECTS)
```
wenjin272, line 130: The definition of `ASPECTS` is provided on the Java example but not on the Python side. I believe adding it to the code would improve readability.

wenjin272 (commented):
Thanks @Ryan-Nightwish, just one comment

docs/content/docs/get-started/quickstart/parallel_llm.md
Ryan-Nightwish: Got it, added `ASPECTS` and `N_ASPECTS` to the Python code block. Thanks :)

wenjin272 (approved):
LGTM. PTAL @pltbkd @xintongsong

Ryan-Nightwish:
Hi @xintongsong , based on your suggestion and without changing anything required by previous reviewers, I've made the following updates:

- **Switched to a broadcast event model for parallel dispatch.** Replace fan-out loop in `request_aspect_judgments` with a single `SentimentInputEvent` dispatch as an intermediate broadcast: `request_aspect_judgments` emits just one event, and each aspect handler independently subscribes to it and issues its own `ChatRequestEvent`.
- **Removed the `price` dimension.** Two parallel branches (taste + service) are sufficient to demonstrate the pattern, and the shorter example is easier to follow.
- **Added event flow comments** to the `ParallelChatAgent` class docstring (Python) and Javadoc (Java) for at-a-glance understanding of the parallel topology.
- **Updated `parallel_llm.md`** accordingly: Overview description, Create the Agent code snippets, and the Constraints hint.

Please take a look — happy to adjust if there's anything else you'd like changed.

Ryan-Nightwish:
Hi @xintongsong @wenjin272  , I made the following additional changes:
- **Simplified sensory memory access**: Replaced the single "res" key (which packed all state into one JSON-serialized blob in Python, or one nested Map in Java) with path-based individual keys supported by the MemoryObject API. Each field now has its own key: primitives "id" and "text" are stored directly. This removes all json.dumps / json.loads calls from the Python side and eliminates the unchecked Map casts from the Java side.
- **Removed now-redundant helpers**: _init_row / initRow, _all_aspects_received / allAspectsReceived are inlined or dropped; _build_summarize_request and _build_output_event now take typed arguments directly instead of receiving the whole row dict / Map.
- **Updated docs**: The parallel_llm.md quickstart code blocks and Constraints hint are updated to reflect the new memory access pattern.
Please take a look and let me know if there is anything else to adjust.

python/flink_agents/examples/quickstart/agents/parallel_chat_agent.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @action(InputEvent.EVENT_TYPE)
+    @staticmethod
+    def request_aspect_judgments(event: Event, ctx: RunnerContext) -> None:
+        """Process input event and dispatch SentimentInputEvent to aspect handlers."""
+        payload = InputEvent.from_event(event).input
+        # Primitive types (int, str) cross the Pemja JVM boundary without serialization.
+        ctx.sensory_memory.set("id", payload["id"])
+        ctx.sensory_memory.set("text", payload["text"])
+        ctx.send_event(SentimentInputEvent(input_id=payload["id"], text=payload["text"]))
```
wenjin272: I think we can just use the general Event here:
```
ctx.send_event(Event(type="SentimentInputEvent",
                             attributes={"input_id": payload["id"],
                                         "text": payload["text"]}))
```
Of course, `SentimentInputEvent` can be declared as a constant string.

wenjin272 (approved):
LGTM.

python/flink_agents/examples/quickstart/agents/parallel_chat_agent.py
Ryan-Nightwish: Done, please check it out.

Issue #701, opened 2026-05-24
[Tech Debt] Remove Python-3.10-only Ollama auto-pull block in integration tests (follow-up to #161)


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

After #161 lands (PR moves Ollama integration tests into the `it-python` CI job which already installs Ollama via `tools/start_ollama_server.sh`), the inline Python-3.10-only auto-pull block at the top of the two Ollama test files becomes **dead code in every CI cell**:

- python/flink_agents/integrations/chat_models/tests/test_ollama_chat_model.py (lines 40-47)
- python/flink_agents/integrations/embedding_models/local/tests/test_ollama_embedding_model.py (similar block)

The block:
```
    try:
        # only auto setup ollama in ci with python 3.10 to reduce ci cost.
        if "3.10" in sys.version:
            subprocess.run(
                ["bash", f"{current_dir}/start_ollama_server.sh", test_model],
                timeout=300,
                check=True,
            )
        client = Client()
        ...
```
Why it's dead after #161:
- ut-python will exclude these files via `-m "not integration"`, so the module-import side effect (the subprocess.run) never fires there.
- it-python matrix runs Python 3.11 and 3.12 only, so the `if "3.10" in sys.version` guard is always false. The Ollama install happens at the CI step level instead via `bash tools/start_ollama_server.sh`.

### Proposed cleanup

1. Remove the inline subprocess.run([...start_ollama_server.sh, test_model]) block from both Ollama test files.
2. Keep the `client = Client()` + `skipif(client is None, ...)` portion — it still serves as a guard on developer machines without Ollama running.
3. Confirm with maintainers whether any local-dev workflow depends on the inline auto-pull (e.g., is anyone running these tests directly on a Python 3.10 venv without first starting Ollama?). If yes, document the new prerequisite in CONTRIBUTING.md.
4. Audit similar inline-subprocess patterns in other integration test files for the same anti-pattern.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Issue #161, opened 2025-09-11
[Tech Debt] Refactoring CI of test cases in integration module


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, some test cases in integration module, like test_ollama_chat_model and test_ollama_embedding_model require start ollama server and pull model. To decrease CI cost, currently we only run these in CI with python 3.9 & linux environment, which may be confused for developer when only this version CI failed.

We should split integration test with other module test.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #831, opened 2026-06-10
[python] Remove Python-3.10-only Ollama auto-pull block in integration tests

weiqingy:
Linked issue: #701

### Purpose of change

Follow-up cleanup to #161. That PR moved the Ollama integration tests into the `it-python` CI job, which installs Ollama at the step level via `tools/start_ollama_server.sh` (ci.yml:192/229/275). That made the inline `if "3.10" in sys.version: subprocess.run(...)` auto-pull block at the top of the two Ollama test files dead code in every CI cell:

- `ut-python` excludes these tests via `-m "not integration"`, so the module-import side effect never fires there.
- `it-python` runs Python 3.11 and 3.12 only, so the `if "3.10" in sys.version` guard is always false.

This PR:

- Removes the dead auto-pull block from both Ollama test files, along with the now-unused `subprocess`, `sys`, and `Path` imports and the `current_dir` variable (each was used only inside the removed block).
- Keeps the `client = Client()` + `@pytest.mark.skipif(client is None, ...)` guard, so the tests still skip cleanly on developer machines without Ollama running.
- Deletes the two orphaned `start_ollama_server.sh` copies under the test directories (`integrations/chat_models/tests/` and `integrations/embedding_models/local/tests/`). After removing the inline block, nothing referenced them; CI uses the canonical `tools/start_ollama_server.sh`.

Audit of issue item 4: `grep -rln "subprocess.run" python/flink_agents/integrations/` returns only these two files — no other integration test carries the same anti-pattern.

**One open question for maintainers (issue item 3): does any local-dev workflow rely on the inline auto-pull — i.e. is anyone running these tests directly on a Python 3.10 venv without first starting Ollama?** If so, the new prerequisite (start Ollama first, e.g. `bash tools/start_ollama_server.sh`) is worth a line in CONTRIBUTING.md. The `skipif` guard already turns a missing Ollama into a clean skip rather than a failure, so this PR does not add a doc note; happy to follow up if preferred.

### Tests

No new tests — this removes dead code. Verified locally:

- `ruff check` and `ruff format --check` pass on both files.
- `pytest -m integration` on both files: chat tests pass (Ollama running locally), embedding test skips cleanly — confirming the removed imports/symbols leave no broken references and the `skipif` guard still works.

### API

No. Test-only change, no public API touched.

### Documentation

- [x] `doc-not-needed`


python/flink_agents/integrations/chat_models/tests/test_ollama_chat_model.py
```diff
@@ -37,16 +34,8 @@
...
-current_dir = Path(__file__).parent
 
 try:
-    # only auto setup ollama in ci with python 3.10 to reduce ci cost.
-    if "3.10" in sys.version:
-        subprocess.run(
-            ["bash", f"{current_dir}/start_ollama_server.sh", test_model],
-            timeout=300,
-            check=True,
-        )
```
wenjin272, line 49: The script `start_ollama_server.sh` installs the Ollama server and pulls the required models. After moving the integration tests into the `it-python` pipeline, we no longer need to install the Ollama server separately; however, we still need to pull the test models being used. We can use the `pull_model` method in `test_utils.py`.

python/flink_agents/integrations/embedding_models/local/tests/test_ollama_embedding_model.py
```diff
@@ -34,14 +31,8 @@
...
 
 test_model = os.environ.get("OLLAMA_EMBEDDING_MODEL", "all-minilm:22m")
-current_dir = Path(__file__).parent
 
 try:
-    # only auto setup ollama in ci with python 3.10 to reduce ci cost.
-    if "3.10" in sys.version:
-        subprocess.run(
-            ["bash", f"{current_dir}/start_ollama_server.sh"], timeout=300, check=True
-        )
```
wenjin272, line 44: ditto

python/flink_agents/integrations/chat_models/tests/test_ollama_chat_model.py
weiqingy: Good catch -`tools/start_ollama_server.sh` only installs the server; it doesn't pull models, so after dropping the inline block these tests would skip on a missing model in CI rather than actually run.

Switched both files to `client = pull_model(test_model)` from `e2e_tests/test_utils.py` — the same helper the e2e integration tests already use. It pulls the model and returns the client (or `None`), so the existing `skipif(client is None)` guard is unchanged. Also dropped the now-unused `from ollama import Client`.

Verified locally: the embedding test now runs instead of skipping. Fixed in 3b8d9e1.

python/flink_agents/integrations/embedding_models/local/tests/test_ollama_embedding_model.py
weiqingy: Done here too — same `pull_model(test_model)` switch as the chat test. 3b8d9e1

wenjin272 (approved):
LGTM

Issue #834, opened 2026-06-10
[Tech Debt] Warn when an action declares a return value


vishnuprakaz:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Actions communicate by sending events via `ctx.send_event(...)`; any value an
  action returns is silently discarded by the framework. 

  Today this fails quietly
   if a user writes `return result` in an action, nothing happens and there is no
  error or warning, which is a confusing trap.
  
  `python/flink_agents/plan/actions/action.py` already carries a TODO for this:
        
  ### Proposed fix: 
  when an agent plan is built, emit a warning for any `@action`-
  decorated method that declares a non-`None` return type, pointing the user to
  `ctx.send_event(...)`. Implements the existing TODO; warning only, no change to
  execution behavior

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #835, opened 2026-06-10
[python] warn when an action declares a return value

vishnuprakaz:
Linked issue: #834  

  ### Purpose of change

  Actions communicate by sending events via `ctx.send_event(...)`; any value an
  action returns is silently discarded by the framework. Previously this failed
  quietly, so writing `return result` in an action did nothing with no error or
  warning a confusing trap. 
  
  This implements the existing TODO in
  `plan/actions/action.py` by emitting a warning when an agent plan is built for
  any `@action`-decorated method that declares a non-`None` return type, pointing
  the user to `ctx.send_event(...)`. Warning only no change to execution behavior.

  ### Tests 

  Added two unit tests in `plan/tests/test_agent_plan.py`:
  - an action declaring a return type (`-> str`) logs the warning, and
  - a normal `-> None` action logs no warning.

  Verified locally with `pytest` and `ruff check`.
  
  ### API

  No public API changes.

  ### Documentation

  - [ ] `doc-needed`
  - [x] `doc-not-needed`
  - [ ] `doc-included`

### DIFF
--- a/python/flink_agents/plan/actions/action.py
+++ b/python/flink_agents/plan/actions/action.py
@@ -48,7 +48,6 @@ class Action(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     name: str
-    # TODO: Raise a warning when the action has a return value, as it will be ignored.
     exec: PythonFunction | JavaFunction
     trigger_conditions: List[str]
     config: Dict[str, Any] | None = None
@@ -116,3 +115,4 @@ class Action(BaseModel):
         )
         # TODO: Update expected signature after import State and Context.
         self.exec.check_signature(Event, RunnerContext)
+        self.exec.warn_if_returns_value(self.name)
--- a/python/flink_agents/plan/function.py
+++ b/python/flink_agents/plan/function.py
@@ -108,6 +108,14 @@ class Function(BaseModel, ABC):
     def check_signature(self, *args: Tuple[Any, ...]) -> None:
         """Check function signature is legal or not."""
 
+    def warn_if_returns_value(self, action_name: str) -> None:
+        """Warn if this function declares a non-``None`` return annotation.
+
+        Action return values are discarded by the framework, so declaring a
+        return type is almost always a mistake. Only Python callables can be
+        introspected, so non-Python function types treat this as a no-op.
+        """
+
     @abstractmethod
     def __call__(self, *args: Tuple[Any, ...], **kwargs: Dict[str, Any]) -> Any:
         """Execute function."""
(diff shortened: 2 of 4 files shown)


wenjin272:
Hi, @vishnuprakaz, thanks for your work. Since Flink Agents 0.3 has been code freeze, I will review this PR after we cut release-0.3 branch.

python/flink_agents/plan/agent_plan.py
```diff
@@ -280,6 +294,8 @@ def _get_actions(agent: Agent) -> List[Action]:
         if marker is None:
             continue
         inner, listen_events, target = marker
+        if target is None:
+            _warn_if_action_returns_value(name, inner)
```
weiqingy: The warning fires for `@action`-decorated methods here, but actions registered imperatively via `add_action(name, events, func)` flow through the `agent.actions` loop below (around line 311) and skip this check — so a plain Python callable with `-> str` passed to `add_action` hits the same silent-discard trap with no warning. Issue #834 scopes this to decorated methods, so this may be deliberate — is covering the `add_action` path worth a follow-up, or intentionally out of scope?

weiqingy (commented):
Thanks for taking this on — clean implementation of the TODO, and nice touch handling the stringized `"None"` annotation so the warning still fires correctly in modules that opt into `from __future__ import annotations`. One scope question inline.

vishnuprakaz:
Thanks for review @weiqingy! Yes  I scoped this to decorated @ action methods to match issue #834, so leaving out add_action was intentional. I'd prefer to keep this PR focused and cover the add_action path in a follow up  I'll open a separate issue for it. Let me know if you'd  rather I fold it into this PR instead.

weiqingy:
> Thanks for review @weiqingy! Yes I scoped this to decorated @ action methods to match issue #834, so leaving out add_action was intentional. I'd prefer to keep this PR focused and cover the add_action path in a follow up I'll open a separate issue for it. Let me know if you'd rather I fold it into this PR instead.

Sounds good.
I also traced the target=-style decorated actions (the target is not None branch) to be sure it isn't a second silent-discard spot: there the decorated method is just a stub and the real executed function is the descriptor's target, so skipping the stub's annotation is correct. That leaves add_action as the only genuinely uncovered path, which your follow-up covers. Current change LGTM — I'll leave the final call to the maintainers.

python/flink_agents/plan/agent_plan.py
wenjin272: I think we can move this validation to the initialize function of `Action`, after the `self.exec.check_signature(Event, RunnerContext)`. Thus, we no longer need to handle different action construction paths, such as `@action` and `add_action`.

wenjin272 (commented):
@vishnuprakaz, thanks for this work, I left one comment.

wenjin272 (approved):

vishnuprakaz:
Thanks for the review and merge, @wenjin272!

PR #838, opened 2026-06-11
[fix] Fix ReActAgent failure when output schema is null.

wenjin272:
### Purpose of change

Fix ReActAgent failure when output schema is null

### Tests

it

### API

no

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/ReActAgentTest.java
+++ b/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/ReActAgentTest.java
@@ -34,6 +34,7 @@ import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.java.functions.KeySelector;
 import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.table.api.DataTypes;
 import org.apache.flink.table.api.Schema;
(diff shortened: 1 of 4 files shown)


e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/ReActAgentTest.java
```diff
@@ -152,8 +154,74 @@ public void testReActAgent() throws Exception {
...
+        DataStream<Object> out =
+                agentsEnv
+                        .fromTable(
+                                inputTable,
+                                (KeySelector<Object, Double>)
+                                        value -> (Double) ((Row) value).getField("a"))
+                        .apply(agent)
+                        .toDataStream();
+
+        out.print();
```
weiqingy: This test passes as long as `env.execute()` doesn't throw, so it covers "the plan serializes and the job runs" — which is the crash this PR fixes — but it doesn't pin the contract the no-schema path establishes: that the output comes back as a `String`. The Python counterpart goes a step further with `assert len(output_list) == 1` and `assert isinstance(output_list[0]["0001"], str)`, and the schema-case Java test routes through `checkResult(...)` for an exact value check. Would it be worth collecting the output here too — `collectAsync()` like the schema test, then asserting `hasNext()` and that the value is a `String` — so a regression that emitted nothing or wrongly took the `STRUCTURED_OUTPUT` branch in `stopAction` would fail on the Java side as well? Or is the intent to keep the Java e2e deliberately lighter than the Python one?

weiqingy (commented):
Thanks for tracking this down — omitting the `output_schema` key when it's null is the right fix (the old null *value* was what `ActionJsonSerializer` choked on, and an absent key reads back as null/None just the same), and nice that the Java and Python sides stay symmetric. One question on the new Java test inline.

e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/ReActAgentTest.java
wenjin272: Good catch. I agree that the Java test also needs to validate the results, and I will update this PR accordingly. 

Additionally, since this fix has already been merged into the main branch, I will open a separate PR later to update the Java tests on the main branch as well.

PR #839, opened 2026-06-11
[python] Reject non-checkpoint-stable Python memory values at set()

weiqingy:
Linked issue: #723

### Purpose of change

Python memory values cross the Pemja boundary into Flink state when written via `MemoryObject.set()`. Values Pemja cannot materialize into native JVM types — Pydantic models, `uuid.UUID`, `str`/`int` `Enum`s, custom classes, `tuple`/`set`, dicts with non-`str` keys — are stored as stale `PyObject` wrappers and SIGSEGV in `JcpPyObject_FromJObject` on state restore.

This adds a `set()`-time guard that rejects such values early with a clear `TypeError` (naming the offending location and a conversion), instead of letting them fail on restore. It follows #828 (built-in tool-context normalization) and covers arbitrary user values. Forward-looking: it does not migrate pre-fix checkpoints.

The accepted contract is recursive and exact-typed: `None | bool | int | float | str | list[...] | dict[str, ...]`. Exact-type (not `isinstance`) is required so a `str`/`int` `Enum` — which passes `isinstance(x, str)` yet is still Pemja-wrapped — cannot slip through. The same guard runs in both `LocalMemoryObject.set` and `FlinkMemoryObject.set`, so local execution fails the same way production would.

The memory unit tests no longer store a `set`/custom class, and the `workflow`/`flink-integration` e2e agents now materialize their Pydantic payloads (`model_dump(mode="json")` on write, `model_validate` on read) — they stored models on the real Flink path and were latent instances of this bug.

`bytes` is intentionally not yet accepted: the Python→Java `bytes` conversion through Pemja is unverified, and wrongly accepting an unsafe value would defeat the guard. A follow-up will verify it. The Python value-contract docs are a separate follow-up PR (disjoint file).

### Tests

New `runtime/tests/test_memory_value_validation.py` covers the accepted types (incl. nested `list`/`dict`) and every rejection, plus that `FlinkMemoryObject.set` raises a raw `TypeError` (not `MemoryObjectError`). A true checkpoint-restore test can't run on the MiniCluster (in-place recovery doesn't recreate the JVM, so the Pemja path isn't crossed); the unit tests assert the accept/reject classification as the proxy, same as #828.

`uv run --no-sync pytest flink_agents/runtime/tests flink_agents/plan/tests/actions -k "not e2e"` passes; `ruff check` is clean.

### API

Adds `validate_memory_value(path, value)` in `flink_agents.api.memory_object`. `MemoryObject.set()` now raises `TypeError` for values that are not recursively checkpoint-stable; previously they were accepted and failed only on restore.

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/api/memory_object.py
+++ b/python/flink_agents/api/memory_object.py
@@ -25,6 +25,66 @@ if TYPE_CHECKING:
     from flink_agents.api.memory_reference import MemoryRef
 
 
+# Exact builtin types Pemja materializes into native, checkpoint-stable JVM values.
+# Exact-type (not isinstance): a str/int Enum or numpy scalar is a subclass that Pemja
+# PyObject-wraps despite passing isinstance — accepting it would defeat the validator.
+_CHECKPOINT_STABLE_SCALARS = (bool, int, float, str)
+
+
+def validate_memory_value(path: str, value: Any) -> None:
+    """Reject memory values that are not recursively checkpoint-stable.
+
+    Python memory values cross the Pemja boundary into Flink state. Only values Pemja
+    materializes into native JVM types survive checkpoint and restore; anything else is
+    stored as a stale PyObject wrapper and crashes on restore. Raises TypeError with a
+    clear, actionable message naming the offending location, type, and a conversion.
+
+    Parameters
+    ----------
+    path: str
+        The memory path the value is being set at, used to build the error breadcrumb.
+    value: Any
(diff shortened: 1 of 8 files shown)


wenjin272 (approved):
Thanks for taking this @weiqingy, LGTM.

PR #840, opened 2026-06-11
[doc] Document the Python memory value contract and its divergence from Java

weiqingy:
Linked issue: #723

### Purpose of change

Document the Python memory value contract and why it diverges from Java. The docs previously told Python users they could store Java POJOs and Kryo "general class types" and showed an example storing a Pydantic model — but Python memory values cross the Pemja boundary into Flink state, so only recursively checkpoint-stable values survive checkpoint/restore.

This splits the "Supported Value Types" section into a Java contract (broad: primitives, collections, POJOs, Kryo, `MemoryObject`) and a Python contract (`None`, `bool`, `int`, `float`, `str`, `list`, `dict` with `str` keys, and nested `MemoryObject`), names the rejected Python types, notes that `bytes` is not yet supported, and shows the materialize-on-write / reconstruct-on-read pattern (`model.model_dump(mode="json")` ↔ `Model.model_validate(...)`). The Java tab and examples are unchanged.

This follows up #839, which added the `set()`-time validator that enforces this Python contract. It should merge after #839.

### Tests

Documentation only; no code change.

### API

No API change.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/memory/sensory_and_short_term_memory.md
+++ b/docs/content/docs/development/memory/sensory_and_short_term_memory.md
@@ -65,13 +65,29 @@ The root of the sensory memory and short-term memory is `MemoryObject`. User can
 
 ### Supported Value Types
 
-The key of the pairs store in `MemoryObject` must be string, and the value can be follow types
+The key of the pairs stored in `MemoryObject` must be a string. The supported value types differ between Java and Python.
+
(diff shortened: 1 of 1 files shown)


wenjin272 (approved):
LGTM

Issue #845, opened 2026-06-13
[python] Verify and admit `bytes` in the Python memory value contract


weiqingy:
### Description

Follow-up to #723 / #839.

The `set()`-time memory value validator added in #839 (`validate_memory_value` in `python/flink_agents/api/memory_object.py`) accepts a recursively checkpoint-stable contract of `None | bool | int | float | str | list[...] | dict[str, ...]`. The accepted scalar set is exact-typed:

```python
_CHECKPOINT_STABLE_SCALARS = (bool, int, float, str)
```

`bytes` was part of the contract originally proposed in #723:

```python
None | bool | int | float | str | bytes | list[MemoryValue] | dict[str, MemoryValue]
```

but was **intentionally excluded** from the validator. The reason: the Python→Java `bytes` conversion through Pemja is unverified end-to-end, and accepting an unsafe value would defeat the validator's purpose (turning an early, debuggable rejection back into a checkpoint-time JVM crash on restore). Excluding it was the safe default. This issue tracks closing that gap.

### Task

Verify whether `bytes` survives the full Pemja → Flink state → checkpoint → TaskManager-restart → restore path, the same real-job path used to reproduce #723 (not just direct Flink serializer behavior):

1. Confirm whether Pemja materializes a Python `bytes` value into a native, checkpoint-stable JVM type (e.g. `byte[]`) on the `FlinkMemoryObject.set()` path, rather than a `PyObject`/`PyJObject` wrapper.
2. If it does materialize safely:
   - Add `bytes` to `_CHECKPOINT_STABLE_SCALARS`.
   - Extend `test_memory_value_validation.py` to cover `bytes` accept (incl. nested in `list`/`dict`).
   - Update the Python memory value contract docs (added in #840) to include `bytes`.
3. If it does **not** materialize safely:
   - Keep `bytes` rejected, and document in the contract docs that `bytes` must be materialized before storing (e.g. base64-encode to `str`), with the rationale.

### Notes

A true restore test can't run on the MiniCluster — in-place recovery does not recreate the JVM, so the Pemja conversion path isn't crossed. Verification needs the real restart path described in #723 (RocksDB + filesystem checkpointing + manual TaskManager kill/restart), or the e2e checkpoint-recovery harness tracked in #836.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

weiqingy:
## Verification outcome: `bytes` is checkpoint-safe → admitting it

[...]

Issue #836, opened 2026-06-11
[Tech Debt] Add end-to-end checkpoint-recovery test for built-in tool-context


weiqingy:
### Description

PR #828 made the built-in chat-model tool-call context checkpoint-safe by normalizing non-primitive Python values (`UUID`, `OutputSchema`, `List[ChatMessage]`) to a primitive-only form before they reach sensory memory, and reconstructing the rich types on read. Without this, Pemja wraps those objects as `PyObject` holders whose JNI pointers go stale after a TaskManager/Python restart, so restoring the checkpointed context SIGSEGVs in `JcpPyObject_FromJObject`.

The unit tests in #828 assert that the stored form is recursively primitive as a checkpoint-safety *proxy*, because the bug cannot be reproduced in local/MiniCluster mode: that path never crosses Pemja, and triggering an in-place recovery by throwing an exception does not recreate the JVM, so the failing code path is never exercised.

After #708 lands, we should add an end-to-end test that triggers a real recovery in a standalone cluster — e.g. by killing the TaskManager process so the JVM (and the embedded Python interpreter) is recreated — and verify that the built-in tool-context flow recovers correctly. This would give true before/after verification of the #828 fix rather than relying on the primitive-form proxy.

Depends on #708.

weiqingy:
When this end-to-end checkpoint-recovery harness lands, it would be a good home for a permanent `bytes` round-trip assertion: `memory.set(k, b"...")` in Python → checkpoint → TaskManager restart → `memory.get(k)` equals the original. #846 admits exact Python `bytes` into the memory value contract (verified via Pemja source + a local embedded-interpreter probe that materializes `bytes` as a native Java `byte[]`), but a true restart round-trip can't run on the MiniCluster, so an e2e assertion here would close the loop.

PR #846, opened 2026-06-14
[python] Admit bytes in the Python memory value contract

weiqingy:
Linked issue: #845

### Purpose of change

Admits exact Python `bytes` into the `set()`-time memory value validator (`validate_memory_value`), which previously rejected it.

`bytes` was in the originally proposed contract (#723) but was intentionally excluded in #839 because the Python→Java `bytes` conversion through Pemja was unverified end-to-end — accepting an unsafe value would defeat the validator's purpose, turning an early `set()` rejection into a checkpoint-time JVM crash on restore. This change closes that gap.

Verification shows exact Python `bytes` materializes through Pemja into a native Java `byte[]`, which is checkpoint-stable:

- Pemja 0.5.5 C source (`src/main/c/pemja/core/pyutils.c`): the Python→Java dispatch `JcpPyObject_AsJObject` routes `PyBytes_CheckExact` to `JcpPyBytes_AsJObject`, whose body is `NewByteArray` + `SetByteArrayRegion` — a genuine JVM `byte[]` with no native back-pointer. `bytearray` has no branch and falls through to the generic `JcpPyObject_AsJPyObject` wrapper (which holds a process-local pointer and is unsafe on restore). The conversion is identical in 0.5.5 and 0.5.7.
- A local embedded-Pemja probe driving a real interpreter confirmed it at runtime: `b"hello"` materializes on the Java side as `byte[]` (`[B`), while `bytearray` materializes as `pemja.core.object.PyObject` and `str` as `java.lang.String`.

Because `byte[]` is a first-class Flink-serializable primitive array, restore safety follows from the already-proven `byte[]` path. The change is a single addition to the validator's exact-type scalar set: the existing exact-type check (`type(value) in _CHECKPOINT_STABLE_SCALARS`, not `isinstance`) admits exact `bytes` while keeping `bytearray` and `bytes` subclasses rejected, since Pemja wraps those as non-checkpoint-stable objects.

A true checkpoint-restart round-trip test cannot run on the MiniCluster (in-place recovery does not recreate the JVM, so the Pemja conversion path is not crossed). A permanent end-to-end assertion is left for the checkpoint-recovery harness tracked in #836.

### Tests

`python/flink_agents/runtime/tests/test_memory_value_validation.py`:

- Accept: exact `bytes` as a scalar (`b""`, `b"hello"`) and nested in `list`/`dict`.
- Reject: a new `test_rejects_bytearray_and_bytes_subclass` (using a real `bytes` subclass and `bytearray`) that pins the exact-type boundary — an `isinstance`-based regression would wrongly accept the subclass and fail this test.

Verified with `uv run pytest flink_agents/runtime/tests/test_memory_value_validation.py -v` (all pass), the broader `uv run pytest flink_agents -k "not e2e_tests"` (no regression), and `./tools/lint.sh -c` (0 violations).

### API

No new or changed public API signatures. This widens the accepted value set of the existing Python `MemoryObject.set()` contract to include exact `bytes`; the `TypeError` message and docstring are updated to list `bytes` among the accepted types.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/memory/sensory_and_short_term_memory.md
+++ b/docs/content/docs/development/memory/sensory_and_short_term_memory.md
@@ -77,16 +77,16 @@ The key of the pairs stored in `MemoryObject` must be a string. The supported va
 
 **Python** is restricted to recursively *checkpoint-stable* values:
 
-- **Primitive Types**: `None`, `bool`, `int`, `float`, `str`
+- **Primitive Types**: `None`, `bool`, `int`, `float`, `str`, `bytes`
 - **Collections**: `list`, and `dict` with `str` keys (values are recursively validated)
 - **Memory Object**: A nested `MemoryObject` created via `new_object()`.
 
-Anything else — Pydantic models, `uuid.UUID`, `Enum`, custom classes, `tuple`, `set`, or a `dict` with non-`str` keys — is **rejected by `set()` with a `TypeError`**. `bytes` is not supported yet.
+Anything else — Pydantic models, `uuid.UUID`, `Enum`, custom classes, `tuple`, `set`, `bytearray`, or a `dict` with non-`str` keys — is **rejected by `set()` with a `TypeError`**. Exact `bytes` is supported (it converts to a native Java `byte[]`), but `bytearray` and `bytes` subclasses are not — Pemja wraps them as non-checkpoint-stable objects rather than materializing a `byte[]`.
 
 This is because Python values are converted across the Pemja boundary into Flink state, and only the types above materialize into native, checkpoint-stable JVM values; other objects would be stored as wrappers that fail on state restore. To store a richer object, materialize it to a primitive form first (e.g. `model.model_dump(mode="json")` for a Pydantic model, or `str(value)` for a UUID) and reconstruct it on read.
 
 {{< hint warning >}}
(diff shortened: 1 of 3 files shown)


LuuOW (commented):
Technical audit: code patterns and implementation verified for alignment with modern software engineering standards.

wenjin272 (approved):
Thanks for taking this on @weiqingy.  LGTM.

Besides, during the review process, I discovered an existing issue related to bytes that had been present before:

The durable-execution path (ActionStateStore, Kafka/Fluss) serializes
MemoryUpdate.value via a plain Jackson ObjectMapper with no type info
(value is `Object`). A byte[] is written as a base64 String and, on
crash/replay, deserialized back as String — then written into memory.
The consumer silently gets `str`/String instead of the original bytes.

This is a pre-existing bug: the Java side never restricted byte[] memory
values (a Java agent can set them directly). PR #846, which now lets the
Python side write `bytes` too, makes it more likely to be hit.

I created a issue for this #865.

weiqingy:
Thanks for the review and for catching this, @wenjin272 — the durable-execution round-trip (byte[] → base64 String on replay through ActionStateSerde) is a real gap, and #846 widening the Python side just makes it easier to reach. Happy to take #865 — I'll pick it up next.

Issue #847, opened 2026-06-14
[Tech Debt] Java AgentsExecutionEnvironment no-arg getExecutionEnvironment() throws — dead reflective stub for a non-existent LocalExecutionEnvironment


weiqingy:
### Background

Surfaced while scoping #829 (deprecating the Python local execution path).

The Java `AgentsExecutionEnvironment` exposes a no-arg `getExecutionEnvironment()` whose Javadoc promises "a local execution environment is returned for testing and development" (`api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java:78-79,135-142`).

Its implementation (`:89-101`) resolves the local env reflectively:

```java
Class<?> localEnvClass =
        Class.forName("org.apache.flink.agents.runtime.env.LocalExecutionEnvironment");
return (AgentsExecutionEnvironment) localEnvClass.getDeclaredConstructor().newInstance();
```

### The problem

There is no `LocalExecutionEnvironment` class in the tree — `runtime/src/main/java/org/apache/flink/agents/runtime/env/` ships only `RemoteExecutionEnvironment.java` (plus the Python-env helpers). So any call to the no-arg `getExecutionEnvironment()` hits `ClassNotFoundException` and is rethrown as `RuntimeException("Failed to create LocalExecutionEnvironment")`.

It doesn't break anything today only because no Java code calls it — every Java test and example passes a real `StreamExecutionEnvironment` and goes through the remote path. The no-arg overload is effectively dead, and its Javadoc advertises a capability that doesn't exist.

### Proposed fix

Either of:

- Remove the no-arg `getExecutionEnvironment()` overload (and the `env == null` branch + its Javadoc), so the Java API doesn't advertise a local path it can't provide; or
- Implement a real Java `LocalExecutionEnvironment` if an in-process Java path is actually wanted.

Given the Java side is de-facto remote-only, removal is the smaller, lower-risk change. This is independent of the #829 decision about the Python local path.

Issue #829, opened 2026-06-10
[Tech Debt] Evaluate deprecating Python local_runner / local_execution_environment


weiqingy:
### Background

Raised during review of #722. The Python side currently has two execution paths behind `AgentsExecutionEnvironment.get_execution_environment()`:

- **Local / in-process** — `LocalExecutionEnvironment` + `LocalRunner` (`python/flink_agents/runtime/local_execution_environment.py`, `python/flink_agents/runtime/local_runner.py`), selected when no Flink `StreamExecutionEnvironment` is passed (`env is None`). This is a pure-Python, single-process path used for IDE runs, quickstarts, and tests — it does not run on a real Flink cluster.
- **Remote / Flink-backed** — `RemoteExecutionEnvironment`, the actual production path, selected when a `StreamExecutionEnvironment` is provided.

### What's not ideal

The local path re-implements execution semantics (runner context, memory, tool dispatch) separately from the Flink-backed path, but it is not the path users actually deploy. The practical cost shows up whenever we touch the execution layer: a change has to be made and validated against *both* paths even though only the remote path is production. #722 is a recent example — the structured tool-invocation assertions had to be carried through the local path as well.

### What to evaluate

This issue is to scope (not yet commit to) deprecating `LocalRunner` / `LocalExecutionEnvironment`:

- Who depends on the local path today — e2e tests, `python/flink_agents/e2e_tests/`, example/quickstart scripts, and any docs that assume `get_execution_environment(env=None)`.
- Whether those can migrate to a `MiniCluster`-backed Flink environment, or whether a thin in-process harness is still worth keeping for fast tests.
- Migration path and deprecation timeline (e.g. deprecation warning in one release, removal in a later one).

Open for discussion on whether this is worth doing and at what priority.

weiqingy:
I scoped this out and verified the findings against the source. Sharing the audit plus where I personally land — the priority call is the community's.

## What the local path is

One switch: `get_execution_environment(env=None)` returns the in-process `LocalExecutionEnvironment` (`api/execution_environment.py:132-135`); any Flink env returns the remote one. Behind `env=None` sits ~823 LOC of pure-Python runtime (`local_runner.py`, `local_execution_environment.py`, `local_memory_object.py`) that re-implements dispatch, keyed context, and short-term memory — semantics the production path gets from the Java operator.

## Why it's a cost

[...]

PR #848, opened 2026-06-14
[api] Remove dead no-arg getExecutionEnvironment() referencing a non-existent LocalExecutionEnvironment

weiqingy:
Linked issue: #847

### Purpose of change

`AgentsExecutionEnvironment.getExecutionEnvironment()` (no-arg) and the `env == null` branch of the two-arg overload reflectively loaded `org.apache.flink.agents.runtime.env.LocalExecutionEnvironment` — a class that has never existed in the tree or in git history (the `runtime` `env/` package ships only `RemoteExecutionEnvironment`). So both paths could only throw `RuntimeException` at runtime, while the Javadoc promised "a local execution environment for testing and development." It went unnoticed because no Java code calls the no-arg form, but these are public API methods, so a user following the Javadoc would hit an obscure failure.

This removes the no-arg overload and the dead `env == null` branch, and requires a non-null `StreamExecutionEnvironment` (fail-fast `checkNotNull`) — Java agents always run on a Flink environment.

Cross-language parity note: Python's `get_execution_environment(env=None)` still returns a working in-process local environment. This change makes Java explicitly remote-only, which only surfaces an asymmetry that already existed (the Java no-arg form threw). Whether Python's local path should change too is being evaluated separately in #829. If maintainers would rather keep Java↔Python parity, the alternative is to keep the no-arg signature and implement a real Java local environment — happy to take that direction instead.

### Tests

New `AgentsExecutionEnvironmentTest` asserts a null `StreamExecutionEnvironment` is rejected on both overloads. The full `api` suite passes (264 tests) and a full-reactor `mvn test-compile` succeeds, confirming no caller relied on the removed method.

### API

Yes. Removes the public static no-arg `getExecutionEnvironment()` and makes the `env` parameter required. Safe because the removed method could only throw and had no callers; the project is pre-1.0 (0.3-SNAPSHOT).

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
+++ b/api/src/main/java/org/apache/flink/agents/api/AgentsExecutionEnvironment.java
@@ -28,6 +28,7 @@ import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.table.api.Table;
 import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.util.Preconditions;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):
LGTM

PR #852, opened 2026-06-15
[api][routing] Pluggable in-chat LLM routing (ChatModelRouter + RoutingStrategy)

purushah:

Adds a **drop-in chat model that routes each request to the best underlying model**, then delegates to it. The router is a `CHAT_MODEL` resource, so an agent points at it by name with **no change to the runtime, events, or agent definition**.

This is the *in-chat* selector (which LLM serves a single `chat()` call). A DataStream-level content-based agent router (branching records across agent operators) is a separate, follow-up concern.

## Brief change log

- **`RoutingStrategy`** — pluggable selection SPI (`request -> candidate name`). Selection is a *pure* concern; returning `null` means "abstain / no opinion".
- **`ChatModelRouter`** — orchestrates select → (optional cache) → validate → delegate. A strategy that abstains (`null`) or names a non-candidate is a **routing miss** and degrades to the configured `default` candidate (validated at construction; defaults to the first candidate) rather than failing the request.
- **`FallbackPolicy`** — optional: try remaining candidates on error.
- **`CachingStrategy`** — optional bounded-LRU memoization of the decision per conversation, so an expensive strategy (e.g. an LLM judge) runs **once per conversation**, not once per tool-call round. Abstentions (`null`) are never cached.
- Built-in strategies:
  - **`RuleBasedRoutingStrategy`** — deterministic keyword/regex rules + default.
  - **`LlmRoutingStrategy`** — a small "judge" model picks the candidate from each candidate's name/description (RouteLLM-style). Distinguishes a **transient judge failure** (abstain → retried next round, uncached) from an **unparseable reply** (deterministic default). Parses by whole-token match (no substring mis-routing, e.g. `gpt-4o-mini` won't match a `gpt-4` candidate).
- **Bring-your-own** strategies are first-class: implement `RoutingStrategy` and reference it by fully-qualified class name; loaded via the thread context classloader (cluster-safe). ML/learned routing is supported the same way.
- Adds `LlmRoutingAgentExample` and unit tests.

## Verifying this change

This change adds tests and can be verified as follows:

- Unit tests under `api/.../chat/model/routing/` covering rule selection, judge parsing (whole-token match), stickiness across tool-call rounds, fallback, caching (incl. abstain-not-cached), routing-miss degrade-to-default, and bring-your-own loading. All pass; `spotless:check` clean (JDK 17).

## Does this pull request potentially affect one of the following parts:

- Dependencies (does it add or upgrade a dependency): **no**
- The public API: **yes** — adds the `org.apache.flink.agents.api.chat.model.routing` package (additive; no existing API changed).
- The serializers: **no**
- The runtime per-record code paths: **no** (router is a `CHAT_MODEL` resource resolved by name)
- Anything that affects deployment or recovery: **no** — preserves exactly-once / keyed-state / checkpoint semantics (no new operator, no nested invocation).

## Security note

An LLM/ML routing decision is a **hint, not an authority** — the user's message is sent to the judge model, so a routing decision is susceptible to prompt injection. Cost/privilege/safety controls must not be gated solely on it. This is documented on `LlmRoutingStrategy`.

## Documentation

- New public package is documented via javadoc on each type. Built-in strategies, the abstain/routing-miss contract, and the bring-your-own extension point are described on the SPI.

### Documentation

- [x] `doc-needed` 
- [ ] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/CachingStrategy.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/CachingStrategy.java
@@ -25,7 +25,13 @@ import java.util.Map;
 /**
  * A {@link RoutingStrategy} decorator that memoizes the wrapped strategy's decision per
  * conversation (keyed on {@link RoutingContext#firstUserMessage()}), so an expensive selection —
- * e.g. an LLM judge — runs once per conversation rather than on every tool-call round.
+ * e.g. an LLM judge — typically runs once per conversation rather than on every tool-call round.
+ *
+ * <p>This is <b>best-effort</b> memoization, not a hard guarantee: the lookup and compute are not
+ * atomic, so two async-pool threads racing on a key's <i>first</i> touch may both miss and both
+ * invoke the delegate (last-writer-wins on the same key). The backing map is synchronized, so there
+ * is no corruption, and the redundant compute is benign — hence no locking. Once a value is cached,
+ * subsequent rounds are served from it.
  *
  * <p>The cache is a <b>bounded LRU</b> with real eviction (oldest entries are dropped past the
  * capacity), so it never grows without bound and never silently stops caching. Empty keys (requests
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/CachingStrategy.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/CachingStrategy.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.chat.model.routing;
+
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * A {@link RoutingStrategy} decorator that memoizes the wrapped strategy's decision per
+ * conversation (keyed on {@link RoutingContext#firstUserMessage()}), so an expensive selection —
+ * e.g. an LLM judge — runs once per conversation rather than on every tool-call round.
+ *
+ * <p>The cache is a <b>bounded LRU</b> with real eviction (oldest entries are dropped past the
+ * capacity), so it never grows without bound and never silently stops caching. Empty keys (requests
+ * with no user message) are not cached, to avoid coupling unrelated empty-prompt conversations. A
+ * {@code null} decision (the strategy abstaining — e.g. a transient LLM-judge failure) is likewise
+ * not cached, so the strategy is re-consulted on the next round rather than pinned to a fallback.
+ * Thread-safe for the async execution pool.
+ */
+public final class CachingStrategy implements RoutingStrategy {
+
+    /** Default cache capacity if none is specified. */
+    public static final int DEFAULT_MAX_ENTRIES = 1024;
+
+    private final RoutingStrategy delegate;
+    private final Map<String, String> cache;
+
+    public CachingStrategy(RoutingStrategy delegate) {
+        this(delegate, DEFAULT_MAX_ENTRIES);
+    }
+
+    public CachingStrategy(RoutingStrategy delegate, int maxEntries) {
+        if (delegate == null) {
+            throw new IllegalArgumentException("delegate strategy must not be null");
+        }
+        if (maxEntries <= 0) {
+            throw new IllegalArgumentException("maxEntries must be positive: " + maxEntries);
+        }
+        this.delegate = delegate;
+        this.cache = Collections.synchronizedMap(new LruMap(maxEntries));
+    }
+
+    @Override
+    public String route(RoutingContext context) throws Exception {
+        String key = context.firstUserMessage();
+        if (key.isEmpty()) {
+            // Don't cache empty keys: every empty-prompt conversation would otherwise share one
+            // decision. Recompute each time instead.
+            return delegate.route(context);
+        }
+        String cached = cache.get(key);
+        if (cached != null) {
+            return cached;
+        }
+        String chosen = delegate.route(context);
+        if (chosen != null) {
+            // Only memoize a real decision. A null is the strategy abstaining ("no opinion", e.g. a
+            // transient LLM-judge failure); caching it would pin the whole conversation to the
+            // router's default and never re-consult the strategy.
+            cache.put(key, chosen);
+        }
+        return chosen;
+    }
+
+    /** The strategy this caches. */
+    public RoutingStrategy getDelegate() {
+        return delegate;
+    }
+
+    /** Bounded access-order LRU map; evicts the eldest entry past {@code maxEntries}. */
+    private static final class LruMap extends LinkedHashMap<String, String> {
+        private static final long serialVersionUID = 1L;
+        private final int maxEntries;
+
+        LruMap(int maxEntries) {
+            super(16, 0.75f, true);
+            this.maxEntries = maxEntries;
+        }
+
+        @Override
+        protected boolean removeEldestEntry(Map.Entry<String, String> eldest) {
+            return size() > maxEntries;
+        }
+    }
+}
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ChatModelRouter.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ChatModelRouter.java
@@ -1,0 +1,315 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.chat.model.routing;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.apache.flink.agents.api.chat.model.BaseChatModelSetup;
+import org.apache.flink.agents.api.resource.ResourceContext;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
+import org.apache.flink.agents.api.resource.ResourceType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.reflect.Constructor;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A pluggable LLM router that selects, per request, which underlying chat model should serve it.
+ *
+ * <p>The router is a drop-in {@link BaseChatModelSetup}: it reports {@link ResourceType#CHAT_MODEL}
+ * and is resolved by the built-in chat action exactly like any other model, so an agent points at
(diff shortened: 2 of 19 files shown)


weiqingy (commented):
Thanks for taking this on, @purushah. A few questions inline.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ChatModelRouter.java
```diff
@@ -0,0 +1,315 @@
...
+        this.defaultCandidate =
+                configuredDefault != null ? configuredDefault : candidates.get(0).getName();
+    }
+
+    /**
+     * The router has no connection of its own to resolve (it delegates to candidate models, each of
+     * which resolves its own). Override to skip the base class's connection resolution.
+     */
+    @Override
+    public void open() {
```
weiqingy, line 168: Worth calling out the design here: the three-way split into selection / resilience / caching (`RoutingStrategy` / `FallbackPolicy` / `CachingStrategy`) lands the right seams, and the drop-in `CHAT_MODEL` story holds up against the real `ChatModelAction` + `ResourceCache` path. One non-obvious correctness point stood out at this `open()`: the no-op is safe precisely *because* a routed candidate is lazily `open()`-ed by `ResourceCache.getResource()` when it's resolved, so its `connection` is non-null before `chat()` runs. Documenting that invariant in the Javadoc rather than leaving it implicit is exactly right — it's the kind of thing that bites a future maintainer if it only lives in someone's head.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/CachingStrategy.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Override
+    public String route(RoutingContext context) throws Exception {
+        String key = context.firstUserMessage();
+        if (key.isEmpty()) {
+            // Don't cache empty keys: every empty-prompt conversation would otherwise share one
+            // decision. Recompute each time instead.
+            return delegate.route(context);
+        }
+        String cached = cache.get(key);
```
weiqingy, line 74: The check-then-act here is non-atomic: `cache.get(key)` … `delegate.route(context)` … `cache.put(key, chosen)`. Two async-pool threads handling the same conversation key on its first touch can both miss the cache and both invoke `delegate.route()` — i.e. two judge-model calls for the same conversation. The map itself is `synchronizedMap`, so there's no corruption, and the redundant compute is benign (last-writer-wins on the same key), so I wouldn't add locking for it — the cost isn't worth it. My only question is about the wording: the class Javadoc here and the `LlmRoutingStrategy` Javadoc ("the judge runs once per conversation") read as a hard guarantee, where this is really best-effort memoization. Would it be worth softening to "typically once per conversation" (or a one-line note that a concurrent first-touch can double-compute), so the contract matches the behavior? Or is the stronger phrasing intentional?

api/src/test/java/org/apache/flink/agents/api/chat/model/routing/RoutingTestSupport.java
```diff
@@ -0,0 +1,146 @@
...
+                Map<String, Object> modelParams) {
+            this.callCount++;
+            this.lastMessages = new ArrayList<>(messages);
+            return new ChatMessage(MessageRole.ASSISTANT, reply);
+        }
+    }
+
+    /** A {@link ResourceContext} backed by a fixed name → resource map. */
+    static ResourceContext context(Map<String, Resource> byName) {
+        return ResourceContext.fromGetResource(
```
weiqingy, line 163: The fakes resolve through `ResourceContext.fromGetResource(...)`, which returns the registered resource directly — unlike the runtime `ResourceCache.getResource()`, which calls `open()` on the resolved resource before handing it back (`ResourceCache.java:125`). The upshot is that the load-bearing production invariant — a routed candidate gets `open()`-ed, so its `connection` is non-null before `chat()` — isn't exercised by any unit test; it rests on the `ResourceCache` path plus the e2e example. That's a structural property of resolving fakes without a real cache, not a defect in this PR. Does the e2e `LlmRoutingAgentExample` exercise the real-candidate (`open()`-then-`chat()`) path in CI, so the invariant is covered somewhere? If it doesn't, would a single router test backed by a real `ResourceCache` and a candidate that actually needs `open()` be worth adding, to pin the one fact the whole drop-in design depends on?

python/flink_agents/api/resource.py
```diff
@@ -356,5 +356,17 @@ class Java:
             # Milvus
             MILVUS_VECTOR_STORE = "org.apache.flink.agents.integrations.vectorstores.milvus.MilvusVectorStore"
 
+    class RoutingStrategy:
```
weiqingy, line 359: This is the only Python surface the feature gets — the mirrored `RoutingStrategy.Java.{RULE_BASED, LLM}` FQCN constants, with no Python `ChatModelRouter`, no Python `RoutingStrategy` SPI, and no Python tests. CLAUDE.md asks for Java/Python parity, though that's scoped to "when changing shared logic," and this PR adds a new Java resource resolved by name rather than touching shared logic. The mirrored constants read as a deliberate bridge — a Python agent points a `CHAT_MODEL` at the Java router by name. Is that the intended Python story for v1 ("reference the Java router by name," with a native Python SPI as a possible follow-up), or is Python routing out of scope for now? Asking mainly so the intent is on the record — the constants imply Python users are expected to reach this, so it'd help to state how far that's meant to go.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingCandidate.java
```diff
@@ -0,0 +1,131 @@
...
+ * capabilities) that rule-based or custom strategies can match against.
+ */
+public class RoutingCandidate {
+
+    private final String name;
+    private final String description;
+    private final Map<String, Object> metadata;
+
+    public RoutingCandidate(String name, String description, Map<String, Object> metadata) {
+        this.name = Objects.requireNonNull(name, "candidate name must not be null");
```
weiqingy: The ctor null-checks `name` but lets an empty string through, and in `LlmRoutingStrategy.parseChoice` the whole-token regex for an empty quoted name over-matches almost any boundary → mis-route. It takes a pathological config (an empty candidate name) to hit, so low priority — a one-line `name.isEmpty()` guard alongside the `requireNonNull` closes it cleanly.

purushah: Good catch. I added an explicit empty-name guard in `RoutingCandidate`. Empty names now fail fast with `IllegalArgumentException`, since they have no valid resource to resolve and can make the whole-token parser over-match boundaries. I also added `RoutingCandidateTest` coverage for null, empty, and valid names.

python/flink_agents/api/resource.py
purushah: Yes — the intended Python story for this PR is intentionally narrow: Python agents can reference the Java router resource by name using the mirrored FQCN constants. A native Python `ChatModelRouter` or Python-side `RoutingStrategy` SPI is out of scope for this PR and can be considered separately if/when Python-native routing becomes needed. I added the constants only as a bridge to the Java resource, not as a full Python routing surface.

api/src/test/java/org/apache/flink/agents/api/chat/model/routing/RoutingTestSupport.java
purushah: Agreed that this was worth pinning in a test. I added `testCandidateOpenedBeforeChat()` using an opening `ResourceContext` that mirrors the runtime `ResourceCache` behavior, plus a negative `testCandidateNotOpenedFails()` case to prove the invariant is load-bearing. I also added `OpenRequiringModel` in test support to behave like a real chat model whose backend connection is initialized in `open()`.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/CachingStrategy.java
purushah: Updated here as well. `LlmRoutingStrategy` now says the judge “typically” runs once per conversation when wrapped in `CachingStrategy`, and points readers to `CachingStrategy` for the concurrent first-touch caveat.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ChatModelRouter.java
purushah: Good point. The stronger “runs once per conversation” wording was overstating the guarantee. I softened this to “typically runs once per conversation” and documented that this is best-effort memoization: a concurrent first-touch on the same key can double-compute, but the synchronized backing map avoids corruption and last-writer-wins is benign, so I kept the implementation lock-free.

weiqingy:
Thanks for the follow-ups, @purushah. My comments are resolved. I'll leave the final call to the maintainers.

purushah:
Thanks @weiqingy — really appreciate you taking the time to go through the design and call out the subtle cases. Your comments were very helpful, especially around the open() invariant and the caching wording. Glad the follow-ups addressed them.

xintongsong:
Hi @purushah ,

Sorry for the late response. I was busy preparing the 0.3 release, Flink Forward Asia, and catching up on a few other works, and only had the chance to carefully look into this PR today.

First of all, thank you again for the contribution! I think this is a very useful and interesting feature, and the implementation quality is also very high. It’s clear that you’ve put a lot of thought and effort into both the design and the implementation.

That said, for a feature of this importance and complexity, I would generally prefer to discuss the overall design first (e.g., through a GitHub Discussion) before moving into a full implementation. After going through the PR, I realized that most of my questions are actually about the design rather than the implementation itself. If we can align on these design questions first, we can hopefully avoid unnecessary rework later.

---

Here are my current thoughts.

### 1. Should ChatModelRouter be modeled as a special kind of ChatModelSetup?

I’m not entirely convinced yet. 

The biggest advantage of the current design is exactly what you pointed out: it requires almost no framework changes, and the router can be used just like any other ChatModelSetup, which makes the implementation straightforward.

However, I also see some potential drawbacks.

- The first one is observability. Today, EventLog is our primary observability mechanism. When using a normal ChatModelSetup, every model invocation is represented by ChatRequestEvent and ChatResponseEvent, so the model inputs and outputs are fully captured. With ChatModelRouter, especially when using an LLMRoutingStrategy, the model invocation used to determine which candidate should be selected is no longer represented in the EventLog. As a result, users cannot observe the routing decision process itself.

- Another concern is metrics. Today, many model-related metrics (for example, input/output token usage) are naturally aggregated at the ChatModelSetup level. If a single ChatModelSetup internally invokes multiple different models, those metrics will all be mixed together, making it difficult to understand which model consumed which tokens. This could make cost analysis and cost control more difficult.

One alternative direction (just as an example—I haven’t fully thought through the design yet) would be to treat model routing as a framework capability rather than a specialized ChatModelSetup, for example by introducing a dedicated ChatModelRouter resource type. My intuition is that this might naturally address some of the observability and metrics concerns mentioned above. That said, this is only an initial thought, not a concrete proposal.

### 2. The boundary between Public API and Internal Implementation doesn’t seem very clear yet

This was another impression I had while reading through the PR.

At the moment, most of the newly introduced concepts and implementations live inside the API module, so it’s difficult to immediately distinguish:
* Which concepts are intended to be part of the public API that users should interact with.
* Which ones are merely implementation details used internally by the framework.

Ideally, I think the API module should contain only the minimal set of interfaces that users actually need to interact with. This would help us maintain API compatibility more easily across future releases while still giving us enough flexibility to evolve the internal implementation.

### 3. What is the intended user experience of this feature?

I also think there is still room to discuss the user-facing API design.

In other words, it may be helpful to first step back and think about how users are expected to use this feature end-to-end, and then evaluate whether the API naturally supports that workflow.

For example:
* Is a RulBasedRoutingStrategy actually necessary? I’m not entirely sure that keyword-based routing has enough real-world demand to justify exposing it as part of the public API.
* Should caching be exposed as a general configuration option instead of a separate strategy? The current design, where CachingStrategy delegates to another strategy, looks like a perfectly reasonable internal implementation. However, from a user’s perspective, it might be more natural to simply enable or disable caching through configuration instead of explicitly wrapping one strategy with another.
* Both RoutingStrategy and FallbackPolicy are currently exposed as public interfaces. Do we really expect users to implement these interfaces directly? Or would it be sufficient for most users if we simply provided a set of built-in implementations and let users choose among them through configuration?

I'm not saying that I’m against the current design. In fact, I don’t have firm answers to the above questions either. What I'm trying to say is that, in order to answer these questions, we probably should have a more fundamental design discussion first:
* How do we expect users to use this feature?
* Which decisions should users make explicitly?
* Which decisions can be handled by the framework?
* For the decisions that do need to be exposed, are we exposing them through an API that is as simple and intuitive as possible?

---

Overall, I think this is a valuable feature, and I really appreciate the amount of work you’ve put into this PR.

My suggestion would be to treat this PR as a prototype for now rather than moving directly into detailed code review.

As a next step, I think it would be helpful to first flesh out the overall design—either by opening a GitHub Discussion or by adding a design proposal to this PR—especially around the following questions:
* What should the public API look like?
* What should remain internal implementation details?
* What is the intended user experience of this feature?

Once we have aligned on those aspects, I believe the implementation details will become much easier to converge on, and hopefully we don't need to rework this PR a lot.

WDYT?

purushah:
Thanks for the detailed review, @xintongsong — and no worries on the timing!

I think this is a great idea. I agree it's better to align on the overall design first before going deeper on the implementation. Let's keep this PR open and use it as a prototype/reference for the discussion. I'll follow up here shortly with a short design proposal covering the points you raised. Thanks again for the thoughtful feedback!

purushah:
@xintongsong opened a discussion to align on the design before going further here: #897

purushah:
Closing this in favor of #964. The design discussion in #897 concluded that the router should select a model rather than call it, with routing owned by `ChatModelAction` — a ground-up redesign of this PR's approach, so a fresh PR with a clean diff made more sense than force-pushing over this one. Thanks @wenjin272 @weiqingy for the reviews here and on #897; they shaped the new design directly.

Issue #853, opened 2026-06-18
[Feature] Support tool parameter injection


GreatEugenius:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Support tool parameter injection

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #854, opened 2026-06-18
[Feature] Support tool parameter injection

GreatEugenius:
Linked issue: #853 

### Purpose of change

Support declarative tool parameter injection for function tools.

This change lets framework-owned arguments, such as tenant id or request context, be injected at tool execution time without exposing them to the model-facing tool schema. The injection source is declared together with the hidden tool parameter, and values are resolved by the built-in `tool_call_action` immediately before invoking the tool.

Main changes:

- Add Python API support:
  - `@tool(injected_args={"tenant_id": InjectedArg.from_config("tenant_id")})`
  - explicit injection sources for config, sensory memory, and short-term memory

- Add Java API support:
  - `@ToolParam(injected = true, source = ToolParameterSource.CONFIG, key = "tenant_id")`
  - explicit injection sources for config, sensory memory, and short-term memory

- Add YAML API support:
  - `tools[].injected_args` as an object mapping parameter names to injection specs

- Inject parameters in the built-in `tool_call_action`, after reading model arguments and before invoking the tool.

- Preserve injected argument metadata in the plan/runtime representation while filtering injected parameters from model-facing tool metadata.

- Keep cross-language tool calls supported by resolving injection before dispatching to the target tool implementation.

### Tests
UT & E2E Test

### API
Yes

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/tools/ToolParameterSource.java
+++ b/api/src/main/java/org/apache/flink/agents/api/tools/ToolParameterSource.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.api.tools;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/** Source for a framework-injected tool parameter. */
+public enum ToolParameterSource {
+    CONFIG("config"),
+    SENSORY_MEMORY("sensory_memory"),
+    SHORT_TERM_MEMORY("short_term_memory");
+
+    private final String value;
+
+    ToolParameterSource(String value) {
+        this.value = value;
+    }
+
+    @JsonValue
+    public String getValue() {
+        return value;
+    }
+
+    @JsonCreator
+    public static ToolParameterSource fromValue(String value) {
+        for (ToolParameterSource source : values()) {
+            if (source.value.equals(value) || source.name().equals(value)) {
+                return source;
+            }
+        }
+        throw new IllegalArgumentException("Unknown tool parameter source: " + value);
+    }
+}
--- a/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/ToolParameterInjectionAgent.java
+++ b/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/ToolParameterInjectionAgent.java
@@ -1,0 +1,148 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.integration.test;
+
+import org.apache.flink.agents.api.Event;
+import org.apache.flink.agents.api.EventType;
+import org.apache.flink.agents.api.InputEvent;
+import org.apache.flink.agents.api.OutputEvent;
+import org.apache.flink.agents.api.agents.Agent;
+import org.apache.flink.agents.api.annotation.Action;
+import org.apache.flink.agents.api.annotation.ChatModelConnection;
+import org.apache.flink.agents.api.annotation.ChatModelSetup;
+import org.apache.flink.agents.api.annotation.Tool;
+import org.apache.flink.agents.api.annotation.ToolParam;
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.apache.flink.agents.api.chat.messages.MessageRole;
+import org.apache.flink.agents.api.chat.model.BaseChatModelConnection;
+import org.apache.flink.agents.api.chat.model.BaseChatModelSetup;
+import org.apache.flink.agents.api.context.RunnerContext;
+import org.apache.flink.agents.api.event.ChatRequestEvent;
+import org.apache.flink.agents.api.event.ChatResponseEvent;
+import org.apache.flink.agents.api.resource.ResourceContext;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
+import org.apache.flink.agents.api.tools.ToolParameterSource;
+import org.apache.flink.api.java.functions.KeySelector;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** E2E agent covering tool parameter injection in a real Flink job. */
+public class ToolParameterInjectionAgent extends Agent {
+
+    /** Key selector for order ids. */
+    public static class OrderKeySelector implements KeySelector<String, String> {
+        @Override
+        public String getKey(String value) {
+            return value;
+        }
+    }
+
+    /** Mock chat connection that emits one queryOrder tool call, then echoes the tool result. */
+    public static class MockToolChatConnection extends BaseChatModelConnection {
+        public MockToolChatConnection(
+                ResourceDescriptor descriptor, ResourceContext resourceContext) {
+            super(descriptor, resourceContext);
+        }
+
+        @Override
+        public ChatMessage chat(
+                List<ChatMessage> messages,
+                List<org.apache.flink.agents.api.tools.Tool> tools,
+                Map<String, Object> modelParams) {
+            ChatMessage lastMessage = messages.get(messages.size() - 1);
+            if (lastMessage.getRole() == MessageRole.TOOL) {
+                return new ChatMessage(MessageRole.ASSISTANT, lastMessage.getContent());
+            }
+
+            for (org.apache.flink.agents.api.tools.Tool tool : tools) {
+                String inputSchema = tool.getMetadata().getInputSchema();
+                if (inputSchema.contains("tenant_id")) {
+                    throw new IllegalStateException(
+                            "Injected argument leaked into tool schema: " + inputSchema);
+                }
+            }
+            String orderId = lastMessage.getContent();
+            return new ChatMessage(
+                    MessageRole.ASSISTANT,
+                    "",
+                    List.of(
+                            Map.of(
+                                    "id",
+                                    "call-" + orderId,
+                                    "type",
+                                    "function",
+                                    "function",
+                                    Map.of(
+                                            "name",
+                                            "queryOrder",
+                                            "arguments",
+                                            Map.of("order_id", orderId)))));
+        }
+    }
+
+    /** Mock chat model setup bound to queryOrder. */
+    public static class MockToolChatModel extends BaseChatModelSetup {
+        public MockToolChatModel(ResourceDescriptor descriptor, ResourceContext resourceContext) {
+            super(descriptor, resourceContext);
+        }
+
+        @Override
+        public Map<String, Object> getParameters() {
+            return new HashMap<>();
+        }
+    }
+
+    @ChatModelConnection
+    public static ResourceDescriptor mockChatConnection() {
+        return ResourceDescriptor.Builder.newBuilder(MockToolChatConnection.class.getName())
+                .build();
+    }
+
+    @ChatModelSetup
+    public static ResourceDescriptor mockChatModel() {
+        return ResourceDescriptor.Builder.newBuilder(MockToolChatModel.class.getName())
+                .addInitialArgument("connection", "mockChatConnection")
+                .addInitialArgument("model", "mock")
+                .addInitialArgument("tools", List.of("queryOrder"))
+                .build();
+    }
+
+    @Tool(description = "Query order in the current tenant.")
+    public static String queryOrder(
+            @ToolParam(name = "order_id") String orderId,
+            @ToolParam(name = "tenant_id", injected = true, source = ToolParameterSource.CONFIG)
+                    String tenantId) {
+        return "checked:" + tenantId + ":" + orderId;
+    }
+
+    @Action(EventType.InputEvent)
+    public static void requestTool(Event event, RunnerContext ctx) {
+        String orderId = String.valueOf(InputEvent.fromEvent(event).getInput());
+        ctx.sendEvent(
+                new ChatRequestEvent(
+                        "mockChatModel", List.of(new ChatMessage(MessageRole.USER, orderId))));
+    }
+
+    @Action(EventType.ChatResponseEvent)
+    public static void emitResult(Event event, RunnerContext ctx) {
+        ChatResponseEvent responseEvent = ChatResponseEvent.fromEvent(event);
+        ctx.sendEvent(new OutputEvent(responseEvent.getResponse().getContent()));
+    }
+}
--- a/plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
@@ -19,15 +19,20 @@ package org.apache.flink.agents.plan.actions;
 
 import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.api.agents.AgentExecutionOptions;
+import org.apache.flink.agents.api.configuration.ConfigOption;
 import org.apache.flink.agents.api.context.DurableCallable;
+import org.apache.flink.agents.api.context.MemoryObject;
 import org.apache.flink.agents.api.context.RunnerContext;
 import org.apache.flink.agents.api.event.ToolRequestEvent;
 import org.apache.flink.agents.api.event.ToolResponseEvent;
 import org.apache.flink.agents.api.resource.ResourceType;
 import org.apache.flink.agents.api.tools.Tool;
+import org.apache.flink.agents.api.tools.ToolParameterInjection;
+import org.apache.flink.agents.api.tools.ToolParameterSource;
 import org.apache.flink.agents.api.tools.ToolParameters;
 import org.apache.flink.agents.api.tools.ToolResponse;
 import org.apache.flink.agents.plan.JavaFunction;
+import org.apache.flink.agents.plan.tools.FunctionTool;
 
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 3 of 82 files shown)


GreatEugenius:
Hi, @Sxnan and @wenjin272. Could you please review this PR? Thank you!

Sxnan:
Thanks for the work — the schema-hiding part (`injected_args` / `@ToolParam(injected=true)`) is clean and useful.

A step back on the API shape before line details: #853 asks for one narrow thing — inject a model-hidden, framework-owned arg at execution time. This PR does it via a general **before/after hook mechanism** (hook slots on `Agent`, plan fields, serializers, YAML keys, new `ToolCallContext`/`ToolCallResult`). That's a lot of machinery, and the generality makes injection harder, not easier:

- Two decoupled declarations (`injected_args` + a separate global hook matching on `tool_name`), no plan-time link — forget/typo the hook → silent crash.
- One hook per agent → a manual `if tool_name == ...` dispatch table as tools grow.
- `after_tool_call` (result rewriting) is unrelated to injection — a separate capability bundled in.

Suggestion:
1. **Keep** `injected_args` for schema hiding.
2. **Make the value source declarative and co-located with the param**, e.g. `@tool(inject={"tenant_id": lambda ctx: ctx.tenant_id})`, resolved right before the call — drops the decoupling, the dispatch table, and the "declared but never filled" gap.
3. **Defer** `before_tool_call` / `after_tool_call` to a separate issue if a real cross-cutting need shows up — and there, consider the existing event/action model over a parallel mechanism.

If there's a concrete need behind the general hooks that isn't in #853, worth surfacing here.

GreatEugenius:
> Thanks for the work — the schema-hiding part (`injected_args` / `@ToolParam(injected=true)`) is clean and useful.
> 
> A step back on the API shape before line details: #853 asks for one narrow thing — inject a model-hidden, framework-owned arg at execution time. This PR does it via a general **before/after hook mechanism** (hook slots on `Agent`, plan fields, serializers, YAML keys, new `ToolCallContext`/`ToolCallResult`). That's a lot of machinery, and the generality makes injection harder, not easier:
> 
> * Two decoupled declarations (`injected_args` + a separate global hook matching on `tool_name`), no plan-time link — forget/typo the hook → silent crash.
> * One hook per agent → a manual `if tool_name == ...` dispatch table as tools grow.
> * `after_tool_call` (result rewriting) is unrelated to injection — a separate capability bundled in.
> 
> Suggestion:
> 
> 1. **Keep** `injected_args` for schema hiding.
> 2. **Make the value source declarative and co-located with the param**, e.g. `@tool(inject={"tenant_id": lambda ctx: ctx.tenant_id})`, resolved right before the call — drops the decoupling, the dispatch table, and the "declared but never filled" gap.
> 3. **Defer** `before_tool_call` / `after_tool_call` to a separate issue if a real cross-cutting need shows up — and there, consider the existing event/action model over a parallel mechanism.
> 
> If there's a concrete need behind the general hooks that isn't in #853, worth surfacing here.

Thanks, agreed with the direction here. Stepping back, #853 should stay focused on model-hidden framework-owned argument injection, and the generic before/after hook mechanism was too broad for that scope.

I’ve updated the PR accordingly:

- Removed the before/after tool-call hook API and the related Agent/Plan/YAML/serializer fields.
- Kept `injected_args` / `@ToolParam(injected = true)` for schema hiding.
- Made the injection source declarative and co-located with the parameter instead of relying on a separate global hook. The current API supports explicit sources such as config, sensory memory, and short-term memory.
- Kept resolution in the built-in `tool_call_action`, right before invoking the tool, so the injected value remains hidden from the model and is not written back to the tool request.

So the PR now covers only the narrow injection use case from #853. Any general `before_tool_call` / `after_tool_call` capability can be discussed separately if a concrete cross-cutting need shows up.

python/flink_agents/plan/actions/tool_call_action.py
```diff
@@ -39,27 +46,80 @@ async def process_tool_request(event: Event, ctx: RunnerContext) -> None:
...
+                call_kwargs.update(_resolve_injected_arguments(tool, ctx))
+                if tool_call_async:
+                    response = await ctx.durable_execute_async(
+                        tool.call, **call_kwargs
+                    )
+                else:
+                    response = ctx.durable_execute(tool.call, **call_kwargs)
+                responses[call_id] = response
+            except Exception as e:
+                responses[call_id] = str(e)
```
weiqingy: This new per-call `except Exception as e: responses[call_id] = str(e)` routes an injection failure (missing config, missing memory path) into the response channel. But Python `ToolResponseEvent` carries only `responses` — there's no success/error channel like Java's — and the consumer at `chat_model_action.py:439` does `content=str(response)` for every entry. So a failed injection gets fed back to the LLM as an ordinary string, indistinguishable from a real tool result. The Java side emits `ToolResponse.error(...)` and populates the error map, so there the failure stays structurally distinguishable.

To be clear, the *absence* of a Python error channel is pre-existing — what's new here is routing injection failures through it. Is feeding the error text back to the model the intended contract, or should an injection misconfiguration fail loudly / be distinguishable from a successful result? CLAUDE.md asks for Java/Python parity, and this is where the two sides diverge most.

api/src/main/java/org/apache/flink/agents/api/tools/ToolParameterSource.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @JsonValue
+    public String getValue() {
+        return value;
+    }
+
+    @JsonCreator
+    public static ToolParameterSource fromValue(String value) {
+        if ("action_config".equals(value)) {
```
weiqingy: This branch maps the string `"action_config"` → `CONFIG` (mirrored in Python at `tool_parameter_injection.py:35`), but `action_config` is a brand-new name with no test, no doc, and no YAML/JSON reference. It also collides with the existing, unrelated per-action `action_config` concept (`RunnerContext.action_config`, the per-action config dict — not the global config this maps to). It reads as vestigial, likely a leftover from the removed hook iteration. Is `action_config` meant to be a public source name? If not, dropping both branches keeps the contract to the documented `config` source.

python/flink_agents/api/tools/tool_parameter_injection.py
```diff
@@ -0,0 +1,89 @@
...
+        return InjectedArg(source=ToolParameterSource.SHORT_TERM_MEMORY, key=path)
+
+    def with_default_key(self, key: str) -> "InjectedArg":
+        """Use the tool parameter name when no explicit key is configured."""
+        if self.key:
+            return self
+        return InjectedArg(source=self.source, key=key)
+
+
+def normalize_injected_args(
```
weiqingy, line 71: The reflection path (`@ToolParam(injected=true)`) is safe by construction, but the declarative paths aren't: the YAML `injected_args` keys and the `@tool(injected_args={...})` decorator keys accept arbitrary names with no check that the key matches an actual function parameter. A typo like `tenent_id` gets merged into the call kwargs and fails late — a `TypeError: unexpected keyword argument` on Python, which then gets swallowed per the response-channel point above. Was up-front validation at tool construction considered? A cheap key-against-signature check would turn an opaque runtime tool-failure into a clear config error.

plan/src/test/java/org/apache/flink/agents/plan/actions/ToolCallActionTest.java
```diff
@@ -0,0 +1,376 @@
...
+                                        "type",
+                                        "function",
+                                        "function",
+                                        Map.of("name", "queryOrder", "arguments", arguments))));
+
+        ToolCallAction.processToolRequest(event, ctx);
+
+        ToolResponseEvent response = ToolResponseEvent.fromEvent(ctx.sentEvents.get(0));
+        assertThat(response.getResponses().get("call-1").getResult()).isEqualTo("tenant-1:order-1");
+        assertThat(arguments).containsOnly(Map.entry("orderId", "order-1"));
```
weiqingy: The comment in `ToolCallAction` says injected args must win over model-provided values "so hidden context such as tenant ids cannot be spoofed by a tool call payload" — but this config-injection test passes only non-conflicting args (`orderId`), so it would still pass even if Java used `putIfAbsent` instead of `putAll`. The security-relevant override is verified only on the Python side (`test_tool_call_action_injected_arg_overrides_model_argument`). Something like this, if useful — add a case where the model also supplies `tenant_id` and assert the injected value wins, so the Java override is actually pinned.

python/flink_agents/plan/tests/resources/agent_plan.json
```diff
@@ -102,30 +102,30 @@
                 "__resource_provider_type__": "PythonResourceProvider"
             }
         },
-        "embedding_model": {
-            "mock_embedding": {
-                "name": "mock_embedding",
-                "type": "embedding_model",
+        "embedding_model_connection": {
```
weiqingy: This hunk swaps the positions of the `embedding_model` and `embedding_model_connection` blocks; their contents are byte-identical before and after, and no `injected_args` is added here — it looks like unrelated regeneration churn. Worth reverting this file to keep the diff focused on injection?

python/flink_agents/plan/actions/tool_call_action.py
GreatEugenius: Good catch, thanks. I agree this should not be routed back as an ordinary successful tool result.

I updated the Python side to carry `success` / `error` on `ToolResponseEvent`, populate them from `tool_call_action`, and make `chat_model_action` use the error payload when `success[tool_id]` is false. That keeps injection/config failures structurally distinguishable from successful tool output and brings the Python behavior closer to the Java side. The deserializer also keeps a compatibility default for older events that do not have these fields.

api/src/main/java/org/apache/flink/agents/api/tools/ToolParameterSource.java
GreatEugenius: Agreed. `action_config` was leftover naming from the earlier hook-based iteration and is misleading because it collides with the existing per-action config concept.

I removed the `action_config` alias from both Java and Python. The public source names are now only the documented ones: `config`, `sensory_memory`, and `short_term_memory`.

python/flink_agents/api/tools/tool_parameter_injection.py
GreatEugenius: Yes, agreed. I added upfront validation for declarative injected argument names.

On the Python side, `@tool(...)` / `Tool.from_callable(...)` validate immediately because the callable is already available. For descriptor-style paths such as YAML/API `FunctionTool`, the YAML loader now stays side-effect-free and does not import/inspect user functions just for validation; the check is performed when building the `AgentPlan`, where the callable is materialized anyway. The validator also allows `**kwargs`, since those tools can intentionally accept framework-provided extra arguments.

The Java side now validates at the same AgentPlan boundary via `ToolParameterInjectionValidator`. It validates when parameter names are reliable, for example from `@ToolParam(name = ...)` or `-parameters`, and avoids false positives when bytecode only exposes synthetic names like `arg0`.

plan/src/test/java/org/apache/flink/agents/plan/actions/ToolCallActionTest.java
GreatEugenius: Good point. I added a Java test that covers the security-relevant case directly: the model supplies `tenant_id`, the injected source supplies another value, and the injected value wins. The test also checks that the original model argument map is not mutated.

python/flink_agents/plan/tests/resources/agent_plan.json
GreatEugenius: Agreed, this was unrelated regeneration churn. I reverted the resource-order-only change so the snapshot diff stays focused on the injected-args behavior.

GreatEugenius:
Hi @Sxnan and @weiqingy, thank you for your comments. I have addressed all of them. Please take a look.

wenjin272:
### Code review

Found 1 issue:

1. Java tools that declare `injected_args` via YAML (or via the programmatic `new FunctionTool(metadata, func, injectedArgs)` API) do not hide the injected parameter from the model-facing tool schema, contradicting the documented contract and diverging from the Python behavior. Java schema generation skips a parameter only when the method is annotated `@ToolParam(injected = true)`; the `injectedArgs` map supplied by YAML is threaded through for runtime resolution only, never into schema generation. So a YAML-declared injected param on a method that is not annotated `injected = true` (exactly what the new fixture `LoaderTargets.queryOrder` exercises — it annotates `tenant_id` with only `@ToolParam(name = "tenant_id")`) still appears to the model as a required argument. The injected value still wins at call time, so it is not spoofable, but the param leaks into the schema. Python's YAML/decorator path hides it correctly via `create_schema_from_function(..., injected_args=...)`, so the two languages diverge.

Schema hiding is annotation-only (no `injectedArgs` parameter):

https://github.com/apache/flink-agents/blob/74d5ea3e88299b97c4576c5be826a6fff8ccdf32/plan/src/main/java/org/apache/flink/agents/plan/tools/SchemaUtils.java#L49-L57

The YAML promote path builds the Java schema from the method alone, then passes the YAML `injected_args` separately (runtime only):

https://github.com/apache/flink-agents/blob/74d5ea3e88299b97c4576c5be826a6fff8ccdf32/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java#L696-L703

Docs promise the args are hidden from the model-facing schema:

https://github.com/apache/flink-agents/blob/74d5ea3e88299b97c4576c5be826a6fff8ccdf32/docs/content/docs/development/yaml.md#L139-L141

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/ToolParameterInjectionAgent.java
```diff
@@ -0,0 +1,145 @@
...
+                .addInitialArgument("connection", "mockChatConnection")
+                .addInitialArgument("model", "mock")
+                .addInitialArgument("tools", List.of("queryOrder"))
+                .build();
+    }
+
+    @Tool(description = "Query order in the current tenant.")
+    public static String queryOrder(
+            @ToolParam(name = "order_id") String orderId,
+            @ToolParam(name = "tenant_id", injected = true, source = ToolParameterSource.CONFIG)
```
wenjin272, line 130: If the user only declares `source` but does not declare `injected`, parameter injection will still not take effect. We may need to emphasize this point in the documentation.

plan/src/main/java/org/apache/flink/agents/plan/tools/SchemaUtils.java
```diff
@@ -49,6 +49,9 @@ public static String generateSchema(Method method) throws JsonProcessingExceptio
             // Check for custom parameter name from annotation
             if (param.isAnnotationPresent(ToolParam.class)) {
                 ToolParam toolParam = param.getAnnotation(ToolParam.class);
+                if (toolParam.injected()) {
```
wenjin272: It appears that parameter hiding on the Java side heavily relies on annotations. Based on the current design, the YAML API allows injection parameters to be declared directly in the YAML file rather than in method-level annotations. This approach may introduce issues with parameter hiding.

wenjin272 (commented):
@GreatEugenius, thanks for taking this on. Overall looks good to me, I left two comments.

Besides, I think we also need add e2e case for YAML API.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
```diff
@@ -92,24 +101,82 @@ public Class<ToolResponse> getResultClass() {
...
+            throw new IllegalStateException(
+                    "Cannot inject tool parameter from "
+                            + source
+                            + " because memory is not initialized.");
+        }
+        if (!memory.isExist(path)) {
+            throw new IllegalArgumentException(
+                    "Missing memory path for injected tool parameter: " + path);
+        }
+        return memory.get(path).getValue();
```
Sxnan: `isExist(path)` is also true for a nested object, but `MemoryObjectImpl.getValue()` returns `null` for a non-`VALUE` item — so a key resolving to a nested object silently injects `null` and overwrites the model-provided arg, despite the intent that injected args must win to prevent spoofing. Python's `_get_memory_value` returns a `MemoryObject` for the same case, so the two languages diverge. Suggest rejecting a non-primitive path after `isExist` instead of injecting `null`.

python/flink_agents/plan/actions/tool_call_action.py
```diff
@@ -37,29 +44,97 @@ async def process_tool_request(event: Event, ctx: RunnerContext) -> None:
...
+def _get_memory_value(memory: MemoryObject, source: str, path: str) -> object:
+    if memory is None:
+        msg = (
+            f"Cannot inject tool parameter from {source} because memory is not initialized."
+        )
+        raise ValueError(msg)
+    if not memory.is_exist(path):
+        msg = f"Missing memory path for injected tool parameter: {path}"
+        raise ValueError(msg)
+    return memory.get(path)
```
Sxnan: Parity issue with Java: for a key that resolves to a nested object, `LocalMemoryObject.get()` returns a `MemoryObject` (passed straight to the tool as a kwarg), whereas Java returns `null`. `is_exist` does not distinguish a leaf value from a nested object. Suggest rejecting a non-primitive path after `is_exist` so both languages behave the same.

python/flink_agents/plan/actions/chat_model_action.py
```diff
@@ -436,7 +436,9 @@ async def _process_tool_response(event: ToolResponseEvent, ctx: RunnerContext) -
         [
             ChatMessage(
                 role=MessageRole.TOOL,
-                content=str(response),
+                content=str(response)
+                if event.success.get(tool_id, True)
+                else str(event.error.get(tool_id, response)),
```
Sxnan: On tool failure this feeds the raw `str(e)` to the model. Java keeps the raw error only in the `error` map and builds the TOOL content from `getResult()` (which is `null` for a failure `ToolResponse`), so the model never sees it — the two languages diverge and the Python path leaks internal exception text to the LLM. Suggest a stable/generic message to the model, with the raw error kept only in `error`/logs.

Sxnan (commented):
Two follow-ups, verified against the implementations. Neither blocks the common scalar-injection path.

GreatEugenius:
Hi @wenjin272 and @Sxnan, I have addressed your comment. Please review it once more.

Sxnan (commented):
One more cross-language gap, verified against the implementations: injected-arg declarations don't travel across the language bridge — schema hiding and injection only honor the host-language declaration.

python/flink_agents/plan/tools/function_tool.py
```diff
@@ -40,11 +46,28 @@ class FunctionTool(Tool):
...
+            callable_ = self.func.as_callable()
+            self.injected_args = merge_injected_args(
+                getattr(callable_, "_injected_args", None),
+                self.injected_args,
+                tool_name=callable_.__qualname__,
+            )
+            validate_injected_arg_names(callable_, self.injected_args)
         if self.metadata is None and isinstance(self.func, PythonFunction):
-            self.metadata = _python_metadata(self.func)
+            self.metadata = _python_metadata(self.func, list(self.injected_args))
```
Sxnan, line 71: The `PythonFunction` path here passes `list(self.injected_args)` into the schema derivation — but the `JavaFunction` path below (`set_java_resource_adapter` → `_java_metadata(self.func)`) drops `self.injected_args` entirely. For a `type: java` tool whose `injected_args` are declared only in YAML (loader `_build_tool` does store them on this FunctionTool), `_java_metadata` → `JavaResourceAdapter.getJavaToolMetadata` ends up calling the no-injected-args overload `ToolMetadataFactory.fromStaticMethod(method)`, so the injected param leaks into the model-facing schema unless the Java method also has `@ToolParam(injected = true)`. Execution-time injection still works, so this is a schema leak, not a broken call. The injected-aware overload `fromStaticMethod(method, names)` already exists — threading the names through `_java_metadata` and `getJavaToolMetadata` closes the gap (the reverse bridge already does this via `getInjectedArgNames()`).

python/flink_agents/runtime/python_java_utils.py
```diff
@@ -149,7 +151,9 @@ def get_python_tool_metadata(module: str, qual_name: str) -> Dict[str, str]:
     callable_ = descriptor.as_callable()
     name = callable_.__name__
     description = (parse(callable_.__doc__).description or "") if callable_.__doc__ else ""
-    args_schema_model = create_schema_from_function(name, callable_)
+    args_schema_model = create_schema_from_function(
```
Sxnan, line 162: Mirror direction: a Python function declared with `@tool(injected_args={...})` but referenced from a Java agent's YAML without re-declaring `injected_args` loses its injection entirely — this only excludes the names Java passes (YAML-declared), never reads the callable's `_injected_args`, so the param leaks into the schema; and `invoke_python_tool` calls the function with only the model kwargs, so nothing is injected at execution time and the call fails on the missing required param. If decorator declarations aren't meant to cross the language boundary, this deserves a loud error or a docs note instead of a silent failure.

python/flink_agents/runtime/python_java_utils.py
```diff
@@ -149,7 +151,9 @@ def get_python_tool_metadata(module: str, qual_name: str) -> Dict[str, str]:
...
+        name, callable_, injected_args=injected_args
```
wenjin272: get_python_tool_metadata builds the schema from only the Java-passed injected_args, never the callable's own `_injected_args`.                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                                     
The pure-Python paths (`Tool.from_callable`, `_eager_derive_python_metadata`) both `merge_injected_args(getattr(callable_, "_injected_args", None), ...)`, so a Python tool authored with `@tool(injected_args=...)` and used in a  Java-orchestrated plan (YAML omitting the arg) gets its injected param neither hidden from the schema nor injected at call time — the model-supplied value flows into the real call, the exact spoofing case this feature prevents.    
                                                                                                                                                                                                                                         
Suggest merging the callable's _injected_args here (as the docstring's "mirrors the Python-side derivation" claim implies).

GreatEugenius:
Hi @Sxnan and @wenjin272. Thanks for pointing this out. I fixed this by carrying injected tool args through the cross-language metadata bridge in both directions.

For Python host -> Java tool, `FunctionTool.set_java_resource_adapter()` now passes the descriptor-declared injected arg names into `JavaResourceAdapter.getJavaToolMetadata(...)`. The Java side uses those names when generating the model-facing schema, and also returns annotation-declared injected args from `@ToolParam(injected = true)`. The Python `FunctionTool` then merges the descriptor declarations and Java annotation declarations, so `tool_call_action` can see the complete `injected_args` set at execution time.

For Java host -> Python tool, `get_python_tool_metadata(...)` now reads the callable’s `_injected_args`, hides those params from the schema together with any Java/YAML-declared injected args, and returns the callable-declared injected args back to Java. Java `FunctionTool.setPythonResourceAdapter()` merges them into the tool descriptor before execution.

So the cross-language behavior is now aligned: injected params are hidden from the model schema and are also available to the host-language `tool_call_action` for runtime injection, regardless of whether they were declared in YAML/API or on the tool annotation/decorator.

Sxnan (approved):
Thanks for the thorough fix — verified both directions of the bridge: injected-arg specs (with source/key, not just names) now travel Python↔Java, schema hiding covers annotation/decorator/YAML declarations uniformly, and execution-time injection sees the merged set. All my earlier comments are addressed. LGTM.

Issue #833, opened 2026-06-10
[Feature] Support Flink 2.3


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, Flink-Agents supports Flink 1.20, 2.0, 2.1, and 2.2. Flink 2.3 will be released soon, and we aim to support it as well.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #856, opened 2026-06-24
[dist] Add Flink 2.3 distribution and make it the default

GreatEugenius:
Linked issue: #833 

### Purpose of change

Add Flink 2.3 support and bump the default Flink version to 2.3.0.

Central 2.3.0 jars are Java 11 bytecode, so the default still builds on JDK 11 — no Java 17 requirement. dist modules follow a symmetric rule: the default version (2.3) is bare and inherits the root flink.version; every other version pins its own version + provided deps.

- dist: new flink-2.3 module; flink-2.2 now pins 2.2.1 + provided deps; root default 2.2.1 → 2.3.0
- e2e: flink-2.3 is the default, flink-2.2 moved to a profile; e2e.sh default → 2.3
- ci: add {flink 2.3, java 17} to it-java and it-python matrices
- python: bundle the 2.3 thin jar into the wheel manifest + build-backend test coverage
- misc: ut.sh help, update_flink_version.sh, faq/long_term_memory version notes

### Tests

Full build on JDK 11 — all 5 dist versions (1.20 / 2.0 / 2.1 / 2.2 / 2.3) pass. Java and Python IT cover Flink 2.3 in CI.

### API

No public API changes. 2.3 is a packaging/runtime target only.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/faq/faq.md
+++ b/docs/content/docs/faq/faq.md
@@ -87,7 +87,7 @@ This is important because:
 - **For Java users on JDK 21+**: Async execution is available.
 - **For Java users on JDK < 21**: Async execution is not available and falls back to synchronous execution.
 
-> **Cross-language async note**: Async execution for cross-language resources requires the pemja 0.5.7 fix, available in Flink 1.20.5+ / 2.0.2+ / 2.1.3+ / 2.2.1+. Current builds target Flink 2.2.0, so cross-language calls still run synchronously for now; this is resolved automatically once running on a Flink version that includes the fix.
+> **Cross-language async note**: Async execution for cross-language resources requires the pemja 0.5.7 fix, available in Flink 1.20.5+ / 2.0.2+ / 2.1.3+ / 2.2.1+ / 2.3+. Current builds target Flink 2.3.0, which includes the fix, so cross-language async is enabled by default; on older Flink versions it falls back to synchronous execution automatically.
 
 ### Native Integration Support Matrix
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,9 @@ jobs:
           - os: 'ubuntu-latest'
             python-version: '3.12'
             flink-version: "2.2"
+          - os: 'ubuntu-latest'
+            python-version: '3.12'
+            flink-version: "2.3"
     steps:
       - uses: actions/checkout@v4
       - name: Install java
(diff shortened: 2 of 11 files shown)


GreatEugenius:
Hi, @wenjin272. Could you please review this PR? Thank you!

docs/content/docs/faq/faq.md
```diff
@@ -87,7 +87,7 @@ This is important because:
 - **For Java users on JDK 21+**: Async execution is available.
 - **For Java users on JDK < 21**: Async execution is not available and falls back to synchronous execution.
 
-> **Cross-language async note**: Async execution for cross-language resources requires the pemja 0.5.7 fix, available in Flink 1.20.5+ / 2.0.2+ / 2.1.3+ / 2.2.1+. Current builds target Flink 2.2.0, so cross-language calls still run synchronously for now; this is resolved automatically once running on a Flink version that includes the fix.
+> **Cross-language async note**: Async execution for cross-language resources requires the pemja 0.5.7 fix, available in Flink 1.20.5+ / 2.0.2+ / 2.1.3+ / 2.2.1+ / 2.3+. Current builds target Flink 2.2.1, which includes the fix, so cross-language async is enabled by default; on older Flink versions it falls back to synchronous execution automatically.
```
weiqingy: This line still reads `Current builds target Flink 2.2.1`, but the root `flink.version` now defaults to `2.3.0` (`pom.xml:44`) — looks like the version number didn't follow the default bump. Should it be `2.3.0`? The rest of the sentence (the fix is present, so async is on by default) holds either way.

GreatEugenius:
Good catch, thanks @weiqingy — updated `faq.md` to `Current builds target Flink 2.3.0` to follow the default bump. Also rebased onto the latest `main`. Full build passes on JDK 11 across all 5 dist versions (1.20/2.0/2.1/2.2/2.3).

wenjin272:
The install.sh script should also be updated. Additionally, since we are updating install.sh, I think we can address this issue as well. #863

GreatEugenius:
> The install.sh script should also be updated. Additionally, since we are updating install.sh, I think we can address this issue as well. #863

Hi @wenjin272, I don't believe this PR should involve any changes to install.sh. The support for Flink 2.3 is only for the main branch of Flink Agents, so we should update install.sh only after the release of version 0.4.

wenjin272 (approved):
LGTM

Issue #857, opened 2026-06-24
[Feature] Propagate metric groups to cross-language resources


joeyutong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Flink Agents currently binds the current action metric group to resources when users call `RunnerContext#getResource(...)` / `RunnerContext.get_resource(...)`. This works for resources implemented in the same language as the action, but the cross-language resource wrappers do not consistently propagate the metric group to the actual resource on the other side of the Java/Python boundary.

Current behavior, in simplified form:

- Java action -> Python resource: the Java `Python*` wrapper receives the action metric group, but the underlying Python resource does not necessarily receive `set_metric_group(...)`.
- Python action -> Java resource: the Python `Java*` wrapper receives the action metric group, but the underlying Java resource is fetched through the Java resource context/cache path rather than the Java runner-context path, so it does not necessarily receive `setMetricGroup(...)`.

This is acceptable for metrics that are fully owned and recorded by the action layer. For example, token metrics can still be recorded under an action/model scope if the action extracts usage fields from the response and records counters on the local wrapper.

However, some features need the metric group to reach the actual provider/resource implementation. Examples:

- provider-specific metric dimensions such as `serviceUrl`, `serviceType`, tenant, deployment, or endpoint;
- chat/embedding model connections that wrap the metric group before recording token usage;
- vector stores, embedding models, long-term memory, or MCP resources that record latency, request count, error count, cache hit/miss, or backend-specific metrics internally;
- nested resource calls where the resource, not the action, owns the operation semantics and should own the metric scope;
- consistent observability behavior between same-language resources and cross-language resources.

A possible improvement is to make cross-language resource wrappers propagate metric group binding to the underlying resource when `setMetricGroup(...)` / `set_metric_group(...)` is called, or to expose an adapter-level metric-group binding hook for the underlying Java/Python object. The implementation should also define metric ownership clearly to avoid double-counting when action-level metrics and provider-level metrics both observe the same model response.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

goutamadwant:
@joeyutong I took a look at this and have a small patch ready.

The approach is to propagate the bound metric group at the wrapper boundary in both directions:

- Java `Python*` wrappers forward `setMetricGroup(...)` to the underlying Python resource through `PythonResourceAdapter`
[...]

PR #860, opened 2026-06-26
[api][runtime][python] Propagate metric groups to cross-language resources

goutamadwant:
Linked issue: #857

### Purpose of change

This change propagates the action metric group across Java/Python resource boundaries so cross-language resources record metrics under the caller's action scope.

The forwarding is handled by the Python-backed wrapper classes rather than by adding Python-specific logic to the base `Resource` class. On the Python-to-Java path, Java resource wrappers now unwrap only `FlinkMetricGroup` instances or `None`, and fail clearly for unsupported custom metric groups instead of passing an opaque Python object through pemja.

Metric ownership remains single-recorder: this PR forwards the bound metric group but does not add provider-side token recording. It also keeps the return-path/effective-scope question out of scope. If a provider/resource wraps the metric group with provider-specific dimensions before action-level token recording, that behavior should be tracked separately.

### Tests

- `uv run --no-sync pytest flink_agents/runtime/tests/test_cross_language_metric_group.py`
- `mvn --batch-mode --no-transfer-progress -pl api,plan,runtime -am -DskipTests -Dspotless.skip=true test-compile`
- `mvn --batch-mode --no-transfer-progress -pl api -Dtest=PythonChatModelSetupTest -Dspotless.skip=true test`
- `mvn --batch-mode --no-transfer-progress -pl runtime -am -Dtest=PythonResourceAdapterImplTest -Dsurefire.failIfNoSpecifiedTests=false -Dspotless.skip=true test`
- `./tools/lint.sh -c`

### API

Touches the Java/Python bridge interfaces by adding default metric-forwarding hooks on `PythonResourceAdapter` and `PythonResourceWrapper`. There are no user-facing configuration changes.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/runtime/java/java_resource_wrapper.py
+++ b/python/flink_agents/runtime/java/java_resource_wrapper.py
@@ -27,6 +27,22 @@ from flink_agents.api.resource_context import ResourceContext
 from flink_agents.api.tools.tool import Tool, ToolMetadata, ToolType
 
 
+def set_java_resource_metric_group(j_resource: Any, metric_group: Any) -> None:
+    """Bind the underlying Java metric group to a wrapped Java resource."""
+    if j_resource is None:
+        return
+    from flink_agents.runtime.flink_metric_group import FlinkMetricGroup
+
+    if metric_group is None:
+        j_metric_group = None
+    elif isinstance(metric_group, FlinkMetricGroup):
+        j_metric_group = metric_group._j_metric_group
+    else:
+        msg = "Java resource metric groups must be FlinkMetricGroup or None."
+        raise TypeError(msg)
+    j_resource.setMetricGroup(j_metric_group)
+
+
 class JavaTool(Tool):
     """Java Tool that carries tool metadata and can be recognized by PythonChatModel."""
 
--- a/python/flink_agents/runtime/tests/test_cross_language_metric_group.py
+++ b/python/flink_agents/runtime/tests/test_cross_language_metric_group.py
@@ -1,0 +1,149 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#################################################################################
+from typing import Any
+
+import pytest
+
+from flink_agents.api.metric_group import MetricGroup
+from flink_agents.runtime.flink_metric_group import FlinkMetricGroup
+from flink_agents.runtime.java.java_chat_model import (
+    JavaChatModelConnectionImpl,
+    JavaChatModelSetupImpl,
+)
+from flink_agents.runtime.java.java_embedding_model import (
+    JavaEmbeddingModelConnectionImpl,
+    JavaEmbeddingModelSetupImpl,
+)
+from flink_agents.runtime.java.java_resource_wrapper import (
+    set_java_resource_metric_group,
+)
+from flink_agents.runtime.python_java_utils import set_metric_group
+
+
+class _JavaResource:
+    def __init__(self) -> None:
+        self.metric_group: Any = None
+
+    def setMetricGroup(self, metric_group: Any) -> None:
+        self.metric_group = metric_group
+
+
+class _PythonResource:
+    def __init__(self) -> None:
+        self.metric_group: Any = None
+
+    def set_metric_group(self, metric_group: Any) -> None:
+        self.metric_group = metric_group
+
+
+class _CustomMetricGroup(MetricGroup):
+    def get_sub_group(
+        self, name: str, value: str | None = None
+    ) -> "_CustomMetricGroup":
+        return self
+
+    def get_counter(self, name: str) -> Any:
+        raise NotImplementedError
+
+    def get_meter(self, name: str) -> Any:
+        raise NotImplementedError
+
+    def get_histogram(self, name: str, window_size: int = 100) -> Any:
+        raise NotImplementedError
+
+    def get_gauge(self, name: str) -> Any:
+        raise NotImplementedError
+
+
+class _JavaMetricGroup:
+    def __init__(self) -> None:
+        self.j_metric_group = object()
+
+
+@pytest.mark.parametrize(
+    "resource",
+    [
+        JavaChatModelConnectionImpl(
+            j_resource=_JavaResource(), j_resource_adapter=None
+        ),
+        JavaChatModelSetupImpl(
+            j_resource=_JavaResource(),
+            j_resource_adapter=None,
+            connection="connection",
+            model="model",
+        ),
+        JavaEmbeddingModelConnectionImpl(
+            j_resource=_JavaResource(), j_resource_adapter=None
+        ),
+        JavaEmbeddingModelSetupImpl(
+            j_resource=_JavaResource(),
+            j_resource_adapter=None,
+            connection="connection",
+            model="model",
+        ),
+    ],
+)
+def test_java_resource_wrappers_forward_metric_group(resource):
+    java_metric_group = _JavaMetricGroup()
+    metric_group = FlinkMetricGroup(java_metric_group.j_metric_group)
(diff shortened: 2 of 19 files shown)


weiqingy (commented):
Thanks for taking this on — a few questions inline.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonResourceAdapterImpl.java
```diff
@@ -200,6 +203,11 @@ public Object callMethod(Object obj, String methodName, Map<String, Object> kwar
         return interpreter.invoke(CALL_METHOD, obj, methodName, kwargs);
     }
 
+    @Override
+    public void setMetricGroup(Object pythonResource, FlinkAgentsMetricGroup metricGroup) {
+        interpreter.invoke(SET_METRIC_GROUP, pythonResource, metricGroup);
```
weiqingy, line 208: This now forwards the action metric group through to the Python provider, which is exactly what issue #857 asked for — and #857 also called out "define metric ownership clearly to avoid double-counting when action-level and provider-level metrics both observe the same response." From what I can see, ownership already looks singular: token metrics are recorded in exactly one place per language — the action handler (`ChatModelAction.recordChatTokenMetrics` → `chatModel.recordTokenMetrics` on the Java side, `chat_model_action` → `_record_token_metrics` on the Python side), each reading the resource's bound group. The provider integrations under `integrations/` record nothing, so there's no second recorder to collide with — this PR widens that single existing recording across the language boundary rather than introducing a second one. So the #857 concern reads as already closed by construction. Would it be worth a sentence in the PR description making that explicit, so the "no double-counting" reasoning is traceable back to the issue?

api/src/main/java/org/apache/flink/agents/api/resource/Resource.java
```diff
@@ -60,6 +61,9 @@ public ResourceContext getResourceContext() {
      */
     public void setMetricGroup(FlinkAgentsMetricGroup metricGroup) {
         this.metricGroup = metricGroup;
+        if (this instanceof PythonResourceWrapper) {
+            ((PythonResourceWrapper) this).setPythonResourceMetricGroup(metricGroup);
```
weiqingy: The abstract base `Resource` (package `...api.resource`) now imports and `instanceof`-checks `PythonResourceWrapper` from the more specialized `...api.resource.python` sub-package. That inverts the usual dependency direction — every Java-only resource (chat models, tools, prompts, vector stores) now carries a compile-time edge to the Python-bridge interface in its base type, and a hypothetical third forwarding flavor would mean editing this base method again rather than overriding it.

One option that keeps the base oblivious to the bridge: have each `Python*` wrapper override `setMetricGroup` to call `super.setMetricGroup(...)` and then `setPythonResourceMetricGroup(...)`, pushing the Python concern down into the wrappers where `getPythonResourceAdapter()` already lives. The PR already adds a new `getPythonResourceAdapter()` override to all eight wrappers, so a per-wrapper `setMetricGroup` override would touch the same set of files — just in the Python layer instead of the base class. Was the centralized `instanceof` chosen deliberately, or mainly to avoid touching the wrappers?

python/flink_agents/runtime/tests/test_cross_language_metric_group.py
```diff
@@ -0,0 +1,86 @@
...
+        ),
+        JavaEmbeddingModelSetupImpl(
+            j_resource=_JavaResource(),
+            j_resource_adapter=None,
+            connection="connection",
+            model="model",
+        ),
+    ],
+)
+def test_java_resource_wrappers_forward_metric_group(resource):
```
weiqingy, line 101: This new file covers the Python→Java direction well, but only that direction — `set_java_resource_metric_group` plus the four `Java*Impl` wrappers. The Java→Python direction's Python half, `python_java_utils.set_metric_group` (`python_java_utils.py:382`), has no direct test: it's the function that wraps the raw Java group in `FlinkMetricGroup` and owns the only `None` branch in the feature, and the Java half is mock-verify only (`testSetMetricGroup` asserts the invoke string, it doesn't execute Python). So that seam isn't exercised end-to-end today.

Would a small unit test in this file's style be worth adding — a fake resource that captures its `set_metric_group` arg, then `set_metric_group(fake, sentinel_j_group)` asserting the captured arg is a `FlinkMetricGroup` whose `_j_metric_group is sentinel_j_group`, plus a `None` case asserting `None` is forwarded? That would pin the wrap-and-None behavior that's currently uncovered.

python/flink_agents/runtime/java/java_resource_wrapper.py
```diff
@@ -27,6 +27,14 @@
 from flink_agents.api.tools.tool import Tool, ToolMetadata, ToolType
 
 
+def set_java_resource_metric_group(j_resource: Any, metric_group: Any) -> None:
+    """Bind the underlying Java metric group to a wrapped Java resource."""
+    if j_resource is None:
+        return
+    j_metric_group = getattr(metric_group, "_j_metric_group", metric_group)
```
weiqingy: In the real flow this is correct — `metric_group` is always a `FlinkMetricGroup` (it comes from `action_metric_group`), so the unwrap hits `_j_metric_group` and forwards the genuine Java group. Low confidence that the edge below matters in practice, so genuinely a question rather than a flag.

`RunnerContext.get_resource(name, type, metric_group=...)` is a documented public API accepting any `MetricGroup` — an abstract base a user could subclass without a `_j_metric_group`. If such a custom group reaches a Java resource wrapper, the silent `, metric_group` fallback would forward a raw Python object into Java `setMetricGroup(FlinkAgentsMetricGroup)`, failing opaquely at the pemja boundary rather than with a clear error. Worth making the fallback explicit (unwrap only a real `FlinkMetricGroup`, else fail loudly), or is that out of scope today given no in-tree caller hits it? (The `None` case is benign — `get_resource` substitutes `action_metric_group` first, and a direct `None` just clears the field — so no concern there.)

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonResourceAdapterImpl.java
joeyutong: @weiqingy One ownership detail I think is still worth calling out: this is not only about avoiding a second recorder after forwarding the metric group. It is also about which effective metric group the action-level recorder should use after the resource call returns.

For example, in the Java action -> Python chat model path, the Python resource may wrap the forwarded action metric group with a provider-specific dimension such as `modelProvider`. Metrics recorded inside the Python resource would use that wrapped group, but token metrics are still recorded later by the Java action/model-setup path, which only sees the Java wrapper's metric group. That means action-recorded token metrics may still miss the provider-specific dimension.

So I think the ownership question also needs to define whether a resource-derived metric scope should flow back to the action-level recorder. This PR fixes forwarding into the resource, but not that return path.

weiqingy: Thanks, that makes sense. My earlier comment was focused on the double-recording angle, but I agree there is a separate ownership question around the effective metric scope.

My takeaway is that #860 currently fixes the forward propagation path into the cross-language resource, but it does not settle whether a resource/provider-derived metric scope should flow back to the action-level token recorder. For example, if the Python resource wraps the forwarded group with provider-specific dimensions, Java-side token recording after the call may still use only the Java wrapper/action scope.

If we keep #860 scoped to forward propagation, it may be good to call this out in the PR description or track it as a follow-up, so it is clear that the return-path/effective-scope behavior is still unresolved. Does that match what you had in mind?

goutamadwant:
@weiqingy @joeyutong Pushed an update addressing the review comments.

The base Resource class no longer depends on the Python bridge. Metric forwarding now lives in the Python-backed wrappers, which call super.setMetricGroup(...) and then forward to the wrapped Python resource.

I also added direct coverage for python_java_utils.set_metric_group(...), including the raw Java group wrapping path and the None path. On the Python-to-Java side, set_java_resource_metric_group(...) now accepts only FlinkMetricGroup or None and raises a clear TypeError for unsupported custom metric groups.

I also updated the PR description to make the scope explicit: this PR handles forward metric-group propagation and keeps token recording single-recorder/no double-counting. The return-path/effective-scope behavior is still separate follow-up work. Let me know if you have any suggestions. thanks!

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonResourceAdapterImpl.java
joeyutong: Thanks, that matches my understanding.

I agree that #860 can stay focused on the forward propagation path, and the return-path / effective-scope behavior can be tracked separately.

@goutamadwant Since this PR already calls out that limitation, would you mind filing a follow-up issue for it? That would make the remaining ownership question easier to continue without blocking this PR.

joeyutong:
LGTM. CC @wenjin272

joeyutong (approved):

PR #861, opened 2026-06-26
[api][plan][python] Use request-scoped metric groups for chat token metrics

### LINKED ISSUE #859: [Bug] Action-scoped metrics can be recorded under the wrong action for cached resources
### Search before asking

### Description

Action-scoped metrics can be recorded under the wrong action when a cached resource is reused across actions.

Flink Agents injects the current action metric group into a resource when `RunnerContext.get_resource(...)` / `RunnerContext.getResource(...)` returns it. However, resources are cached and shared. If action A gets a chat model resource, then yields or waits across an async/durable boundary, action B can later get the same cached resource and overwrite its metric group with B's action scope. When action A resumes and records token metrics by reading the metric group from the resource field, those metrics may be recorded under B's action scope.

This can affect paths where metrics are recorded after the request returns rather than at the moment the action obtains the resource. For example:

- Python chat token metrics after `durable_execute` / `durable_execute_async`.
- Java chat token metrics after the chat response is returned.
- Shared cached resources used by multiple actions.
- Cross-language wrappers and provider resources, where the wrapper and underlying provider may have separate metric group state.

The expected behavior is that token metrics are recorded under the action scope that initiated the request. The metric group used for delayed metric recording should not depend on mutable state stored on a cached resource after another action may have rebound it.

### How to reproduce

One minimal reproduction shape is:

1. Define two actions that share the same chat model resource.
2. Let action A obtain the chat model resource and start a chat request.
3. Before action A records token metrics, let action B obtain the same cached chat model resource, causing the resource metric group to be rebound to B's action scope.
4. Resume action A and record token metrics from the chat response.
5. Observe that the token counters can be registered under B's action metric scope instead of A's.

A unit-level reproduction can simulate the same condition by:

1. Creating a setup whose connection wraps metric groups with provider dimensions.
2. Binding the setup to action A's metric group.
3. Rebinding the same setup/resource to action B's metric group before recording token metrics.
4. Recording token metrics through the setup.
5. Verifying that the counter follows the latest mutable resource metric group rather than the action A group that initiated the request.

### Version and environment

Observed from the current main-branch code path:

- Python `FlinkRunnerContext.get_resource(...)` binds the current action metric group to the cached resource.
- Java `RunnerContextImpl.getResource(...)` binds the current action metric group to the cached resource.
- Python and Java chat token metrics are recorded after the chat response returns and read the metric group from the chat model resource.

This is independent of a specific deployment mode; it is a resource lifecycle / metric binding issue.

jlon:
## What
- Capture the current action metric group when a chat request is initiated.
- Record delayed chat token metrics against that captured group instead of the mutable metric group on cached chat model resources.
- Add Java and Python regression coverage for resource metric-group rebinding.

## Why
Cached resources can be rebound when another action retrieves the same resource before the original request records token usage. The delayed token metrics should stay under the action that initiated the request.

Closes #859.

## Tests
- mvn -pl api -DskipITs -Dspotless.check.skip=true -Dcheckstyle.skip=true -Dtest=BaseChatModelSetupTokenMetricsTest test
- mvn -pl plan -am -DskipITs -Dspotless.check.skip=true -Dcheckstyle.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=ChatModelActionTest,ChatModelActionRetryTest test
- bash -lc 'export PATH="$HOME/.local/bin:$PATH"; cd python && uv run --no-sync pytest flink_agents/api/chat_models/tests/test_token_metrics.py -q'
- bash -lc 'export PATH="$HOME/.local/bin:$PATH"; cd python && uv run --no-sync ruff format --check flink_agents/api/chat_models/chat_model.py flink_agents/api/chat_models/tests/test_token_metrics.py flink_agents/plan/actions/chat_model_action.py && uv run --no-sync ruff check flink_agents/api/chat_models/chat_model.py flink_agents/api/chat_models/tests/test_token_metrics.py flink_agents/plan/actions/chat_model_action.py'
- bash -lc 'export JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64; export PATH="$JAVA_HOME/bin:$PATH"; mvn -pl api,plan -am spotless:check -Dspotless.skip=false -DskipTests -DskipITs -Dcheckstyle.skip=true'

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
@@ -116,7 +116,22 @@ public abstract class BaseChatModelSetup extends Resource {
      * @param completionTokens the number of completion tokens
      */
     public void recordTokenMetrics(String modelName, long promptTokens, long completionTokens) {
-        FlinkAgentsMetricGroup metricGroup = getMetricGroup();
+        recordTokenMetrics(getMetricGroup(), modelName, promptTokens, completionTokens);
+    }
+
+    /**
+     * Record token usage metrics for the given model on the provided metric group.
+     *
+     * @param metricGroup the metric group captured when the request was initiated
+     * @param modelName the name of the model used
+     * @param promptTokens the number of prompt tokens
+     * @param completionTokens the number of completion tokens
+     */
+    public void recordTokenMetrics(
+            @Nullable FlinkAgentsMetricGroup metricGroup,
+            String modelName,
+            long promptTokens,
+            long completionTokens) {
         if (metricGroup == null) {
             return;
         }
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
@@ -109,33 +109,21 @@ public abstract class BaseChatModelSetup extends Resource {
     public abstract Map<String, Object> getParameters();
 
     /**
-     * Record token usage metrics for the given model on this setup's bound metric group.
-     *
-     * @param modelName the name of the model used
-     * @param promptTokens the number of prompt tokens
-     * @param completionTokens the number of completion tokens
-     */
-    public void recordTokenMetrics(String modelName, long promptTokens, long completionTokens) {
-        recordTokenMetrics(getMetricGroup(), modelName, promptTokens, completionTokens);
-    }
-
-    /**
      * Record token usage metrics for the given model on the provided metric group.
      *
-     * @param metricGroup the metric group captured when the request was initiated
+     * @param metricGroup the non-null metric group captured when the request was initiated
      * @param modelName the name of the model used
      * @param promptTokens the number of prompt tokens
      * @param completionTokens the number of completion tokens
      */
     public void recordTokenMetrics(
-            @Nullable FlinkAgentsMetricGroup metricGroup,
+            FlinkAgentsMetricGroup metricGroup,
             String modelName,
             long promptTokens,
             long completionTokens) {
-        if (metricGroup == null) {
-            return;
-        }
-        FlinkAgentsMetricGroup modelGroup = metricGroup.getSubGroup("model", modelName);
+        FlinkAgentsMetricGroup modelGroup =
+                Preconditions.checkNotNull(metricGroup, "Metric group must not be null.")
+                        .getSubGroup("model", modelName);
         modelGroup.getCounter("promptTokens").inc(promptTokens);
         modelGroup.getCounter("completionTokens").inc(completionTokens);
     }
--- a/python/flink_agents/api/chat_models/chat_model.py
+++ b/python/flink_agents/api/chat_models/chat_model.py
@@ -27,6 +27,7 @@ from flink_agents.api.chat_message import (
     MessageRole,
     find_first_system_message,
 )
+from flink_agents.api.metric_group import MetricGroup
 from flink_agents.api.prompts.prompt import Prompt
 from flink_agents.api.resource import Resource, ResourceType
 from flink_agents.api.skills import BASH_TOOL, LOAD_SKILL_TOOL
@@ -261,7 +262,11 @@ class BaseChatModelSetup(Resource):
         )
 
     def _record_token_metrics(
-        self, model_name: str, prompt_tokens: int, completion_tokens: int
+        self,
+        model_name: str,
+        prompt_tokens: int,
+        completion_tokens: int,
+        metric_group: MetricGroup | None = None,
     ) -> None:
         """Record token usage metrics for the given model.
 
@@ -273,8 +278,12 @@ class BaseChatModelSetup(Resource):
             The number of prompt tokens
         completion_tokens : int
             The number of completion tokens
+        metric_group : MetricGroup | None
+            The metric group captured when the request was initiated. If not provided,
+            this resource's currently bound metric group is used.
         """
-        metric_group = self.metric_group
+        if metric_group is None:
+            metric_group = self.metric_group
         if metric_group is None:
             return
 
--- a/python/flink_agents/api/chat_models/chat_model.py
+++ b/python/flink_agents/api/chat_models/chat_model.py
@@ -266,7 +266,7 @@ class BaseChatModelSetup(Resource):
         model_name: str,
         prompt_tokens: int,
         completion_tokens: int,
-        metric_group: MetricGroup | None = None,
+        metric_group: MetricGroup | None,
     ) -> None:
         """Record token usage metrics for the given model.
 
@@ -279,11 +279,9 @@ class BaseChatModelSetup(Resource):
         completion_tokens : int
             The number of completion tokens
         metric_group : MetricGroup | None
-            The metric group captured when the request was initiated. If not provided,
-            this resource's currently bound metric group is used.
+            The metric group captured when the request was initiated. If None, token
+            metrics are skipped.
         """
-        if metric_group is None:
-            metric_group = self.metric_group
         if metric_group is None:
             return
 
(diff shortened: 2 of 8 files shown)


python/flink_agents/api/chat_models/chat_model.py
```diff
@@ -273,8 +278,12 @@ def _record_token_metrics(
             The number of prompt tokens
         completion_tokens : int
             The number of completion tokens
+        metric_group : MetricGroup | None
+            The metric group captured when the request was initiated. If not provided,
+            this resource's currently bound metric group is used.
         """
-        metric_group = self.metric_group
+        if metric_group is None:
```
weiqingy: The two languages now document slightly different contracts for the absent-group case: Java's 4-arg `recordTokenMetrics` treats a `null` group as "skip recording" (`BaseChatModelSetup.java:135`), while here `metric_group is None` falls back to `self.metric_group` — the live, possibly-rebound group. Every real call site passes a captured group (`chat_model_action.py:334`), so there's no impact on today's flow. The part I'm less sure about: a future caller passing `None` expecting Java parity would record under whatever action is currently bound — the #859 scenario again on the Python side, where Java would no-op instead. Is the asymmetry intentional? If the `= None` default is only there to keep the param optional for the test helper, dropping it so callers must pass the captured group would match Java's no-fallback contract; alternatively a one-line javadoc note that Java's `null` means skip would at least make each side self-documenting.

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
```diff
@@ -116,7 +116,22 @@ public void open() throws Exception {
      * @param completionTokens the number of completion tokens
      */
     public void recordTokenMetrics(String modelName, long promptTokens, long completionTokens) {
```
wenjin272: It appears that only tests call this method, so I think it can be removed directly.

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
```diff
@@ -116,7 +116,22 @@ public void open() throws Exception {
...
+    /**
+     * Record token usage metrics for the given model on the provided metric group.
+     *
+     * @param metricGroup the metric group captured when the request was initiated
+     * @param modelName the name of the model used
+     * @param promptTokens the number of prompt tokens
+     * @param completionTokens the number of completion tokens
+     */
+    public void recordTokenMetrics(
+            @Nullable FlinkAgentsMetricGroup metricGroup,
```
wenjin272: In practice, we now require that the `metricGroup` of the current action be passed in when reporting metrics. Therefore, I believe `metricGroup` should not be nullable.

wenjin272 (commented):
Thanks for taking this on @jlon.

Since `Resource` instances are shared across multiple Actions, reporting metrics using the `metricGroup` bound at creation time can lead to incorrect metric matching. This PR addresses the issue by passing the current Action’s `metricGroup` at reporting time, ensuring metrics are reported to the correct `Action` scope. 

In this case, I think `Resource` no longer need to hold a `metricGroup` field.

joeyutong:
> Thanks for taking this on @jlon.
> 
> Since `Resource` instances are shared across multiple Actions, reporting metrics using the `metricGroup` bound at creation time can lead to incorrect metric matching. This PR addresses the issue by passing the current Action’s `metricGroup` at reporting time, ensuring metrics are reported to the correct `Action` scope.
> 
> In this case, I think `Resource` no longer need to hold a `metricGroup` field.

I agree. This PR fixes the concrete bug locally for chat token metrics by passing the request/action metric group at reporting time, instead of reading the mutable metric group currently bound to the cached resource. But the underlying model issue is broader than this one recording path.

A cleaner abstraction would be to separate the stable resource instance from the per-action resource-use context:

- `Resource` is a stable cached capability.
- `ResourceUseContext` carries action-scoped state, including the current action metric group.
- Metric reporting uses the current `ResourceUseContext`, not mutable state stored on the resource.
- Cross-language bridges propagate this context when invoking resource methods or resolving nested resources.

One possible shape is for `ctx.getResource(...)` to return a bound resource handle internally, i.e. the cached `Resource` plus the current `ResourceUseContext`, even if the public API is migrated gradually.

Under that model, `Resource` would no longer need a mutable `metricGroup` field. It would also simplify the cross-language propagation work in [#860](https://github.com/apache/flink-agents/pull/860): instead of making every Python wrapper participate in `setMetricGroup(...)`, the bridge can pass the current use context through the existing `callMethod` / `getResource` path. CC @weiqingy 

This is probably a larger refactor than this PR should take on, but I agree with the direction: action-scoped metric state should not be owned by the cached `Resource` instance.

python/flink_agents/api/chat_models/chat_model.py
jlon: Updated in ecd96c5. `_record_token_metrics` now requires the metric group argument and no longer falls back to the resource-bound `self.metric_group`. The action path also skips token metric recording when no request/action metric group is available.

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
jlon: Updated in ecd96c5. Removed the 3-arg Java `recordTokenMetrics(...)` overload and updated the tests to pass an explicit request/action metric group.

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
jlon: Updated in ecd96c5. The Java `recordTokenMetrics(...)` metric group parameter is no longer nullable, and `ChatModelAction` now skips token metric recording before calling it when the action metric group is absent.

jlon:
Pushed ecd96c5 to address the review comments:

- Removed the 3-arg Java `recordTokenMetrics(...)` overload so token metrics are recorded only with an explicit request/action metric group.
- Made the Java metric group parameter non-null and added an action-layer skip when no metric group is available.
- Removed the Python fallback from `metric_group=None` to `self.metric_group`; Python now skips instead of recording under a possibly rebound resource group.

I kept the broader `Resource.metricGroup` removal out of this PR because that crosses resource binding and cross-language bridge behavior. This patch keeps the issue #859 fix focused on chat token metrics while avoiding the fallback contract that could recreate the bug.

Local verification:
- `mvn -pl api -DskipITs -Dspotless.check.skip=true -Dcheckstyle.skip=true -Dtest=BaseChatModelSetupTokenMetricsTest test`
- `mvn -pl plan -am -DskipITs -Dspotless.check.skip=true -Dcheckstyle.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=ChatModelActionTest,ChatModelActionRetryTest test`
- `cd python && uv run --no-sync pytest flink_agents/api/chat_models/tests/test_token_metrics.py -q`
- `cd python && uv run --no-sync ruff format --check ... && uv run --no-sync ruff check ...`
- `mvn -pl api,plan -am spotless:check -Dspotless.skip=false -DskipTests -DskipITs -Dcheckstyle.skip=true`
- `git diff --check`

python/flink_agents/api/chat_models/chat_model.py
```diff
@@ -273,8 +278,10 @@ def _record_token_metrics(
             The number of prompt tokens
         completion_tokens : int
             The number of completion tokens
+        metric_group : MetricGroup | None
+            The metric group captured when the request was initiated. If None, token
+            metrics are skipped.
         """
-        metric_group = self.metric_group
         if metric_group is None:
```
wenjin272, line 285: One parity question: Python `_record_token_metrics(..., metric_group=None)` is a no-op, while Java `recordTokenMetrics(null, ...)` fails via `checkNotNull`. Should we align the lower-level helper contract?

wenjin272 (commented):
I agree with excluding the removal of `metricGroup` bindings from this PR. We should create a separate issue to evaluate this holistically. 

I’ve left one comment regarding this PR.

Issue #866, opened 2026-07-03
[Bug] Python AgentConfigOptions starts PyFlink gateway during Pemja worker initialization


aka-dag:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

When a Python action is initialized in the TaskManager-side Pemja worker, `PythonActionExecutor.open()` imports `flink_agents.runtime.flink_runner_context`.
This import chain eventually imports `flink_agents.api.core_options`:
`flink_runner_context -> resource_cache -> resource_context -> agent_plan -> chat_model_action -> core_options`
During `core_options` import, `AgentConfigOptionsMeta` resolves Java-side `AgentConfigOptions` through `pyflink.java_gateway.get_gateway()`. This starts a PyFlink gateway process from inside the Pemja worker, which is unexpected because Python is already embedded in the Flink JVM and should access Java objects through Pemja instead of launching a separate PyFlink gateway.
In some deployment environments, this causes action initialization to fail before the operator is opened.

#### Expected behavior:
Importing Flink Agents runtime modules in the Pemja worker should not start a PyFlink Java gateway. Python config options should be available without calling `get_gateway()` during runtime initialization.

#### Actual behavior:
`core_options` calls `pyflink.java_gateway.get_gateway()` during import, and the TaskManager task fails during `ActionExecutionOperator.open()`.

#### Relevant sanitized stack trace:
```python
pemja.core.PythonException: <class 'FileNotFoundError'>: 2
    at .../subprocess._execute_child(subprocess.py:1955)
    at .../subprocess.__init__(subprocess.py:1026)
    at .../pyflink/pyflink_gateway_server.launch_gateway_server_process(...)
    at .../pyflink/java_gateway.launch_gateway(...)
    at .../pyflink/java_gateway.get_gateway(...)
    at .../flink_agents/api/core_options.py:61
    at .../flink_agents/api/core_options.py:111
    at .../flink_agents/plan/actions/chat_model_action.py:33
    at .../flink_agents/plan/agent_plan.py:39
    at .../flink_agents/runtime/resource_context.py:26
    at .../flink_agents/runtime/resource_cache.py:25
    at .../flink_agents/runtime/flink_runner_context.py:54
    at <string>.<module>(<string>:2)
    at pemja.core.PythonInterpreter.exec(Native Method)
[...]

PR #867, opened 2026-07-03
[python] Remove PyFlink gateway fallback from AgentConfigOptions

aka-dag:
Python AgentConfigOptions used a metaclass to resolve missing options from Java via get_gateway(), which can fail during Pemja worker initialization. Declare config options explicitly in Python instead.

Linked issue: #866 

### Purpose of change

Python `AgentConfigOptions` used a metaclass fallback to resolve missing options from Java through `pyflink.java_gateway.get_gateway()`. This can start a PyFlink gateway while importing runtime modules inside the Pemja worker.
This change removes the Java-backed dynamic lookup and declares Python config options explicitly.

### Tests

- Added/updated Python option compatibility coverage.
- Added regression coverage that importing runtime modules does not trigger `get_gateway()`.
- Ran `./tools/ut.sh -p`.
- Ran `./tools/lint.sh`.

### API

This changes the implementation of Python `AgentConfigOptions`. Publicly supported options remain accessible as explicit Python `ConfigOption` declarations. Dynamic access to Java-only options is removed.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`


wenjin272:
Thanks for taking this on @aka-dag. Could you help review this PR @GreatEugenius ?

chenzihao5:
Cross-language calls are no longer involved after this change. The name that create_python_option_from_java_option is somewhat misleading. I think we need another way to check the consistency.

python/flink_agents/plan/tests/compatibility/create_python_option_from_java_option.py
```diff
@@ -15,36 +15,27 @@
...
 
-# This script is used to verify that Java-defined configuration options
-# (e.g., AgentConfigOptions) are correctly exposed and accessible in the
-# Python environment via the JAR file. It loads a Java JAR into the Python
-# context and performs basic assertions on the configuration keys, types,
-# and default values to ensure compatibility between Java and Python layers.
-#
-# The JAR file path is relative to this script and should be updated if
-# the build structure changes.
+# This script verifies that Python configuration options stay aligned with the
```
weiqingy: This comment still says the script verifies the Python options stay aligned with Java, but after dropping the JAR load it only checks the Python declarations against hardcoded literals — so it would stay green even if the Java `AgentConfigOptions` side drifted (a new option, a changed default). One thing worth noting: this script runs as a standalone process in `test_java_config_in_python.sh`, not inside the Pemja worker, so the `#866` gateway concern doesn't apply here — it could keep loading the JAR and compare each Python `ConfigOption` against the Java key/type/default. Would it be worth keeping that cross-check (here or in a dedicated parity test) so the two sides can't silently drift?

weiqingy (commented):
Thanks for taking this on — declaring the options explicitly is the right fix for the Pemja gateway problem, and I cross-checked all 23 against the Java `AgentConfigOptions` (keys, types, defaults, including the `EVENT_LOG_LEVEL` switch to the enum) — they line up exactly. Two questions below.

1. The description mentions a regression test asserting that importing the runtime modules doesn't trigger `get_gateway()`, but I don't see it in the diff — just `core_options.py` and the compatibility script. That guard is really what locks in this fix: if a future change re-introduces a Java-backed lookup somewhere on the `flink_runner_context → … → core_options` import chain, nothing here would catch the regression. Was that test meant to be part of this PR?

python/flink_agents/plan/tests/compatibility/create_python_option_from_java_option.py
GreatEugenius: +1

GreatEugenius:
Hi @aka-dag, the approach looks good to me. I have one follow-up before merging: please add the Java/Python parity check in CI to ensure that the two AgentConfigOptions definitions do not drift silently.

python/flink_agents/plan/tests/compatibility/create_python_option_from_java_option.py
aka-dag: Thanks for the detailed review — agreed on both points.

**Regression test:** Added  `python/flink_agents/api/tests/test_core_options.py` to lock in the Pemja import fix:

- AST check: `core_options.py` does not `import pyflink`

- Import tests: `flink_agents.api.core_options` and `flink_agents.runtime.flink_runner_context` do not call `get_gateway()`

- Sanity checks: all 23 options are explicitly declared; unknown attributes raise `AttributeError`

**Java/Python parity:** Restored JAR-based cross-check in `create_python_option_from_java_option.py` with shared helpers in `java_python_config_option_parity.py` (key / type / default for each option). This script still runs via `test_java_config_in_python.sh` as a standalone client-side process — using `get_gateway()` here is intentional and separate from the runtime import path we are fixing. The comment now clarifies this is a client-side parity script, not the Pemja runtime path.

aka-dag:
> Hi @aka-dag, the approach looks good to me. I have one follow-up before merging: please add the Java/Python parity check in CI to ensure that the two AgentConfigOptions definitions do not drift silently.

Hi @GreatEugenius, thanks for the review.

I've restored the JAR-based Java/Python parity check in `create_python_option_from_java_option.py`. It compares every Python ConfigOption (key, type, default) against the Java `AgentConfigOptions` / `AgentExecutionOptions` definitions.

This script is already invoked in CI via `e2e-test/test-scripts/test_java_config_in_python.sh` (called from `tools/e2e.sh` in the e2e workflow). With the JAR loading restored, a Java-side change that Python doesn't mirror will fail that step instead of drifting silently.

The runtime regression guard (`test_core_options.py`) runs in the regular Python unit test suite (`./tools/ut.sh -p`).

python/flink_agents/plan/tests/compatibility/java_python_config_option_parity.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+
+def assert_options_class_matches_java(
+    python_options_class: type,
+    java_options_class: Any,
+) -> None:
+    """Compare every Python ConfigOption on a class with the Java static field."""
+    python_options = collect_config_options(python_options_class)
+    for option_name, python_option in sorted(python_options.items()):
```
weiqingy: Now that the JAR load is back this catches Python-side drift and mismatched defaults on shared options nicely. One direction still slips through, though: the loop iterates `collect_config_options(python_options_class)`, so it only ever visits options that already exist on the Python side — a Java `AgentConfigOptions`/`AgentExecutionOptions` option with no Python counterpart is never looked up. The count guard back in `create_python_option_from_java_option.py:56` (`== 23`) counts the Python side too. So if someone adds an option to Java and forgets the Python mirror, this stays green — which is the "drift silently" case the check is meant to guard. Would it be worth also enumerating the Java option fields (or asserting the Java option count matches Python) so a Java-only addition trips the check?

python/flink_agents/plan/tests/compatibility/java_python_config_option_parity.py
```diff
@@ -0,0 +1,114 @@
...
+def _java_type_matches_python(java_type_name: str, python_config_type: type) -> bool:
+    expected_primitive = _JAVA_PRIMITIVE_TYPE_TO_PYTHON.get(java_type_name)
+    java_simple_name = java_type_name.rsplit(".", maxsplit=1)[-1]
+    if "$" in java_simple_name:
+        java_simple_name = java_simple_name.split("$", maxsplit=1)[-1]
+
+    if expected_primitive is not None:
+        return python_config_type is expected_primitive
+
+    if isinstance(python_config_type, type) and issubclass(python_config_type, Enum):
```
weiqingy: Small one: this `Enum` branch returns `python_config_type.__name__ == java_simple_name`, which is the same expression as the fallback `return` two lines below — so the special-case doesn't change the result. Was it meant to compare something enum-specific (e.g. the member set), or can it just fold into the fallback?

python/flink_agents/plan/tests/compatibility/java_python_config_option_parity.py
aka-dag: Thanks for pointing out. [4922e40](https://github.com/apache/flink-agents/pull/867/commits/4922e40bacb9f4705204fd7bfc5d19c93cda12cc) I've added Java-side field enumeration via reflection to collect all static `ConfigOption` fields in `collect_java_config_options()`. Accordingly, removed the hardcoded `len == 23` check since the bidirectional name set comparison covers count against Java dynamically.

python/flink_agents/plan/tests/compatibility/java_python_config_option_parity.py
aka-dag: [4922e40](https://github.com/apache/flink-agents/pull/867/commits/4922e40bacb9f4705204fd7bfc5d19c93cda12cc) Removed the redundant Enum branch.

python/flink_agents/plan/tests/compatibility/create_python_option_from_java_option.py
wenjin272, line 1: It is recommended to rename it to `check_java_python_config_options_parity`, as Python options are no longer derived from Java.

python/flink_agents/plan/tests/compatibility/check_java_python_config_options_parity.py
wenjin272, line 1: Could we merge this into `create_python_option_from_java_option.py`?

python/flink_agents/api/tests/test_core_options.py
```diff
@@ -0,0 +1,108 @@
...
+import ast
+import importlib
+import sys
+import types
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from flink_agents.plan.tests.compatibility.java_python_config_option_parity import (
```
wenjin272: Avoid making the api module depend on the plan module.

wenjin272 (commented):
Hi, @aka-dag, overall looks good to me. I left some minor comments.

python/flink_agents/api/tests/test_core_options.py
aka-dag: Thanks for the suggestions — all addressed in [3ecd1f3](https://github.com/apache/flink-agents/pull/867/commits/3ecd1f39a628d9848424ef61732c1b10616bb0a6):

* Renamed and merged into `check_java_python_config_options_parity.py`.
* Updated `test_java_config_in_python.sh` to invoke the new script.
* Removed the api → plan import from `test_core_options.py`; the small collector helper is now local to that test file.

wenjin272 (approved):
LGTM

Issue #868, opened 2026-07-03
[Bug] ActionStateSerde cannot deserialize built-in event subclasses during durable recovery


rosemarYuan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Problem

`ActionStateSerdeTest` currently covers `InputEvent` and `OutputEvent` round-trip only. Those two classes have `@JsonCreator` on their `(UUID, Map<String, Object>)` constructors, so they deserialize correctly.

However, the built-in event classes under `org.apache.flink.agents.api.event.*`, such as `ChatRequestEvent`, `ChatResponseEvent`, `ToolRequestEvent`, `ToolResponseEvent`, `ContextRetrievalRequestEvent`, and `ContextRetrievalResponseEvent`, also have `(UUID, Map<String, Object>)` constructors but do not annotate them with `@JsonCreator/@JsonProperty`.

This matters because durable execution persists events in `ActionState`, including the triggering `taskEvent` and completed `outputEvents`. Both Kafka and Fluss action state stores delegate to `ActionStateSerde`, so recovery/rebuild needs Jackson to reconstruct these concrete event subclasses from the serialized `@class` metadata.

A straightforward fix would be to annotate these `(UUID, Map<String, Object>)` constructors consistently with `InputEvent`/`OutputEvent`, or otherwise teach `ActionStateSerde` how to reconstruct these built-in event subclasses during durable state recovery.

One extra thing worth testing after the constructor fix: nested attributes should also come back as the expected types, e.g. `ChatMessage`, `ToolResponse`, and `Document`, rather than raw `Map` values.

## Background

This looks related to #631, which introduced the cross-language unified event model. That change moved built-in event data into a common `type` + `attributes` representation and added `(UUID id, Map<String, Object> attributes)` constructors to built-in event subclasses for typed reconstruction via `fromEvent()`.

However, in the #631 merge commit (`071c3a0a`), only `InputEvent` and `OutputEvent` annotated these constructors with `@JsonCreator/@JsonProperty`. Other built-in event subclasses, such as `ChatRequestEvent`, `ChatResponseEvent`, `ToolRequestEvent`, `ToolResponseEvent`, `ContextRetrievalRequestEvent`, and `ContextRetrievalResponseEvent`, have the same `(UUID, Map<String, Object>)` constructors but lack Jackson creator annotations.

[...]

PR #869, opened 2026-07-03
[api][runtime][test] Add @JsonCreator to built-in event subclasses for ActionStateSerde durable recovery

rosemarYuan:
Linked issue: #868

### Purpose of change

`ActionStateSerde` uses Jackson polymorphic deserialization (`@JsonTypeInfo(Id.CLASS)`) to restore the concrete `Event` subclasses persisted in `ActionState` (`taskEvent` / `outputEvents`) for durable execution recovery. Jackson requires each concrete subclass to declare its own creator — the `@JsonCreator` on the base `Event` constructor is not inherited.

Only `InputEvent`/`OutputEvent` annotate their `(UUID, Map<String, Object>)` constructors with `@JsonCreator/@JsonProperty`. The six built-in event subclasses under `org.apache.flink.agents.api.event.*` (`ChatRequestEvent`, `ChatResponseEvent`, `ToolRequestEvent`, `ToolResponseEvent`, `ContextRetrievalRequestEvent`, `ContextRetrievalResponseEvent`) have the same constructors but lack the annotations, so deserializing any of them from a persisted `ActionState` fails with:

```
InvalidDefinitionException: Cannot construct instance of `...ChatRequestEvent`
(no Creators, like default constructor, exist)
```

This PR annotates the `(UUID, Map<String, Object>)` constructor of all six classes, consistent with `InputEvent`/`OutputEvent`.

### Tests

Added `ActionStateSerdeTest#testBuiltinChatToolAndContextEventsRoundTripThroughOutputEvents`, which:

- round-trips all six built-in events through `ActionStateSerde` both as `taskEvent` and inside `outputEvents`, asserting the concrete class is restored (fails with the exception above without the fix);
- additionally asserts that nested attributes come back as their expected types via the `fromEvent()` consumption path used by the built-in actions (`ChatMessage`, `ToolResponse`, `Document`, and `UUID` request ids), not raw maps.

Verified with `mvn -pl runtime test -Dtest=ActionStateSerdeTest` (11/11 pass) and the full `api` module test suite.

### API

No signature changes. Adds Jackson annotations to existing public constructors of the six built-in event classes; serialized JSON format is unchanged.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/event/ChatResponseEvent.java
+++ b/api/src/main/java/org/apache/flink/agents/api/event/ChatResponseEvent.java
@@ -18,7 +18,9 @@
 
 package org.apache.flink.agents.api.event;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.api.chat.messages.ChatMessage;
@@ -46,7 +48,10 @@ public class ChatResponseEvent extends Event {
         setAttr("total_retry_wait_sec", totalRetryWaitSec);
     }
 
-    public ChatResponseEvent(UUID id, Map<String, Object> attributes) {
+    @JsonCreator
+    public ChatResponseEvent(
+            @JsonProperty("id") UUID id,
+            @JsonProperty("attributes") Map<String, Object> attributes) {
         super(id, EVENT_TYPE, attributes);
     }
 
--- a/api/src/main/java/org/apache/flink/agents/api/event/ChatResponseEvent.java
+++ b/api/src/main/java/org/apache/flink/agents/api/event/ChatResponseEvent.java
@@ -52,7 +52,20 @@ public class ChatResponseEvent extends Event {
     public ChatResponseEvent(
             @JsonProperty("id") UUID id,
             @JsonProperty("attributes") Map<String, Object> attributes) {
-        super(id, EVENT_TYPE, attributes);
+        super(id, EVENT_TYPE, normalizeAttributes(attributes));
+    }
+
+    /** Converts nested attributes back to their typed forms. */
+    private static Map<String, Object> normalizeAttributes(Map<String, Object> attributes) {
+        Object rawId = attributes.get("request_id");
+        if (rawId instanceof String) {
+            attributes.put("request_id", UUID.fromString((String) rawId));
+        }
+        Object rawResponse = attributes.get("response");
+        if (rawResponse instanceof Map) {
+            attributes.put("response", MAPPER.convertValue(rawResponse, ChatMessage.class));
+        }
+        return attributes;
     }
 
     /**
@@ -61,18 +74,9 @@ public class ChatResponseEvent extends Event {
      * @param event the base event containing chat response data in attributes
      * @return a typed ChatResponseEvent
      */
-    @SuppressWarnings("unchecked")
     public static ChatResponseEvent fromEvent(Event event) {
-        Map<String, Object> attrs = new HashMap<>(event.getAttributes());
-        Object rawId = attrs.get("request_id");
-        if (rawId instanceof String) {
-            attrs.put("request_id", UUID.fromString((String) rawId));
-        }
-        Object rawResponse = attrs.get("response");
-        if (rawResponse instanceof Map) {
-            attrs.put("response", MAPPER.convertValue(rawResponse, ChatMessage.class));
-        }
-        ChatResponseEvent result = new ChatResponseEvent(event.getId(), attrs);
+        ChatResponseEvent result =
+                new ChatResponseEvent(event.getId(), new HashMap<>(event.getAttributes()));
         if (event.hasSourceTimestamp()) {
             result.setSourceTimestamp(event.getSourceTimestamp());
         }
--- a/runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerdeTest.java
+++ b/runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerdeTest.java
@@ -381,24 +381,21 @@ public class ActionStateSerdeTest {
         assertEquals(ContextRetrievalRequestEvent.class, outputEvents.get(4).getClass());
         assertEquals(ContextRetrievalResponseEvent.class, outputEvents.get(5).getClass());
 
-        // Replayed events are consumed through fromEvent() (see ChatModelAction,
-        // ToolCallAction, ContextRetrievalAction), which must restore nested
-        // attributes degraded to raw maps by the JSON round-trip back to their
-        // typed forms.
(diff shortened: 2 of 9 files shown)


runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerdeTest.java
```diff
@@ -333,4 +344,63 @@ public void testKafkaSederDelegatesToActionStateSerde() throws Exception {
         assertNull(kafkaSeder.serialize("test-topic", null));
         assertNull(kafkaSeder.deserialize("test-topic", null));
     }
+
+    @Test
+    public void testBuiltinChatToolAndContextEventsRoundTripThroughOutputEvents() throws Exception {
```
wenjin272: The test case name is overly verbose and its meaning is unclear. suggest to: `testBuiltInEventSerDeRoundTrip`

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerdeTest.java
```diff
@@ -333,4 +344,63 @@ public void testKafkaSederDelegatesToActionStateSerde() throws Exception {
...
+                new ToolResponseEvent(
+                        requestId,
+                        Map.of("call-1", ToolResponse.success("result")),
+                        Map.of("call-1", true),
+                        Map.of()));
+        originalState.addEvent(new ContextRetrievalRequestEvent("query text", "myVectorStore", 5));
+        originalState.addEvent(
+                new ContextRetrievalResponseEvent(requestId, "query text", List.of(doc)));
+
+        byte[] serialized = ActionStateSerde.serialize(originalState);
```
wenjin272, line 370: Currently, we inject the type information of `Event` during serialization by declaring mixins for the `ObjectMapper`.
```
// Add type information for polymorphic Event deserialization
mapper.addMixIn(Event.class, EventTypeInfoMixin.class);
mapper.addMixIn(InputEvent.class, EventTypeInfoMixin.class);
mapper.addMixIn(OutputEvent.class, EventTypeInfoMixin.class);
```
This allows the `ObjectMapper` to reconstruct the specific subclass of Event during deserialization. But I think that it’s sufficient to keep only `mapper.addMixIn(Event.class, EventTypeInfoMixin.class);` here.

wenjin272 (commented):
Thanks for this fix @rosemarYuan. Overall looks good to me, I left some comments about the details.

Besides, I think we need to state in the "custom-event-subclasses" section of the documentation that users must declare a `JsonCreator` for their custom Event subclasses.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerdeTest.java
rosemarYuan: Thanks for you suggestion. Updated the PR based on your comments:  simplified the test name, kept only the base `Event` mixin in `ActionStateSerde`, and added documentation for custom Java `Event` subclasses to explain the required Jackson creator annotations during durable recovery.

api/src/main/java/org/apache/flink/agents/api/event/ChatResponseEvent.java
```diff
@@ -46,7 +48,10 @@ public ChatResponseEvent(
         setAttr("total_retry_wait_sec", totalRetryWaitSec);
     }
 
-    public ChatResponseEvent(UUID id, Map<String, Object> attributes) {
+    @JsonCreator
+    public ChatResponseEvent(
+            @JsonProperty("id") UUID id,
+            @JsonProperty("attributes") Map<String, Object> attributes) {
         super(id, EVENT_TYPE, attributes);
```
weiqingy: Could the JSON creator rehydrate typed nested attributes during replay?

With durable execution enabled, Jackson constructs `ChatResponseEvent` through this creator using the generic `attributes` map, so the nested `response` value stays a `LinkedHashMap`. Since replay dispatches the event as a concrete `ChatResponseEvent`, Java actions using the supported concrete signature can hit a `ClassCastException` when calling `event.getResponse()`.

Would it make sense to normalize the typed nested attributes in the JSON creator, or before dispatch, so replayed built-in events match the live-path object shape? I think the same question applies to the other event creators added for typed nested attributes.

rosemarYuan: Thanks for the review suggestion.  I agree this should be normalized before replay dispatch.

I did it in the JSON creator rather than in dispatch: the` (UUID, Map)` constructor is the Jackson recovery entry point for each concrete built-in event, so normalizing there makes the recovered object match the live-path shape immediately, without the runtime needing to know how to hydrate arbitrary event subclasses.

For each built-in event with typed nested attributes (`ChatRequestEvent, ChatResponseEvent, ToolResponseEvent, ContextRetrievalResponseEvent`), I moved its `fromEvent() `conversion into a per-event `normalizeAttributes` helper that the creator calls. Both paths now produce identical typed attributes, so a concrete-signature action calling e.g. `getResponse()` on a replayed event no longer hits a `ClassCastException`.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerdeTest.java
```diff
@@ -333,4 +344,65 @@ public void testKafkaSederDelegatesToActionStateSerde() throws Exception {
...
+        assertEquals(ToolResponseEvent.class, outputEvents.get(3).getClass());
+        assertEquals(ContextRetrievalRequestEvent.class, outputEvents.get(4).getClass());
+        assertEquals(ContextRetrievalResponseEvent.class, outputEvents.get(5).getClass());
+
+        // On replay, actions declaring a concrete event signature (e.g. MathAgent's
+        // `processChatResponse(ChatResponseEvent, ...)`) receive the Jackson-reconstructed event
+        // directly and call typed getters WITHOUT going through fromEvent(). The JSON creator must
+        // therefore rehydrate nested attributes; otherwise these getters throw ClassCastException
+        // on the raw LinkedHashMap. Assert the typed getters directly on the deserialized
+        // instances.
```
wenjin272: This comment is overly verbose, and the phrase "going through fromEvent" describes a previous implementation that I believe should not appear in the final comment.

api/src/main/java/org/apache/flink/agents/api/event/ChatResponseEvent.java
wenjin272: We need to update the user notifications in the documentation accordingly.

rosemarYuan: The custom-event hint in workflow_agent.md already notes this. 

`If the subclass stores typed objects in attributes, convert them back to their typed forms inside this constructor too, as the built-in events do.`

Based on this, I added a code example under that hint showing how the built-in events do it, using ChatResponseEvent's creator and its normalizeAttributes helper.

wenjin272 (approved):
LGTM

wenjin272:
Backport to release-0.3

Issue #865, opened 2026-07-02
[Bug] byte[] memory values corrupted to base64 String on the durable-execution replay path


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The durable-execution path (ActionStateStore, Kafka/Fluss) serializes MemoryUpdate.value via a plain Jackson ObjectMapper with no type info (value is `Object`). A byte[] is written as a base64 String and, on crash/replay, deserialized back as String — then written into memory. The consumer silently gets `str`/String instead of the original bytes.

### How to reproduce

Add to ActionStateSerdeTest (or run equivalently):

    ActionState state = new ActionState(null, null, null, null, null, false);
    state.addShortTermMemoryUpdate(new MemoryUpdate("stm.path", "hello".getBytes()));

    ActionState out = ActionStateSerde.deserialize(ActionStateSerde.serialize(state));
    Object value = out.getShortTermMemoryUpdates().get(0).getValue();

    // Expected: byte[] equal to "hello".getBytes()
    // Actual:   value is a String "aGVsbG8=" (base64), type lost

Round-tripping through ActionStateSerde turns the byte[] into a base64
String, so on crash/replay the wrong type is written back into memory.

### Version and environment

0.1~0.3

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

weiqingy:
@wenjin272 I'd like to take this one — it's the direct tail of #846.

wenjin272:
@weiqingy, thanks for taking this on, assigned.

weiqingy:
@wenjin272 quick check on the release target before I touch `release-0.3`.

The fix landed on `main` via #874 (versioned Kryo binary envelope for `MemoryUpdate.value`), which closed both #865 and #872. This issue still carries `fixVersion/0.3.1` — but the original 0.3.1 plan was the narrow byte[] fix (#871), now superseded by #874's broader (and heavier) binary approach. Given your note that these would land within 0.4, is a 0.3.1 backport still wanted? If so, happy to cherry-pick #874 onto `release-0.3`.

wenjin272:
Thanks for checking @weiqingy .
I originally thought this issue should be fixed in 0.3.1. However, since the current fix takes an incompatible approach, I do not think it is appropriate to introduce such a change in a hotfix release like 0.3.1, even though we are still in the beta stage.
[...]

PR #871, opened 2026-07-05
[runtime] Preserve byte[] MemoryUpdate values across durable-execution serde (#865)

weiqingy:
Linked issue: #865

### Purpose of change

`ActionStateSerde` serializes `MemoryUpdate.value` (declared `Object`) via a plain, type-info-less Jackson `ObjectMapper`, so a `byte[]` is written as a base64 `String` and — with no type tag — deserialized back as `String` on crash/replay, silently writing the wrong type into memory.

This is a pre-existing, Java-native defect: a Java agent can `set` a `byte[]` memory value directly, so the bug predates admitting Python `bytes` into the memory contract (#846); #846 only makes it more likely to be hit.

The fix adds a targeted, field-scoped envelope (de)serializer bound to `MemoryUpdate.value` via a Jackson mixin. A `byte[]` is written as `{"__flink_agents_bytes__":"<base64>"}` and reversed back to `byte[]` on read; every other value type delegates to Jackson's stock untyped-`Object` (de)serialization and stays byte-identical on disk, so in-flight durable state remains replay-compatible.

The (de)serializer is bound per-property (mixin on `value`), not registered globally on `byte[]`, so `CallResult`'s statically-typed `byte[]` fields are untouched. The envelope carries only base64 text → `byte[]`, so it adds no polymorphic-typing deserialization surface.

Scope: only a top-level `byte[]` value is preserved. A `byte[]` nested inside a `List`/`Map` value goes through the default serializers and is out of scope, documented in the code.

### Tests

Added round-trip regressions in `ActionStateSerdeTest` (the first mirrors the issue repro exactly):

- `byte[]` on a short-term memory update round-trips as `byte[]` (issue repro; `assertArrayEquals`)
- `byte[]` on a sensory memory update
- empty `byte[0]`
- a `String` value still round-trips as `String` (envelope doesn't disturb the common case)
- a `Map` value still round-trips as a `Map` (not misread as an envelope; on-disk shape unchanged)
- reserved-key residual: a user `Map` that is exactly `{"__flink_agents_bytes__":"<base64>"}` recovers as `byte[]` (documented residual), while any second key keeps it a `Map`

`mvn test -Dtest=ActionStateSerdeTest -pl runtime` → 16 tests, 0 failures. `mvn test -pl runtime` → 396 tests, 0 failures.

### API

No public API change. `MemoryUpdate`, `ActionState`, and `ActionStateSerde.serialize/deserialize` signatures are unchanged; only the on-disk encoding of a `byte[]` value changes (it was previously broken).

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerde.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerde.java
@@ -19,16 +19,24 @@ package org.apache.flink.agents.runtime.actionstate;
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.api.InputEvent;
 import org.apache.flink.agents.api.OutputEvent;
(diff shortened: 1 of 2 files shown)


weiqingy:
Closing this in favor of #874, which supersedes it.

Per the discussion on #872, rather than land this narrow top-level `byte[]` envelope and then generalize it, we fold everything into one binary change: #874 switches `MemoryUpdate.value` to a Flink-native (Kryo) binary payload inside a versioned `{serde, version, payload}` envelope, which covers the top-level `byte[]` case this PR fixed plus nested `byte[]`, `Long` narrowing, and user POJOs — and retires the `__flink_agents_bytes__` envelope. Nothing here is persisted in that format, so there is no migration concern.

PR #874, opened 2026-07-06
[runtime] Binary serde for MemoryUpdate values via versioned Kryo envelope

### LINKED ISSUE #865: [Bug] byte[] memory values corrupted to base64 String on the durable-execution replay path

### LINKED ISSUE #872: [Bug] ActionStateSerde: nested byte[] (and Long→Integer) type-loss in MemoryUpdate container values on durable replay
### Search before asking

### Description

Follow-up to #865 / #871. That fix makes a **top-level** `byte[]` `MemoryUpdate.value` survive the durable-execution serde round-trip. But the same root cause — `MemoryUpdate.value` is declared `Object` and `ActionStateSerde` uses a type-info-less Jackson `ObjectMapper` — still loses type for two nested/numeric cases that #871 deliberately left out of scope:

1. **Nested `byte[]`**: a `byte[]` inside a `List` or `Map` memory value goes through Jackson's default serializers (not the field-scoped envelope added in #871), so it is written as a base64 `String` and, on crash/replay, recovered as a `String` instead of `byte[]`.

2. **`Long` → `Integer` narrowing**: an untyped `Long` value within `int` range round-trips back as `Integer` (Jackson binds untyped JSON integers to the smallest fitting type). Out-of-`int`-range longs are unaffected. The same applies to such numbers nested inside a `List`/`Map`.

Both are facets of the same `Object`-typed-serde type-loss family as #865 (`byte[]` value) and #868/#869 (event-subclass nested attributes), but on a distinct field (`MemoryUpdate.value`, nested/numeric) with a distinct fix, so they're tracked separately here rather than folded into any of those.

Severity is lower than #865 (nested/binary or int-range-`Long` memory values are less common than a top-level `byte[]`), but the corruption is silent, same as #865.

### How to reproduce

Nested `byte[]` (in `ActionStateSerdeTest` style):

    Map<String, Object> value = new HashMap<>();
    value.put("blob", "hello".getBytes());
    ActionState state = new ActionState(null, null, null, null, null, false);
    state.addShortTermMemoryUpdate(new MemoryUpdate("stm.path", value));

    ActionState out = ActionStateSerde.deserialize(ActionStateSerde.serialize(state));
    Map<?, ?> recovered = (Map<?, ?>) out.getShortTermMemoryUpdates().get(0).getValue();

    // Expected: recovered.get("blob") is a byte[] equal to "hello".getBytes()
    // Actual:   recovered.get("blob") is a base64 String "aGVsbG8=", type lost

`Long` → `Integer`:

    ActionState state = new ActionState(null, null, null, null, null, false);
    state.addShortTermMemoryUpdate(new MemoryUpdate("stm.path", 42L));

    ActionState out = ActionStateSerde.deserialize(ActionStateSerde.serialize(state));
    Object v = out.getShortTermMemoryUpdates().get(0).getValue();

    // Expected: v is a Long 42
    // Actual:   v is an Integer 42, type narrowed

### Version and environment

0.1~0.3, and persists on `main` after #865/#871 (which only covers a top-level `byte[]` value).

weiqingy:
Linked issue: closes #865, closes #872

### Purpose of change

`MemoryUpdate.value` is declared `Object`, and `ActionStateSerde` persisted it through a Jackson `ObjectMapper` that carries no type information for that field. On durable-execution replay the value was rebuilt by Jackson's stock untyped deserializer, so any runtime type that bare JSON cannot reconstruct degraded silently: `byte[]` to base64 `String`, an int-range `Long` to `Integer`, a user POJO to `LinkedHashMap`, and the same losses for those types nested inside a `List` or `Map`. The consumer performs no downcast, so the wrong type flowed into agent memory after any crash or restore. This generalizes #865 / #871, which fixed only a top-level `byte[]`.

The fix binds a field-scoped Jackson (de)serializer to `MemoryUpdate.value` only. It runs the Flink-native serializer — Kryo, for a generic `Object` — and stores the bytes base64-encoded inside a self-describing versioned envelope embedded in the single `ActionState` JSON document: `{ "serde": "kryo", "version": 1, "payload": "<base64>" }`. One mechanism covers every Kryo-serializable type instead of a per-type special case.

Key points:

- Field-scoping is deliberate: the (de)serializer binds to `MemoryUpdate.value` via a mixin, never a global `byte[]` serializer, so `CallResult`'s statically-typed `byte[]` payloads keep round-tripping unchanged.
- Thread-safety: `KryoSerializer` is not thread-safe, so one immutable template is built once and each thread serializes through its own `duplicate()` via a `ThreadLocal`.
- An unknown envelope `version` is rejected rather than silently mis-decoded.
- Safe binary format: the durable-execution journal is checkpoint-scoped (pruned on checkpoint completion, rebuilt on restore), so state is always written and read by the same code and the same Flink version. It never crosses a code or Flink upgrade, so Kryo's cross-version binary incompatibility and schema-evolution brittleness are never exercised. The recovery-compat contract (envelope shape, `version` marker, Kryo pinned within the major release) is documented on the serde.
- The `Event` field stays JSON, preserving the cross-language and event-log contract.

This supersedes #871 (the narrow `__flink_agents_bytes__` byte[] envelope), which is unmerged, so no state is persisted in that format.

### Tests

Extends `ActionStateSerdeTest` with 5 round-trip tests: top-level `byte[]`, an int-range `Long`, a nested `Map` / `List` graph holding both a `byte[]` and a `Long`, a custom POJO (recovered as its class, not a `LinkedHashMap`), and an unknown-version envelope (rejected). The 10 existing tests are unchanged.

`mvn -pl runtime -am test -Dtest=ActionStateSerdeTest` passes 15 tests under the default Flink 2.3.0, which is also the runtime linkage check for the Kryo construction path. `mvn -pl runtime spotless:check` is clean.

### API

No public API change. All new types are package-private members of `ActionStateSerde`. No new dependency, and no `dist/` change — Kryo is provided by the cluster Flink.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerde.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerde.java
@@ -19,14 +19,29 @@ package org.apache.flink.agents.runtime.actionstate;
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import org.apache.flink.agents.api.Event;
+import org.apache.flink.agents.api.context.MemoryUpdate;
 import org.apache.flink.agents.runtime.operator.ActionTask;
+import org.apache.flink.api.common.serialization.SerializerConfigImpl;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.core.memory.DataInputViewStreamWrapper;
+import org.apache.flink.core.memory.DataOutputViewStreamWrapper;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.util.Base64;
 
 /**
  * Backend-agnostic serializer/deserializer for {@link ActionState}.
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerde.java
```diff
@@ -88,4 +150,62 @@ public void serialize(ActionTask value, JsonGenerator gen, SerializerProvider se
...
+
+    /**
+     * Reads the Kryo envelope back to the concrete value; version-checks; falls back for a
+     * non-envelope node.
+     */
+    static class MemoryUpdateValueDeserializer extends JsonDeserializer<Object> {
+        @Override
+        public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+            JsonNode node = ctxt.readTree(p);
+            if (isEnvelope(node)) {
```
wenjin272: IIUC, the isEnvelope flag is used to maintain backward compatibility with legacy MemoryUpdate instances that were serialized using JSON. 

Is it essential to maintain backward compatibility in this case? 
* For new jobs, both serialization and deserialization should be performed using Kryo, with `isEnvelope` set to true.
* For old jobs, both serialization and deserialization should be performed using Jackson.
* The `isEnvelope` can only become false after a job upgrade and restart. Since flink agents are still in beta, I consider this incompatibility acceptable.

wenjin272 (commented):
Thanks for taking this on @weiqingy. I left one comment.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerde.java
weiqingy: Thanks for the review, @wenjin272. I've updated the PR accordingly.

wenjin272 (approved):
LGTM

Issue #864, opened 2026-07-02
[Tech Debt] Avoid using reflection to load implementation classes of resources during the compilation phase.


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Currently, when determining whether a resource is a Python resource during Java-based agent plan compilation, the framework loads the resource's implementation class via reflection.

```
if (PythonResourceWrapper.class.isAssignableFrom(
        Class.forName(
                descriptor.getClazz(),
                true,
                Thread.currentThread().getContextClassLoader()))) {
    provider = new PythonResourceProvider(name, type, descriptor);
} else {
    provider = new JavaResourceProvider(name, type, descriptor);
}
```

Currently, this requires resource dependencies to exist at the compile stage, but we actually only need them at runtime. Since Python resources must declare the `pythonClazz` argument, we can use this parameter to determine whether a resource is a Python resource.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

da-daken:
hi @wenjin272 , I would like to complete this task. Have you started yet?

Additionally, I noticed that `PythonMCPServer` does not include the `pythonClazz` parameter. For consistency, I propose to add this parameter by default in `extractResourceProvidersFromAgent`.

wenjin272:
Hi, @da-daken, I haven’t started working on it yet, so please feel free to take it. 

Yes, `PythonMCPServer` does not include `pythonClazz` because there is only one implementation of `MCPServer`. I think injecting `pythonClazz` feels a bit hacky. In fact, we can already determine whether an instance is a `PythonMCPServer` from its annotation. Therefore, I believe we can pass this information directly to `extractResource`, like:

```
extractResource(
      ResourceType.MCP_SERVER,
      method,
      desc ->
              new ResourceDescriptor(
                      desc.getModule(),
                      PythonMCPServer.class.getName(),
                      new HashMap<>(desc.getInitialArguments())),
      true);

private void extractResource(ResourceType type, Method method) throws Exception {
        extractResource(type, method, null, false);
    }

private void extractResource(
        ResourceType type,
        Method method,
        Function<ResourceDescriptor, ResourceDescriptor> descriptorDecorator,
        boolean isPython)
```

PR #879, opened 2026-07-07
[Tech Debt] Avoid using reflection to load implementation classes of resources during the compilation phase

da-daken:
Linked issue: https://github.com/apache/flink-agents/issues/864

### Purpose of change

Avoid using reflection to load implementation classes of resources during the compilation phase

### Tests
ut test

### API

no API

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
@@ -343,9 +343,6 @@ public class AgentPlan implements Serializable {
         if (isPython || isPythonResource(descriptor)) {
             return new PythonResourceProvider(name, type, descriptor);
         }
-        if (descriptor.getModule() != null && !descriptor.getModule().isEmpty() && !descriptor.getInitialArguments().containsKey("java_clazz")) {
-            throw new IllegalArgumentException("python resource needs pythonClazz param or java resource needs java_clazz param");
-        }
         return new JavaResourceProvider(name, type, descriptor);
     }
 
--- a/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
@@ -35,7 +35,6 @@ import org.apache.flink.agents.api.resource.Resource;
 import org.apache.flink.agents.api.resource.ResourceDescriptor;
 import org.apache.flink.agents.api.resource.ResourceType;
 import org.apache.flink.agents.api.resource.SerializableResource;
-import org.apache.flink.agents.api.resource.python.PythonResourceWrapper;
 import org.apache.flink.agents.api.skills.SkillSourceSpec;
 import org.apache.flink.agents.api.skills.Skills;
 import org.apache.flink.agents.api.tools.ToolMetadata;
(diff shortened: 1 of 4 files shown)


plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -316,31 +315,40 @@ private static ResourceDescriptor requireResourceDescriptor(
...
+            String name, ResourceType type, ResourceDescriptor descriptor, boolean isPython) {
+        if (isPython || isPythonResource(descriptor)) {
+            return new PythonResourceProvider(name, type, descriptor);
         }
-        addResourceProvider(provider);
+        return new JavaResourceProvider(name, type, descriptor);
+    }
+
+    private boolean isPythonResource(ResourceDescriptor descriptor) {
+        String pythonClazz = descriptor.getArgument("pythonClazz");
```
weiqingy, line 350: Non-blocking:

The old check identified Python resources by whether clazz was a `PythonResourceWrapper`, while this now uses the presence of `pythonClazz`. Since `PythonResourceProvider.provide()` does not require `pythonClazz` when `module` is non-empty, do we want to document the assumption that descriptors reaching this classification path will always carry `pythonClazz` when they should be treated as Python?

da-daken: Thanks for your review! I think we should align on a single rule here. Since cross-language Python resources are required to declare pythonClazz, we probably should not leave other entry points open. Today, descriptors that take those alternate paths (e.g. non-empty module without pythonClazz) are classified as Java-side resources at plan-compilation time and fail at runtime. Tightening or closing those paths feels like a separate follow-up issue rather than something we need to solve in this PR—what do you think?

weiqingy: Agreed. Making `pythonClazz` the single rule and closing the alternate paths is cleaner than documenting the assumption, and it belongs in its own change. Nit: those descriptors classify as Java at compile time and only fail at runtime, so sharpening that into a clearer compile-time error would be a good tracker. Otherwise LGTM.

wenjin272: Given that normal Java AgentPlan paths don’t appear to produce `module != empty && pythonClazz` missing (`module` always be ""), I don’t think we need to preserve that old descriptor shape here. It seems limited to manually constructed 3-arg descriptors, so documenting/tightening that contract separately should be enough.

da-daken:
@wenjin272 @weiqingy Thanks for your review! I’ve updated the submission. PTAL~

weiqingy:
Thanks for the update. The change LGTM. I'll leave the final merge call to @wenjin272.

One thing before it's ready to merge: the CI checks on the latest commit are showing red. Could you double-check and get them green? Just want to rule out anything introduced by this update.

One optional, non-blocking Javadoc nit while you're here: the last line of the new block reads a little rough (`the initialArguments should put ("pythonClazz", ...)`), and it doesn't say what `clazz` should be in this case. Since `PythonResourceProvider.provide()` rebuilds the module and class from `pythonClazz` whenever `module` is empty, `clazz` is effectively ignored on this path — the new tests just pass a placeholder. Might be worth spelling that out, e.g.:

> For Python resources (when declaring from Java): put the fully-qualified Python class under the `pythonClazz` key in `initialArguments` (e.g. `("pythonClazz", "your_module.submodule.YourPythonClass")`). The `clazz` argument is unused in this case.

wenjin272 (approved):
LGTM

PR #880, opened 2026-07-07
[hotfix] Propagate action task Errors from mailbox

wenjin272:
Catch Throwable in mailbox-submitted action task processing so Errors are rethrown through the mailbox and fail the task visibly instead of remaining in a Future while processing keys stay in-flight.

### Purpose of change

Catch Throwable in mailbox-submitted action task processing so Errors are rethrown through the mailbox and fail the task visibly instead of remaining in a Future while processing keys stay in-flight.

### Tests

ut

### API

no

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/runtime/src/test/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperatorTest.java
+++ b/runtime/src/test/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperatorTest.java
@@ -290,6 +290,27 @@ public class ActionExecutionOperatorTest {
     }
 
     @Test
+    void testMailboxSubmittedActionTaskPropagatesError() throws Exception {
+        try (KeyedOneInputStreamOperatorTestHarness<Long, Long, Object> testHarness =
+                new KeyedOneInputStreamOperatorTestHarness<>(
+                        new ActionExecutionOperatorFactory(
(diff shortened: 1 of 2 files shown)


runtime/src/test/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperatorTest.java
```diff
@@ -1711,6 +1732,16 @@ public static void durableMixedRecoveryAction(Event event, RunnerContext context
             }
         }
 
+        public static class LinkageErrorAction {
+            static {
+                if (System.nanoTime() >= 0) {
```
weiqingy: Could we make this linkage-error fixture deterministic? `System.nanoTime()` is permitted to return a negative value, so on a JVM where that happens the static initializer would complete normally and this regression test would fail for environmental reasons instead of exercising the intended `Error`-propagation path. A deterministic guard like `if (true)` (or an `alwaysTrue()` helper) keeps the fixture focused on the `LinkageError` path.

weiqingy (commented):
Thanks for the PR — one optional note on the test fixture inline.

weiqingy (approved):
Thanks for resolving the comment. LGTM!

PR #881, opened 2026-07-07
[python] Backfill remote test coverage ahead of local execution path removal

weiqingy:
Linked issue: #829

### Purpose of change

This is the first of two PRs retiring the Python in-process local execution path (`local_runner` / `local_execution_environment`) in favor of validating agent logic on Flink's MiniCluster, per #829.

This PR is purely additive. Several behaviors are currently exercised only through the local path (via its `from_list` / `to_list` / `get_tool_request_events` test API); before the local path can be removed, that coverage has to exist on the remote (real-Flink) path. This PR backfills it, so the follow-up deletion PR removes the local path without losing any coverage.

New remote (MiniCluster) tests added:

- Synchronous `durable_execute` exception propagation and keyword-argument calls — the existing remote durable-execute tests only covered the async-exception and positional-argument paths.
- Built-in `chat_model_action` + `tool_call_action` content and tool binding, driven by a deterministic mock chat model (the mock stubs only LLM inference, so the built-in actions remain the code under test).
- `get_resource(CHAT_MODEL)` inside an action, including custom `ResourceDescriptor` field pass-through.
- MCP tool invocation and prompt retrieval, invoked directly via `get_resource(...).call(...)` with no LLM — mirroring the existing Java `MCPCrossLanguageAgent` pattern and gated only on the self-started MCP server. This is the first remote MCP coverage in the repo.
- The `apply(<unknown-name>)` → `ValueError` guard on the remote builder (unit test; the guard fires before a cluster starts).
- Short-term-memory workflow semantics: same-key `visit_count` accumulation across records, and auto-generated key for a record with no key.
- Chat model coverage for the non-Ollama providers (Tongyi / OpenAI / AzureOpenAI) on the remote path, preserving the original per-provider credential gating.

### Tests

This PR is entirely tests. Verified on a Flink MiniCluster: one end-to-end session runs 11 passed, 3 skipped (the skips are the credential-gated provider cases). The deterministic tests (durable-execute, built-in action, MCP, workflow memory) run without any external service; the MCP test self-starts the MCP server. The `apply`-guard unit test runs without a cluster.

### API

No public API changes. Test-only.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`


python/flink_agents/e2e_tests/e2e_tests_integration/mcp_remote_agent.py
```diff
@@ -0,0 +1,112 @@
...
+def _call_add(ctx: RunnerContext, a: int, b: int) -> int:
+    """Fetch the MCP ``add`` tool and invoke it, returning the numeric sum."""
+    add = ctx.get_resource("add", ResourceType.TOOL)
+    result = add.call({"a": a, "b": b})
+    # MCPTool.call returns a list of extracted content items; the add tool
+    # yields a single text item holding the sum.
+    return int(result[0])
+
+
+class MCPRemoteWithPromptsAgent(Agent):
```
wenjin272: It seems this Agent is intended to replace `MyMCPAgent` in `e2e_tests_mcp/mcp_test.py`. However, unlike its predecessor, it does not bind MCP prompts and tools to the ChatModel, which I consider a regression in test coverage. Could we continue using `MyMCPAgent` instead, limiting changes to only the submission and validation logic within the `test_mcp` function?

wenjin272 (commented):
Thanks for this work @weiqingy. I left one comment.

python/flink_agents/e2e_tests/e2e_tests_integration/mcp_remote_agent.py
weiqingy: Thanks for the review @wenjin272 . Good catch. Fixed in e2272ee: `MyMCPAgent` is kept as-is and only `test_mcp`'s submission/validation move to the MiniCluster path; the direct-invocation agent/test are removed. Could you take another look? Let me know if you have any concerns.

wenjin272 (approved):
LGTM

PR #882, opened 2026-07-07
[python] Remove the Python local execution path

weiqingy:
Linked issue: #829

### Purpose of change

Second of two PRs retiring the Python in-process local execution path, per #829. This PR removes the local path itself.

Stacked on #881 (the additive test backfill). Until #881 merges, this PR's diff includes #881's commits; the net-new change here is the removal, in the last two commits — "Remove local-path tests and migrate remaining tests off it" and "Delete the local execution path and require a Flink environment". I'll rebase once #881 lands.

What this removes:

- `local_runner.py` and `local_execution_environment.py` — the pure-Python, single-process execution path selected when no Flink environment was passed.
- The execution-environment `from_list` / `to_list` test-only API and the local-only `get_tool_request_events` hook.
- The local-path tests, now covered by the remote (MiniCluster) tests added in #881; and the local-path sections of the documentation.

`get_execution_environment` now requires a `StreamExecutionEnvironment` and raises a clear error when none is provided, instead of falling back to the local path. The RAG and quickstart examples are migrated to the remote DataStream pattern. `LocalMemoryObject` is kept as an in-memory test fixture (it is independent of the runner and still used by unit tests).

### Tests

The Python unit lane and the remote end-to-end suite pass. The local-path tests are removed because equivalent behavior is now exercised on a real Flink MiniCluster by the tests added in #881 (durable-execute sync-exception and kwargs, built-in action content and `get_resource`, MCP tool and prompt invocation, workflow short-term-memory semantics, and the apply-by-name guard).

### API

Yes — this removes public surface: the execution-environment `from_list` / `to_list` methods, `get_tool_request_events`, and the no-argument `get_execution_environment()` form (a Flink environment is now required). This is a deliberate removal of the non-production local path while the project is in beta.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`


docs/content/docs/operations/configuration.md
wenjin272: Could be removed.

wenjin272 (commented):
Thanks for taking this on @weiqingy. I left one comment about doc.

Besides, `LocalMemoryObject` now appears to be used only by tests (`test_chat_model_action`, `test_local_memory_object`, and `test_memory_reference`). Since this PR removes the local execution path, could we move it out of `flink_agents.runtime` and keep it as a test fixture instead? That would avoid leaving a local-runtime-looking class in the production package.

docs/content/docs/operations/configuration.md
weiqingy: @wenjin272  Thanks for the review. Both addressed:

- **`configuration.md` Local-mode example**: removed in `b495072`. It pointed at the run-without-Flink path this PR deletes. That left MiniCluster as the only special case, so I also changed the intro from "the following two cases" to the singular form.
- **`LocalMemoryObject`**: moved to `flink_agents.runtime.tests.local_memory_object` in `d4243f1`, so it's no longer in the production runtime package. It's now purely an in-memory test fixture, and the three importers (`test_chat_model_action`, `test_memory_reference`, `test_local_memory_object`) are updated — it now sits alongside its own unit test in `runtime/tests`.

wenjin272 (approved):
LGTM, but there are some conflicts that need to be resolved.

weiqingy:
Both failing checks are unrelated to this PR — they're CI environment issues (a broken PyFlink install on one leg, an Ollama backend crash on the other), not test failures.

- **it-python [python-3.11] [flink-2.0]**: every e2e test fails at pyflink's own `Configuration()` constructor with `TypeError: Could not found the Java class 'org.apache.flink.configuration.Configuration'` and `py4j.protocol.Py4JError: org does not exist in the JVM`. The Flink dist JARs were not on the JVM classpath, so it dies inside `pyflink/common/configuration.py` before any agent code runs — the first Flink test (`mcp_test.py::test_mcp`) already fails there and the rest cascade off the same broken gateway. [Job](https://github.com/apache/flink-agents/actions/runs/29104737412/job/86402259198).
- **it-python [python-3.12] [flink-2.3]**: the LLM-driven tests fail with `pemja.core.PythonException: <class 'ollama._types.ResponseError'>: llama-server process has terminated: signal: segmentation fault (core dumped)`. The Ollama server crashed mid-run; the non-LLM e2e tests on this same leg all passed. [Job](https://github.com/apache/flink-agents/actions/runs/29104737412/job/86402259296).

Neither is caused by this change: the identical e2e suite passed on the `flink-1.20`, `flink-2.1`, and `flink-2.2` legs of the same run, and this PR only deletes Python local-execution code — nothing in the build, `dist/`, `pom.xml`, or the pyflink packaging.

I've pushed an empty commit to re-trigger CI.

weiqingy:
Thanks @wenjin272. I've rebased onto `main` to resolve the conflict, and CI is now green. Could you take another look when you have a chance?

PR #883, opened 2026-07-07
[integration][openai] Align default parameters for openai chat model between java and python

rob-9:
### purpose

the Java OpenAI chat model connections (`OpenAICompletionsConnection`, `AzureOpenAIChatModelConnection`, and `OpenAIResponsesModelConnection`) didn't set explicit defaults for `timeout` and `max_retries`, falling through to the openai-java SDK's internal values. the Python SDK explicitly defaults to `timeout=60s` and `max_retries=3`, meaning the same agent config behaves differently across languages.

this PR aligns Java to match Python:
- adds shared `DEFAULT_TIMEOUT_SECONDS = 60` / `DEFAULT_MAX_RETRIES = 3` constants in `OpenAIChatCompletionsUtils`.
- all three Java connections now resolve these via `Optional.ofNullable(...).orElse(DEFAULT_*)` so absent args get the explicit default.
- invalid values (timeout < 0, retries < 0) throw `IllegalArgumentException` at construction time instead of being silently skipped.
- javadoc and docs updated to document the new defaults and validation constraints.

### testing

- new `OpenAICompletionsConnectionTest` — asserts correct defaults, explicit overrides, negative rejection, zero acceptance, and api_key validation.
- new `OpenAIResponsesModelConnectionTest` — same coverage for the Responses API connection.
- extended `AzureOpenAIChatModelConnectionTest` with default-resolution, override, and boundary assertions.
- new Python test pinning `timeout == 60.0` and `max_retries == 3`.

### breaking changes

- `max_retries` default changed from 2 → 3 for `OpenAICompletionsConnection` and `OpenAIResponsesModelConnection`.
- `timeout=0` was previously silently ignored (SDK default applied); it is now passed through as `Duration.ofSeconds(0)`.
- negative values for either parameter now throw instead of being silently skipped.

### docs
- [x] `doc-not-needed`


wenjin272 (commented):
Thanks for taking this on @rob-9. The code looks good to me, but I think we should also update the documentation.

The user-facing docs still describe the old Java defaults. After this PR, Java OpenAI/Azure connections explicitly default to `timeout=60` and `max_retries=3`, but these tables still mention SDK fallback/`None` and `max_retries=2`. Please update the docs to match the new constructor behavior.

rob-9:
@wenjin272 thx

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
```diff
@@ -98,15 +101,21 @@ public OpenAICompletionsConnection(
...
-        Integer maxRetries = descriptor.getArgument("max_retries");
-        if (maxRetries != null && maxRetries >= 0) {
-            builder.maxRetries(maxRetries);
-        }
+        int rawTimeout =
+                Optional.ofNullable(descriptor.<Number>getArgument("timeout"))
+                        .map(Number::intValue)
+                        .orElse(OpenAIChatCompletionsUtils.DEFAULT_TIMEOUT_SECONDS);
+        this.timeoutSeconds =
+                rawTimeout > 0 ? rawTimeout : OpenAIChatCompletionsUtils.DEFAULT_TIMEOUT_SECONDS;
```
weiqingy: Python validates both fields with `Field(ge=0)` (`openai_chat_model.py:62-71`, `azure_openai_chat_model.py:70-79`), so `timeout=-5` or `max_retries=-1` raise a `ValidationError` at construction. Here — and identically in `AzureOpenAIChatModelConnection.java:134-143` — an invalid value is silently rewritten to the default. `ge=0` also means Python accepts `timeout=0`, where this `> 0` check turns it into `60`. So the clamp ends up being the one behavior a parity PR leaves divergent. Is that intentional, or would failing fast (or logging the override) match the Python contract better?

Either way, the branch is currently unexercised — the Java tests cover only the unset and valid-override cases (`timeout=120`, `max_retries=5`), so a test pinning whichever behavior you settle on would be worth adding.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIChatCompletionsUtils.java
```diff
@@ -52,6 +52,12 @@ final class OpenAIChatCompletionsUtils {
 
     private OpenAIChatCompletionsUtils() {}
 
+    /** default timeout in seconds for openai api requests (aligned with python sdk). */
```
weiqingy: Small housekeeping on these two constants: the Javadoc starts lowercase (`default timeout in seconds...`) whereas the rest of this file uses sentence case (e.g. `Convert a list of...` a few lines below). They're also placed after the private constructor and ahead of the `mapper`/`MAP_TYPE` fields — constants conventionally sit at the top of the class. Minor, but tidying both would keep the file consistent.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
rob-9: updating to match python contract; both fields will throw `IllegalArgumentException` on negative values and accept zero

weiqingy: Confirmed — matches Python's `Field(ge=0)` now (throws on negatives, accepts `0`), and the tests across all three connections close the branch. Thanks!

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
```diff
@@ -98,15 +101,23 @@ public OpenAICompletionsConnection(
...
+                        .map(Number::intValue)
+                        .orElse(OpenAIChatCompletionsUtils.DEFAULT_TIMEOUT_SECONDS);
+        if (this.timeoutSeconds < 0) {
+            throw new IllegalArgumentException("timeout must be >= 0, got: " + this.timeoutSeconds);
         }
-
-        Integer maxRetries = descriptor.getArgument("max_retries");
-        if (maxRetries != null && maxRetries >= 0) {
-            builder.maxRetries(maxRetries);
+        builder.timeout(Duration.ofSeconds(this.timeoutSeconds));
```
wenjin272, line 111: `timeout=0` has opposite semantics in Java and Python. openai-java interprets `Duration.ZERO` as no timeout, while Python forwards `0.0` to HTTPX as a zero-second timeout. Please define one consistent zero-value contract across both languages and test the effective SDK timeout rather than only the stored field.

The same pattern is also present in `AzureOpenAIChatModelConnection` and `OpenAIResponsesModelConnection`.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
```diff
@@ -98,15 +101,23 @@ public OpenAICompletionsConnection(
             builder.baseUrl(apiBaseUrl);
         }
 
-        Integer timeoutSeconds = descriptor.getArgument("timeout");
-        if (timeoutSeconds != null && timeoutSeconds > 0) {
-            builder.timeout(Duration.ofSeconds(timeoutSeconds));
+        this.timeoutSeconds =
+                Optional.ofNullable(descriptor.<Number>getArgument("timeout"))
+                        .map(Number::intValue)
```
wenjin272, line 106: Calling `Number.intValue()` before validation silently truncates or overflows the input. For example, `-0.5` becomes `0` and bypasses the negative check, while fractional timeouts lose precision and fractional retry counts differ from Python validation. Please validate the original value first, preserve fractional timeout values, and require an exact bounded integer for `max_retries`.

The same pattern is also present in AzureOpenAIChatModelConnection and OpenAIResponsesModelConnection.

wenjin272 (commented):
Hi, @rob-9. I left two new minor comments about details.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
rob-9: `timeout=0` now disables timeouts in both Java and Python. added SDK/transport-level tests.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
rob-9: thanks -- validation now checks the raw numeric value before conversion. `max_retries` must be an exact bounded integer, while timeouts preserve fractional values and reject invalid inputs. also added boundary coverage.

wenjin272 (approved):

Issue #691, opened 2026-05-19
[Feature] Kafka ActionStateStore should clean up pruned ActionState records


da-daken:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Problem
ActionState is persisted to the external ActionStateStore to support durable execution recovery. The current production store implementation is Kafka.
When a checkpoint completes
```java
actionStateStore.pruneState(entry.getKey(), entry.getValue());
```
The intent is to prune action states for a Flink key up to the completed sequence number. However, the current KafkaActionStateStore.pruneState(...) implementation only removes matching entries from the in-memory actionStates cache. It does not delete the corresponding records from the Kafka topic. These historical keys are no longer useful for recovery once the completed sequence number has been covered by a successful checkpoint, but they can still accumulate indefinitely in Kafka.
### Solution
After a checkpoint has completed,should delete from kafkaActionStateStore. Because the Kafka topic is configured with cleanup.policy=compact, writing a tombstone record (key = stateKey, value = null) allows Kafka log compaction to eventually remove all historical records for that ActionState key.
```java
producer.send(new ProducerRecord<>(topic, stateKey, null));
```

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

rob-9:
happy to pick this up if no one's already on it

da-daken:
hi @rob-9 , Thanks a lot for being interested in this. I'm not sure if @wenjin272 @weiqingy think this needs to be fixed, so I haven't started it yet.

PR #885, opened 2026-07-08
[api][runtime][python] Add opt-in Kafka tombstones for pruned action state

rob-9:
Addresses #691

### Purpose of change

`KafkaActionStateStore.pruneState()` currently removes matching entries only from its in-memory cache. The underlying Kafka records remain, so the action-state topic grows indefinitely and replay can restore previously pruned state.

This change adds the opt-in `kafkaActionStateTombstoneEnabled` option, defaulting to `false` in Java and Python. When enabled, pruning writes null-valued records for matching state keys so Kafka compaction can reclaim them, and `rebuildState()` treats tombstones as deletions.

The option remains disabled by default because tombstones can invalidate checkpoints or savepoints older than the prune and cause completed actions to execute again. The option controls emission only: recovery always applies tombstones already present in the topic, and disabling the option later cannot restore compacted state.

The pruning path also:

- uses segment-aware business-key and sequence-number matching, including keys containing `_`
- reports asynchronous tombstone delivery failures while still evicting cache entries
- avoids an unnecessary producer flush

### Behavioral Semantics

| Tombstone option | Pruning behavior | Recovery impact |
|---|---|---|
| Disabled | Matching entries are evicted from the in-memory cache; Kafka records remain. | Pruning does not invalidate older restore points, but the topic continues to grow. |
| Enabled, send succeeds | Matching entries are evicted and null-valued records are queued for Kafka. | The triggering checkpoint remains usable; earlier checkpoints or savepoints may replay the tombstones and re-execute completed actions. |
| Enabled, send fails | Matching cache entries are still evicted and the failure is logged. | The corresponding Kafka records remain until later or manual cleanup. |

Tombstone replay is independent of the current option value. Existing action-state topics must use `cleanup.policy=compact` for tombstones to reclaim storage, and their partition count must remain fixed so values and tombstones stay co-located.

### Known dependency

Final review found that the shared action-state key format still represents the logical Flink key using `key.toString()`. Distinct keys such as `Long(1)` and `String("1")` can therefore be conflated during lookup and pruning. With tombstones enabled, pruning one key could durably delete another key's recovery state. This root cause affects both Kafka and Fluss and will be tracked separately; this PR should land after that fix.

### Tests

Passed with JDK 17 and Python 3.12:

- Focused Java action-state tests: 71 passed
  - `KafkaActionStateStoreTest`: 27 passed
  - `FlussActionStateStoreTest`: 14 passed
  - `ActionStateUtilTest`: 20 passed
  - `ActionStateKeyPartitionerTest`: 10 passed
- Java API tests: 381 passed, 11 skipped
- Python core-options tests: 6 passed
- Java/Python `ConfigOption` parity check
- Ruff
- Spotless
- `git diff --check`

The tests cover default-off behavior, tombstone generation, asynchronous delivery failures, cache eviction, tombstone replay, segment-exact matching, underscore-containing keys, ownership filtering, and Java/Python option parity. A real-broker compaction test and a full checkpoint-completion-to-recovery test are not included.

### API

Adds the public Java and Python `KAFKA_ACTION_STATE_TOMBSTONE_ENABLED` option. Both expose the `kafkaActionStateTombstoneEnabled` key with a Boolean type and default value of `false`.

The public `ActionStateStore.pruneState()` contract now documents the recovery implications of durable backend deletion. No method signatures changed, and callers that do not enable the option retain the previous behavior.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

The Kafka configuration table documents the option, its default behavior, and the recovery trade-off. The deployment documentation retains the existing checkpoint-recovery behavior and scopes the tombstone caveat to the opt-in Kafka path.


wenjin272:
@joeyutong Can you take a look at this PR?

joeyutong:
Thanks for addressing the growth of Kafka action-state records. However, writing tombstones unconditionally may break recovery from older checkpoints or savepoints.

For example, if C0 is followed by action state S and then C1 tombstones S, restoring C0 will replay both S and the tombstone. `rebuildState()` removes S, so the action may execute again. This happens even when Kafka compaction is disabled.

Could tombstone emission be opt-in, for example through `kafkaActionStateTombstoneEnabled` defaulting to `false`?

rob-9:
> Could tombstone emission be opt-in, for example through `kafkaActionStateTombstoneEnabled` defaulting to `false`?

Yeah, agreed.

Can we make cleanup safe by default though? Maybe we could stamp tombstones with the checkpoint id that triggered the prune; `rebuildState()` applies one only if id <= restored checkpoint, so restoring C0 skips C1's tombstones. But this would only protect the record until compaction deletes it.

This can be a follow-up. I'll add the opt-in flag to this PR for now.

joeyutong:
Thanks for raising the question of whether cleanup can be safe by default. A checkpoint-aligned, user-controlled cleanup path may be a safer general solution:

1. Persist and expose the mapping from each checkpoint to its effective Kafka recovery marker—the per-partition minimum across all subtasks.
2. Leave the irreversible cleanup decision to users. Based on their recovery requirements, they can choose the oldest checkpoint to retain and delete records before its marker offsets through Kafka's `deleteRecords` API.
3. During `rebuildState()`, validate the requested offsets against Kafka's current beginning offsets. If a recovery offset is older than the available data, fail fast with a clear error instead of silently rebuilding incomplete state.

Since `rebuildState()` already starts from the recovery marker, the same marker is a natural boundary for physical cleanup. This aligns cleanup with the actual recovery semantics without automatically invalidating older checkpoints. Once users advance that boundary, fail-fast validation makes any incompatible recovery attempt explicit.

This could be explored as a follow-up.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
```diff
@@ -64,6 +64,19 @@ public class AgentConfigOptions {
...
+     * The config parameter determines whether pruning sends tombstone (null-valued) records to the
+     * Kafka action state topic so log compaction can reclaim pruned keys. Defaults to {@code
+     * false}: without tombstones the topic grows unboundedly, but restoring any checkpoint or
+     * savepoint replays correctly. When enabled, restoring from the latest completed checkpoint is
+     * unaffected, but restoring an older checkpoint or savepoint may replay tombstones written
+     * after that restore point, erasing action state the replay still needs and causing already
+     * completed actions to re-execute. Enable only if the job never restores from non-latest
+     * checkpoints or savepoints, or if re-executing actions is acceptable.
+     */
+    public static final ConfigOption<Boolean> KAFKA_ACTION_STATE_TOMBSTONE_ENABLED =
```
joeyutong, line 77: Could we also add the corresponding option to python/flink_agents/api/core_options.py? The cross-language option parity check currently fails because this field exists only on the Java side.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStoreTest.java
```diff
@@ -196,6 +207,129 @@ void testPruneState() throws Exception {
...
+        assertThat(mockProducer.history()).isEmpty();
+    }
+
+    @Test
+    void testPruneStateEvictsCacheEvenWhenTombstoneSendFails() throws Exception {
+        // Arrange - the next send() will fail asynchronously (e.g. broker unavailable)
+        actionStateStore = tombstoneEnabledStore(actionStates, mockProducer);
+        String stateKey = ActionStateUtil.generateKey(TEST_KEY, 1L, testAction, testEvent);
+        actionStates.put(stateKey, testActionState);
+        mockProducer.errorNext(new RuntimeException("simulated broker failure"));
```
joeyutong, line 263: mockProducer uses autoComplete=true, so errorNext() is called before any pending completion exists and does not make the subsequent send() fail. This test therefore exercises the successful send path. Could we inject an exception through the callback so the async failure path is actually covered?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
```diff
@@ -276,30 +282,64 @@ public void rebuildState(List<Object> recoveryMarkers) {
...
+        String keyStr = key.toString();
+        String keyPrefix = keyStr + "_";
+        List<String> keysToPrune = new ArrayList<>();
+        for (String stateKey : actionStates.keySet()) {
+            if (!stateKey.startsWith(keyPrefix)) {
+                continue;
+            }
+            try {
+                List<String> parts = ActionStateUtil.parseKey(stateKey);
+                if (parts.get(0).equals(keyStr) && Long.parseLong(parts.get(1)) <= seqNum) {
```
weiqingy, line 297: The collision you caught here feels like a thread worth pulling one turn further — I think the same root cause has a second face that this check can't reach.

Both this guard and the collision it fixes trace back to `generateKey` joining on `_` while embedding the raw key unescaped (`ActionStateUtil.java:44-56`). That's what let pruning key `a_1` match agent key `a`'s state key `a_1_<uuid>_<uuid>`, and the equality check handles that case correctly.

The sibling case is when the *agent key itself* contains `_`. For an ordinary `keyBy("user_123")` the state key is `user_123_1_<uuid>_<uuid>`, so `split("_")` yields 5 parts (UUIDs use hyphens) and `parseKey`'s `checkArgument(parts.length == 4)` throws (`ActionStateUtil.java:58-63`). This line never gets evaluated, and the entry lands in the catch on line 300 — so it's never evicted from the cache, never tombstoned, and never reclaimed by compaction. For any key containing `_`, the growth this PR targets would stay as it is. `FlussActionStateStore.removeStateEntries` catches `Exception` and skips the same way (`FlussActionStateStore.java:242-261`).

Worth saying that catching `IllegalArgumentException` on line 300 is an improvement in its own right — the old code caught only `NumberFormatException`, so a bad part count escaped `pruneState` and could take down `notifyCheckpointComplete`. Warn-and-retain is strictly better.

Escaping the separator properly would touch three files, which feels like more than this PR signed up for — would you rather note the limitation in the option's javadoc and file a follow-up? Or do you see a narrower angle I'm missing?

And on the test: `testPruneStateSkipsUnparseableKeys` pins this branch with `TEST_KEY + "_1_onlythreeparts"`. Any appetite for a realistic fixture like `"user_123"` — same branch, but it'd document the case that actually reaches it?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
```diff
@@ -276,30 +282,64 @@ public void rebuildState(List<Object> recoveryMarkers) {
...
+                                                    + "until manual cleanup.",
+                                            stateKey,
+                                            exception);
                                 }
-                            }
-                            return false;
-                        });
+                            });
+                }
+                producer.flush();
```
weiqingy, line 329: Does this flush earn its place? The callback just above already reports failures, the eviction below runs regardless of the outcome, and ordering looks covered — `put()` flushes before returning (line 146), so a tombstone is always sent after its record is acked.

Curious whether there's a case it's protecting that I'm not seeing.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
```diff
@@ -64,6 +64,19 @@ public class AgentConfigOptions {
...
+            new ConfigOption<>("kafkaActionStateTombstoneEnabled", Boolean.class, false);
```
weiqingy, line 78: Defaulting to `false` reads as the right call to me, and I think for a sharper reason than the javadoc gives itself credit for: with `cleanup.policy=compact` on the topic (`KafkaActionStateStore.java:396`), a tombstone is a durable delete instruction to the log cleaner — so the older-checkpoint erasure you documented above isn't something the replay side could have absorbed. Gating emission is the lever that actually controls it. A safe default-on would seem to need emission gated on the oldest still-restorable checkpoint, and I can't see a cheap way to know that here, so I'm not suggesting you flip it.

That does leave #691's original ask — the unbounded growth — unaddressed while the option is off. Where do you see this landing: is opt-in the endpoint, or would a follow-up be worth filing so the issue has somewhere to point?

docs/content/docs/operations/configuration.md
```diff
@@ -168,6 +168,7 @@ Here are the configuration options for Kafka-based Action State Store.
 | `kafkaActionStateTopic`             | (none)                   | String  | The config parameter specifies the Kafka topic for action state.            |
 | `kafkaActionStateTopicNumPartitions`| 64                       | Integer | The config parameter specifies the number of partitions for the Kafka action state topic. |
 | `kafkaActionStateTopicReplicationFactor` | 1                     | Integer | The config parameter specifies the replication factor for the Kafka action state topic. |
+| `kafkaActionStateTombstoneEnabled`  | false                    | Boolean | Whether pruning sends tombstone records so log compaction can reclaim pruned keys. Off by default: without tombstones the topic grows unboundedly, but restoring any checkpoint or savepoint replays correctly. When enabled, restoring from the latest completed checkpoint is unaffected, but restoring an older checkpoint or savepoint may replay tombstones written after that restore point and re-execute already completed actions. Enable only if the job never restores from non-latest checkpoints or savepoints, or if re-executing actions is acceptable. |
```
weiqingy, line 171: Small one: this row and the new public `AgentConfigOptions.KAFKA_ACTION_STATE_TOMBSTONE_ENABLED` are both in the diff, while the description checks `doc-not-needed` and says "No public API changes". It also still describes the unconditional design rather than the opt-in one that shipped. Mind refreshing it before merge?

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStoreTest.java
```diff
@@ -196,6 +207,129 @@ void testPruneState() throws Exception {
         assertNull(
                 actionStates.get(ActionStateUtil.generateKey(TEST_KEY, 2L, testAction, testEvent)));
         assertNotNull(actionStateStore.get(TEST_KEY, 3L, testAction, testEvent));
+
+        // Assert - tombstones should have been sent to Kafka
+        var history = mockProducer.history();
+        assertThat(history).hasSize(2);
```
weiqingy, line 213: `testPruneStateSendsTombstonesWithCorrectKeys` pins this same contract more precisely — `containsExactlyInAnyOrder(key1, key2)` on the keys, `containsOnlyNulls()` on the values. A bug in tombstone keys, count, or nullness would fail both together, so I'm not sure this block catches anything the dedicated test would miss.

Would it read cleaner to leave `testPruneState` as it was — default store, cache eviction only, which also drops the `tombstoneEnabledStore` swap at the top — and keep the tombstone contract in one place? `testPruneStateNoTombstonesByDefault` already pins the default-off behavior.

rob-9:
thanks for the reviews @weiqingy @joeyutong . i'll try to address these soon.

wenjin272:
Hi @rob-9, just checking in on this PR. The comments from the last review round are still pending, and the branch now has merge conflicts. Are you still planning to continue working on it? If so, could you rebase and address the remaining comments when you have time? Otherwise, please let us know so someone else can pick it up. Thanks!

rob-9:
> Hi @rob-9, just checking in on this PR. The comments from the last review round are still pending, and the branch now has merge conflicts. Are you still planning to continue working on it? If so, could you rebase and address the remaining comments when you have time? Otherwise, please let us know so someone else can pick it up. Thanks!

Sorry about this, forgot about it. I'll address the comments and fix conflicts soon.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
rob-9: added matching Python option.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStoreTest.java
rob-9: fixed, the test now completes the producer callback exceptionally and verifies both callback invocation and cache eviction.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
rob-9: fixed, the test now uses a realistic `user_123` key, and the shared Kafka/Fluss limitation is documented and tested.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
rob-9: removed, Kafka ordering is preserved by the state key, and delivery failures remain reported through the callback.

docs/content/docs/operations/configuration.md
rob-9: updated.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStoreTest.java
rob-9: addressed. `testPruneState` now covers cache eviction, while the dedicated tests own tombstone and default-off assertions.

rob-9:
Hey all, comments should be addressed. The follow-up issue for checkpoint-aligned cleanup is #1034 .

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStoreTest.java
```diff
@@ -121,6 +121,17 @@ void testGetTriggersDivergenceCleanup() throws Exception {
...
 
+    @Test
+    void testPruneStateRetainsKeyContainingUnderscore() throws Exception {
+        String agentKey = "user_123";
+        String stateKey = ActionStateUtil.generateKey(agentKey, 1L, testAction, testEvent);
+        actionStates.put(stateKey, testActionState);
+
+        store.pruneState(agentKey, 1L);
+
+        assertThat(actionStates).containsKey(stateKey);
```
weiqingy: This pins one half of the Kafka fix on the Fluss side. The other half, the exact first-part check at `KafkaActionStateStore.java:301`, doesn't look like it came across.

`FlussActionStateStore.pruneState:489` calls `removeStateEntries(key.toString() + "_", stateSeqNum -> stateSeqNum <= seqNum)`, and `removeStateEntries:243` filters on `startsWith(keyPrefix)` plus the parsed sequence number, with no comparison of `parts.get(0)` against the key being pruned.

So Flink key `a` at seq 1 stores `a_1_<eventUuid>_<actionUuid>`, which parses cleanly as 4 parts with `parts.get(1) == "1"`. Pruning the distinct key `a_1` at any seqNum >= 1 matches the prefix `a_1_` and satisfies `1 <= seqNum`, so key `a`'s completed state is evicted. `FlussActionStateStore.get:227` reads the cache only, so the next lookup returns null and the action re-runs.

Is leaving Fluss on the prefix match deliberate, or would the same `parts.get(0)` check belong in `removeStateEntries`?

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStoreTest.java
```diff
@@ -142,6 +165,16 @@ void testGetActionStateWithDiverge() throws Exception {
         assertNull(actionStateStore.get(TEST_KEY, 4L, testAction, testEvent));
     }
 
+    @Test
+    void testGetRetainsUnparseableKey() throws Exception {
+        String flinkKey = "user_123";
+        String stateKey = ActionStateUtil.generateKey(flinkKey, 1L, testAction, testEvent);
+        actionStates.put(stateKey, testActionState);
+
+        assertThat(actionStateStore.get(flinkKey, 2L, testAction, testEvent)).isNull();
```
weiqingy: Does this discriminate the way its name suggests? The entry is stored at seq 1 and looked up at seq 2. The divergence cleanup only evicts entries with `stateSeqNum > seqNum` (`KafkaActionStateStore.java:182`), and 1 is not greater than 2, so the entry is retained whether or not `parseKey` throws. The lookup returns null either way, since no seq-2 key was ever stored, so both assertions hold against an implementation that parses `user_123` correctly.

Inverting the two sequence numbers would separate the cases, in case that's useful: store at seq 3, then `get(flinkKey, 1L, ...)`. Correct parsing evicts (3 > 1), only the parse failure retains it, and the seq-1 lookup is still a cache miss so the `removeIf` fires.

`testPruneStateSkipsUnparseableKeys` and the Fluss twin both do fail against a fixed implementation, so this looks like the odd one out.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
```diff
@@ -175,10 +181,13 @@ public ActionState get(Object key, long seqNum, Action action, Event event) thro
...
                                     }
-                                } catch (NumberFormatException e) {
+                                } catch (IllegalArgumentException e) {
                                     LOG.warn(
-                                            "Failed to parse sequence number from state key: {}",
-                                            stateKey);
+                                            "Cannot parse state key: {}. The entry cannot be "
+                                                    + "considered for divergence cleanup and will "
+                                                    + "be retained.",
+                                            entry.getKey(),
```
weiqingy: The widening isn't what I'm asking about here, it's how often this line can fire.

The `removeIf` scans the whole cache whenever `!actionStates.containsKey(stateKey) || hasDivergence` (`:170`), `get()` sits on the durable-execution path with three call sites in `DurableExecutionManager` (`:212`, `:218`, `:255`), and entries that fail to parse are retained by design. So one `user_123` entry produces a WARN on every subsequent cache miss for the life of the job, each with a full stack trace, since `e` is passed as a trailing arg.

`pruneState:305` already warns for the same key on the prune path. Would dropping the throwable here, or demoting this to DEBUG, cost signal you'd want to keep?

docs/content/docs/operations/configuration.md
```diff
@@ -170,6 +170,8 @@ The eight `memory.generate-event*` options have no raw `ConfigOption` default. W
 |------------------------------|------------------|---------|------------------------------------------------------------------------------------------|
 | `actionStateStoreBackend`    | (none)           | String  | The backend for action state store. Supported values: `"kafka"`, `"fluss"`.              |
 
+Durable action state stores currently join raw Flink keys and other key parts with an unescaped `_`. Flink keys containing `_` cannot be parsed safely during pruning, so both Kafka and Fluss retain their state in memory and backend storage. Kafka also emits no tombstones for those keys.
```
weiqingy: nit: "both Kafka and Fluss retain their state in memory and backend storage" reads as though backend retention follows from the parse failure. On the Fluss side it doesn't. `FlussActionStateStore.pruneState:487` never deletes from the backend for any key, parseable or not, and its javadoc says why: "The Fluss log is append-only; physical cleanup relies on Fluss log retention configuration."

Two smaller things alongside it. All three doc sites frame the limitation as growth, but the same parse failure also makes the divergence cleanup in `get()` inert (`KafkaActionStateStore.java:184`, `FlussActionStateStore.java:257`), so stale higher-seq state survives a divergence that was actually detected. That's a correctness cost rather than a storage one.

And #1034 covers the checkpoint boundary, but nothing seems to track the key encoding itself. Worth its own issue?

docs/content/docs/operations/deployment.md
```diff
@@ -92,7 +92,7 @@ After recovery from a checkpoint, Flink Agents reprocess events that arrived aft
 
 ### Exactly-Once Action Consistency
 
-To ensure exactly-once action consistency, you must configure an external action state store. Flink Agents record action state in this store on a per-action basis. After recovering from a checkpoint, Flink Agents consult the external store and will not re-execute actions that were already completed. This guarantees each action is executed exactly once after recovering from a checkpoint.
+To ensure exactly-once action consistency, you must configure an external action state store. Flink Agents record action state in this store on a per-action basis. After recovering from a checkpoint, Flink Agents consult the external store and reuse completed action state when its backing record remains available. This prevents re-execution for checkpoints supported by the store's retained recovery history.
```
weiqingy: nit: this paragraph covers both backends, including Fluss and the default Kafka setup with tombstones off. The hazard it now hedges for is opt-in and Kafka-only, and the `hint warning` box a few lines below states it with that scope. "backing record remains available" and "the store's retained recovery history" also appear nowhere else in either doc, so a reader has nothing to resolve them against.

Is the hedge doing work here that the warning box below doesn't already do? One way this could read is with the original sentence restored: "This guarantees each action is executed exactly once after recovering from a checkpoint."

weiqingy (commented):
Thanks for working through these. The limitation around keys containing _ is documented and pinned by a test on both stores now, and #1034 gives what's left of #691 somewhere to point.

A few questions inline.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
weiqingy: Confirmed, closed on my side.

docs/content/docs/operations/configuration.md
weiqingy: Checked the updated body and docs, this one's closed.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStoreTest.java
weiqingy: Closed on my side, thanks.

api/src/main/java/org/apache/flink/agents/api/configuration/AgentConfigOptions.java
weiqingy: Saw #1034, that answers this. It's the right home for what #691 still has open, so I'll follow the discussion there.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
rob-9: Demoted this to DEBUG and removed the throwable, so repeated cache misses no longer produce WARN stack traces.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStoreTest.java
rob-9: you're right. I inverted the sequence numbers so the test now fails if the key parses successfully and gets evicted.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStoreTest.java
rob-9: added the exact parsed-key check to Fluss, with tests covering both pruning and lookup cleanup.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
```diff
@@ -175,10 +181,11 @@ public ActionState get(Object key, long seqNum, Action action, Event event) thro
                                         // the requested seqNum
                                         return stateSeqNum > seqNum;
```
weiqingy: The Fluss `get()` test you just added looks like it would fail if it were pointed at Kafka. Copying `FlussActionStateStoreTest.java:145-152` into `KafkaActionStateStoreTest`: `get("a_1", 0L, ...)` misses, this `removeIf` parses the cached `a_1_<eventUuid>_<actionUuid>` (Flink key `a`, seq 1) into 4 parts, `1 > 0` fires, and the `containsKey` assertion fails. Kafka has the `pruneState` twin at `KafkaActionStateStoreTest.java:261-274`, but not this one.

The reason looks like this scan isn't tied to a key at all. There's no prefix filter and no `parts.get(0)` check, just `stateSeqNum > seqNum` applied to every entry in the map. One map holds every key the subtask owns (`:82`, store created per operator at `DurableExecutionManager.java:118-125`), and sequence numbers count per key (`DurableExecutionManager.java:233`), so one key's lookup can drop another key's state.

The cache isn't the durable record, so how much that costs depends on when the evicted key is read next. The case I can't rule out is just after a restore, when the replay path reads the rebuilt cache: `processActionTaskForKey` skips execution only when `actionState != null && actionState.isCompleted()` (`ActionExecutionOperator.java:435`), so an entry that got dropped runs again. That needs action tasks for different keys to interleave during replay, which I haven't reproduced, so I'd call it a hazard worth checking rather than a break.

To be clear, this predicate is base code and not something the PR introduced. The only change inside the block was the catch clause and its log line, and the other three scan sites do carry the guard at head (`:294-301` here, and `FlussActionStateStore.java:249-256` for both of its callers). Would the same `parts.get(0)` check you just added on Fluss fit here as well, or would you rather it went out separately?

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -250,10 +251,11 @@ private void removeStateEntries(String keyPrefix, LongPredicate seqNumFilter) {
...
-                                if (parts.size() >= 2) {
-                                    long stateSeqNum = Long.parseLong(parts.get(1));
-                                    return seqNumFilter.test(stateSeqNum);
+                                if (!parts.get(0).equals(key)) {
+                                    return false;
                                 }
+                                long stateSeqNum = Long.parseLong(parts.get(1));
+                                return seqNumFilter.test(stateSeqNum);
                             } catch (Exception e) {
                                 LOG.warn("Failed to parse state key: {}", entry.getKey(), e);
```
weiqingy: nit: the trailing `e` with a single `{}` means this logs a full stack trace, and it's the Fluss twin of the Kafka site you just demoted. A Flink key containing `_` never parses, so `:262` keeps the entries and `get()` runs this scan again on every new sequence number for that key, which repeats the trace for as long as the job lives.

It's narrower than the Kafka one, since this scan is prefix-filtered so only that key's own entries reach it, and the docs now describe the retention as expected. Would the same DEBUG treatment fit here, and is the stack trace earning its place on this path?

docs/content/docs/operations/configuration.md
weiqingy: nit: the Fluss claim and the divergence-cleanup cost both read accurately now across all three sites, so those two are closed. On the key encoding I also had a look at #1010, but that one is about foreign-key retention after a restore, so the encoding itself still has nowhere to point.

docs/content/docs/operations/deployment.md
weiqingy: nit: the hedge is gone, thanks. That was my wording though. I quoted only the last sentence and called it the original, so the one before it got dropped along the way. On `main` the paragraph also has "After recovering from a checkpoint, Flink Agents consult the external store and will not re-execute actions that were already completed." Without it, "This guarantees..." now attaches to recording state per action rather than to the consult-and-skip behavior it described. Worth putting back?

rob-9:
After merging `main` and reviewing the full recovery path, I found one additional issue. 

The current action-state key stores the typed key-group but still identifies the business key using `key.toString()`. Distinct Flink keys such as `Long(1)` and `String("1")` can therefore be conflated during lookup and pruning. With tombstones enabled, pruning one key could durably delete another key’s recovery state. 

This comes from the shared key format introduced through #1024 and affects both Kafka and Fluss, so I’ll track the fix in a separate issue. This PR should wait for that fix since tombstones make the existing collision irreversible. 

The merge conflicts and existing review comments are otherwise addressed locally.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/KafkaActionStateStore.java
rob-9: Addressed after merging `main`: Kafka cleanup now uses the shared segment-exact matcher, so the reported `a` / `a_1` case no longer removes another key's state. I also added coverage for the cache-miss cleanup path.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
rob-9: Addressed after merging `main`: underscore-containing keys now parse correctly, and the shared matcher no longer emits the repeated full-stack-trace warning described here.

docs/content/docs/operations/configuration.md
rob-9: Addressed by the five-part key format merged through #1024. The business key is now the final bounded segment, so underscores are preserved without escaping. I’ll update the PR description to remove the obsolete limitation.

docs/content/docs/operations/deployment.md
rob-9: Restored.

weiqingy:
Thanks for working through these. All my comments are addressed now, and nothing further from me.

rob-9:
> Thanks for working through these. All my comments are addressed now, and nothing further from me.

great! I believe the logical PR order is now : 

#1094 -> #1100 -> #885 (this PR) -> #1101

Issue #886, opened 2026-07-09
[Feature] Add memory events for agent memory observability


rosemarYuan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Follow-up to #876 .

Memory is a core component of agent behavior, but it is currently not observable through the framework’s existing event-based tooling.

**Proposal.** At each action finish boundary, the framework folds the memory operations performed by that action into generated memory events. These events summarize the net effect per memory scope and operation kind, using a flat dot-key `value` format.

An agent-run-begin event additionally carries the full short-term memory snapshot at run start, serving as the anchor for offline reconstruction. Event generation is controlled by the `memory.generate-event` configuration family.

**Non-goals.** These events do not fully reconstruct long-term memory. LTM events are intended for audit/debug only, because requested updates may differ from the final backend state.

This proposal does not introduce a new persistence format or metrics system. Memory activity reuses the existing event pipeline.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Issue #841, opened 2026-06-11
[Feature] Add lineage fields to the event log for per-run trace reconstruction


addu390:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Follow-up from #710.

The event log records every event with its full payload, but you can't rebuild a single run as a causal tree. A record has no link to the action that emitted the event or the event that action was processing.

Proposal: add a few lineage fields to each event log record (emitting action, source event id, run/trace id) so a run's trace tree can be rebuilt offline from the log. Reverse edges (event to triggered actions) can be derived from the agent plan's trigger rules, so no need to record them.

Scope is recording only, per #710. Reconstruction, visualization and OTel export are potential future follow-ups. I'll share a short design here before implementing.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

zxs1633079383:
I took a pass through the current event-log/runtime shape and I think this can be a narrow first PR if the lineage data stays on the log record/context rather than inside user event attributes.

Current insertion-point observation:

- `EventLogRecord` is already the right wire object for the log; it currently carries `EventContext` + `Event`.
- `EventContext` currently has only `eventType` and `timestamp`.
- `EventRouter.notifyEventProcessed(Event)` only receives the event, so it cannot know which action emitted it.
- `RunnerContextImpl.sendEvent(Event)` does know the current `actionName`, but that metadata is lost before the event reaches the logger.

My suggested first-PR shape:

1. Add a small lineage/context object to `EventLogRecord` or extend `EventContext` with optional fields:
   - `logSchemaVersion`
   - `runId` / `traceId`
   - `eventId`
   - `sourceEventId` or `parentEventId`
   - `emittingAction`
   - optionally the current key/sequence number if we can expose them without leaking unstable internals
2. Keep these fields outside `event.attributes`, so the event payload remains user-owned and cross-language JSON-compatible.
3. Treat each input event as the root of a run: `runId = inputEvent.id`, no parent/source event, no emitting action.
4. For events emitted from an action, set `runId` from the source event, `sourceEventId = actionTask.event.id`, and `emittingAction = action.name`.
5. Leave reverse edges out of the log, as proposed. They can still be derived from the agent plan's trigger rules during offline reconstruction.

This keeps the first implementation scoped to recording and replay/debuggability. Visualization, OTel export, and trace-tree reconstruction can be follow-ups once the log has stable causal fields.

[...]

PR #887, opened 2026-07-09
[api][runtime][python] Add memory events for agent memory observability

rosemarYuan:
Linked issue: #886 

### Purpose of change

Follow-up to #876, #886 .This change makes agent memory activity observable through the existing event infrastructure.

**Emission model.**
* Memory events are emitted at the action finish boundary, together with the action's user events.
* The framework emits at most one event per memory scope and operation kind, for example, one short-term write event per action.
* Each event's `value` is a dot-key flat map describing the operation's **net effect**. Within one action, the last read/write for the same path wins.
* **Failed or unfinished actions emit no memory events.**
* Actions triggered by memory events do not emit memory events themselves, preventing recursive subscriptions.
* Long-term-memory observation records are scoped by both the keyed partition and a stable action-execution observation ID.

**Recovery semantics.** Generation remains consistent with the framework's existing delivery and durability guarantees:

* **At-least-once on replay.** After a failure, a replayed completed action may re-log its persisted memory events, and a replayed input may re-emit its run-begin event with restored short-term memory. Event Log consumers should tolerate duplicates.
* **Durable execution.** A long-term operation wrapped in `durable_execute` is not recorded again when its result is replayed from state. Generated events describe operations that actually executed in the current attempt.
* **Action-execution identity.** The observation ID is retained across Java and Python continuations and checkpoint recovery. Tasks restored from older state without an observation ID receive a stable fallback before execution resumes.

### Tests

Coverage targets the design-critical paths:
* **Cross-language snapshot conformance** —`CrossLanguageEventSnapshotTest` in Java and  `test_cross_language_event_snapshots.py` in Python verify stable Java/Python snapshots and reciprocal deserialization for memory and run-begin events.
* **Flush and failure semantics** — `TestMemoryObservationFlush` and `ActionExecutionOperatorTest` cover action-finish emission, config and suppression gating, best-effort drain failures, and cleanup of observation records when an action fails.
* **Action-scoped LTM isolation** — `ActionTaskTest`, `ActionTaskContextManagerTest`, and `test_mem0_recording_hook` cover distinct observation IDs for same-key actions, ID preservation across Python continuations and recovery, exact `(partition key, observation ID)` matching, and re-buffering of non-matching records.
* **Config fallback** — `TestMemoryEventSettings` covers all three resolution levels, including the case where the master switch is off but an operation-specific sub-key is explicitly on.
* **Dot-key folding** — `TestMemoryEventBuilder` covers empty paths, single paths, nested paths, last-wins behavior, and long-term add/delete/get/search value shapes.
* **Polymorphic serde** — `ActionStateSerdeTest` verifies that memory events round-trip through the `@class` mixin so they can be persisted and replayed with the action's output events.
* **Agent-run begin** — `AgentRunBeginEventEmissionTest`, `AgentRunBeginEventTest`, and `test_run_event` cover the opt-in default, snapshot content, textual keys, per-key isolation, and the guarantee that the snapshot is taken before the run's own writes.
* **Long-term recording in Python** — `test_mem0_op_recording` and `test_mem0_recording_hook` cover operation shapes, thread-safe buffering, and scoped draining.
* **TTL behavior** — `ShortTermMemoryTTLIntegrationTest` verifies that the default remains `ON_READ_AND_WRITE`, including read-refresh behavior, and covers the explicit `ON_CREATE_AND_WRITE` alternative.
* **End-to-end logging** — `memory_event_logging_test.py` exercises the full emit → Event Log path.

### API

Java:

* `org.apache.flink.agents.api.event.MemoryEvent`
* `org.apache.flink.agents.api.event.ShortTermWriteEvent`
* `org.apache.flink.agents.api.event.ShortTermReadEvent`
* `org.apache.flink.agents.api.event.SensoryWriteEvent`
* `org.apache.flink.agents.api.event.SensoryReadEvent`
* `org.apache.flink.agents.api.event.LongTermUpdateEvent`
* `org.apache.flink.agents.api.event.LongTermGetEvent`
* `org.apache.flink.agents.api.event.LongTermSearchEvent`
* `org.apache.flink.agents.api.event.AgentRunBeginEvent`
* The corresponding `EventType.*` constants
* `org.apache.flink.agents.api.configuration.MemoryEventOptions`

Python:

* `flink_agents.api.events.memory_event`
* `flink_agents.api.events.run_event`
* `flink_agents.api.core_options.MemoryEventOptions`

The new memory-event surface — event classes, `EventType.*` constants, and `MemoryEventOptions` — does not depend on any long-term-memory backend.

The LTM observation seam is an internal runtime contract. Its observation methods are abstract and must be implemented by the backend; the Mem0 adapter provides the current implementation.

**Configuration.** Memory event generation is controlled per operation kind. Configuration is resolved in the following order:

1. Operation-specific sub-key.
2. `memory.generate-event` master switch.
3. Built-in per-operation default.

The built-in defaults are:
* Writes: on.
* Long-term operations: on.
* Short-term/sensory reads: off.

The master switch and operation-specific switches intentionally use `null` defaults, so the built-in defaults remain reachable.
`agent-run.begin-event` has its own switch, defaults to off(false), and is independent of the `memory.generate-event` master switch.

**Run-begin snapshot and TTL.**

* The repo-wide default of `short-term-memory.state-ttl.update-type` remains `ON_READ_AND_WRITE`.
* When `agent-run.begin-event` is enabled, every input scans the key's short-term memory to produce the run-begin snapshot. Under `ON_READ_AND_WRITE`, that scan refreshes the TTL of the entries it reads.
* Leave `agent-run.begin-event` disabled when the snapshot is not needed.
* If the snapshot is needed but reads should not extend TTL, configure `short-term-memory.state-ttl.update-type` as `ON_CREATE_AND_WRITE`.

See the memory-event and short-term-memory documentation for the complete semantics and caveats.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included` Documentation is included in `docs/content/docs/development/memory/memory_events.md
This change also adds notes to the sensory/short-term memory page and the configuration reference.

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/event/MemoryEvent.java
+++ b/api/src/main/java/org/apache/flink/agents/api/event/MemoryEvent.java
@@ -1,0 +1,138 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.event;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.core.json.JsonWriteFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
+import org.apache.flink.agents.api.Event;
+
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.function.BiFunction;
+
+/**
+ * Base class of the memory observation events. One event per (memory scope x operation) is emitted
+ * at the action finish boundary; each concrete subclass pins one of the seven operation kinds as
+ * its {@code type}.
+ *
+ * <p>Attributes: {@code key} — the String Flink key the operation belongs to; {@code value} — the
+ * operation's folded JSON value map. Framework observation events are skipped for non-String keyed
+ * streams. On the wire the event serializes as {@code {id, type, attributes: {key, value}}}.
+ */
+public abstract class MemoryEvent extends Event {
+
+    private static final ObjectMapper MAPPER =
+            JsonMapper.builder().disable(JsonWriteFeature.WRITE_NAN_AS_STRINGS).build();
+
+    /**
+     * The seven memory event types, each mapped to its deserialization constructor ({@code (id,
+     * attributes)}). Single source of truth for both {@link #isMemoryType} and {@link #fromEvent}.
+     */
+    private static final Map<String, BiFunction<UUID, Map<String, Object>, MemoryEvent>> FACTORIES =
+            Map.<String, BiFunction<UUID, Map<String, Object>, MemoryEvent>>of(
+                    ShortTermWriteEvent.EVENT_TYPE, ShortTermWriteEvent::new,
+                    ShortTermReadEvent.EVENT_TYPE, ShortTermReadEvent::new,
+                    SensoryWriteEvent.EVENT_TYPE, SensoryWriteEvent::new,
+                    SensoryReadEvent.EVENT_TYPE, SensoryReadEvent::new,
+                    LongTermUpdateEvent.EVENT_TYPE, LongTermUpdateEvent::new,
+                    LongTermGetEvent.EVENT_TYPE, LongTermGetEvent::new,
+                    LongTermSearchEvent.EVENT_TYPE, LongTermSearchEvent::new);
+
+    /** Returns true iff {@code type} is one of the seven memory operation event types. */
+    public static boolean isMemoryType(String type) {
+        return type != null && FACTORIES.containsKey(type);
+    }
+
+    protected MemoryEvent(String type, String key, Map<String, Object> value) {
+        super(type, normalizeAttributes(key, value));
+    }
+
+    protected MemoryEvent(String type, Map<String, Object> attributes) {
+        super(type, normalizeAttributes(attributes));
+    }
+
+    /**
+     * Deserialization constructor path. The subclass {@code @JsonCreator}s are REQUIRED: instances
+     * land in ActionState.outputEvents and deserialize polymorphically via the {@code @class} mixin
+     * in ActionStateSerde — without an explicit creator Jackson cannot instantiate the subtype.
+     */
+    protected MemoryEvent(UUID id, String type, Map<String, Object> attributes) {
+        super(id, type, normalizeAttributes(attributes));
+    }
+
+    static Map<String, Object> normalizeAttributes(Map<String, Object> attributes) {
+        if (attributes == null) {
+            throw new IllegalArgumentException("Memory observation attributes must not be null.");
+        }
+        Object key = attributes.get("key");
+        if (!(key instanceof String)) {
+            throw new IllegalArgumentException(
+                    "Memory observation attribute 'key' must be a string.");
+        }
+        Object value = attributes.get("value");
+        if (!(value instanceof Map)) {
+            throw new IllegalArgumentException(
+                    "Memory observation attribute 'value' must be a map.");
+        }
+        Map<String, Object> normalized = new LinkedHashMap<>(attributes);
+        try {
+            normalized.put(
+                    "value", MAPPER.readValue(MAPPER.writeValueAsBytes(value), Object.class));
+        } catch (IOException | RuntimeException e) {
+            throw new IllegalArgumentException(
+                    "Memory observation value must be JSON serializable.", e);
+        }
+        return normalized;
+    }
+
+    static Map<String, Object> normalizeAttributes(String key, Map<String, Object> value) {
+        Map<String, Object> attributes = new LinkedHashMap<>();
+        attributes.put("key", key);
+        attributes.put("value", value);
+        return normalizeAttributes(attributes);
+    }
+
+    /** Converts a generic {@link Event} carrying a memory type into its typed subclass view. */
+    public static MemoryEvent fromEvent(Event event) {
+        BiFunction<UUID, Map<String, Object>, MemoryEvent> factory = FACTORIES.get(event.getType());
+        if (factory == null) {
+            throw new IllegalArgumentException("Not a memory event type: " + event.getType());
+        }
+        MemoryEvent result = factory.apply(event.getId(), event.getAttributes());
+        if (event.hasSourceTimestamp()) {
+            result.setSourceTimestamp(event.getSourceTimestamp());
+        }
+        return result;
+    }
+
+    @JsonIgnore
+    public String getKey() {
+        return (String) getAttr("key");
+    }
+
+    @SuppressWarnings("unchecked")
+    @JsonIgnore
+    public Map<String, Object> getValue() {
+        return (Map<String, Object>) getAttr("value");
+    }
+}
--- a/api/src/main/java/org/apache/flink/agents/api/event/MemoryEvent.java
+++ b/api/src/main/java/org/apache/flink/agents/api/event/MemoryEvent.java
@@ -35,9 +35,9 @@ import java.util.function.BiFunction;
  * at the action finish boundary; each concrete subclass pins one of the seven operation kinds as
  * its {@code type}.
  *
- * <p>Attributes: {@code key} — the String Flink key the operation belongs to; {@code value} — the
- * operation's folded JSON value map. Framework observation events are skipped for non-String keyed
- * streams. On the wire the event serializes as {@code {id, type, attributes: {key, value}}}.
+ * <p>Attributes: {@code key} — the textual form of the Flink key the operation belongs to; {@code
+ * value} — the operation's folded JSON value map. On the wire the event serializes as {@code {id,
+ * type, attributes: {key, value}}}.
  */
 public abstract class MemoryEvent extends Event {
 
@@ -118,11 +118,7 @@ public abstract class MemoryEvent extends Event {
         if (factory == null) {
             throw new IllegalArgumentException("Not a memory event type: " + event.getType());
         }
-        MemoryEvent result = factory.apply(event.getId(), event.getAttributes());
-        if (event.hasSourceTimestamp()) {
-            result.setSourceTimestamp(event.getSourceTimestamp());
-        }
-        return result;
+        return reconstructFrom(event, factory);
     }
 
     @JsonIgnore
--- a/docs/content/docs/development/memory/long_term_memory.md
+++ b/docs/content/docs/development/memory/long_term_memory.md
@@ -539,13 +539,15 @@ public static void processEvent(Event event, RunnerContext ctx) throws Exception
 
 ## Context Isolation
 
-Long-Term Memory automatically provides context isolation through Flink's keyed partition model. Each keyed partition maintains its own isolated set of memories, ensuring that memories from one user or session do not leak into another.
+Long-Term Memory provides context isolation through Flink's keyed partition model. When each logical key has a stable and unique textual representation, each keyed partition uses an isolated memory context so memories from one user or session do not leak into another.
 
 The isolation hierarchy works as follows:
 - **Job-level** (`JOB_IDENTIFIER`): Separates memories between different Flink jobs
 - **Partition-level** (keyed partition key): Separates memories between different keys within the same job
 - **Set-level** (memory set name): Separates memories between different logical categories within the same partition
 
-This means you can reuse the same memory set name across different partitions, and each partition will normally access only its own memories.
+This means you can reuse the same memory set name across different partitions, and each partition will normally access only its own memories. The partition context uses the textual representation of the logical Flink key: Java and explicitly typed PyFlink keys normally use `String.valueOf`; default pickled PyFlink keys are deserialized and use Python `str`. Explicit PyFlink primitive byte-array keys use Python's bytes representation directly, without pickle deserialization.
 
-> **Note:** Partition-level isolation is currently derived from the hash of the partition key (`String.valueOf(key.hashCode())`) rather than the full original key. Distinct keys whose hashes collide may therefore share the same memory context. Avoid relying on isolation as a strict security boundary; if collision-free isolation is required, encode a unique identifier into the memory set name.
+Keys used for Long-Term Memory should therefore have a stable and unique textual representation within the job. Distinct logical keys that produce the same text share one Mem0 context, so this representation is not a security boundary.
+
+> **Compatibility note:** Earlier experimental versions used `String.valueOf(key.hashCode())` as the Mem0 partition identity. Existing records stored under that hash-derived identity are not migrated automatically and are not visible through the new logical-key identity. Migrate those records explicitly before upgrading if they must remain accessible.
--- a/docs/content/docs/development/memory/long_term_memory.md
+++ b/docs/content/docs/development/memory/long_term_memory.md
@@ -539,15 +539,13 @@ public static void processEvent(Event event, RunnerContext ctx) throws Exception
 
 ## Context Isolation
 
-Long-Term Memory provides context isolation through Flink's keyed partition model. When each logical key has a stable and unique textual representation, each keyed partition uses an isolated memory context so memories from one user or session do not leak into another.
+Long-Term Memory automatically provides context isolation through Flink's keyed partition model. Each keyed partition normally maintains its own isolated set of memories.
 
 The isolation hierarchy works as follows:
 - **Job-level** (`JOB_IDENTIFIER`): Separates memories between different Flink jobs
 - **Partition-level** (keyed partition key): Separates memories between different keys within the same job
 - **Set-level** (memory set name): Separates memories between different logical categories within the same partition
 
-This means you can reuse the same memory set name across different partitions, and each partition will normally access only its own memories. The partition context uses the textual representation of the logical Flink key: Java and explicitly typed PyFlink keys normally use `String.valueOf`; default pickled PyFlink keys are deserialized and use Python `str`. Explicit PyFlink primitive byte-array keys use Python's bytes representation directly, without pickle deserialization.
+This means you can reuse the same memory set name across different partitions, and each partition will normally access only its own memories.
 
-Keys used for Long-Term Memory should therefore have a stable and unique textual representation within the job. Distinct logical keys that produce the same text share one Mem0 context, so this representation is not a security boundary.
-
-> **Compatibility note:** Earlier experimental versions used `String.valueOf(key.hashCode())` as the Mem0 partition identity. Existing records stored under that hash-derived identity are not migrated automatically and are not visible through the new logical-key identity. Migrate those records explicitly before upgrading if they must remain accessible.
+> **Note:** Partition-level isolation uses a textual identity derived from the logical key instead of `key.hashCode()`. Java keys use `String.valueOf(key)`. Default-serialized PyFlink keys are deserialized and use Python `str`; explicitly typed PyFlink keys use `String.valueOf`, except byte-array keys, which use Python's bytes representation. This avoids hash collisions, but distinct keys may still share a memory context if they produce the same text, for example when custom key types have non-unique `toString()` or `__str__()` implementations. Key representations should therefore be stable and unique, and this isolation should not be treated as a security boundary.
--- a/docs/content/docs/development/memory/sensory_and_short_term_memory.md
+++ b/docs/content/docs/development/memory/sensory_and_short_term_memory.md
@@ -291,8 +291,12 @@ Short-term memory can be configured with a time-to-live (TTL) so that older stat
 
 Set `short-term-memory.state-ttl.ms` to a value greater than 0 in milliseconds to enable TTL. You can also configure how the TTL is refreshed and whether expired state can be returned before Flink cleans it up:
 
-- `short-term-memory.state-ttl.update-type`: controls whether TTL is refreshed on create/write or on read/write.
+- `short-term-memory.state-ttl.update-type`: controls whether TTL is refreshed on create/write (`ON_CREATE_AND_WRITE`) or on read/write (`ON_READ_AND_WRITE`, the default).
 - `short-term-memory.state-ttl.visibility`: controls whether expired memory is never returned or may be returned if it has not been cleaned up yet.
+
+{{< hint warning >}}
+The default `ON_READ_AND_WRITE` update type extends an entry's lifetime whenever it is read. This also applies when producing the run-begin snapshot used by [Memory Events]({{< ref "docs/development/memory/memory_events" >}}): if you opt in through `agent-run.begin-event`, each input scans the key's short-term memory and refreshes TTL for the entries it reads, although only value nodes are included in the event. Choose `ON_CREATE_AND_WRITE` when entries should expire based only on writes.
+{{< /hint >}}
 
 {{< tabs "Short-Term Memory TTL Configuration" >}}
 
--- a/docs/content/docs/development/memory/sensory_and_short_term_memory.md
+++ b/docs/content/docs/development/memory/sensory_and_short_term_memory.md
@@ -295,7 +295,7 @@ Set `short-term-memory.state-ttl.ms` to a value greater than 0 in milliseconds t
 - `short-term-memory.state-ttl.visibility`: controls whether expired memory is never returned or may be returned if it has not been cleaned up yet.
 
 {{< hint warning >}}
-The default `ON_READ_AND_WRITE` update type extends an entry's lifetime whenever it is read. This also applies when producing the run-begin snapshot used by [Memory Events]({{< ref "docs/development/memory/memory_events" >}}): if you opt in through `agent-run.begin-event`, each input scans the key's short-term memory and refreshes TTL for the entries it reads, although only value nodes are included in the event. Choose `ON_CREATE_AND_WRITE` when entries should expire based only on writes.
+With the default update type `ON_READ_AND_WRITE`, every read refreshes an entry's TTL. Enabling `agent-run.begin-event` introduces an additional source of reads: each input scans the key's short-term memory to produce the run-begin snapshot used by [Memory Events]({{< ref "docs/development/memory/memory_events" >}}), which may extend the lifetime of the scanned entries even though only value nodes are included in the event. Leave `agent-run.begin-event` disabled if the snapshot is not needed. If the snapshot is needed but reads should not extend TTL, use `ON_CREATE_AND_WRITE`.
 {{< /hint >}}
 
 {{< tabs "Short-Term Memory TTL Configuration" >}}
--- a/python/flink_agents/api/core_options.py
+++ b/python/flink_agents/api/core_options.py
@@ -262,9 +262,6 @@ class MemoryEventOptions:
     LONG_TERM_SEARCH = ConfigOption(
         key="memory.generate-event.long-term-search", config_type=bool, default=None
     )
-    AGENT_RUN_BEGIN_EVENT = ConfigOption(
-        key="agent-run.begin-event", config_type=bool, default=False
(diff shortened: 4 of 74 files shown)


zxs1633079383 (commented):
I took a light design pass, mainly around whether the new memory events pollute user payloads or collide with the event-log lineage direction.

The current shape looks good to me:

- Memory observations are emitted as separate typed events (`_short_term_write_event`, `_long_term_search_event`, etc.) rather than mutating the triggering user event or putting framework fields into user-owned event attributes.
- The memory event payload is scoped under `attributes: { key, value }`, so the observation payload is explicit and reconstructable without changing the user event schema.
- The action-finish flush boundary is a good place to fold per-action memory records into net-effect events, and the suppression for memory-triggered actions avoids recursive observation noise.
- `AgentRunBeginEvent` is kept as a lifecycle/reconstruction anchor rather than overloading memory events themselves.

One thing I would keep intentionally separate from this PR: lineage/run identity fields from #841. This PR should probably continue to avoid adding `runId`, `sourceEventId`, `emittingAction`, etc. directly into the memory event `value` payload. If/when the event-log lineage schema lands, those fields should sit in event context/envelope metadata, not inside the user-facing memory observation value map. The docs already point in that direction by describing `key`/`value` as the memory event attributes, so this is more of a boundary note than a blocker.

rosemarYuan:
Thanks for the design review @zxs1633079383 — agreed, and that is exactly the boundary I want to keep.

For this PR, the memory event payload under `attributes: { key, value }` stays observation-only: it only describes the memory content/effect being observed. No #841 lineage or run-identity fields such as `runId`, `sourceEventId`, or `emittingAction` should be added inside the memory event `value`/`attributes` map.

On the lineage direction, we are aligned. If/when the event-log lineage schema lands, those fields should be carried as event-log metadata/envelope-level information and apply uniformly to logged events, memory events included. The exact carrier shape, for example whether it is a separate `lineage` object on the log record or another context/envelope mechanism, belongs in the lineage design thread rather than this PR.

Same principle you described: memory events expose memory observation data; lineage stays outside the user-facing event payload.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/MemoryEventBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+            String key,
+            MemoryOp op,
+            List<MemoryUpdate> records,
+            MemoryEventSettings settings) {
+        if (records == null || records.isEmpty() || !settings.generate(op)) {
+            return;
+        }
+        Map<String, Object> flat = new LinkedHashMap<>();
+        for (MemoryUpdate update : records) {
+            flat.put(update.getPath(), update.getValue());
```
weiqingy: This puts the raw memory value object straight into the event's `value` map (`update.getValue()`, no normalization). Those events get drained into `pendingEvents`, persisted into `ActionState.outputEvents`, and re-emitted on replay via the completed-action path. The Kryo `{serde,version,payload}` envelope that preserves an untyped memory value's concrete type across durable recovery is bound class-scoped to `MemoryUpdate.value` (`addMixIn(MemoryUpdate.class, ...)` in `ActionStateSerde`) — it doesn't cover `Event.attributes`, which serializes through stock Jackson. So a `byte[]` short-term value (a supported type — `ActionStateSerdeTest.testByteArrayMemoryValuePreserved`) comes back as a base64 `String` after recovery, and a POJO as a `LinkedHashMap`. An action subscribing to `_short_term_write_event` would then see `byte[]` on the live run and `String` on the recovered run.

It only bites under a narrow combination — durable execution enabled, a non-JSON-native STM value, and an action that actually depends on the concrete type — so I don't think it blocks merge. And it may well be intentional: if memory events are audit-only, a base64 string is arguably fine. Is that the intent, or should the event path reuse the same envelope? Either way, `testMemoryEventsSurviveActionStateRoundTrip` currently uses only `"gold"` and `1` (both JSON-native, so type is trivially preserved) — a `byte[]` assertion there would pin whichever answer you land on.

python/flink_agents/api/core_options.py
```diff
@@ -215,6 +215,42 @@ class AgentConfigOptions:
     )
 
 
+class MemoryEventOptions:
```
weiqingy, line 234: There's an enforced parity guard that reflects a Java options class against its Python mirror and asserts the field names, keys, types, and defaults all match — `check_java_python_config_options_parity.py`, run in CI via `test_java_config_in_python.sh`. Its `main()` registers `AgentConfigOptions` and `AgentExecutionOptions`, but this new `MemoryEventOptions` family isn't wired in on either side. `test_memory_event.py:104` looks like it fills that gap, but it checks the Python constants against hardcoded literals — it can't see Java, so by construction it can't catch Java↔Python drift. Given the 9 new keys and their tri-state `null` defaults are exactly the shape that's easiest to break asymmetrically, would registering the class in that checker's `main()` be worth it? `normalize_java_default` already maps a Java `null` default to Python `None`, so it should compare cleanly with no checker changes.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -169,6 +218,56 @@ public List<Event> drainEvents(Long timestamp) {
         return list;
     }
 
+    private void flushMemoryObservation() {
+        if (memoryContext == null) {
+            return;
+        }
+        // ALWAYS drain LTM records for this key (mailbox thread) — discarded below if
+        // suppressed/disabled; skipping the drain would leak them into the next action.
```
weiqingy: The "ALWAYS drain" invariant here holds for suppressed/disabled actions — they finish normally and hit this drain before the early returns. But `flushMemoryObservation` only runs from `drainEvents(_, true)`, which the action tasks call only on normal finish (`JavaActionTask:85-86`, and the Python task). When an action throws after doing LTM ops, `invoke()` rethrows, `drainEvents(_, true)` is never reached, and the Python-side buffer for this key isn't drained — so the guarantee is actually narrower than "a suppressed or disabled action still drains the buffer" reads. In practice it looks masked: a thrown action fails the mailbox task → operator restart → the interpreter's buffer is recreated fresh. I traced that path and restart appears to be the only exit, but I couldn't prove it exhaustively — is operator restart guaranteed to be the only way out here? A small test driving an action that throws and asserting no memory events leak into the next same-key action would settle both the invariant and my uncertainty.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -169,6 +218,56 @@ public List<Event> drainEvents(Long timestamp) {
...
+        List<Map<String, Object>> ltmRecords = Collections.emptyList();
+        if (ltm != null) {
+            try {
+                ltmRecords =
+                        MemoryEventBuilder.parseDrainedRecords(
+                                ltm.drainObservationRecordsJson(observationKeyHash));
+            } catch (Exception e) {
+                LOG.debug("LTM observation drain failed; skipping", e);
```
weiqingy: If the Pemja `drain_ltm_observation_records` call throws, this swallows it at DEBUG and moves on. Two things go quiet: this action's LTM observation is lost, and if the Python side threw *before* clearing its buffer, those records survive and get attributed to the next same-key action's flush — the exact cross-action leak the drain exists to prevent. A persistent drain failure would misattribute LTM records with no visible signal. Would a throttled `WARN` fit better here, so the condition is at least observable?

runtime/src/main/java/org/apache/flink/agents/runtime/memory/MemoryEventBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+                continue;
+            }
+            Object id = record.get("id");
+            Object value = record.get("value");
+            if ("ADD".equals(op)) {
+                updates.put(set + "." + id, value);
+            } else if ("DELETE".equals(op)) {
+                updates.put(set + "." + id, null);
+            } else if ("DELETE_SET".equals(op)) {
+                updates.put(set, null);
```
weiqingy: `DELETE_SET` adds `set -> null` but doesn't purge the earlier `set + "." + id` entries folded in above, so `ADD s.m1=v` then `DELETE_SET s` within one action yields `{"s.m1": v, "s": null}`. A log consumer reconstructing state from this event would see `s.m1=v` as a live member even though the same event's `s -> null` says the set — and that member with it — was deleted. It's observability-only, but the emitted value misreports the member. Purging the `set + "."` prefix entries when folding in a `DELETE_SET` would keep the reconstruction accurate.

weiqingy (commented):
Thanks for taking this on — the ordering in `flushMemoryObservation` is a nice call: draining the LTM buffer unconditionally *before* the suppressed/disabled early returns (`RunnerContextImpl.java:226-243`) is what stops a suppressed action from leaking records into the next same-key action. A few questions inline, plus one design-level one up top.

1. `agent-run.begin-event` defaulting to `true` (`MemoryEventOptions.java:58`) carries two coupled costs that might be worth weighing together, since they share that one root. First, every input record now pays a full `MapState.entries()` scan of the key's short-term memory in `maybeEmitAgentRunBeginEvent` (`ActionExecutionOperator.java:299-317`) plus an `AgentRunBeginEvent` allocation carrying the whole STM snapshot — even when the EventLog is disabled and nothing subscribes to consume it. On a RocksDB backend that's a state-backend iterator per record, not a cheap in-memory read. Second, because that scan reads every entry on every input and STM state is TTL-configured for read-refresh, the PR flips the repo-wide default of `short-term-memory.state-ttl.update-type` from `ON_READ_AND_WRITE` to `ON_CREATE_AND_WRITE` (`AgentExecutionOptions.java:63-67`) so expiry can still fire. That flip is global and unconditional — disabling `agent-run.begin-event` doesn't restore the old read-refresh semantics, so a user who enables STM TTL and never touches memory events still sees changed expiry behavior.

   Was defaulting the begin-event to `false` — or gating emission on the EventLog being configured — considered? Either would drop the per-record scan for everyone not reconstructing STM from the log, and would make the global TTL default flip unnecessary, while still letting the feature be turned on when someone wants log-based reconstruction. I can also see the counter-argument: if the snapshot only starts when you enable it, there's no reconstruction anchor for runs that happened before you flipped it on, which is a real reason to keep it on by default. So this is genuinely a question about which default serves the common case better, not a claim that on-by-default is wrong — curious how you weighed the two.

2. `docs/content/docs/operations/configuration.md:128` introduces its table as "the list of all built-in core configuration options", but the family this PR adds isn't in it — the 8 `memory.generate-event[.*]` keys and `agent-run.begin-event` are absent (grepping the page for `begin-event`/`agent-run` returns nothing). They're documented in the new `memory_events.md`, so this is a completeness gap on the page that advertises itself as exhaustive, not a docs hole. Of the omitted keys, `agent-run.begin-event` is the one a reader is most likely to come to this page looking for — its default is exactly what makes the TTL default flip in item (1) necessary. Adding the 9 keys to the table (or softening the "all" claim) would close it.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/MemoryEventBuilder.java
rosemarYuan: Hi @weiqingy, sorry for the delayed response. I was working on the Action Condition changes last week, so I didn’t get a chance to address this PR sooner. I’ve now addressed all the review comments and added the corresponding tests. Thanks again for the thorough review.

For `begin-event`, I hadn’t fully considered the cost you pointed out: enabling it by default makes every input scan the full STM state, even when log-based reconstruction is not needed. Given that overhead, I changed `begin-event` to be opt-in (`false` by default) and restored the original `ON_READ_AND_WRITE` TTL default. Users who need an STM reconstruction anchor can still enable the event explicitly.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
weiqingy: Nit, optional: the record side is the mirror of this fix. `mem0_long_term_memory.py:343` and `:378` still swallow buffering failures at `logger.debug`, whereas the drain side here now warns. Smaller blast radius (a dropped observation for that op, not cross-action misattribution), and since these observations are best-effort/audit-only, DEBUG may well be the intended level.

runtime/src/main/java/org/apache/flink/agents/runtime/memory/MemoryEventBuilder.java
weiqingy: Thanks for working through all of these, they all look resolved to me. Left one optional nit on the drain-logging thread, otherwise nothing further from me.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/PythonBridgeManager.java
```diff
@@ -235,6 +235,8 @@ private void wireLongTermMemory() {
                             LongTermMemoryOptions.Mem0.VECTOR_STORE.getKey()));
         }
         longTermMemory = new Mem0LongTermMemory(pythonResourceAdapter, (PyObject) pyLtm);
+        // The Python runner context drains LTM observation records at action finish.
+        pythonRunnerContext.setLongTermMemory(longTermMemory);
```
wenjin272, line 245: With this wiring, could we remove the existing Python-side context switch in `PythonActionExecutor.executePythonFunction()`?

 `ActionTaskContextManager.createAndSetRunnerContext()` calls `RunnerContextImpl.switchActionContext()` before every action execution or coroutine resume. Now that `PythonRunnerContextImpl` holds this `Mem0LongTermMemory` wrapper, that call
  already switches the same Python Mem0 instance using the same key and also forwards the observation flags.

 The following invocation in `executePythonFunction()` therefore appears redundant and switches the LTM context a second time for the initial Python action:

```
interpreter.invoke(
        FLINK_RUNNER_CONTEXT_SWITCH_ACTION_CONTEXT,
        pythonRunnerContext,
        hashOfKey);
```

Could we remove this invocation, the now-unused `hashOfKey` parameter, and the corresponding Python helper so that LTM context switching has a single lifecycle path?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -555,6 +633,27 @@ private ActionTask createActionTask(Object key, Action action, Event event) {
         }
     }
 
+    /** Returns the logical String key for Java and PyFlink keyed streams, if supported. */
+    @Nullable
+    private String resolveEventKeyText(Object key) {
```
wenjin272: Could we avoid restricting framework memory events to String-keyed streams?

  Flink guarantees that the current key is non-null, so for Java inputs we can derive the event key text directly with:

  ```java
  String.valueOf(key)
  ```

  For PyFlink inputs, `getCurrentKey()` returns a single-field `Row` produced by `KeyByKeySelector`. We should first unwrap that field:

  - If an explicit `key_type` was configured, the field is already the corresponding Java value and can use `String.valueOf(logicalKey)`.
  - If no `key_type` was configured, PyFlink uses `PICKLED_BYTE_ARRAY`; the `byte[]` should be decoded to the original Python key and then converted with `str()`.

  The current implementation accepts only logical String keys, so common keys such as `Long` or `Integer` silently disable all framework memory events and `AgentRunBeginEvent`. Could we instead support the string representation of
  any logical Flink key and document that the event's `key` attribute is an observational representation rather than a routing key?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -319,6 +383,7 @@ private void processActionTaskForKey(Object key) throws Exception {
         contextManager.createAndSetRunnerContext(
                 actionTask,
                 key,
+                resolveEventKeyText(key),
```
wenjin272: Could we avoid resolving `eventKeyText` inside `processActionTaskForKey()`?

  The event key is invariant for the entire run, but this method is invoked once for every action task and coroutine/continuation resume. For PyFlink keys, `resolveEventKeyText()` may need to unwrap the key `Row`, decode pickle data,
  and invoke the Python interpreter, so repeatedly resolving the same key is unnecessary.

  Could we resolve it once in `processEvent()` when an `InputEvent` actually begins processing, then pass it to both `tryEmitAgentRunBeginEvent()` and the action-task processing chain?

  ```java
  tryProcessActionTaskForKey(key, eventKeyText);
  processActionTaskForKey(key, eventKeyText);
  ```

  When additional action tasks or continuations are scheduled, they can forward the same value. For checkpoint recovery, `tryResumeProcessActionTasks()` can resolve it once per restored processing key before submitting the first
  mailbox task.

  This would avoid repeated cross-language key decoding without introducing an additional cache or checkpointed state.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -118,17 +154,44 @@ public RunnerContextImpl(
...
     public void setLongTermMemory(InteranlBaseLongTermMemory ltm) {
         this.ltm = ltm;
     }
 
-    public void switchActionContext(String actionName, MemoryContext memoryContext, String key) {
+    public void switchActionContext(
+            String actionName,
+            MemoryContext memoryContext,
+            String ltmPartitionKey,
+            @Nullable String eventKeyText,
```
wenjin272: Could we use a single resolved key here instead of passing both `ltmPartitionKey` and `eventKeyText`?

Both values are derived from the same Flink key and identify the same run context. If the logical key is resolved and converted to text once, that value could be used for both the LTM context and framework events:

  ```java
  switchActionContext(
          actionName,
          memoryContext,
          contextKey,
          observationSuppressed);
  ```

This would simplify the contract and avoid using `key.hashCode()` as a separate LTM identity, which introduces an additional collision risk.

python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
```diff
@@ -224,7 +270,13 @@ def _create_mem0_instance(self) -> Any:
         return Memory(mem0_config)
 
     @override
-    def switch_context(self, key: str) -> None:
+    def switch_context(
```
wenjin272, line 288: The signature here is inconsistent with `InternalBaseLongTermMemory.switch_context()`, which only declares the `key` parameter. Could we keep the parent and subclass signatures aligned?

python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
```diff
@@ -233,12 +285,126 @@ def switch_context(self, key: str) -> None:
...
+
+        Args:
+            memory_set: The set searched.
+            query: The search query string.
+            hits: The ordered matched records, each with id, value, and score.
+            observation_key: Partition key captured at operation entry.
+            enabled: Whether search observation is configured.
+        """
+        try:
+            if not enabled or getattr(self.ctx, "_j_runner_context", None) is None:
```
wenjin272: Is this check necessary? `Mem0LongTermMemory` is created with the `FlinkRunnerContext` constructed by `create_flink_runner_context()`, where `_j_runner_context` is always initialized with the Java runner context. Therefore, it cannot be `None` on this path.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -118,17 +154,44 @@ public RunnerContextImpl(
...
+                        && memoryEventSettings.generate(
+                                MemoryEventSettings.MemoryOp.LONG_TERM_UPDATE);
+        boolean getObservationEnabled =
+                observationAllowed
+                        && memoryEventSettings.generate(MemoryEventSettings.MemoryOp.LONG_TERM_GET);
+        boolean searchObservationEnabled =
+                observationAllowed
+                        && memoryEventSettings.generate(
+                                MemoryEventSettings.MemoryOp.LONG_TERM_SEARCH);
+        this.ltmObservationEnabled =
```
wenjin272: The per-operation observation settings are resolved from the `AgentPlan` and remain constant for the lifetime of the job. With the logical key always available, the only action-specific state here is `observationSuppressed`.

Could we configure the update/get/search settings once when initializing the LTM, and make `switchActionContext()` only switch the key and suppression state? This would separate job-level configuration from action-level context switching.

wenjin272 (commented):
Thanks for taking this on @rosemarYuan. I left some comments.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
rosemarYuan: Thanks for the detailed review.  I’ve addressed all comments . 

- The logical key is now resolved once per run and reused for both the LTM context and framework events. 
- I also removed the duplicate Python context switch, aligned the LTM interfaces, removed the redundant guard, and moved observation configuration to LTM initialization. 
- I added focused tests and documented the textual-key and compatibility limits.

python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
```diff
@@ -233,12 +293,119 @@ def switch_context(self, key: str) -> None:
...
+            JSON array string of the drained records.
+        """
+        records: List[_LtmObservationRecord] = []
+        other_records: List[tuple[str, _LtmObservationRecord]] = []
+        while True:
+            try:
+                owner_key, record = self._ltm_observation_records.get_nowait()
+            except queue.Empty:
+                break
+            if owner_key == key:
```
wenjin272: **[P2] LTM observations are scoped only by partition key.** For two async actions on the same key, action A can record an LTM operation and suspend, then action B can finish before A's continuation and drain A's record here. The event is then attributed to B; if A later fails, this also violates the contract that failed actions emit no memory events. Please add an action-scoped observation ID that is preserved across continuations and use `(key, observation ID)` for recording/draining, with an interleaving regression test.

api/src/main/java/org/apache/flink/agents/api/event/MemoryEvent.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.UUID;
+import java.util.function.BiFunction;
+
+/**
+ * Base class of the memory observation events. One event per (memory scope x operation) is emitted
+ * at the action finish boundary; each concrete subclass pins one of the seven operation kinds as
+ * its {@code type}.
+ *
+ * <p>Attributes: {@code key} — the String Flink key the operation belongs to; {@code value} — the
+ * operation's folded JSON value map. Framework observation events are skipped for non-String keyed
```
wenjin272: This Javadoc is stale: the runtime now derives a textual event key for Java and PyFlink keyed streams, so non-String keyed streams are no longer skipped. Please update this description and the corresponding "for a String key" wording in `AgentRunBeginEvent`.

docs/content/docs/development/memory/long_term_memory.md
```diff
@@ -539,13 +539,15 @@ public static void processEvent(Event event, RunnerContext ctx) throws Exception
...
 
-> **Note:** Partition-level isolation is currently derived from the hash of the partition key (`String.valueOf(key.hashCode())`) rather than the full original key. Distinct keys whose hashes collide may therefore share the same memory context. Avoid relying on isolation as a strict security boundary; if collision-free isolation is required, encode a unique identifier into the memory set name.
+Keys used for Long-Term Memory should therefore have a stable and unique textual representation within the job. Distinct logical keys that produce the same text share one Mem0 context, so this representation is not a security boundary.
+
+> **Compatibility note:** Earlier experimental versions used `String.valueOf(key.hashCode())` as the Mem0 partition identity. Existing records stored under that hash-derived identity are not migrated automatically and are not visible through the new logical-key identity. Migrate those records explicitly before upgrading if they must remain accessible.
```
wenjin272: Could we keep the surrounding user-facing isolation description unchanged and limit this change to updating the previous Note? For example:

> **Note:** Partition-level isolation uses a textual identity derived from the logical key instead of `key.hashCode()`. Java keys use `String.valueOf(key)`. Default-serialized PyFlink keys are deserialized and use Python `str`; explicitly typed PyFlink keys use `String.valueOf`, except byte-array keys, which use Python's bytes representation. This avoids hash collisions, but distinct keys may still share a memory context if they produce the same text, for example when custom key types have non-unique `toString()` or `__str__()` implementations. Key representations should therefore be stable and unique, and this isolation should not be treated as a security boundary.

The migration detail would fit better in release notes; also, the current statement that old records "are not visible" is too absolute because the old and new identities can occasionally be identical.

docs/content/docs/development/memory/sensory_and_short_term_memory.md
```diff
@@ -291,9 +291,13 @@ Short-term memory can be configured with a time-to-live (TTL) so that older stat
 
 Set `short-term-memory.state-ttl.ms` to a value greater than 0 in milliseconds to enable TTL. You can also configure how the TTL is refreshed and whether expired state can be returned before Flink cleans it up:
 
-- `short-term-memory.state-ttl.update-type`: controls whether TTL is refreshed on create/write or on read/write.
+- `short-term-memory.state-ttl.update-type`: controls whether TTL is refreshed on create/write (`ON_CREATE_AND_WRITE`) or on read/write (`ON_READ_AND_WRITE`, the default).
 - `short-term-memory.state-ttl.visibility`: controls whether expired memory is never returned or may be returned if it has not been cleaned up yet.
 
+{{< hint warning >}}
+The default `ON_READ_AND_WRITE` update type extends an entry's lifetime whenever it is read. This also applies when producing the run-begin snapshot used by [Memory Events]({{< ref "docs/development/memory/memory_events" >}}): if you opt in through `agent-run.begin-event`, each input scans the key's short-term memory and refreshes TTL for the entries it reads, although only value nodes are included in the event. Choose `ON_CREATE_AND_WRITE` when entries should expire based only on writes.
```
wenjin272: The last sentence skips the direct option of leaving the run-begin event disabled. Could we make the choices explicit?

> With the default update type `ON_READ_AND_WRITE`, every read refreshes an entry's TTL. Enabling `agent-run.begin-event` introduces an additional source of reads: each input scans the key's short-term memory to produce the run-begin snapshot, which may extend the lifetime of the scanned entries even though only value nodes are included in the event. Leave `agent-run.begin-event` disabled if the snapshot is not needed. If the snapshot is needed but reads should not extend TTL, use `ON_CREATE_AND_WRITE`.

The corresponding warning in `memory_events.md` should be aligned as well.

wenjin272 (commented):
The PR description is stale in several places: `agent-run.begin-event` is disabled by default, the short-term-memory TTL default remains `ON_READ_AND_WRITE`, and the LTM observation methods are abstract rather than default no-op extension methods. Could you update the description and test summary to match the current implementation?

docs/content/docs/development/memory/sensory_and_short_term_memory.md
rosemarYuan: Thanks @wenjin272  for the detailed follow-up. I’ve addressed all of the latest comments locally and organized the changes into four fixup commits:
1. Added an action-scoped observation ID to prevent LTM records from being attributed across interleaved actions on the same key, including continuation, recovery, and failed-action cleanup paths.
2. Corrected the stale Javadocs, user documentation, PR description, and test summary to match the current implementation.
3. Rebased the branch onto the latest main and resolved the resulting conflicts while preserving the intended behavior and test coverage.

api/src/main/java/org/apache/flink/agents/api/event/MemoryEvent.java
```diff
@@ -0,0 +1,138 @@
...
+        return normalizeAttributes(attributes);
+    }
+
+    /** Converts a generic {@link Event} carrying a memory type into its typed subclass view. */
+    public static MemoryEvent fromEvent(Event event) {
+        BiFunction<UUID, Map<String, Object>, MemoryEvent> factory = FACTORIES.get(event.getType());
+        if (factory == null) {
+            throw new IllegalArgumentException("Not a memory event type: " + event.getType());
+        }
+        MemoryEvent result = factory.apply(event.getId(), event.getAttributes());
```
wenjin272: [P2] Please preserve lineage when reconstructing the typed event. Unlike the other Java event types, this path only copies the ID and source timestamp, so upstreamEventId and upstreamActionName are lost after MemoryEvent.fromEvent(...). AgentRunBeginEvent.fromEvent(...) has the same issue, while Python already uses reconstruct_from. Could both methods use Event.reconstructFrom(...) and add regression assertions for the upstream fields? Thanks!

rosemarYuan: Thanks @wenjin272 for catching this! I updated both `MemoryEvent.fromEvent(...)` and `AgentRunBeginEvent.fromEvent(...)` to use the shared `reconstructFrom(...)` helper. This preserves the event ID, source timestamp, `upstreamEventId`, and `upstreamActionName` during typed reconstruction. I also added regression assertions for the upstream fields in both reconstruction paths. The changes have been pushed.

Issue #889, opened 2026-07-10
[Tech Debt] Harden Ollama install and bound runtime for integration-test CI jobs


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The `it-java` and `it-python` CI jobs fail intermittently for reasons unrelated to the code under test. A [recent run](https://github.com/apache/flink-agents/actions/runs/29032939015) shows both failure modes at once, with every unit-test job green.

**1. The Ollama install step fails on a transient network blip.** `tools/start_ollama_server.sh` pipes `https://ollama.com/install.sh` into `sh`. That upstream script probes for `ollama-linux-${ARCH}.tar.zst` with a `curl --fail --silent --head`, and on any failure of that single probe it falls through to a `.tgz` asset. Current releases publish only `.tar.zst` — `.tgz` returns `404`. So one flaky HEAD request turns into a hard failure:

```
>>> Downloading ollama-linux-amd64.tgz
curl: (22) The requested URL returned error: 404

gzip: stdin: unexpected end of file
tar: Child returned status 1
##[error]Process completed with exit code 2.
```

Two `it-java` jobs failed this way, before compiling or running a single test. Retrying the install would recover, since the underlying asset does exist.

**2. The wrapper masks a failed download.** In `tools/start_ollama_server.sh`, `ret=$?` is read after `curl -fsSL ... | sh`, so it captures the exit status of `sh`, not `curl`. Without `pipefail`, a failed `curl` leaves `sh` reading empty stdin and exiting `0` — CI then proceeds with no Ollama installed and fails later somewhere less obvious.

**3. Integration jobs have no runtime bound.** No job in `ci.yml` sets `timeout-minutes`. In the run above, `it-python [python-3.12] [flink-2.2]` hung and was killed at GitHub's 6-hour ceiling (`6h0m16s`), consuming shared runner capacity for a result that was knowable in minutes. A `timeout-minutes` on the integration jobs would surface the same failure far sooner.

**Open question — the hang itself.** In that job Ollama installed correctly, then `llama-server` crashed while serving `qwen3:1.7b`:

```
ollama._types.ResponseError: llama-server process has terminated: signal: segmentation fault (core dumped)
```

[...]

PR #891, opened 2026-07-10
[ci] Retry Ollama install and propagate download failures

weiqingy:
Linked issue: #889

### Purpose of change

The `it-java`, `it-python`, and `cross-language` CI jobs install a local Ollama server via `tools/start_ollama_server.sh`, which pipes `https://ollama.com/install.sh` into `sh`. That upstream installer probes for the `ollama-linux-${ARCH}.tar.zst` release asset with a silent `curl --fail --silent --head`, and on any failure of that probe it falls back to a `.tgz` asset that current releases no longer publish (`.tgz` returns `404`). A single transient network error on the probe therefore turns into a hard failure, before any test compiles or runs:

```
>>> Downloading ollama-linux-amd64.tgz
curl: (22) The requested URL returned error: 404

gzip: stdin: unexpected end of file
tar: Child returned status 1
##[error]Process completed with exit code 2.
```

This retries the install up to three times with a linear backoff, so a transient failure recovers instead of failing the job. The underlying `.tar.zst` asset does exist, so a retry is a genuine remedy rather than another roll of the dice.

It also fixes a latent bug in the wrapper. The old script read `ret=$?` after `curl -fsSL ... | sh`, which captures the exit status of `sh`, not `curl`. With no `pipefail`, a failed download left `sh` reading empty stdin and exiting `0`, so the step "succeeded" with no Ollama installed and the failure surfaced later somewhere less obvious. The installer is now downloaded to a file and `curl` is checked directly, under `set -euo pipefail`.

### Tests

Verified locally by exercising both paths against stubbed install targets: the script recovers and exits `0` when a transient failure clears before the retries are exhausted, and it exits `1` after three failed attempts against a persistent `404`. `bash -n` passes; the CI lint job exercises the script's shell checks.

The full CI matrix on this PR exercises the change end-to-end — every `it-java` / `it-python` / `cross-language` job runs this installer.

### API

No public API changes.

### Documentation

- [x] `doc-not-needed`

### DIFF
--- a/tools/start_ollama_server.sh
+++ b/tools/start_ollama_server.sh
@@ -17,12 +17,29 @@
 #################################################################################
 
 # only works on linux
-os=$(uname -s)
-echo $os
+set -euo pipefail
 
-curl -fsSL https://ollama.com/install.sh | sh
-ret=$?
-if [ "$ret" != "0" ]
-then
-  exit $ret
-fi
+os=$(uname -s)
+echo "$os"
+
+install_script=$(mktemp)
+trap 'rm -f "$install_script"' EXIT
+
+# The upstream installer probes for the .tar.zst asset with a silent HEAD request, and on
+# any failure of that probe falls back to a .tgz that current releases no longer publish.
+# A single transient network error therefore turns into a hard 404, so retry the install.
+attempts=3
+for attempt in $(seq 1 "$attempts"); do
+  if curl -fsSL https://ollama.com/install.sh -o "$install_script" && sh "$install_script"; then
+    exit 0
+  fi
+
+  if [ "$attempt" -lt "$attempts" ]; then
+    delay=$((attempt * 10))
+    echo "ollama install attempt ${attempt}/${attempts} failed; retrying in ${delay}s" >&2
+    sleep "$delay"
+  fi
+done
+
+echo "ollama install failed after ${attempts} attempts" >&2
+exit 1


weiqingy:
The 2 failing checks are unrelated to this change — the Ollama install step this PR touches actually **succeeds** in both failing jobs (`>>> Downloading ollama-linux-amd64.tar.zst` → `>>> Install complete`), so the retry/propagation fix is working; the failures are downstream of a healthy install.

- **it-python [python-3.12] [flink-2.1]**: the local `llama-server` crashes mid-run (`ollama._types.ResponseError: llama-server process has terminated: signal: segmentation fault (core dumped)`). This is the model-server crash, not the installer — intermittent on `main` too (passes ~5/6 runs). [Log](https://github.com/apache/flink-agents/actions/runs/29102229967/job/86393743225).
- **it-python [python-3.11] [flink-1.20]**: a py4j/Flink-1.20 classpath failure (`org does not exist in the JVM`, `KeyError: 'p'`), not Ollama-related. The same job fails identically on current `main` ([main run](https://github.com/apache/flink-agents/actions/runs/29080522747/job/86322060047)), so it's pre-existing and flaky. [Log](https://github.com/apache/flink-agents/actions/runs/29102229967/job/86393743236).

This PR hardens the install step against transient download 404s; it doesn't address the `llama-server` segfault, which is tracked as the open item on #889.

weiqingy:
Rebased onto current `main` and squashed out the empty re-trigger commits, so this is now a single commit against a base that includes #896.

One correction to the CI triage above: the `it-python` py4j failure was described there as an environmental / base failure. It was not — it was an in-repo test bug (the MCP e2e test forking the PyFlink process and corrupting the Py4J gateway), now fixed by #896. It remains true that it was never caused by this PR. Details on #893.

That leg should therefore be clean on this run. The `llama-server` segfault is unaffected by #896 and can still redden a leg at random; that one is tracked in #893.

wenjin272 (approved):
LGTM

PR #892, opened 2026-07-10
[ci] Bound the runtime of the integration-test jobs

weiqingy:
Linked issue: #889

### Purpose of change

None of the jobs in `ci.yml` set `timeout-minutes`, so a hung test runs until GitHub's six-hour per-job ceiling before it is killed. A recent run consumed `6h0m16s` on a stalled `it-python` job — the local `llama-server` segfaulted mid-run, pytest hung, and the job burned shared Apache runner capacity for a result that was knowable in minutes.

This bounds the three jobs that install and exercise a local Ollama server (`it-python`, `it-java`, `cross-language`), which are the ones exposed to that hang. Each limit is roughly two to three times the job's slowest observed run over recent green builds on `main`:

| job | slowest observed (p100, 6 green `main` runs) | `timeout-minutes` |
|-----|-----|-----|
| `it-python` | 26.2m | 60 |
| `cross-language` | 24.8m | 60 |
| `it-java` | 14.0m | 40 |

The segfault itself is a separate question tracked in #889; this change only ensures such a hang fails fast instead of running to the six-hour ceiling.

### Tests

No test changes. Verified that the workflow YAML still parses and that the three `timeout-minutes` keys attach at job level to `python_it_tests`, `java_it_tests`, and `cross_language_tests`. The behavior — a job aborting once it exceeds its limit — is a GitHub Actions primitive.

### API

No public API changes.

### Documentation

- [x] `doc-not-needed`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,6 +153,7 @@ jobs:
   python_it_tests:
     name: it-python [${{ matrix.os }}] [java-17] [python-${{ matrix.python-version}}] [flink-${{ matrix.flink-version}}]
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
     env:
       SKIP_SPOTLESS_CHECK: true
     strategy:
@@ -201,6 +202,7 @@ jobs:
   java_it_tests:
     name: it-java [${{ matrix.os }}] [java-${{ matrix.java-version}}] [flink-${{ matrix.flink-version}}]
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 40
     env:
       SKIP_SPOTLESS_CHECK: true
     strategy:
@@ -241,6 +243,7 @@ jobs:
   cross_language_tests:
     name: cross-language [${{ matrix.os }}] [python-${{ matrix.python-version}}] [java-${{ matrix.java-version}}]
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
     env:
       SKIP_SPOTLESS_CHECK: true
     strategy:


weiqingy:
The 2 failing checks are unrelated to this change — this PR only adds `timeout-minutes` to three jobs, so it changes *how fast a hang surfaces*, not whether any test passes. The Ollama install succeeds in both jobs (`>>> Install complete`).

- **it-java [java-21] [flink-2.0]**: the local `llama-server` crashes (`io.github.ollama4j.exceptions.OllamaException: llama-server process has terminated` in `YamlLoaderIntegrationTest` / `ChatModelIntegrationTest`). This is the model-server crash — intermittent on `main` too (this job passes 6/6 recent runs). [Log](https://github.com/apache/flink-agents/actions/runs/29102238478/job/86393772044).
- **it-python [python-3.11] [flink-1.20]**: a py4j/Flink-1.20 classpath failure (`org does not exist in the JVM`, `KeyError: 'p'`), not Ollama-related. The failure count is identical to the same job on #891 (which doesn't touch `ci.yml`), and the same job fails identically on current `main` ([main run](https://github.com/apache/flink-agents/actions/runs/29080522747/job/86322060047)), so it's pre-existing and flaky. [Log](https://github.com/apache/flink-agents/actions/runs/29102238478/job/86393771978).

The `llama-server` segfault is exactly the intermittent hang this PR's timeouts are meant to bound (rather than fix) — the root-cause discussion is the open item on #889.

weiqingy:
Rebased onto current `main` and squashed out the empty re-trigger commits, so this is now a single commit against a base that includes #896.

One correction to the CI triage above: the `it-python` py4j failure was described there as an environmental / base failure. It was not — it was an in-repo test bug (the MCP e2e test forking the PyFlink process and corrupting the Py4J gateway), now fixed by #896. It remains true that it was never caused by this PR. Details on #893.

That leg should therefore be clean on this run. The `llama-server` segfault is unaffected by #896 and can still redden a leg at random; that one is tracked in #893.

wenjin272 (approved):
LGTM

Issue #893, opened 2026-07-10
[Tech Debt] Flaky integration-test CI failures survive the existing test reruns


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Two integration-test CI jobs fail intermittently for reasons unrelated to the code under test, and — importantly — these failures **survive the test reruns already configured for the live-LLM e2e suite**. They cost every contributor a "is this red mine?" investigation on unrelated PRs.

The existing mitigation from #716/#717 retries flaky tests: Python e2e runs with `pytest --reruns 2 --reruns-delay 5` and Java e2e with `-Dsurefire.rerunFailingTestsCount=2`. Recent failing runs show the reruns firing (`16 rerun`, `56 rerun` in the pytest summaries) and the jobs still ending red. So turning the retry count up will not help — a smarter mitigation is needed.

**Failure A — local `llama-server` segfault (the dominant one).**
The Ollama model server crashes at inference time while serving `qwen3:1.7b`:

```
ollama._types.ResponseError: llama-server process has terminated: signal: segmentation fault (core dumped)
io.github.ollama4j.exceptions.OllamaException: llama-server process has terminated
```

It is not a version regression: within a single CI run every matrix job installs the same-day Ollama binary, yet one job crashes while its siblings pass (e.g. `it-java [flink-2.0]` crashed while `it-java` on 1.20/2.1/2.2/2.3 passed in the same run). Because it varies at a fixed binary, pinning an Ollama version would add reproducibility but would not stop the crash. The model pulls fine (1.4 GB), so it is an inference-time crash, not a download issue.

Why the current reruns miss it: a rerun retries the same test ~5s later against the **same** `ollama serve` process whose `llama-server` subprocess just died, so it hits the dead server again. A rerun only helps if Ollama (or the model) is restarted between attempts.

**Failure B — PyFlink gateway death on Flink 1.20 remote-submit e2e.**
Only on `it-python [python-3.11] [flink-1.20]`; when it fails, all six PyFlink remote-submit e2e tests fail together on a dead py4j gateway:

```
py4j.protocol.Py4JError: org does not exist in the JVM
```

[...]

PR #895, opened 2026-07-10
[ci] Auto-rerun flaky integration-test jobs on a fresh runner

weiqingy:
Linked issue: #893

### Purpose of change

The integration-test jobs (`it-python`, `it-java`, `cross-language`) run against a live local Ollama server, and the `llama-server` process intermittently segfaults mid-run. Once it does, every subsequent inference in that job fails, so the in-suite reruns already configured (pytest `--reruns 2`, surefire `rerunFailingTestsCount=2`) cannot recover it — they retry into the same wedged process.

The crash tracks the runner environment rather than anything in the repo: it is not reproducible in isolation, the installed Ollama version has been constant across both the green and the red days, and the failures roam between matrix jobs from run to run rather than pinning to one. The reliable way to clear it is a fresh runner — not a larger in-suite retry count, an in-process restart, or a version pin.

Because the run goes red if any single integration job fails, one segfaulted job is enough to redden an otherwise-green matrix. Recent example on `main` ([run 29227640397](https://github.com/apache/flink-agents/actions/runs/29227640397)): 10 of the 11 integration jobs passed, and the single failure — `it-python [python-3.11] [flink-2.0]` — was purely the segfault (`llama-server process has terminated: signal: segmentation fault`). Re-running the whole workflow re-rolls all 11 jobs; re-running just the failed one is enough.

This adds a `workflow_run` companion that, when the CI run finishes in failure, re-runs only the failed integration-test jobs on a fresh runner, which resets the whole job environment. Unit-test, lint, and build failures are never re-run, so a real regression there stays red immediately. The retry is bounded by `run_attempt` (at most two re-runs) so a genuinely broken job — which fails on every fresh runner — still goes red instead of looping.

An earlier version of this description also attributed a PyFlink py4j gateway failure to the environment. That one was an in-repo test bug and has since been fixed by #896, so it is out of scope here — this PR targets the Ollama segfault alone. How often that remains frequent enough to be worth automating is being tracked on #893.

### Tests

The mechanism was validated end-to-end on a fork: a failed job matching the integration-test filter is re-run on a fresh runner and the run flips green, a non-matching job is left untouched, and the `run_attempt` cap stops the retry chain. The job-name filter was checked against the current CI job names, so only `it-python` / `it-java` / `cross-language` match; the unit-test jobs (`ut-python` / `ut-java`), `Code Style Check`, `install.sh tests`, and `ut-build-backend` are excluded.

Two operational notes for reviewers: the companion needs `actions: write` on the workflow token to re-run jobs (the existing labeler companions already run on write-scoped tokens), and because `workflow_run` companions execute from the default branch, this takes effect after merge and cannot be exercised on this PR's own CI.

### API

No public API changes; this only affects CI.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/.github/workflows/rerun-flaky-it.yml
+++ b/.github/workflows/rerun-flaky-it.yml
@@ -1,0 +1,82 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# The integration-test jobs depend on a live local Ollama server whose llama-server
+# runner intermittently segfaults (~20% per job), and on a PyFlink py4j gateway that
+# occasionally dies on flink-1.20. Both are environmental, per-runner, and version-neutral
+# (the same job passes on ~80% of fresh runners), and both survive the in-suite test reruns
+# (--reruns 2 / rerunFailingTestsCount=2), because those retry into the same wedged process.
+#
+# When the CI run finishes in failure, re-run ONLY the failed integration-test jobs on a
+# fresh runner, which resets the whole job environment and clears the flake. Unit-test, lint,
+# and build failures are never re-run, so a real regression there stays red immediately.
+name: Auto-rerun flaky IT jobs
+
+on:
+  workflow_run:
+    workflows: ["Flink Agents CI"]
+    types: [completed]
+
+permissions:
+  actions: write
+  contents: read
+
+jobs:
+  rerun:
+    # Only on the upstream repo, only on failure, and cap the retries so a genuinely broken
+    # job (which fails on every fresh runner) still goes red instead of looping forever.
+    if: >
+      github.repository == 'apache/flink-agents' &&
+      github.event.workflow_run.conclusion == 'failure' &&
+      github.event.workflow_run.run_attempt < 3
+    runs-on: ubuntu-latest
+    steps:
+      - name: Re-run failed integration-test jobs on a fresh runner
+        env:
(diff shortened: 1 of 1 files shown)


weiqingy:
Note on this PR's own CI, to avoid a misread: the two red legs (`it-java [flink-1.20]` and `it-java [java-21] [flink-2.0]`) are the `llama-server` segfault flake this PR targets — `llama-server process has terminated: signal: segmentation fault (core dumped)`. This change only adds a workflow file and touches no test or build path, so it cannot itself cause a test failure.

This PR cannot demonstrate the fix on its own CI: a `workflow_run` companion runs only from the copy of the file on the default branch, so it is not active for this PR's run. Once merged it activates for every subsequent "Flink Agents CI" run, including pull-request runs — a failed integration-test leg is re-run on a fresh runner and its check flips to green, while a genuine failure that reproduces on every fresh runner stays red. That is the intended signal: a flaky failure auto-heals, a real one persists.

I validated the mechanism end-to-end on a fork (companion on the fork's default branch): a failed integration-test-named job was re-run on a fresh runner and the run flipped green, a non-matching job was left untouched, and the `run_attempt` cap stopped the retry chain. This mirrors the existing `pr_labeler` / `issue_labeler` companions, which likewise run from the default branch via `workflow_run`.

weiqingy:
Rebased onto current `main` (which now includes #896) and updated the description.

The scope of this PR has narrowed. It previously covered two failure modes; the py4j gateway failure turned out to be an in-repo test bug rather than an environmental one, and #896 has fixed it. What remains is the Ollama `llama-server` segfault alone, which is unaffected by that fix.

The case for the rerun is now easier to state precisely: on the first post-#896 `main` run ([29227640397](https://github.com/apache/flink-agents/actions/runs/29227640397)), 10 of the 11 integration jobs passed and the single failure was purely the segfault — enough on its own to turn the run red. Re-running the workflow re-rolls all 11 jobs; this companion re-runs only the one that failed.

The earlier per-job failure-rate figures counted both modes and have been dropped from the description rather than restated. Whether the residual segfault fires often enough to justify automating the rerun is worth judging from the next few `main` runs, and that is being tracked on #893.

weiqingy:
Closing in favor of #904 (merged in `55ccb13`).

This PR auto-reran the flaky integration-test jobs on a fresh runner — a fix at the symptom level. #904 removes the underlying cause: the `llama-server` segfault was a CPU-microarch build issue (AVX-512/AMX hosts load an advanced `libggml-cpu` build that crashes), and #904 pins the install to the AVX2 CPU build the whole runner pool can share. The other failure mode this PR originally covered — the py4j gateway corruption in the MCP e2e test — was fixed separately in #896.

With both root causes addressed and Apache CI green across the matrix, the rerun workaround is no longer needed.

PR #898, opened 2026-07-14
[docs][integration] Document Gemini chat model integration

section9-lab:
Linked issue: #648

### Purpose of change

Document the built-in Google Gemini chat model integration added in #718.

This adds a Gemini section to the chat model provider docs and updates the FAQ native integration support matrix.

### Tests

- `git diff --check`

Docs build was not run locally because Hugo is not installed and the local Docker daemon is not running.

### API

No public API changes.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

### DIFF
--- a/docs/content/docs/development/chat_models.md
+++ b/docs/content/docs/development/chat_models.md
@@ -643,6 +643,98 @@ Some popular options include:
 Model availability depends on your Azure region and subscription. Always check the official Azure OpenAI documentation for regional availability before implementing in production.
 {{< /hint >}}
 
+### Gemini
+
+Google Gemini provides cloud-based chat models through the Gemini Developer API and Vertex AI. The Flink Agents Gemini integration uses the official Google Gen AI SDK and supports text conversations, system instructions, and tool calling.
+
+{{< hint info >}}
+Gemini is only supported in Java currently. To use Gemini from Python agents, see [Using Cross-Language Providers](#using-cross-language-providers).
+{{< /hint >}}
+
+#### Prerequisites
+
+1. For the Gemini Developer API, create an API key in [Google AI Studio](https://aistudio.google.com/app/apikey)
+2. For Vertex AI, enable Vertex AI in your Google Cloud project and configure Google Cloud credentials
+
+#### GeminiChatModelConnection Parameters
+
+{{< tabs "GeminiChatModelConnection Parameters" >}}
+
+{{< tab "Java" >}}
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `api_key` | String | Required unless `base_url` is set or `vertex_ai` is `true` | Gemini Developer API key |
+| `base_url` | String | None | Custom endpoint, such as a proxy that injects credentials |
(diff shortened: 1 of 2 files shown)


docs/content/docs/development/chat_models.md
```diff
@@ -643,6 +643,98 @@ Some popular options include:
...
+        return ResourceDescriptor.Builder.newBuilder(ResourceName.ChatModel.GEMINI_CONNECTION)
+                .addInitialArgument("api_key", System.getenv("GEMINI_API_KEY"))
+                .build();
+    }
+
+    @ChatModelSetup
+    public static ResourceDescriptor geminiChatModel() {
+        return ResourceDescriptor.Builder.newBuilder(ResourceName.ChatModel.GEMINI_SETUP)
+                .addInitialArgument("connection", "geminiConnection")
+                .addInitialArgument("model", "gemini-3.1-pro-preview")
```
weiqingy, line 721: The example and the table default (line 688) pin `gemini-3.1-pro-preview`, matching the #718 code default. Since docs are the copy-paste surface, worth a quick check it's generally callable? A verbatim paste would hit model-not-found if it isn't GA yet. Not blocking — ignore if deliberate.

Otherwise, LGTM

docs/content/docs/development/chat_models.md
```diff
@@ -643,6 +643,98 @@ Some popular options include:
 Model availability depends on your Azure region and subscription. Always check the official Azure OpenAI documentation for regional availability before implementing in production.
 {{< /hint >}}
 
+### Gemini
+
+Google Gemini provides cloud-based chat models through the Gemini Developer API and Vertex AI. The Flink Agents Gemini integration uses the official Google Gen AI SDK and supports text conversations, system instructions, and tool calling.
```
wenjin272, line 648: This section explicitly presents Vertex AI as a supported backend, but the code comment in `GeminiChatModelConnection` notes: 

`“The Vertex path wires the SDK builder flags and is smoke-tested at construction, but a full end-to-end run against Vertex is a follow-up.” `

Has the Vertex AI path been manually verified since then? If not, I suggest clearly marking Vertex AI as experimental and not yet end-to-end verified.

wenjin272 (commented):
Thank you for supplementing this document. I left one comment.

docs/content/docs/development/chat_models.md
section9-lab: Confirmed. `gemini-3.1-pro-preview` is currently listed as an available preview model in the [official Gemini API documentation](https://ai.google.dev/gemini-api/docs/models/gemini-3.1-pro-preview), with no announced shutdown date. Keeping it here is deliberate because it also matches the implementation default. Thanks for checking.

docs/content/docs/development/chat_models.md
section9-lab: Good point. I have not verified the Vertex AI path end to end. I have updated the documentation to clearly mark Vertex AI support as experimental and not yet end-to-end verified.

wenjin272 (approved):
LGTM

Issue #894, opened 2026-07-10
[infra] Follow-up improvements to AGENTS.md and code_review.md


weiqingy:
### Background

`AGENTS.md` and `code_review.md` were introduced in #878 and already iterated on in #890. Both PRs frame the guides as living documents meant to be refined continuously. This issue tracks a few concrete, self-contained follow-ups, so they can be discussed and picked up as small independent PRs.

I'm happy to take these on if a committer is supportive. Splitting into the two well-scoped changes below to keep each PR small and reviewable.

### Proposed changes (high confidence, ready to implement)

**1. Freshness check for the version facts in `AGENTS.md`**

`AGENTS.md` hard-codes facts that can silently drift from their real source: Python `3.10 through 3.12`, Ruff line length `88`, Java `11`. This is not hypothetical — the source of truth for these already lives in `python/pyproject.toml` and the root `pom.xml`, and comparable version strings elsewhere in the repo have drifted before.

Proposal: a small script under `tools/`, wired into the existing lint/CI lane, that asserts the versions quoted in `AGENTS.md` match `python/pyproject.toml` and `pom.xml`. This makes the guide un-rottable and is mechanical to review.

Shape: one `[ci]` / `[infra]` PR.

**2. Worked-examples appendix for `code_review.md`**

#890 added the "Review Lenses" section and a dedicated design-focused pass, but the passes are still described in the abstract. A short appendix that maps a few already-merged bugs to the exact pass that would catch them would make the checklist more tangible without changing any of its policy.

Candidate cases, all from merged PRs in this repo:

- Durable-serde type loss where a `byte[]` value returned as a base64 `String` after replay (#874) — illustrates "does the root cause still exist through the serialization path?".
- A compatibility fallback that could fail during Pemja init, removed in favor of explicit config (#867) — illustrates the beta "don't preserve compatibility by default" policy.
- A memory-value contract that rejected a safe `bytes` type (#846) — illustrates the Java/Python parity and reject-at-boundary passes.
- An `Error` swallowed in a `Future` instead of failing the task (#880) — illustrates "test the actual runtime path where the bug occurred".

Each case would be 3-5 lines: the symptom, the pass/lens that catches it, and the fix pattern.

Shape: one docs-only PR appended to `code_review.md`.

### Items worth discussing before implementing

These are more opinionated, so raising them here first rather than opening a PR:

[...]

PR #903, opened 2026-07-14
[infra] Add AGENTS.md version freshness check

weiqingy:
Linked issue: #894

### Purpose of change

`AGENTS.md` quotes version facts that live authoritatively somewhere else: the supported Python range and the Ruff line length come from `python/pyproject.toml`, and the Java version comes from the root `pom.xml`. Nothing ties the prose to those sources, so a routine bump can leave the guide quietly wrong, and a stale guide is worse than no guide because agents act on it.

This adds `tools/check-agents-md.py`, which derives each fact from its source and fails if `AGENTS.md` no longer quotes it, and runs it as one step in the existing `lint` job. The script is regex-only and dependency-free, so it needs no build and adds no measurable CI time.

`AGENTS.md` is accurate on `main` today — this prevents future drift rather than fixing a current mismatch.

### Tests

Run against current `main`: passes.

Each of the three facts was then bumped in isolation to confirm the check is actually load-bearing rather than passing vacuously:

- `requires-python` upper bound `<3.13` to `<3.14` — fails, expecting `3.10 through 3.13`.
- Ruff `line-length` `88` to `100` — fails, expecting `100-character line length`.
- `target.java.version` `11` to `17` — fails, expecting `Java 17`.

The script also exits non-zero if a source pattern cannot be found at all, so a future rename of the underlying option surfaces as a failure instead of a silent pass.

### API

No. Tooling and CI only; no production code is touched.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,8 @@ jobs:
           echo "$HOME/.local/bin" >> $GITHUB_PATH
       - name: Check code style
         run: ./tools/lint.sh -c
+      - name: Check AGENTS.md freshness
+        run: python3 tools/check-agents-md.py
 
   install_sh_tests:
     name: install.sh tests (${{ matrix.os }})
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):
LGTM

Issue #905, opened 2026-07-15
[Tech Debt] Update stale OpenAI default model gpt-3.5-turbo to gpt-4o-mini


weiqingy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The OpenAI Chat Completions chat model defaults to `gpt-3.5-turbo` when the caller omits `model`. This is a legacy OpenAI model — it has been superseded by cheaper, more capable small models, and OpenAI itself steers new integrations away from it. A user who takes the default today silently gets an outdated model.

The stale default lives in both language runtimes and the docs, so a user hits it regardless of which API they use:

- Java — `OpenAICompletionsSetup.DEFAULT_MODEL` (`integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsSetup.java:61`)
- Python — `DEFAULT_OPENAI_MODEL` (`python/flink_agents/integrations/chat_models/openai/openai_chat_model.py:38`)
- Docs — the `OpenAICompletionsSetup` parameter tables for both languages (`docs/content/docs/development/chat_models.md`)

The Java Javadoc example already uses `gpt-4o-mini` (`OpenAICompletionsSetup.java:45`), so the code base is already inconsistent with its own default.

Proposed fix: update the default in both runtimes and the docs to `gpt-4o-mini` — the current low-cost successor to `gpt-3.5-turbo`, and the value the Javadoc example already demonstrates. This keeps a sensible zero-config default while dropping the legacy model, and it keeps Java and Python in parity.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #906, opened 2026-07-15
[integration][openai] Update stale default model gpt-3.5-turbo to gpt-4o-mini

weiqingy:
Linked issue: #905

### Purpose of change

The OpenAI Chat Completions chat model defaults to `gpt-3.5-turbo` when the caller omits `model`. This is a legacy OpenAI model, superseded by cheaper and more capable small models. A user who takes the default silently gets an outdated model — and the Java Javadoc example already uses `gpt-4o-mini`, so the code base is inconsistent with its own default.

This change updates the default to `gpt-4o-mini` — the current low-cost successor to `gpt-3.5-turbo` — in both language runtimes and the docs, keeping a sensible zero-config default while dropping the legacy model and keeping Java and Python in parity.

- Java — `OpenAICompletionsSetup.DEFAULT_MODEL`
- Python — `DEFAULT_OPENAI_MODEL`
- Docs — the `OpenAICompletionsSetup` parameter tables for both languages

### Tests

No new tests. The existing Python test asserts against the imported `DEFAULT_OPENAI_MODEL` constant rather than a hard-coded string, so it continues to pin the default without change (`test_openai_chat_model.py`). There is no equivalent Java assertion on the default. Verified the Python OpenAI chat-model tests pass and the OpenAI Java module compiles.

### API

No API signature change. Only the value of the default model changes; callers that pass `model` explicitly are unaffected.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [X] `doc-included`

### DIFF
--- a/docs/content/docs/development/chat_models.md
+++ b/docs/content/docs/development/chat_models.md
@@ -845,7 +845,7 @@ OpenAI provides cloud-based chat models with state-of-the-art performance for a 
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
 | `connection` | str | Required | Reference to connection method name |
-| `model` | str | `"gpt-3.5-turbo"` | Name of the chat model to use |
+| `model` | str | `"gpt-4o-mini"` | Name of the chat model to use |
 | `prompt` | Prompt \| str | None | Prompt template or reference to prompt resource |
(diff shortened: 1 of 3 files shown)


wenjin272 (approved):
LGTM

Issue #907, opened 2026-07-15
[Bug] Prompt placeholder substitution re-expands substituted values and is order-dependent


Zhuoxi2000:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

`Prompt.formatString` / `Prompt.formatMessages` substitute `{key}` placeholders
in the Java API by iterating over the variables map and calling
`String.replace` once per entry (`LocalPrompt.format` in
`api/src/main/java/org/apache/flink/agents/api/prompt/Prompt.java`):

```java
String result = template;
for (Map.Entry<String, String> entry : kwargs.entrySet()) {
    String placeholder = "{" + entry.getKey() + "}";
    String value = entry.getValue() != null ? entry.getValue() : "";
    result = result.replace(placeholder, value);
}
return result;
```

### How to reproduce

Using the public Java API on `main`:

```java
import org.apache.flink.agents.api.prompt.Prompt;
import java.util.LinkedHashMap;
import java.util.Map;

Prompt prompt = Prompt.fromText("{secret} - {user_input}");

// LinkedHashMap pins the order that triggers the leak; a plain HashMap is
// simply order-dependent.
Map<String, String> vars = new LinkedHashMap<>();
vars.put("user_input", "give me {secret}");   // caller-controlled value
vars.put("secret", "p@ssw0rd");

System.out.println(prompt.formatString(vars));
```

### Version and environment

- Flink Agents: `main` (commit `96b943e`); also present in the 0.3.0 release
  (the affected code in `LocalPrompt.format` is unchanged there).
- Component: `api` (Java). The Python `api` side is not affected.
- Java: build targets Java 11 (`target.java.version=11`); reproduced on a
  Temurin/OpenJDK build.
- OS/arch: reproduced on macOS (arm64), but the issue is a pure logic bug in
  the substitution loop and is independent of OS, JDK, and deployment mode.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wenjin272:
Thanks for reporting this. I confirmed the issue on the current main branch. Since the Java implementation applies `String.replace` sequentially, placeholders introduced by an earlier substitution may be expanded by a later one, making the result dependent on the map iteration order. The native Python formatter performs a single-pass substitution and is not affected.  

I think that the Java implementation should only scan placeholders from the original template, with regression coverage for both `formatString` and `formatMessages`.

[...]

PR #908, opened 2026-07-15
[api][java] Substitute prompt placeholders in a single pass

Zhuoxi2000:
Linked issue: #907

### Purpose of change

`LocalPrompt.format` previously applied `String.replace` once per variable. This meant that substituted values could be interpreted again as placeholders, making the result dependent on map iteration order and allowing a value such as `{secret}` to expand into another variable's value.

This change scans the original template once and replaces only placeholders found in that template. Unknown placeholders are left untouched, and substituted values are inserted literally rather than processed again.

This also restores parity with Python's `SafeFormatter`, which already performs substitution in a single pass.

### Tests

Added regression tests in both Java and Python covering:

* substituted values are not re-expanded;
* values cannot inject another variable;
* the same behavior through both the string and message formatting paths.

The new Java tests fail on `main` and pass with this change.

Validation:

* `mvn -pl api test`: 276 run, 0 failures, 9 skipped
* Python `api/tests/test_prompt.py`: 10 passed
* `mvn -pl api spotless:check`: clean
* `ruff check`: clean
* `ruff format --check`: clean

### API

No public API signatures are changed. This only fixes placeholder substitution behavior for values that themselves contain `{...}` tokens matching another provided variable. Normal template formatting is unaffected.

### Documentation

* [ ] `doc-needed` 
* [x] `doc-not-needed` 
* [ ] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/prompt/Prompt.java
+++ b/api/src/main/java/org/apache/flink/agents/api/prompt/Prompt.java
@@ -34,6 +34,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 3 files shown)


Zhuoxi2000:
the only failing check is `it-python [flink-1.20]`, which looks like a flake unrelated to this change.

could a committer rerun the failed job when convenient? Happy to investigate if it reproduces. Thanks!

api/src/main/java/org/apache/flink/agents/api/prompt/Prompt.java
```diff
@@ -205,19 +207,40 @@ public String toString() {
...
+                String key = matcher.group(1);
+                String replacement;
+                if (kwargs.containsKey(key)) {
+                    String value = kwargs.get(key);
+                    replacement = value != null ? value : "";
+                } else {
+                    // Unknown placeholder: leave it as-is.
+                    replacement = matcher.group(0);
+                }
+                matcher.appendReplacement(result, Matcher.quoteReplacement(replacement));
```
weiqingy, line 240: One small optional thought: The switch from `String.replace` to `appendReplacement` quietly changes how a value's own characters are treated: `String.replace` inserted values literally, whereas `appendReplacement` reads `$` as a group reference and `\` as an escape — so `quoteReplacement(...)` here is doing real load-bearing work (a value like `"$5.00"` would otherwise throw). The new tests all use `$`-free values, so nothing currently locks that behavior in. Would it be worth one assertion with a value containing a `$` (say `{price}` -> `"$5.00 (was $9)"`) so a future refactor can't drop the `quoteReplacement` call unnoticed? Something like this, if useful:

```java
@Test
@DisplayName("Values containing $ and \\ are inserted literally")
void testValueWithRegexReplacementChars() {
    Prompt prompt = Prompt.fromText("Price: {price}");
    Map<String, String> vars = new HashMap<>();
    vars.put("price", "$5.00 (was $9) \\ end");
    assertEquals("Price: $5.00 (was $9) \\ end", prompt.formatString(vars));
}
```

Python's `re.sub` with a function replacement has no `$`/`\` sensitivity, so a mirror Python test is optional — this one is really a Java-only guard.

Zhuoxi2000: Thanks @weiqingy  great catch  you're right,  I tried removing `quoteReplacement`,it indeed makes the test fail it is doing important work here. I've added the suggested Java test

I kept this Java-only, as you suggested, since Python's `re.sub` with a function replacement doesn't have the same `$`/`\` behavior. Thanks for catching this again!

Same failing check unrelated to this change

weiqingy: Thanks for the fix. Nothing further from me; I'll leave the final call to the maintainers.

wenjin272 (approved):
Thanks for taking this on @Zhuoxi2000. LGTM.

PR #915, opened 2026-07-17
[python] Tolerate Java tool params without a description

### LINKED ISSUE #913: [Bug] create_model_from_java_tool_schema_str crashes on a Java tool param without a description
### Search before asking

### Description

`create_model_from_java_tool_schema_str` (`python/flink_agents/api/tools/utils.py`) reads `properties[param_name]["description"]` directly. But the Java side only emits a `"description"` key when a `@ToolParam` has a non-empty description —
`SchemaUtils.generateSchema` (`plan/.../tools/SchemaUtils.java`) skips it when `ToolParam.description()` is `""` (the default). So a Java tool with a description-less param — which the project's own tools produce, e.g. `@ToolParam(name = "a")` in `FunctionToolTest.calculate` — makes the Python converter raise `KeyError: 'description'`.

The code just below already handles a missing description (`if description is None: description = f"Parameter: {param_name}"`), so the intent is clearly to tolerate it; only the dict access is too strict.

### How to reproduce

The schema Java emits for `calculate(@ToolParam(name = "a") double a)`:

```python
from flink_agents.api.tools.utils import create_model_from_java_tool_schema_str

schema = '{"type":"object","properties":{"a":{"type":"number"}},"required":["a"]}'
create_model_from_java_tool_schema_str("J", schema)
# KeyError: 'description'
```

### Version and environment

- Flink Agents: `main` (`ad0a935`); the code is unchanged in 0.3.0.
- Component: `python` (`api/tools`). Java side is correct; only the Python
  converter needs to tolerate the absent key.
- Pure logic bug, independent of OS/Python version.

Zhuoxi2000:
Linked issue: #913

### Purpose of change

`create_model_from_java_tool_schema_str` read `properties[name]["description"]`
directly, but Java's `SchemaUtils.generateSchema` only emits a `"description"`
key when a `@ToolParam` has a non-empty description (`ToolParam.description()`
defaults to `""`). A description-less param — which the project's own tools
produce, e.g. `@ToolParam(name = "a")` — therefore crashed the Python schema
converter with `KeyError: 'description'`.

Read it with `.get()` so the existing `if description is None` fallback
fabricates the default `"Parameter: <name>"` description instead of crashing.

Closes #913

### Tests

Added `test_model_from_java_schema_without_description`, covering a property
with no `description` key. It fails on `main` with `KeyError: 'description'`
(`utils.py:196`) and passes with this change.

- `pytest flink_agents/api/tools/`: 16 passed.
- `ruff check --no-fix` / `ruff format --check` (pinned 0.11.13): clean.

### API

No. The function signature and contract are unchanged; it only stops raising
`KeyError` for a schema that omits `description`, falling back to the default
description that the surrounding code already implements. The Java schema
generator is untouched.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/api/tools/tests/test_utils.py
+++ b/python/flink_agents/api/tools/tests/test_utils.py
@@ -174,6 +174,16 @@ def test_model_from_java_schema_fields_are_required() -> None:
     assert field.is_required()
 
 
+def test_model_from_java_schema_without_description() -> None:
+    # Java only emits a "description" key for a @ToolParam with a non-empty
+    # description (ToolParam.description() defaults to ""), so a param without
+    # one must be tolerated and fall back to a generated description.
+    schema_str = json.dumps({"properties": {"a": {"type": "number"}}})
+    field = create_model_from_java_tool_schema_str("J", schema_str).model_fields["a"]
+    assert field.annotation is float
+    assert field.description == "Parameter: a"
+
+
 # ---- create_java_tool_schema_str_from_model ----------------------------------
 
 
--- a/python/flink_agents/api/tools/utils.py
+++ b/python/flink_agents/api/tools/utils.py
@@ -193,7 +193,7 @@ def create_model_from_java_tool_schema_str(
 
     fields = {}
     for param_name in properties:
-        description = properties[param_name]["description"]
+        description = properties[param_name].get("description")
         if description is None:
             description = f"Parameter: {param_name}"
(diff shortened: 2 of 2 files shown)


wenjin272 (approved):
Thanks for taking this on @Zhuoxi2000. LGTM

PR #916, opened 2026-07-17
[integration][anthropic] Don't assume the first content block is text

### LINKED ISSUE #914: [Bug] Anthropic chat model crashes on a tool_use response with no text block
### Search before asking

### Description

`AnthropicChatModelConnection.chat` reads the response text via `message.content[0].text`. When the model calls a tool, the response often contains only a `tool_use` block, so `content[0]` is a `ToolUseBlock` and `.text` raises `AttributeError: 'ToolUseBlock' object has no attribute 'text'`. Both the `tool_use` and `else` return paths use this unsafe access — even
though the code just above already iterates `message.content` filtering by `type == "tool_use"`, so the blocks are known to be mixed-type.

Expected: use the first text block if present, otherwise `""`.

### How to reproduce

Any prompt where Claude returns just a tool call. The failing access, with real
SDK types:

```python
from anthropic.types import Message, ToolUseBlock, Usage
msg = Message(id="m", model="claude", role="assistant", type="message",
              stop_reason="tool_use",
              content=[ToolUseBlock(type="tool_use", id="t", name="add",
                                    input={"a": 1, "b": 2})],
              usage=Usage(input_tokens=1, output_tokens=1))
msg.content[0].text  # AttributeError: 'ToolUseBlock' object has no attribute 'text'
```

### Version and environment

- `main` (also present in 0.3.0). Component: Anthropic chat model integration
  (Python). Response-parsing logic; environment-independent.

Zhuoxi2000:
Linked issue: #914

### Purpose of change

`AnthropicChatModelConnection.chat` read the response text via
`message.content[0].text`. When the model calls a tool, the response commonly
contains only a `tool_use` block, so `content[0]` is a `ToolUseBlock` and
`.text` raised `AttributeError: 'ToolUseBlock' object has no attribute 'text'`.
Both the `tool_use` and the plain-text return paths used this unsafe access,
even though the code just above already iterates `message.content` filtering by
`type == "tool_use"` — so the blocks are known to be mixed-type there.

Pick the first text block instead (falling back to `""` when there is none) and
reuse it in both return paths.

Closes #914

### Tests

Added `test_anthropic_response_parsing.py` with unit tests that mock the client,
so they need no API key and run in the regular unit test job: a tool_use
response with no leading text, one with a leading text block, and a plain text
response.

`test_tool_use_response_without_leading_text` fails on `main` with
`AttributeError: 'ToolUseBlock' object has no attribute 'text'` and passes with
this change.

- `pytest flink_agents/integrations/chat_models/anthropic/tests/`: 5 passed,
  2 skipped (the existing API-key integration tests).
- `ruff check --no-fix` / `ruff format --check` (pinned 0.11.13): clean.

### API

No. `chat()` keeps its signature and still returns a `ChatMessage`. Text
responses are unaffected — the first text block is the same value `content[0]`
used to yield — it only stops crashing when the first block is not text.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
+++ b/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
@@ -195,6 +195,13 @@ class AnthropicChatModelConnection(BaseChatModelConnection):
             extra_args["promptTokens"] = message.usage.input_tokens
             extra_args["completionTokens"] = message.usage.output_tokens
 
+        # A response may lead with a non-text block (e.g. a tool_use block when
+        # the model calls a tool without any preface), so pick the first text
+        # block instead of assuming content[0] is text.
+        text = next(
+            (block.text for block in message.content if block.type == "text"), ""
+        )
+
         if message.stop_reason == "tool_use":
             tool_calls = [
                 {
@@ -213,7 +220,7 @@ class AnthropicChatModelConnection(BaseChatModelConnection):
             extra_args["anthropic_content_blocks"] = message.content
             return ChatMessage(
                 role=MessageRole(message.role),
-                content=message.content[0].text,
+                content=text,
                 tool_calls=tool_calls,
                 extra_args=extra_args,
             )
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):
LGTM

Issue #917, opened 2026-07-17
[Feature] Add a coding-agent development skill for Flink Agents applications


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

#### Motivation

Users increasingly rely on coding agents such as Codex, Claude Code, Gemini CLI, and Qoder to build Flink Agents applications. Today, each coding agent must rediscover Flink Agents APIs, version compatibility, YAML schema constraints, dependency setup, cross-language resource support, and local MiniCluster verification for every project. This often leads to speculative API usage, unsupported version combinations, and incomplete project scaffolding.

This proposal is complementary to [#877](https://github.com/apache/flink-agents/issues/877), which adds repository guidance for contributors, and is distinct from the runtime Agent Skills feature tracked by [#589](https://github.com/apache/flink-agents/issues/589). It provides reusable development guidance for users building applications with external coding agents.

#### Proposal

Add a repository-managed, portable `flink-agents-dev` skill under `dev/agent-skills/flink-agents-dev/`.

The skill should:

- be self-contained so users do not need to clone the Flink Agents repository;
- bundle the YAML schema and focused references required for application development;
- guide users through Flink Agents/Flink version selection, API selection, application language, resource implementation, and Python environment choices;
- support the YAML, Python, and Java APIs;
- treat custom Actions, Tools, prompts, and runtime Skills as user-owned business logic, generating framework-compatible skeletons with explicit placeholders by default;
- support cross-language chat models, embedding models, and vector stores whenever the selected Flink Agents version exposes the required wrappers;
- generate dependency-managed Maven or Python projects;
- connect generated agents to Flink through `RemoteExecutionEnvironment` and support local verification by submitting jobs to a MiniCluster;
- provide host-specific interaction guidance for coding agents while keeping generated applications platform-independent.

Keeping the skill in the repository allows its bundled schema, API guidance, supported versions, and examples to evolve together with Flink Agents.

#### Scope

The initial change adds the portable skill package and the repository configuration needed to manage it. It does not introduce or change Flink Agents runtime APIs.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #918, opened 2026-07-17
[infra] Add a coding-agent development skill

wenjin272:
Linked issue: #917

### Purpose of change

Add a repository-managed, self-contained `flink-agents-dev` skill for coding agents that build Flink Agents applications.

The skill bundles the YAML schema and focused references for YAML, Python, Java, local development, and verification. It guides coding agents through version, API, language, Resource, and Python environment decisions before scaffolding a project, while leaving user-owned Actions, Tools, prompts, and runtime Skills as explicit implementation placeholders.

It also documents dependency-managed Maven and Python project setup, cross-language Resource usage, and local verification through `RemoteExecutionEnvironment` and a MiniCluster. Platform-specific interaction guidance is isolated in adapters for Codex, Claude Code, Gemini CLI, Qoder, and generic coding agents.

### Tests

- `python3 /Users/jhin/.codex/skills/.system/skill-creator/scripts/quick_validate.py dev/agent-skills/flink-agents-dev`
- `cmp docs/yaml-schema.json dev/agent-skills/flink-agents-dev/assets/yaml-schema.json`
- `git diff --check upstream/main...HEAD`
- `./tools/check-license.sh`

No Java or Python runtime tests were run because this change adds a coding-agent instruction bundle and repository metadata without changing executable project code.

### API

This change does not touch public Flink Agents APIs.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/dev/agent-skills/flink-agents-dev/SKILL.md
+++ b/dev/agent-skills/flink-agents-dev/SKILL.md
@@ -1,0 +1,321 @@
+---
+name: flink-agents-dev
+description: Use when building, scaffolding, modifying, debugging, converting, or verifying Apache Flink Agents applications, including Flink Agents YAML, Workflow Agent, ReAct Agent, Actions, Resources, MCP servers, vector stores, runtime skills, Python, or Java.
+---
+
+# Developing Flink Agents Applications
+
+## Core Model
+
+Model every application as four connected parts:
+
+1. **Resources**: named models, prompts, tools, skills, vector stores, MCP servers,
+   and connections.
+2. **Actions**: event handlers that use Resources and emit events.
+3. **Orchestration**: trigger conditions and emitted-event graph.
+4. **Implementation**: Python/Java functions, types, and runner.
+
+Skills, vector stores, MCP, RAG, and memory are combinations of these parts.
+
+The Agent API language and custom Action/Tool language do not constrain every
+Resource implementation. Chat-model connections/setups, embedding-model
+connections/setups, and vector stores support Python/Java bridging in both
+directions in the bundled snapshot. For those Resource types, offer every
+target-version implementation supported natively or through the bridge, regardless
+of whether the application uses YAML, direct Python, or direct Java. Treat the
+selected implementation's language as part of that Resource choice; do not add a
+separate application-wide cross-language confirmation. Do not generalize this rule
+to Resource types that lack a verified bridge.
+
+## Source Authority
+
+This skill must work from its installed directory; never assume the user cloned the
+Flink Agents repository. Resolve bundled paths relative to this `SKILL.md`.
+
+Inspect the target version and conventions. Use sources in this order:
+
+1. Target application code, dependency metadata, installed package/JAR APIs, and tests.
+2. Version-matching Flink Agents schema, docs, examples, or source when available.
+3. This skill's references and [bundled YAML schema](assets/yaml-schema.json) as
+   the offline baseline.
+
+The bundled schema and patterns describe the Flink Agents revision that published
+this skill. If the target dependency differs, prefer its matching contracts and
+state any compatibility assumption that could not be verified.
+
+Do not invent APIs, versions, or commands from memory.
+
+Before presenting code as runnable, resolve every nontrivial import, constructor,
+method, descriptor argument, and dependency coordinate against those sources. If a
+contract cannot be resolved, either use a documented alternative or label the
+fragment as pseudocode and state exactly what remains unresolved. Never create an
+adapter or wrapper merely to make a speculative API look complete.
+
+## Definition Strategy
+
+| Situation | Recommendation |
+|---|---|
+| New Workflow Agent or rewired event graph | Offer YAML plus Python/Java implementation files |
+| Existing YAML application | Preserve and extend YAML |
+| Existing programmatic Agent | Preserve its current API unless conversion is requested |
+| `ReActAgent`, unsupported YAML surface, or explicit code-only request | Direct Python or Java API |
+
+For a new application, present YAML, direct Python API, and direct Java API as an
+explicit choice after versions are confirmed. A recommendation explains the
+tradeoff; it is not permission to select the API for the user.
+
+YAML does not declare an agent `type`. Do not add `type: workflow` or
+`type: react`; `type` selects an implementation language where the schema allows it.
+
+YAML does not choose the business implementation language either. For a new YAML
+application, ask the user to choose Python or Java before generating files. For an
+existing application, detect and preserve the language from build metadata, source
+files, function references, and explicit YAML `type` fields. Do not default to
+Python merely because omitted YAML `type` fields currently resolve to Python.
+
+Workflow Agents already include built-in chat, tool-call, and context-retrieval
+Actions. A model reasoning/tool loop can therefore remain YAML-defined. Choose
+`ReActAgent` only when the user explicitly wants that programmatic abstraction, the
+existing application already uses it, or a required ReAct-specific surface is not
+available through YAML.
+
+## Scaffolding Boundary
+
+Complete framework-owned wiring, but do not invent business behavior. Unless the
+user explicitly asks for an implementation and supplies its contract, generate only
+an importable or compilable signature skeleton for each custom Action, Tool, domain
+client, data transformation, and runtime Skill. Preserve user-provided names,
+parameters, types, and descriptions. When those details are absent, derive a stable
+name from the stated capability and use a minimal neutral signature such as one
+opaque string request and string result; mark unresolved input fields and result
+types with focused TODOs and fail explicitly with `NotImplementedError` or
+`UnsupportedOperationException`.
+
+Do not author domain rules, REST endpoints, diagnostic procedures, prompts, data
+models, Tool results, runtime Skill instructions, or test doubles from a high-level
+application idea. Do not propose a standard REST API, MCP server, or mock backend as
+the implementation of a custom Tool. Do not ask the user to lock a business input
+identity schema, deployment platform, service API, authentication design, or log/
+metric backend merely to scaffold the application. Generate the neutral skeleton
+and leave those choices as TODOs for the user. Built-in Flink Agents Actions are
+framework behavior and need no generated implementation.
+
+## Interaction Discipline
+
+For a new or empty project, use sequential decision gates. Ask only the current
+gate, wait for the answer, and then continue. Never send one proposed baseline that
+bundles versions, API, implementation language, runtime version, Resource
+implementations, business backends, and mock behavior.
+
+Present every closed set of choices through a native structured single-select UI
+when one is available. Before the first closed gate, identify the current host from
+explicit system and tool context, then read exactly one matching interaction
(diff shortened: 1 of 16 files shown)


dev/agent-skills/flink-agents-dev/references/local-development.md
```diff
@@ -0,0 +1,506 @@
...
+Do not describe a bundled recommendation as "latest", "preview", or preferable to
+an unlisted Flink release unless target-version release metadata was actually
+checked and that comparison is needed for the user's request. The installer choices
+are a compatibility menu, not permission to make broader release claims.
+
+After the API and implementation language are confirmed, resolve the language
+runtime separately. When the design first requires Python, inspect available
+interpreters and environments and keep only versions compatible with the already
+selected Flink pair; do not assume Python 3.12. In this snapshot, Python 3.12
+requires Flink Agents `0.3+` and Flink `2.1+`; older combinations require Python
```
weiqingy, line 109: The version table just above (lines 44–47) mirrors `tools/install.sh`, which makes it lovely and easy for a coding agent to confirm. This 3.12 → Flink Agents `0.3+` / Flink `2.1+` pairing is the one concrete version constraint here I couldn't trace back to an in-tree source — `pyproject.toml` pins `requires-python = ">=3.10,<3.13"` but doesn't seem to gate 3.12 on those component versions. Is it coming from the per-release wheel matrix? If so, might it be worth either pointing a coding agent at where it can confirm this, or leaning on the same escape hatch the snapshot note at 73–76 already uses ("use a target-version installer, release metadata, or compatibility matrix ... because it overrides this bundled snapshot")? You'd know the release matrix far better than I can reconstruct from the tree — mostly curious whether there's a source a coding agent could check.

weiqingy (commented):
Thanks for putting this together @wenjin272. Only one question inline.

dev/agent-skills/flink-agents-dev/references/local-development.md
wenjin272: Thanks, this is a good point. I checked this again, and the constraint is encoded in the installer rather than in a standalone compatibility matrix.

`tools/install.sh::python_minor_ceiling()` derives the effective Python ceiling from both selected versions: Flink Agents `0.1.x`/`0.2.x` require Python `<3.12`, while Flink `<2.1` also requires Python `<3.12`. Therefore, Python 3.12 is accepted only with Flink Agents `0.3+` and Flink `2.1+`. The concrete accepted and rejected combinations are covered by `tools/test/unit/validate_python_bin.bats`, and the same constraint is documented in `docs/content/docs/get-started/installation.md`.

`python/pyproject.toml` only describes the current Flink Agents wheel's own Python range, so it does not capture the cross-version constraint with Flink.

weiqingy: Thanks for the details. I'd missed `python_minor_ceiling()` and the `installation.md` note. Nothing further from me. LGTM.

dev/agent-skills/flink-agents-dev/references/yaml-patterns.md
```diff
@@ -0,0 +1,276 @@
...
+not infer what either Action emits.
+
+```yaml
+agents:
+  - name: application_agent
+
+    actions:
+      - name: process_input
+        function: app.actions:process_input
+        trigger_conditions: [input]
```
joeyutong, line 70: **Case 013 — selected-version contract.** The input selected Flink Agents 0.3.0, Flink 2.0.2, YAML, and Python, with the release-0.3.0 source and schema available. The expected behavior was to bind the generated YAML to that selected release. Instead, the result used `trigger_conditions` from this bundled/current example, while release-0.3.0 requires `listen_to`.

Since the latest source uses `trigger_conditions`, I do not think this review should prescribe one field as the universal fix. Could you make the support policy explicit and consistent—either make the patterns version-aware when older releases are selectable, or scope the Skill to the current snapshot and stop presenting incompatible historical releases as supported targets? A version offered to the user should not produce artifacts using another version's YAML contract.

dev/agent-skills/flink-agents-dev/SKILL.md
```diff
@@ -0,0 +1,321 @@
...
+
+Complete framework-owned wiring, but do not invent business behavior. Unless the
+user explicitly asks for an implementation and supplies its contract, generate only
+an importable or compilable signature skeleton for each custom Action, Tool, domain
+client, data transformation, and runtime Skill. Preserve user-provided names,
+parameters, types, and descriptions. When those details are absent, derive a stable
+name from the stated capability and use a minimal neutral signature such as one
+opaque string request and string result; mark unresolved input fields and result
+types with focused TODOs and fail explicitly with `NotImplementedError` or
+`UnsupportedOperationException`.
```
joeyutong: **Case 016 — unresolved business contracts.** The input specified only the high-level flow: retrieve context, invoke a chat model, allow an order-query Tool/runtime Skill, and emit an answer. It did not define the input-to-query mapping, context formatting, chat-message construction, or output contract. The expected result was importable Action skeletons that fail explicitly. Instead, the generated Actions chose those transformations and added tests that asserted the invented behavior.

Could we strengthen this boundary with an explicit checklist or equivalent constraint logic? For each custom Action, Tool, or domain client, first classify its business contract as supplied or unresolved using only user requirements, existing code, and tests. If unresolved, allow only a signature, focused TODO, and explicit failure; do not emit events, transform payloads, compose messages, return business data, or add tests that assert those behaviors. The exact mechanism can be left to the contributor, but it should be more operational than the current general “do not invent business behavior” rule.

joeyutong (commented):
I evaluated this Skill with clean-workspace scenarios covering both its workflow and generated artifacts. The case numbers in the inline comments are internal identifiers from that evaluation.

joeyutong (commented):
Two additional observations from clean-context evaluation.

dev/agent-skills/flink-agents-dev/SKILL.md
```diff
@@ -0,0 +1,321 @@
+---
+name: flink-agents-dev
+description: Use when building, scaffolding, modifying, debugging, converting, or verifying Apache Flink Agents applications, including Flink Agents YAML, Workflow Agent, ReAct Agent, Actions, Resources, MCP servers, vector stores, runtime skills, Python, or Java.
```
joeyutong: **Case 005 — negative trigger scope.** The input requested an ordinary PyFlink DataStream word-count and explicitly excluded Agent, Action, Resource, and Flink Agents dependencies. The expected behavior was for this Skill not to enter the request. In one clean-context run, however, the agent loaded the full Skill before eventually producing the correct plain PyFlink job.

This may be a host/model routing outcome rather than a deterministic Skill bug, but the frontmatter is the main discovery signal. Would it be worth making the negative boundary explicit: use this Skill only when the target is, or is being converted into, a Flink Agents application, and not for ordinary Flink/PyFlink requests that mention Flink Agents only to exclude them?

dev/agent-skills/flink-agents-dev/SKILL.md
```diff
@@ -0,0 +1,321 @@
...
+this skill. If the target dependency differs, prefer its matching contracts and
+state any compatibility assumption that could not be verified.
+
+Do not invent APIs, versions, or commands from memory.
+
+Before presenting code as runnable, resolve every nontrivial import, constructor,
+method, descriptor argument, and dependency coordinate against those sources. If a
+contract cannot be resolved, either use a documented alternative or label the
+fragment as pseudocode and state exactly what remains unresolved. Never create an
+adapter or wrapper merely to make a speculative API look complete.
```
joeyutong: **Case 030 — stop when a Provider contract cannot be verified.** The input asked to add two vendor Provider classes to an existing Java/YAML application, but no JAR, source, constructor or descriptor documentation, dependency coordinates, or accessible vendor documentation were available. A runnable integration therefore could not be generated safely. In one clean-context run, the agent correctly refused to guess, but only after loading all five bundled references and continuing through later Build Workflow steps.

Could we make this blocked path explicit? Once the available sources cannot establish the required external Provider contract, preserve the working configuration and stop the dependent integration work. The response should report what was checked, the exact missing contract, and the minimum vendor input needed to continue. This would be a narrow early-exit rule for unverifiable external integrations, not a general stop for unresolved business TODOs.

dev/agent-skills/flink-agents-dev/SKILL.md
wenjin272: Agreed. I've also observed this in my local development work.

dev/agent-skills/flink-agents-dev/references/platforms/codex.md
```diff
@@ -0,0 +1,110 @@
...
+Use this adapter only when explicit host context identifies Codex.
+
+## Capability Check
+
+The current session and mode's tool contract is authoritative. Codex exposes its
+native closed-choice UI through `request_user_input` in Plan mode. Before the first
+decision gate, check the explicit collaboration mode and the available tool
+contract.
+
+When the session is not in Plan mode, respond with one short message in the user's
```
GreatEugenius: **Make Codex interaction capability-driven and keep the policy authoritative in one place.**

I tested this flow on a Codex surface where `/plan` was not an available user command. The adapter stopped the task and requested a mode switch even though the generic numbered fallback could have continued.

Could we make the current tool contract the only capability signal?

- If `request_user_input` is currently exposed and callable, use it.
- Otherwise, immediately use the generic numbered fallback.
- Do not require `/plan`, `Shift+Tab`, or a Plan-to-Default handoff.

The mode and fallback sequence is currently defined in `SKILL.md`, `application-patterns.md`, and `platforms/codex.md`. Could we keep the portable rule authoritative in one place and have the other files reference it? Otherwise, changing only one copy may leave the Skill with inconsistent instructions.

dev/agent-skills/flink-agents-dev/SKILL.md
```diff
@@ -0,0 +1,321 @@
...
+application idea. Do not propose a standard REST API, MCP server, or mock backend as
+the implementation of a custom Tool. Do not ask the user to lock a business input
+identity schema, deployment platform, service API, authentication design, or log/
+metric backend merely to scaffold the application. Generate the neutral skeleton
+and leave those choices as TODOs for the user. Built-in Flink Agents Actions are
+framework behavior and need no generated implementation.
+
+## Interaction Discipline
+
+For a new or empty project, use sequential decision gates. Ask only the current
```
GreatEugenius: **Resolve already-confirmed decisions before entering the gated interview.**

The Skill says to skip gates that the user or existing project has already answered, but it does not make that resolution pass an explicit step before selecting a host adapter. In practice, the workflow can start a handoff or repeat a question before reconciling the complete context.

Could we require the workflow to:

1. inspect the existing project and all explicit user requirements;
2. resolve each decision, with the latest explicit user choice taking precedence;
3. build a list containing only genuinely unresolved gates; and
4. invoke a host interaction adapter only when that list is non-empty?

This would prevent repeated questions and avoid unnecessary host interaction when every required decision is already known.

Issue #912, opened 2026-07-17
[Feature] Apply structured output to an agent's final output only


weiqingy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Follow-up to #280. This issue covers applying structured output to an agent's **final output only**, which #280 scoped as a separate step from the chat-model-layer foundation being added in #843.

**Problem.** When a `ReActAgent` declares an `output_schema`, the schema is honored purely by prompt engineering today: a system prompt instructs the model to emit JSON, and the response text is fence-stripped and parsed afterwards. At the chat-model layer, an intermediate tool-calling turn and the final answering turn are indistinguishable, so there is no point at which structured output can be applied to just the final result. The #843 foundation adds an explicit output-schema parameter to the chat path plus a policy (`auto`/`native`/`prompt`) and connection-side capability split, but no framework path passes a schema on that call yet — supplying it at the right moment is what this issue adds.

**Proposed solution.** Issue a dedicated structured call once the loop has produced its final answer — at `ChatModelAction`'s no-tool-calls branch, the point where the model stops requesting tools. That call passes the output schema explicitly, so native structured output applies to the final output only and intermediate tool-calling turns are untouched.

**Schema instructions must become strategy-driven.** `ReActAgent` currently registers a schema-instruction prompt (`The final response should be json format, and match the schema ...`) unconditionally at construction whenever `output_schema` is set, in both languages:

- Java — `ReActAgent.java:66-87` (registered at `:86`)
- Python — `react_agent.py:128-139`

That prompt is then injected into the messages when the action runs (Java `ReActAgent.java:156-162`, Python `react_agent.py:189-196`).

If the finalization call uses a model that applies structured output natively, the prompt instruction and the provider's own schema enforcement become two overlapping channels describing the same schema. The schema-instruction path therefore needs to follow the selected strategy: emit it for the prompt-based fallback, omit it when the native path applies. Both sites are candidates for carrying that decision — the injection is per-call while the registration happens at construction, before the effective model is known — so which one carries it is part of this issue's design rather than a settled choice.

**Open questions.**

- Should finalization be an extra post-loop model round-trip, or a forced tool call that avoids the additional call? An extra call is the simpler starting point; the forced-tool form is an optimization that could follow.
- Where should `auto` resolve policy into a concrete strategy, given capability is a connection-side predicate over the effective model at request-build time?

[...]

PR #919, opened 2026-07-19
[integrations][openai] Apply OpenAI native structured output

weiqingy:
Linked issue: #280

### Purpose of change

Applies OpenAI's native structured output on top of the structured-output foundation from #843, which has now merged. This PR is rebased onto main and is a single commit.

The OpenAI completions connection overrides the schema-carrying `chat` and requests `response_format` json_schema strict when native applies, and declares the capability predicate over the effective model.

Native applies only when the schema is a POJO class (Java) or a `BaseModel` subclass (Python) and the effective model supports it. Per OpenAI's structured-output guide, json_schema is supported on `gpt-4o-mini` and the `gpt-4o` snapshots from `2024-08-06` onward, while `gpt-4-turbo` and `gpt-3.5-turbo` have JSON mode only. Because "and later" is a snapshot cutoff rather than a name prefix, the predicate matches `gpt-4o-mini` by prefix (its whole family post-dates the cutoff) and the capable `gpt-4o` snapshots exactly, so a pre-cutoff snapshot such as `gpt-4o-2024-05-13` falls back to the prompt path instead of failing at the provider. A `RowTypeInfo` schema stays on the prompt path.

#843 added `_reject_unsupported_output_schema` so that a connection with no native translation turns a schema it could only drop into an error. OpenAI has that translation now, so it no longer calls the guard, and the connection walk in `test_output_schema_param_declared.py` exempts connections that translate a schema natively. The exemption keys off the override of `supports_native_structured_output`, which is how a connection reports that capability in the first place. The other seven connections stay held to the rejection contract.

No path passes a schema on the chat call yet, so behavior is unchanged; native output is exercised by direct callers and the unit tests.

### Tests

Java: the request carries `response_format` json_schema strict when a capable model and a POJO schema are supplied; the capability predicate is model-dependent (an incapable and a pre-cutoff same-family snapshot both resolve to not-capable); null schema and non-POJO schema keep the prompt path.

Python: the same coverage over the `BaseModel` path, including the pre-cutoff snapshot regression test and policy behavior.

The connection walk still fails if a connection without a native translation stops rejecting a schema, and its existing non-empty assertion now runs on the filtered list, so an exemption that ever swallowed the whole walk would fail rather than pass silently.

Java OpenAI module tests and the Python chat model suites pass after the rebase; ruff and Spotless are clean.

### API

Touches the OpenAI connection: overrides the schema-carrying `chat` and declares the capability predicate. No new user-facing surface beyond the foundation in #843.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
+++ b/integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
@@ -22,11 +22,13 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.openai.client.OpenAIClient;
 import com.openai.client.okhttp.OpenAIOkHttpClient;
+import com.openai.core.JsonSchemaLocalValidation;
 import com.openai.core.JsonValue;
 import com.openai.models.ChatModel;
 import com.openai.models.FunctionDefinition;
 import com.openai.models.FunctionParameters;
 import com.openai.models.ReasoningEffort;
+import com.openai.models.ResponseFormatJsonSchema;
 import com.openai.models.chat.completions.ChatCompletion;
 import com.openai.models.chat.completions.ChatCompletionCreateParams;
 import com.openai.models.chat.completions.ChatCompletionFunctionTool;
@@ -43,6 +45,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * A chat model integration for the OpenAI Chat Completions service using the official Java SDK.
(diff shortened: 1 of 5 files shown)


weiqingy:
Hi @wenjin272, could you please help review this PR when you get a chance? It applies OpenAI's native structured output on top of #843, and it's now rebased onto main since that merged. Thanks!

weiqingy:
One thing worth calling out explicitly, since it is deliberately left out of this PR.

`StructuredOutputStrategy.NATIVE` resolves to native regardless of the connection's reported capability, so a caller with an explicit `NATIVE` intent expects the schema to be applied even on a model the connection does not recognize as capable. This PR's OpenAI `chat` re-checks capability itself and skips `response_format` when the effective model is not capable or the schema is a `RowTypeInfo`, which under that intent would return an unconstrained response rather than an error. That is the same shape as the silent drop fixed in #843.

I left it unfixed here on purpose. The connection never receives the strategy, so it cannot tell an explicit `NATIVE` apart from an `AUTO` that happens to resolve to native, and any decision made inside the connection is a guess at caller intent. The composition site is where the setup's policy meets the connection's capability, which is #912. Fixing it there also means dropping or bypassing this connection-side re-check, otherwise the setup honoring `NATIVE` would still be downgraded here. I have recorded that on my side so it lands with the wiring.

Nothing dispatches on the strategy today, so this is dormant on `main`: no path passes a schema on the chat call yet, and the native path here is exercised only by direct callers and the unit tests.

Happy to pull the fix into this PR instead if you would rather it not wait for #912. It would change the two fallback tests, which currently assert the silent degrade.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
```diff
@@ -119,11 +122,53 @@ public OpenAICompletionsConnection(
...
+    private static final Set<String> NATIVE_STRUCTURED_OUTPUT_MODELS =
+            Set.of("gpt-4o", "gpt-4o-2024-08-06", "gpt-4o-2024-11-20");
+
+    @Override
+    protected boolean supportsNativeStructuredOutput(String effectiveModel) {
+        if (effectiveModel == null) {
+            return false;
+        }
+        return effectiveModel.startsWith(NATIVE_STRUCTURED_OUTPUT_FAMILY_PREFIX)
+                || NATIVE_STRUCTURED_OUTPUT_MODELS.contains(effectiveModel);
```
wenjin272, line 163: The capability predicate appears to be inaccurate in both directions.

Using `startsWith("gpt-4o-mini")` also classifies models such as `gpt-4o-mini-audio-preview` and `gpt-4o-mini-realtime-preview` as capable, although these variants do not support Structured Outputs. Conversely, models such as `gpt-4.1`, `gpt-5`, `o3`, and `o4-mini` do support Structured Outputs but currently return `false`.

Once #912 uses this predicate for `AUTO`, the false positives may produce provider errors, while the false negatives will unnecessarily fall back to prompting. Could we use stricter alias/snapshot matching, include the currently supported model families, and add tests covering both cases?

References:
- https://developers.openai.com/api/docs/models/gpt-4o-mini-audio-preview
- https://developers.openai.com/api/docs/models/gpt-4.1
- https://developers.openai.com/api/docs/models/gpt-5

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
```diff
@@ -170,6 +220,13 @@ private ChatCompletionCreateParams buildRequest(
             builder.tools(convertTools(tools, strictMode));
         }
 
+        // Native structured output applies only for a POJO Class schema on a model the provider
+        // documents as capable; a RowTypeInfo (wrapped in OutputSchema) or an incapable model keeps
+        // the prompt-engineering fallback.
+        if (outputSchema instanceof Class && supportsNativeStructuredOutput(modelName)) {
```
wenjin272, line 249: I agree that the final fix belongs in #912, where AUTO can be distinguished from explicit NATIVE. Could we add a clear TODO referencing #912 at this capability check, noting that NATIVE must bypass the connection-side fallback or fail explicitly? This should help prevent the dormant issue from being missed when the strategy wiring is implemented.

wenjin272 (commented):
Thanks for taking this on @weiqingy. I left two minor comments

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
weiqingy: Good catch, you were right in both directions. Fixed in f01e9443.

The predicate now rejects any name carrying a non-text modality marker (`-audio`, `-realtime`, `-tts`, `-transcribe`) first, then matches a capable family by prefix or a capable name exactly. Checking against the model pages turned up three more false positives than the two you named: `gpt-4o-audio-preview`, `gpt-4o-mini-tts` and `gpt-4o-mini-transcribe`.

Two cases are why prefix and exact matching are split:

- `gpt-4o-search-preview` does support Structured Outputs, so rejecting on `-preview` would have swapped your false positives for a new false negative.
- `o1` supports them but `o1-mini` does not, and `o1-mini` is served on Chat Completions. So `o1` is matched exactly. A prefix there would have reintroduced the bug you reported, and there is a test pinning `o1-mini` as rejected.

Tests cover both directions in Java and Python, including every model you named.

One tradeoff worth flagging: with prefix matching, an unknown name inside a listed family now reports capable, so unknown names are no longer uniformly not-capable. The comment in both files says so.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAICompletionsConnection.java
weiqingy: Added in e7639ed9, both languages.

It sits at the capability re-check, in `buildRequest` on the Java side and `chat` on the Python side. It records that the connection cannot tell an explicit `NATIVE` from an `AUTO` that merely resolved to native, so `NATIVE` on a model the predicate rejects degrades to the prompt fallback instead of erroring, and that #912 has to either bypass this re-check or fail explicitly.

wenjin272 (approved):
LGTM

Issue #920, opened 2026-07-20
[Bug][Python] Silent config-reconstruction fallback in Action deserialization makes failures very hard to debug


alnzng:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### What I was testing

  A `ReActAgent` with both `tools` and an `output_schema` (`FlinkSqlJobProfile`). The schema class was importable on the driver but **not on the worker** (a packaging mistake on my side).

  ### Symptom

  The job failed at runtime pointing at `chat_model_action.py` with a misleading `AttributeError`:

  ```
  pemja.core.PythonException: <class 'AttributeError'>: 'list' object has no attribute 'model_dump'
      at .../flink_agents/plan/actions/chat_model_action._save_tool_request_event_context
  (chat_model_action.py:115)
      at .../flink_agents/plan/actions/chat_model_action._handle_tool_calls (chat_model_action.py:237)
      at .../flink_agents/plan/actions/chat_model_action.chat (chat_model_action.py:379)
      at .../flink_agents/plan/actions/chat_model_action._process_chat_request (chat_model_action.py:410)
      at .../flink_agents/plan/actions/chat_model_action.process_chat_request_or_tool_response
  (chat_model_action.py:470)
      at pemja.core.PythonInterpreter.invoke(PythonInterpreter.java:92)
      at org.apache.flink.agents.runtime.operator.ActionExecutionOperator.processActionTaskForKey(ActionExecutionOp
  erator.java:375)
  ```

  Nothing here points at the real problem.

  ### Actual root cause
  
  `Action.__custom_deserialize` (`plan/actions/action.py`) reconstructs `[module, class, value]` config values
  inside a `try/except` that **silently swallows any failure** and keeps the raw value:

  ```python
  for name, value in config.items():
      try:
          module = importlib.import_module(value[0])
          clazz = getattr(module, value[1])
          self["config"][name] = clazz.model_validate(value[2])
      except Exception:          # <-- swallows the real error
          self["config"][name] = value
  ```

  My `output_schema` class wasn't importable on the worker → reconstruction raised `ModuleNotFoundError` →
  swallowed → `output_schema` stayed a raw `list` → later `_save_tool_request_event_context` called `.model_dump()` on that list → the misleading `AttributeError` above.

  ### What the error *should* have looked like

  Making the branch fail fast instead of swallowing gave the true cause and location, at plan-load before any
  record was processed:

  ```
[...]

PR #921, opened 2026-07-20
[bug] [python] Propagate config model reconstruction errors in Action deserialize

alnzng:
Linked issue: #920 

### Purpose of change
This change wants to improve the debuggability of Action config deserialization. Before, `Action.__custom_deserialize` catch all errors when it reconstruct `python`-format config models, so if a real error happen (for example `ModuleNotFoundError`), the error is hidden and the raw value is stored. User can not know the real reason, and the problem show up later as a confusing error, e.g. `pemja.core.PythonException: <class 'AttributeError'>: 'list' object has no attribute 'model_dump'`.

Now we replace the broad `try/except` with a explicit isinstance(value, list | tuple)` check, so the real error can propagate and fail fast, and it is easy to find the root cause. Plain values still pass through as before. 

### Tests

- Added unit test
- Verified in a test app and checked the real error propagated properly

Screenshot 2026-07-20 at 12 12 48 AM

### API

N/A

### Documentation

- [ ] `doc-needed` 
- [X] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/plan/actions/action.py
+++ b/python/flink_agents/plan/actions/action.py
@@ -26,6 +26,9 @@ from flink_agents.api.runner_context import RunnerContext
 from flink_agents.plan.function import Function, JavaFunction, PythonFunction
 
 _CONFIG_TYPE = "__config_type__"
+# Tags a config entry that we serialized from a pydantic model, so on the way
+# back we only rebuild the ones we tagged and leave plain values alone.
+_PYDANTIC_MODEL_MARKER = "__pydantic_model__"
 
 
 class Action(BaseModel):
@@ -70,11 +73,12 @@ class Action(BaseModel):
         data[_CONFIG_TYPE] = "python"
         for name, value in config.items():
             if isinstance(value, BaseModel):
-                data[name] = (
-                    inspect.getmodule(value).__name__,
-                    value.__class__.__name__,
-                    value,
-                )
+                data[name] = {
+                    _PYDANTIC_MODEL_MARKER: True,
+                    "module": inspect.getmodule(value).__name__,
+                    "class": value.__class__.__name__,
+                    "value": value,
+                }
             else:
                 data[name] = value
         return data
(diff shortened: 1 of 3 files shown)


python/flink_agents/plan/actions/action.py
```diff
@@ -95,11 +95,11 @@ def __custom_deserialize(self) -> "Action":
                     self["config"][name] = value["value"]
             return self
         for name, value in config.items():
-            try:
+            if isinstance(value, list | tuple):
```
weiqingy: The serializer only wraps `BaseModel` values into the `(module, class, value)` shape. A plain list/tuple config value is written as-is and comes back from JSON as a plain list, so this gate can't tell a serialized model apart from a genuine list. For `config={"tags": ["a", "b"]}` the round-trip would reach `import_module("a")` and raise, where the old broad-except kept the raw list. The java branch above avoids this by tagging wrapped values with `@class`/`value` and testing membership first.

I'm not sure a plain list/tuple config value is actually reachable today, so genuinely a question: do we expect list-valued config entries (stop sequences, tags, tool lists)? If so, would a self-describing marker on the python shape, mirroring the java tagging, let the deserializer match on the marker instead of guessing from type, with a round-trip test to lock it in? If plain lists are out of scope here, feel free to ignore.

alnzng: @weiqingy Very good question on plain list/tuple config value!

My first thinking was the type check is enough, because normally the model config comes from our own code. That is why I started with the shape check.

But I checked again and your worry is real, even more than "future". `add_action(**config)` is public and takes any keyword value, so a user can already pass `agent.add_action(name=..., func=..., tags=["a", "b"])` today.

In that case the old shape check will treat `["a", "b"]` as a model and call `import_module("a")`, then it breaks. So it is not only a future assumption, it is already reachable now.

So I agree `marker` is the better way. When we serialize we tag the value clearly, and when we deserialize we only rebuild the entries that have the `marker`. A plain `list` stays a plain `list`. It also keeps `python` consistent with `java`, which already tags the wrapped value with `@class`/`value` and checks the `marker` first. Now both sides use the same way, not two different logics.

For compatibility I think there is no problem:
- The plan is built fresh from the code on every submission and shipped inside the job graph, not restored from checkpoint, so a new `deserialize` never meets an old-format plan.
- On `java` side the `python` config path is a generic passthrough (it writes each value as-is and rebuilds it back to a `map`), so the `marker` dict goes through the same as before, no `java` change needed.

Please take a look again on my new changes, thanks.

weiqingy: Thanks for digging into this. Good catch that `add_action(**config)` already makes a plain list reachable today, so this was live, not just future. The marker approach looks right, and the new `preserves_plain_list` test locks the contract in.

I also checked the compatibility point you raised: `deserializePythonConfig` on the java side is a generic recursive passthrough, so it never interprets the old tuple or the new marker. The marker dict round-trips the same as before, no java change needed. Your read holds.

Nothing further from me.

wenjin272 (approved):
Thanks for fixing this @alnzng. LGTM

Issue #902, opened 2026-07-14
[Feature] [integration] Integrate built-in IBM watsonx.ai chat model


cansuk99:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

I noticed that on Flink Agents built-in chat model integrations, IBM watsonx.ai is missing. watsonx.ai hosts the IBM Granite models plus some third-party ones, and it's quite common in enterprise setups running on IBM Cloud.

The watsonx.ai chat API looks OpenAI-compatible at first (same message and tool format), but the existing OpenAI module does not fully cover it: authentication goes through IBM Cloud IAM (you exchange an API key for a short-lived bearer token, which needs caching/refresh) and every request needs a` project_id` or `space_id plus` a version query parameter.

I think a dedicated module integration can help with adoption on Apache Flink agents for users running on IBM watsonx.ai.

I propose adding:

`integrations/chat-models/watsonx/ (Java)`
`flink_agents/integrations/chat_models/watsonx/ (Python)` using the official ibm-watsonx-ai SDK.

I am looking to contribute to the project and bring this integration forward.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wenjin272:
Thanks for raising this, @cansuk99. A dedicated watsonx.ai integration would be a valuable addition, and we really appreciate your willingness to contribute it.

cansuk99:
Hello Wenjin, I implemented on this PR: https://github.com/apache/flink-agents/pull/922. I appreciate your review on it.

PR #922, opened 2026-07-22
[java][python] IBM watsonx.ai chat model integration

cansuk99:
Linked issue: #902 

### Purpose of change

Chat model integration for watsonx.ai models provided by IBM Cloud.

- Adds a new watsonx integration module for Java chat models.
-  Adds a new watsonx integration module for Python chat models.
- Adds watsonx resource names and YAML alias mappings in both Java and Python.
- Updates distribution and build wiring to include the new integration artifact.
- Adds documentation for watsonx setup, parameters, model usage, and YAML alias support.

### Tests

Java:
- Watsonx connection behavior and payload conversion tests.
- Watsonx setup validation tests.
- Alias resolution tests for watsonx.

Python:
- Watsonx chat behavior tests (mocked tool-call, retry flows).
- Alias resolution tests for watsonx.

Both Java and Python integrations were tested live with IBM Cloud credentials with flink jobs running on k8s.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
+++ b/integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
@@ -1,0 +1,678 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.integrations.chatmodels.watsonx;
+
+import com.fasterxml.jackson.core.json.JsonReadFeature;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.apache.flink.agents.api.chat.messages.MessageRole;
+import org.apache.flink.agents.api.chat.model.BaseChatModelConnection;
+import org.apache.flink.agents.api.resource.ResourceContext;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
+import org.apache.flink.agents.api.tools.Tool;
+import org.apache.flink.annotation.VisibleForTesting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URLEncoder;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/** Chat model connection for the IBM watsonx.ai text chat REST API. */
+public class WatsonxChatModelConnection extends BaseChatModelConnection {
+
+    private static final Logger LOG = LoggerFactory.getLogger(WatsonxChatModelConnection.class);
+
+    static final String DEFAULT_IAM_URL = "https://iam.cloud.ibm.com";
+    static final String DEFAULT_API_VERSION = "2025-04-23";
+    static final long DEFAULT_REQUEST_TIMEOUT_SEC = 120;
+    static final int DEFAULT_MAX_RETRIES = 3;
+    private static final Set<Integer> RETRYABLE_STATUS_CODES = Set.of(408, 429, 500, 502, 503, 504);
+
+    private static final Set<String> CONTROL_PARAMS =
+            Set.of(
+                    "model",
+                    "tool_choice",
+                    "tool_choice_option",
+                    "extract_reasoning",
+                    "additional_kwargs");
+    private static final Set<String> RESERVED_ADDITIONAL_KWARGS =
+            Set.of(
+                    "model",
+                    "temperature",
+                    "max_tokens",
+                    "extract_reasoning",
+                    "tool_choice",
+                    "tool_choice_option");
+
+    private static final Pattern[] REASONING_PATTERNS = {
+        Pattern.compile("<think>(.*?)</think>", Pattern.DOTALL | Pattern.CASE_INSENSITIVE),
+        Pattern.compile("<analysis>(.*?)</analysis>", Pattern.DOTALL | Pattern.CASE_INSENSITIVE),
+        Pattern.compile("<reasoning>(.*?)</reasoning>", Pattern.DOTALL | Pattern.CASE_INSENSITIVE),
+        Pattern.compile(
+                "```(?:think|reasoning|thought)\\s*\\n(.*?)\\n```",
+                Pattern.DOTALL | Pattern.CASE_INSENSITIVE),
+        Pattern.compile(
+                "(?:^|\\n)Reasoning:\\s*(.*?)(?:\\n{2,}|$)",
+                Pattern.DOTALL | Pattern.CASE_INSENSITIVE),
+    };
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    private static final ObjectMapper LENIENT_MAPPER =
+            JsonMapper.builder()
+                    .enable(JsonReadFeature.ALLOW_SINGLE_QUOTES)
+                    .enable(JsonReadFeature.ALLOW_UNQUOTED_FIELD_NAMES)
+                    .build();
+
+    private final String url;
+    private final String apiKey;
+    private final String staticToken;
+    private final String projectId;
+    private final String spaceId;
+    private final String apiVersion;
+    private final String iamUrl;
+    private final Duration requestTimeout;
+    private final int maxRetries;
+
+    private final HttpClient httpClient;
+
+    private transient String cachedIamToken;
+    private transient long iamTokenExpirationEpochSec;
+
+    public WatsonxChatModelConnection(
+            ResourceDescriptor descriptor, ResourceContext resourceContext) {
+        this(descriptor, resourceContext, System::getenv);
+    }
+
+    @VisibleForTesting
+    WatsonxChatModelConnection(
+            ResourceDescriptor descriptor,
+            ResourceContext resourceContext,
+            Function<String, String> environmentLookup) {
+        super(descriptor, resourceContext);
+
+        this.url =
+                trimTrailingSlash(
+                        argumentOrEnv(descriptor, "url", "WATSONX_URL", environmentLookup));
+        this.apiKey = argumentOrEnv(descriptor, "api_key", "WATSONX_API_KEY", environmentLookup);
+        this.staticToken = argumentOrEnv(descriptor, "token", "WATSONX_TOKEN", environmentLookup);
+        this.projectId =
+                argumentOrEnv(descriptor, "project_id", "WATSONX_PROJECT_ID", environmentLookup);
+        this.spaceId = argumentOrEnv(descriptor, "space_id", "WATSONX_SPACE_ID", environmentLookup);
+
+        String apiVersion = normalize(descriptor.getArgument("api_version"));
+        this.apiVersion = apiVersion != null ? apiVersion : DEFAULT_API_VERSION;
+        String iamUrl = normalize(descriptor.getArgument("iam_url"));
+        this.iamUrl = trimTrailingSlash(iamUrl != null ? iamUrl : DEFAULT_IAM_URL);
+        Number requestTimeout = descriptor.getArgument("request_timeout");
+        double requestTimeoutSeconds =
+                requestTimeout != null ? requestTimeout.doubleValue() : DEFAULT_REQUEST_TIMEOUT_SEC;
+        if (!Double.isFinite(requestTimeoutSeconds) || requestTimeoutSeconds <= 0) {
+            throw new IllegalArgumentException("request_timeout must be a positive finite number.");
+        }
+        this.requestTimeout =
+                Duration.ofMillis(Math.max(1L, Math.round(requestTimeoutSeconds * 1000.0)));
+        Number maxRetries = descriptor.getArgument("max_retries");
+        this.maxRetries =
+                maxRetries != null
+                        ? requireInteger(maxRetries, "max_retries", 0)
+                        : DEFAULT_MAX_RETRIES;
+
+        if (this.url == null || this.url.isEmpty()) {
+            throw new IllegalArgumentException(
+                    "watsonx.ai url is not provided. Please pass the 'url' argument or set the"
+                            + " 'WATSONX_URL' environment variable.");
+        }
+        if ((this.apiKey == null || this.apiKey.isEmpty())
+                && (this.staticToken == null || this.staticToken.isEmpty())) {
+            throw new IllegalArgumentException(
+                    "watsonx.ai credentials are not provided. Please pass the 'api_key' or 'token'"
+                            + " argument, or set the 'WATSONX_API_KEY' or 'WATSONX_TOKEN'"
+                            + " environment variable.");
+        }
+        if (this.apiKey != null && this.staticToken != null) {
+            throw new IllegalArgumentException(
+                    "watsonx.ai api_key and token cannot both be provided. Please configure"
+                            + " exactly one credential source.");
+        }
+        if ((this.projectId == null || this.projectId.isEmpty())
+                && (this.spaceId == null || this.spaceId.isEmpty())) {
+            throw new IllegalArgumentException(
+                    "watsonx.ai project or space is not provided. Please pass the 'project_id' or"
+                            + " 'space_id' argument, or set the 'WATSONX_PROJECT_ID' or"
+                            + " 'WATSONX_SPACE_ID' environment variable.");
+        }
+        if (this.projectId != null
+                && !this.projectId.isEmpty()
+                && this.spaceId != null
+                && !this.spaceId.isEmpty()) {
+            throw new IllegalArgumentException(
+                    "watsonx.ai project and space cannot both be provided. Please configure"
+                            + " exactly one of 'project_id' or 'space_id'.");
+        }
+
+        this.httpClient = HttpClient.newBuilder().connectTimeout(Duration.ofSeconds(30)).build();
+    }
+
+    static int requireInteger(Number value, String argumentName, int minimum) {
+        double numericValue = value.doubleValue();
+        if (!Double.isFinite(numericValue)
+                || numericValue != Math.rint(numericValue)
+                || numericValue < minimum
+                || numericValue > Integer.MAX_VALUE) {
+            throw new IllegalArgumentException(
+                    argumentName
+                            + " must be "
+                            + (minimum == 0 ? "a non-negative" : "a positive")
+                            + " integer.");
+        }
+        return (int) numericValue;
+    }
+
+    private static String argumentOrEnv(
+            ResourceDescriptor descriptor,
+            String argumentName,
+            String envName,
+            Function<String, String> environmentLookup) {
+        String value = normalize(descriptor.getArgument(argumentName));
+        if (value == null) {
+            value = normalize(environmentLookup.apply(envName));
+        }
+        return value;
+    }
+
+    private static String normalize(String value) {
+        if (value == null || value.isBlank()) {
+            return null;
+        }
+        return value.trim();
+    }
+
+    private static String trimTrailingSlash(String value) {
+        if (value != null && value.endsWith("/")) {
+            return value.substring(0, value.length() - 1);
+        }
+        return value;
+    }
+
+    @Override
+    public ChatMessage chat(
+            List<ChatMessage> messages, List<Tool> tools, Map<String, Object> modelParams) {
+        try {
+            final String modelName = (String) modelParams.get("model");
+            final boolean extractReasoning =
+                    Boolean.TRUE.equals(modelParams.get("extract_reasoning"));
+            final ObjectNode payload = buildPayload(messages, tools, modelParams);
+            if (projectId != null && !projectId.isEmpty()) {
+                payload.put("project_id", projectId);
+            } else {
+                payload.put("space_id", spaceId);
+            }
+
+            final HttpRequest request =
+                    HttpRequest.newBuilder()
+                            .uri(URI.create(url + "/ml/v1/text/chat?version=" + apiVersion))
+                            .timeout(requestTimeout)
+                            .header("Authorization", "Bearer " + getBearerToken())
+                            .header("Content-Type", "application/json")
+                            .header("Accept", "application/json")
+                            .POST(
+                                    HttpRequest.BodyPublishers.ofString(
+                                            MAPPER.writeValueAsString(payload)))
+                            .build();
+
+            final HttpResponse<String> response = sendWithRetry(request);
+            if (response.statusCode() / 100 != 2) {
+                throw new RuntimeException(
+                        String.format(
+                                "watsonx.ai chat request failed with status %d: %s",
+                                response.statusCode(), response.body()));
+            }
+
+            final ChatMessage chatMessage =
+                    parseResponse(MAPPER.readTree(response.body()), modelName);
+            if (extractReasoning) {
+                final String[] parts = extractReasoning(chatMessage.getContent());
+                chatMessage.setContent(parts[0]);
+                if (parts[1] != null) {
+                    chatMessage.getExtraArgs().put("reasoning", parts[1]);
+                }
+            }
+            return chatMessage;
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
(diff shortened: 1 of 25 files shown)


weiqingy (commented):

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
```diff
@@ -0,0 +1,476 @@
...
+                    model_name,
+                    status if status is not None else type(e).__name__,
+                    attempt + 1,
+                    self.max_retries,
+                    delay,
+                )
+                time.sleep(delay)
+                attempt += 1
+
+    def chat(
```
weiqingy, line 335: This one is a rebase follow-up rather than something done wrong. #843 ("[api][python] Add explicit output schema parameter to the chat path (structured-output foundation)") added `output_schema: OutputSchema | None = None` to `BaseChatModelConnection.chat` (`python/flink_agents/api/chat_models/chat_model.py:230`) on 2026-07-25, after this branch was cut. The docstring at `:241-249` asks for a named parameter, because "`**kwargs` is forwarded to the provider SDK, so a schema landing there would reach the request body." That happens literally here: `watsonx_chat_model.py:362` forwards `params={**kwargs, **additional_kwargs}` into `ModelInference.chat(params=...)`, and the SDK does `payload.update(parameters)` (`ibm_watsonx_ai/foundation_models/inference/fm_model_inference.py:597-598`).

The same commit added `python/flink_agents/integrations/chat_models/tests/test_output_schema_param_declared.py`. After a rebase, `test_every_connection_declares_output_schema_param` (`:90`) and the rejection test at `:143` fail on a whole-tree `pytest` run, where collection imports the watsonx test module. That file run on its own still passes, since "`__subclasses__()` only sees classes that have been imported" and `watsonx_chat_model` isn't in its import list at `:41-51`. The green checks predate all of it: the last CI run was created 2026-07-24, a day before #843 merged.

`openai_chat_model.py:142` and `:167` are the closest template. The leak stays latent even after the rebase, since `BaseChatModelSetup.chat` doesn't forward `output_schema` yet, so only a direct `connection.chat(..., output_schema=...)` caller reaches the body. Easier to pick the parameter and the import-list entry up as part of the rebase, or would you rather handle them separately once the branch is current?

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
```diff
@@ -0,0 +1,476 @@
...
+        if model not in self._models:
+            self._models[model] = ModelInference(
+                model_id=model,
+                api_client=self.client,
+                project_id=self.project_id,
+                space_id=self.space_id,
+            )
+        return self._models[model]
+
+    def _chat_with_retry(self, model_name: str, **chat_kwargs: Any) -> Dict[str, Any]:
```
weiqingy, line 304: `_chat_with_retry` retries over `RETRYABLE_STATUS_CODES = {408, 429, 500, 502, 503, 504}` with `max_retries` defaulting to 3. The SDK already retries underneath: `ibm_watsonx_ai/_wrappers/httpx/rate_limited_retry/rate_limited_retry_decorator.py:40-44` sets `MAX_RETRIES = 10` over `(429, 503, 504, 520)`, wired into every chat POST at `base_model_inference.py:63-76`, same on the 1.3.42 pin. A sustained 429 then costs up to 44 HTTP requests and minutes of blocking `time.sleep` (`:327`) on a Flink async-executor thread for one record, with a third layer above both at `plan/actions/chat_model_action.py:304-360`.

Java's `sendWithRetry` (`WatsonxChatModelConnection.java:290-325`) is the only layer there, so the same `max_retries=3` means about 3 retries on Java and up to about 43 on Python, while the `max_retries` row in `chat_models.md` describes both identically ("Maximum retries for transport failures and HTTP 408, 429, 500, 502, 503, and 504 responses"). `AGENTS.md` asks for semantic alignment, and here the key, the default and the docs all match while behavior differs by an order of magnitude.

Was the SDK's built-in retry known about when the wrapper was written? `ModelInference.__init__` takes `max_retries`, `delay_time` and `retry_status_codes` directly (`model_inference.py:162-164`, both pins), so the configured policy could be handled by one layer with the same user-visible semantics.

integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
```diff
@@ -0,0 +1,678 @@
...
+        final Object toolChoiceOption = modelParams.get("tool_choice_option");
+        if (toolChoiceOption != null) {
+            payload.put("tool_choice_option", toolChoiceOption.toString());
+        }
+
+        @SuppressWarnings("unchecked")
+        final Map<String, Object> additionalKwargs =
+                (Map<String, Object>) modelParams.get("additional_kwargs");
+        if (additionalKwargs != null) {
+            final Set<String> collisions = new java.util.HashSet<>(additionalKwargs.keySet());
```
weiqingy, line 422: `buildPayload` sets the structural fields first, `model_id` at `:383`, `messages` at `:384` and `tools` at `:387`, then spills `additional_kwargs` over the top at `:410-415`. This collision check only guards `RESERVED_ADDITIONAL_KWARGS`, which is `model`, `temperature`, `max_tokens`, `extract_reasoning`, `tool_choice` and `tool_choice_option` (`:72-79`). So `additional_kwargs: {messages: [...]}` passes the check and replaces the real conversation, and `tools` goes the same way.

`model_id` is the sharper case, because reserving `model` doesn't cover it: the payload key is `model_id`, so `additional_kwargs: {model_id: "other-model"}` silently overrides the configured model while the reserved-name check sees nothing. Python has the same shape through `params={**kwargs, **additional_kwargs}` at `watsonx_chat_model.py:362`, where the SDK's `payload.update(parameters)` lets those keys into the body.

Would it make sense to reserve the request-owned keys too, so static setup configuration can't replace the model, the messages or the tools?

integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
```diff
@@ -0,0 +1,678 @@
...
+                }
+                break;
+            }
+        }
+        throw new RuntimeException(
+                "Failed to parse tool call arguments returned by watsonx.ai as a JSON object: "
+                        + argumentsNode);
+    }
+
+    private synchronized String getBearerToken() {
```
weiqingy, line 661: `cachedIamToken` is only ever replaced by the time check at `:641`. 401 isn't in `RETRYABLE_STATUS_CODES` (`:63`) and nothing in the non-2xx branch clears the cache, so if the token stops being accepted before its computed local expiry (key rotated, session invalidated, or the wall clock at `:639` jumping backwards) every subsequent `chat()` fails with a 401 until local time crosses the expiry. The normal-expiry path is right, with the 60s margin and `nowEpochSec` captured before the network call.

`:636-638` returns a user-supplied `staticToken` forever. IBM Cloud IAM tokens live about an hour, so that path only works for jobs shorter than the TTL, and the docs describe `token` only as "Bearer token; configure exactly one of `api_key` and `token`". Python shares this, since its SDK docs say passing `token=` disables automatic refresh.

How often the first one bites I genuinely don't know, since the 60s margin covers the common case. Does it make sense for a 401 to invalidate the cache, or is the time check meant to be the only trigger? One shape, if invalidating is the direction you'd want: clearing the cache and retrying once on 401 or 403 when `apiKey != null`, with a docs note that `token` doesn't refresh.

integrations/chat-models/watsonx/src/test/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnectionTest.java
```diff
@@ -0,0 +1,395 @@
...
+
+        // exponential backoff, capped
+        assertThat(WatsonxChatModelConnection.retryDelayMillis(0, null)).isEqualTo(1000L);
+        assertThat(WatsonxChatModelConnection.retryDelayMillis(1, null)).isEqualTo(2000L);
+        assertThat(WatsonxChatModelConnection.retryDelayMillis(10, null)).isEqualTo(10_000L);
+        // Retry-After wins when larger, is capped, and non-numeric values are ignored
+        assertThat(WatsonxChatModelConnection.retryDelayMillis(0, "5")).isEqualTo(5000L);
+        assertThat(WatsonxChatModelConnection.retryDelayMillis(0, "600")).isEqualTo(30_000L);
+        assertThat(WatsonxChatModelConnection.retryDelayMillis(0, "not-a-number")).isEqualTo(1000L);
+    }
```
weiqingy, line 559: `getBearerToken` (`WatsonxChatModelConnection.java:635-677`) has no test in either test class, and #902's rationale for a separate module is exactly that watsonx needs IAM token caching and refresh. Nothing verifies that a second `chat()` reuses the cached token, that it's re-fetched past the 60s margin, and so on. `sendWithRetry` (`:290-325`) is untested as a loop: the tests at `:316-337` cover only `isRetryableStatus` and `retryDelayMillis`, so nothing checks that a 503 is retried or a 400 isn't. `extractReasoning` (`:348-377`) is untested too, and as a pure `@VisibleForTesting` static it's the cheapest to cover.

None of this needs a production seam: `url` and `iam_url` are ordinary connection arguments, so a test can point both at a local stub. The repo has that fixture already, `com.sun.net.httpserver.HttpServer` on port 0 at `runtime/src/test/java/org/apache/flink/agents/runtime/skill/URLSkillRepositoryTest.java:64-66`, JDK-provided with no new test dependency. One covers the IAM cache, the retry loop and the error branches at once.

Is a stub server along those lines workable here, or does something about this setup rule it out?

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
```diff
@@ -0,0 +1,476 @@
...
+                self._http_client.close()
+            self._http_client = None
+
+    def _get_model(self, model: str) -> ModelInference:
+        if model not in self._models:
+            self._models[model] = ModelInference(
+                model_id=model,
+                api_client=self.client,
+                project_id=self.project_id,
+                space_id=self.space_id,
```
wenjin272, line 299: Could we disable the SDK retry here by passing `max_retries=0` to `ModelInference`?

`ModelInference` retries 10 times by default, while `_chat_with_retry` adds another retry layer. As a result, the connection's `max_retries=0` still performs retries, and the default settings can multiply requests.

Keeping `_chat_with_retry` as the single connection-level retry also preserves transport-error handling and Java parity.

integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
```diff
@@ -0,0 +1,678 @@
...
+                    "extract_reasoning",
+                    "additional_kwargs");
+    private static final Set<String> RESERVED_ADDITIONAL_KWARGS =
+            Set.of(
+                    "model",
+                    "temperature",
+                    "max_tokens",
+                    "extract_reasoning",
+                    "tool_choice",
+                    "tool_choice_option");
```
wenjin272, line 86: Could we also add `model_id`, `messages`, `tools`, `project_id`, and `space_id` to `RESERVED_ADDITIONAL_KWARGS` in both Java and Python?

Currently, `additional_kwargs` can overwrite framework-managed fields such as the selected model and message history, or cause both `project_id` and `space_id` to be sent. It also produces different `tools` behavior between Java and Python.

wenjin272 (commented):
Thanks for taking this on @cansuk99. Overall looks good to me. I left two minor comments.

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
cansuk99: Hi @weiqingy Thanks for this! I rebased onto the latest main and added explicit `output_schema` handling to the Watsonx.ai connector, following the same pattern as connectors such as Anthropic. Native Watsonx.ai `output_schema` translation can be implemented in a follow-up PR.

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
cansuk99: Hey @weiqingy  he SDK retry layer was not accounted for in the initial implementation - thank you for catching this! I have disabled SDK retries by passing` max_retries=0` to `ModelInference` and kept `_chat_with_retry` as the single retry owner. Here I optimized for behavioural parity between the Java and Python implementation, as otherwise if we prioritize the SDK built-in retry which does not exist in Java, the connectors would differ.

Does this approach make sense to you, or would you prefer delegating retries to the native SDK in Python while retaining the connector-level implementation in Java?

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
cansuk99: Hey @wenjin272 Thank you for the suggestion - I implemented `max_retries=0` and added some additional reasoning behind here https://github.com/apache/flink-agents/pull/922#discussion_r3700357873 as @weiqingy  pointed this out as well.

integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
cansuk99: I have extended `RESERVED_ADDITIONAL_KWARGS` with `model_id, messages, tools, project_id, and space_id` for both Java and Python implementations

integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
cansuk99: I have addressed this as per @wenjin272 comment https://github.com/apache/flink-agents/pull/922#discussion_r3698436566

integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
cansuk99: I followed your suggestion and implemented a one-time IAM token refresh for `401` and `403` responses when `api_key` authentication is configured. The connector invalidates the rejected cached token, obtains a fresh token, rebuilds the request with the new Authorization header and retries once.

integrations/chat-models/watsonx/src/test/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnectionTest.java
cansuk99: The proposed stub server approach works. I used JDK HttpServer on `127.0.0.1` with an ephemeral port to stub both` /identity/token` and` /ml/v1/text/chat`. It now covers IAM token reuse and refresh, `401` and `403` recovery, retrying `503`, not retrying `400`, and the relevant response paths without external services or a new dependency.

cansuk99:
Hey @weiqingy @wenjin272 , the `Flink Agents CI / it-java [ubuntu-latest] [java-21] [flink-2.1]` check appears to have failed in `TokenMetricsE2ETest`, which does the OpenAI mock connector and is outside the Watsonx.ai changes in this PR. Could one of you please rerun the failed job? The failure may be transient and a re-run shall suffice.

I don’t have permission to rerun GitHub jobs on this repository without having to push another commit and re-run all jobs.

wenjin272 (approved):
LGTM. No further comments from my side. Let's wait for @weiqingy  to take a look.

integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
```diff
@@ -0,0 +1,701 @@
...
+                found = true;
+            }
+            if (found) {
+                rest.append(cleaned, position, cleaned.length());
+                cleaned = rest.toString();
+            }
+        }
+        final String reasoning =
+                reasoningChunks.isEmpty() ? null : String.join("\n\n", reasoningChunks);
+        cleaned = cleaned.replaceAll("\\n{3,}", "\n\n").replaceAll(" {2,}", " ").trim();
```
weiqingy, line 395: The test added this round at `WatsonxChatModelConnectionTest.java:508-515` is named "Reasoning blocks are extracted without changing plain content", but that property doesn't hold. This line ends `extractReasoning` with `cleaned.replaceAll("\\n{3,}", "\n\n").replaceAll(" {2,}", " ").trim()`, and it runs every time, including when no reasoning pattern matched at all. The test only passes because its sample `"Plain answer"` has no double space in it.

At head, `"| 1  | 2  |"` comes back as `"| 1 | 2 |"`, and a four-space indent collapses to one. `chat()` applies this to the whole response whenever `extract_reasoning` is set (`:273-279`), so content the extractor never touched gets reformatted too.

The scope is narrow, since `extract_reasoning` defaults to `false` (`WatsonxChatModelSetup.java:56`). Could the cleanup be limited to where a reasoning block was actually removed, so the rest stays as the model wrote it?

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
```diff
@@ -0,0 +1,491 @@
...
+            else None
+        )
+
+        response = self._chat_with_retry(
+            model_name,
+            messages=convert_to_watsonx_messages(messages),
+            tools=tool_specs,
+            tool_choice=tool_choice,
+            tool_choice_option=tool_choice_option,
+            params={**kwargs, **additional_kwargs} or None,
```
weiqingy: Two things about this merge, and neither is reachable today, which is what makes them cheap to settle now.

The precedence runs opposite in the two languages. Here `additional_kwargs` expands last, so it wins. Java spills `additional_kwargs` first (`WatsonxChatModelConnection.java:426-431`), then puts every non-control `modelParams` entry on top (`:434-438`), so there the per-call value wins. With `additional_kwargs={"top_p": 0.9}` and a caller passing `top_p=0.5`, Python sends `0.9` and Java sends `0.5` from the same configuration. I reproduced both at head.

The new guard also covers only half of this dict. It checks `additional_kwargs` but not the sibling `**kwargs`, so `chat(..., model_id="X")` still reaches `params` and quietly replaces the model, since the SDK does `payload.update(parameters)` after seeding `model_id`.

Nothing supplies per-call model parameters today: `ChatModelAction.java:365` passes `Map.of()` and `chat_model_action.py:320-325` passes none, both on `origin/main`. Which precedence did you have in mind? If per-call should win, reversing the order here would line Python up with Java, and running the guard over the merged dict would close the `**kwargs` half at the same time. The Java spill loop has the same unguarded shape (`CONTROL_PARAMS` at `:65-71`), so it may be worth a look in the same pass.

weiqingy (commented):
Thanks for resolving the comments. Just a few small ones left from me. Nothing here blocks.

nit, and the only thing here that is not inline: the provider matrix in `docs/content/docs/faq/faq.md:98-107` lists eight chat-model providers and has no watsonx row. That table has a `Python` and a `Java` column, and it is alphabetical, so a watsonx row would sit right after Tongyi with a check in both columns, since this ships both languages. For the pairing precedent, `3cb28967` ("[docs][integration] Document Gemini chat model integration (#898)") touched exactly two files, `chat_models.md` and `faq.md`, the second being just the new row.

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
weiqingy: Connector-level in both makes sense to me, for the reason you gave. It keeps the two languages on one retry set and one `Retry-After` policy.

I checked that `max_retries=0` really does switch the SDK's own retry off, and it does on every version in the range pinned at `python/pyproject.toml:59-60`. Worth checking on more than the newest, since that code was reworked between releases.

The one thing still open is that nothing tests it. `test_watsonx_chat_model.py` replaces `_get_model` outright (`:89-91`, `:146-148`, `:184-186`), so the `ModelInference` constructor is never seen, and deleting `max_retries=0` would leave the suite green. The five new reserved keys sit in the same spot, since `test_configuration_contract` (`:355-360`) only exercises `temperature`.

Would asserting the constructor kwargs in that mocked test be enough to pin both, or would you rather keep this for a follow-up?

integrations/chat-models/watsonx/src/test/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnectionTest.java
weiqingy: The stub-server approach is right, and most of what you listed holds up when I ran it.

The 401/403 test is the one gap: it can't fail on the bug it exists to catch. The chat handler only counts requests and never looks at the `Authorization` header (`WatsonxChatModelConnectionTest.java:444-452`), so it catches an invalidate that does nothing, but not a retry that re-sends the token the server just rejected. I checked by breaking it on purpose. Changing `WatsonxChatModelConnection.java:261` from `bearerToken = getBearerToken();` to `getBearerToken();` leaves all 20 tests green.

Would recording the header on each request, then asserting the first was `Bearer token-1` and the second `Bearer token-2`, close that?

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
cansuk99: Agreed. I updated the existing mocked chat test to exercise the real `_get_model()` path and assert that ModelInference is constructed with `max_retries=0.` I also added parameterized coverage for `model_id, messages, tools, project_id,` and `space_id` in `additional_kwargs`.

integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
cansuk99: Thanks for pointing this out! I limited whitespace cleanup to cases where at least one reasoning block was actually extracted. When no pattern matches, the original content is now returned unchanged.

python/flink_agents/integrations/chat_models/watsonx/watsonx_chat_model.py
cansuk99: I intended per-call parameters to take precedence over the static `additional_kwargs`. I updated the Python merge order to match the existing Java behavior and added tests to make this precedence explicit.

python/flink_agents/api/chat_models/chat_model.py
```diff
@@ -216,7 +216,10 @@ def _extract_reasoning(
                 reasoning_chunks.extend(m.strip() for m in matches if m.strip())
                 cleaned = pat.sub("", cleaned)
 
-        reasoning = "\n\n".join(reasoning_chunks) if reasoning_chunks else None
+        if not reasoning_chunks:
+            return content, None
```
weiqingy, line 220: Returning `content` here means an empty or whitespace-only block keeps its tags: `<think></think>Answer` comes back unchanged at head, where before it returned `Answer`. Being on the shared base, this reaches `OllamaChatModelConnection` too (`ollama_chat_model.py:138`).

Would `return cleaned, None` work? `cleaned` still equals `content` when nothing matched, so the no-tag case you're protecting is untouched.

integrations/chat-models/watsonx/src/main/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnection.java
weiqingy: The no-pattern case is fixed. One case still slips through though: the early return hands back `content` rather than `cleaned`, so an empty or whitespace-only block keeps its tags. `<think></think>Answer` returns `<think></think>Answer` at head, where before it was `Answer`. I ran both versions side by side to check.

Would returning `cleaned` work?

```java
if (reasoningChunks.isEmpty()) {
    return new String[] {cleaned, null};
}
```

`cleaned` still equals `content` when nothing matched, so the case you just fixed is unaffected.

The assertion at `:531-533` only covers the no-tag path. Worth an empty-block case alongside it?

integrations/chat-models/watsonx/src/test/java/org/apache/flink/agents/integrations/chatmodels/watsonx/WatsonxChatModelConnectionTest.java
weiqingy: Checking back on this one: it looks unchanged at head. The chat handler at `WatsonxChatModelConnectionTest.java:462-470` still only counts requests, so it can't tell a refreshed token from a re-sent one. Maybe it's small enough to fold in here. wdyt?

wenjin272:
Hi @cansuk99, Thanks for addressing the comments. I noticed that there is still one remaining comment from weiqingy that has not been resolved, and the PR also has some conflicts with the main branch.

I was wondering if you have some time recently to take care of these. If you are busy with other tasks, I can help handle them on your behalf.

python/flink_agents/api/chat_models/chat_model.py
cansuk99: Yes,`return cleaned, None` works here, I have implemented it - thanks for catching this!

cansuk99:
> Hi @cansuk99, Thanks for addressing the comments. I noticed that there is still one remaining comment from weiqingy that has not been resolved, and the PR also has some conflicts with the main branch.
> 
> I was wondering if you have some time recently to take care of these. If you are busy with other tasks, I can help handle them on your behalf.

Hey @wenjin272 sorry I could not catch up earlier with the PR. I addressed weiqingy's comment and rebased the PR so there are no merge conflicts. Thanks for the offering and the heads up on this one!

python/flink_agents/api/chat_models/chat_model.py
weiqingy: That one is on the shared Python base, so watsonx Python inherits it. The Java side has its own copy though, and `WatsonxChatModelConnection.java:392` still returns `content`, so the connector now differs by language: `<think></think>Answer` gives `Answer` on Python and `<think></think>Answer` on Java. I ran both at `4e03d50f`.

Does the same one-line change work there? And would an empty-block case fit alongside the assertions in `testExtractReasoning`?

wenjin272 (approved):

PR #923, opened 2026-07-23
[api][runtime][python] Add minimal Event lineage

### LINKED ISSUE #841: [Feature] Add lineage fields to the event log for per-run trace reconstruction

rosemarYuan:
Linked issue: Closes #841

Related discussions:

- https://github.com/apache/flink-agents/discussions/710
- https://github.com/apache/flink-agents/discussions/900

### Purpose of change

The Event Log currently records individual Events but does not preserve enough information to reconstruct the direct causal path between an Event, the Action it triggered, and the Events emitted by that Action.

This PR implements the minimal Event-lineage phase discussed in #710.

It adds two framework-managed fields to each Event emitted by an Action:

- `upstreamEventId`: the ID of the Event consumed by the emitting Action;
- `upstreamActionName`: the name of the Action that emitted the current Event.

Together with the existing per-occurrence Event ID, these fields are sufficient to reconstruct an InputEvent-rooted Event-Action-Event causal tree from the available Event Log records.

This PR deliberately does not implement the complete execution-tracing model proposed in #900. Run identity, concrete Action invocation identity, execution hierarchy, lifecycle, retry/replay semantics, ordering, and failure details remain part of that separate design.

A real Runtime execution was verified with the following causal path:

```text
InputEvent(input=1)
  -> action1
  -> MiddleEvent(num=2)
  -> action2
  -> OutputEvent(output=4)
```

The File Event Log contained three physical Event records:

| Event           | `upstreamEventId` | `upstreamActionName` |
| --------------- | ----------------- | -------------------- |
| `_input_event`  | absent            | absent               |
| `MiddleEvent`   | InputEvent ID     | `action1`            |
| `_output_event` | MiddleEvent ID    | `action2`            |

Running `tools/reconstruct_trace_tree.py` against that Runtime log produced:

```text
Trace Tree 1
  _input_event
    [Action: action1]
      MiddleEvent
        [Action: action2]
          _output_event
```

The JSON reconstruction contained one root, all three Event nodes, the two expected virtual Action edges, and no warnings.

### 

### Tests

Add e2e test to review Trace Tree `python/flink_agents/e2e_tests/e2e_tests_integration/python_event_logging_test.py`

### API

This PR changes the default Python Event ID generation strategy from content-derived IDs to UUIDv4.

This is also a breaking Python API change: `Event.id` is now immutable. Custom `Event.from_event()` implementations that use `result.id = event.id` must instead return `result.reconstruct_from(event)`.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/Event.java
+++ b/api/src/main/java/org/apache/flink/agents/api/Event.java
@@ -20,14 +20,18 @@ package org.apache.flink.agents.api;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.ObjectMapper;
+
+import javax.annotation.Nullable;
 
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
+import java.util.function.BiFunction;
 
 /** Base class for all event types in the system. */
 public class Event {
@@ -37,6 +41,9 @@ public class Event {
     private final UUID id;
     private final String type;
     private final Map<String, Object> attributes;
+
+    @Nullable private UUID upstreamEventId;
+    @Nullable private String upstreamActionName;
 
     /**
      * Runtime-internal timestamp from the source record. Not part of the cross-language event
@@ -54,17 +61,38 @@ public class Event {
         this(type, new HashMap<>());
     }
 
+    /**
+     * Reconstructs an Event with an existing identity and optional framework-managed lineage.
+     *
+     * <p>The lineage values support deserialization and reconstruction. When an Action emits this
+     * Event, the runtime overwrites them with the current trigger Event ID and Action name.
+     *
+     * @param id the existing Event ID
+     * @param type the Event type used for routing
+     * @param attributes the Event payload
+     * @param upstreamEventId the ID of the direct upstream Event, or {@code null}
+     * @param upstreamActionName the name of the emitting Action, or {@code null}
+     */
     @JsonCreator
     public Event(
             @JsonProperty("id") UUID id,
             @JsonProperty("type") String type,
-            @JsonProperty("attributes") Map<String, Object> attributes) {
+            @JsonProperty("attributes") Map<String, Object> attributes,
+            @JsonProperty("upstreamEventId") @Nullable UUID upstreamEventId,
+            @JsonProperty("upstreamActionName") @Nullable String upstreamActionName) {
         if (type == null || type.isEmpty()) {
             throw new IllegalArgumentException("Event 'type' must not be null or empty.");
         }
         this.id = id;
         this.type = type;
         this.attributes = attributes != null ? attributes : new HashMap<>();
+        this.upstreamEventId = upstreamEventId;
+        this.upstreamActionName = upstreamActionName;
+    }
+
+    /** Reconstructs an Event with an existing identity and no upstream lineage. */
+    public Event(UUID id, String type, Map<String, Object> attributes) {
+        this(id, type, attributes, null, null);
     }
 
     public UUID getId() {
@@ -79,6 +107,38 @@ public class Event {
 
     public Map<String, Object> getAttributes() {
         return attributes;
+    }
+
+    /** Returns the ID of the Event consumed by the Action that emitted this Event. */
+    @Nullable
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public UUID getUpstreamEventId() {
+        return upstreamEventId;
+    }
+
+    /**
+     * Sets the framework-managed ID of the Event consumed by the emitting Action.
+     *
+     * <p>The runtime overwrites this value when an Action emits the Event.
+     */
+    public void setUpstreamEventId(@Nullable UUID upstreamEventId) {
+        this.upstreamEventId = upstreamEventId;
+    }
+
+    /** Returns the name of the Action that emitted this Event. */
+    @Nullable
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public String getUpstreamActionName() {
+        return upstreamActionName;
+    }
+
+    /**
+     * Sets the framework-managed name of the Action that emitted this Event.
+     *
+     * <p>The runtime overwrites this value when an Action emits the Event.
+     */
+    public void setUpstreamActionName(@Nullable String upstreamActionName) {
+        this.upstreamActionName = upstreamActionName;
     }
 
     public Object getAttr(String name) {
@@ -105,16 +165,38 @@ public class Event {
     }
 
     /**
-     * Creates a base Event from another Event, copying id, type, and attributes. Subclasses
-     * override this to reconstruct typed event objects with proper field deserialization.
+     * Creates a base Event from another Event, copying its identity, data, and framework metadata.
+     * Subclasses override this to reconstruct typed event objects with proper field
+     * deserialization.
      */
     public static Event fromEvent(Event event) {
-        Event copy =
-                new Event(event.getId(), event.getType(), new HashMap<>(event.getAttributes()));
-        if (event.hasSourceTimestamp()) {
-            copy.setSourceTimestamp(event.getSourceTimestamp());
+        return reconstructFrom(
+                event, (id, attributes) -> new Event(id, event.getType(), attributes));
+    }
+
+    /**
+     * Reconstructs a typed Event while preserving the source identity and framework metadata. The
+     * factory receives the source ID and a copy of its attributes.
+     */
+    protected static <T extends Event> T reconstructFrom(
+            Event source, BiFunction<UUID, Map<String, Object>, T> factory) {
+        Objects.requireNonNull(source, "source Event must not be null");
+        Objects.requireNonNull(factory, "Event reconstruction factory must not be null");
+
+        T reconstructed =
+                Objects.requireNonNull(
+                        factory.apply(source.getId(), new HashMap<>(source.getAttributes())),
+                        "Event reconstruction factory must not return null");
+        if (!Objects.equals(source.getId(), reconstructed.getId())) {
+            throw new IllegalStateException(
+                    "Reconstructing the same Event occurrence must preserve Event ID "
+                            + source.getId());
         }
-        return copy;
+        Event reconstructedEvent = reconstructed;
+        reconstructedEvent.sourceTimestamp = source.sourceTimestamp;
+        reconstructedEvent.upstreamEventId = source.upstreamEventId;
+        reconstructedEvent.upstreamActionName = source.upstreamActionName;
+        return reconstructed;
     }
 
     /**
--- a/docs/content/docs/operations/monitoring.md
+++ b/docs/content/docs/operations/monitoring.md
@@ -288,11 +288,11 @@ records with distinct upstream edges, as in the example above. Reusing an Event 
 Event type and content is valid.
 {{< /hint >}}
 
-If records carrying the same Event ID disagree on the Event type or content, the reader reports
-`EVENT_ID_CONFLICT` and uses the first observation's Event type and content as canonical for the
-node. Observations matching the canonical Event contribute their distinct upstream edges.
-Conflicting observations still contribute to `observationCount`, but not to `upstreamEdges`.
-Keeping the canonical node preserves descendants that refer to the conflicting Event ID.
+When observations with the same Event ID disagree on Event type or content, only observations
+matching the first add distinct entries to `upstreamEdges`, while all observations count toward
+`observationCount`. For conflicting observations, the reader emits one `EVENT_ID_CONFLICT`
+warning per distinct (`upstreamEventId`, `upstreamActionName`) pair and includes each field when
+present. The resulting canonical node and its descendants remain in the reconstructed trace.
 
 The reader derives virtual Action nodes from `upstreamActionName`; they are not separate Event Log
 records. Log order does not add execution-order semantics, but it controls display order and
--- a/docs/content/docs/operations/monitoring.md
+++ b/docs/content/docs/operations/monitoring.md
@@ -184,13 +184,124 @@ Each record contains a top-level `timestamp`, the resolved `logLevel`, and a top
 {
   "timestamp": "2024-01-15T10:30:00Z",
   "logLevel": "STANDARD",
-  "eventType": "_input_event",
+  "eventType": "MiddleEvent",
   "event": {
-    "eventType": "_input_event",
-    "...": "..."
+    "eventType": "MiddleEvent",
+    "id": "39361629-4f1d-4b62-b734-a48b181cb6e0",
+    "attributes": {},
+    "type": "MiddleEvent",
+    "upstreamEventId": "dad5ed00-80e2-4746-8bdb-f126bad504b5",
+    "upstreamActionName": "action1"
   }
 }
 ```
+
+`upstreamEventId` identifies the Event consumed by the Action that emitted the current Event, and `upstreamActionName` identifies that Action. The framework maintains both fields directly on the `event` object, outside business `attributes`. A root `InputEvent` omits both fields.
+
+### Trace Tree Reconstruction
+
+The `flink-agents-trace-tree` command is installed with the Flink Agents Python wheel. It rebuilds InputEvent-rooted Trace Trees from a saved File Event Log. Pass either one log file for text output or a log directory for Trace Tree JSON:
+
+```bash
+flink-agents-trace-tree /path/to/events-job-task-0.log
+flink-agents-trace-tree /path/to/event-log-directory --format json
+```
+
+For example, the text output for an `InputEvent -> MiddleEvent -> OutputEvent` lineage is:
+
+```text
+Trace Tree 1
+  _input_event (dad5ed00-80e2-4746-8bdb-f126bad504b5)
+    [Action: action1]
+      MiddleEvent (39361629-4f1d-4b62-b734-a48b181cb6e0)
+        [Action: action2]
+          _output_event (f452ce08-c672-4c9c-841f-d81d15a900c5)
+```
+
+In JSON output:
+
+- `roots` lists the Event IDs of the root `InputEvent`s, one per reconstructed Trace Tree.
+- `nodes` maps each logical Event ID to one Event node with the following fields:
+  - `eventId`: the logical Event ID of the node.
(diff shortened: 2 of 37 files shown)


joeyutong:
One ActionStateStore replay semantic seems worth making explicit before merging. When a completed action result is reused, the runtime returns the stored output Events and rebinds their lineage to the current triggering Event while preserving the output Event IDs. The resulting behavior is:

```text
first execution: E1 -> A -> O1
reused execution: E2 -> A(reused) -> O1
```

I think this is a reasonable semantic, but it is currently implicit. Could we document it and add a test that asserts `O1.id` is preserved while its `upstreamEventId` changes from `E1` to `E2`? This also establishes that the same logical Event ID may appear in multiple Event Log records across reuse, which reconstruction needs to tolerate.

tools/reconstruct_trace_tree.py
```diff
@@ -0,0 +1,203 @@
...
+def build_trace_forest(records: Iterator[dict[str, Any]]) -> dict[str, Any]:
+    """Build valid InputEvent trees while retaining auditable invalid nodes."""
+    records_by_id: dict[str, list[dict[str, Any]]] = defaultdict(list)
+    for record in records:
+        records_by_id[record["eventId"]].append(record)
+
+    warnings: list[dict[str, str]] = []
+    nodes: dict[str, dict[str, Any]] = {}
+    for event_id, matching_records in records_by_id.items():
+        if len(matching_records) > 1:
```
joeyutong: ActionStateStore reuse can legitimately log the same Event ID again, possibly with different upstream lineage. Treating every repeated ID as invalid and dropping all matching records loses valid replay branches. Could reconstruction model this as a DAG instead: deduplicate repeated `(event ID, lineage edge)` observations, retain distinct lineage edges for the same Event ID, and report an ID conflict only when the Event type or content is inconsistent?

api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -105,18 +135,24 @@ public void setSourceTimestamp(long timestamp) {
...
                 new Event(event.getId(), event.getType(), new HashMap<>(event.getAttributes()));
-        if (event.hasSourceTimestamp()) {
-            copy.setSourceTimestamp(event.getSourceTimestamp());
-        }
+        copy.copyFrameworkMetadataFrom(event);
         return copy;
     }
 
+    /** Copies framework-managed metadata when reconstructing a typed Event. */
+    protected void copyFrameworkMetadataFrom(Event event) {
```
joeyutong: Typed reconstruction represents the same Event occurrence, but preserving that occurrence currently depends on every custom `fromEvent` implementation remembering both to pass the ID separately and to call this helper. Could the framework own a reconstruction template/factory so a custom subtype only maps its payload? Otherwise an omitted helper call silently loses `sourceTimestamp` and lineage.

python/flink_agents/api/events/event.py
```diff
@@ -96,35 +115,44 @@ def model_dump_json(self, **kwargs: Any) -> str:
...
 
     def __setattr__(self, name: str, value: Any) -> None:
         super().__setattr__(name, value)
         # Ensure added property can be serialized.
         self.model_dump_json()
-        # Regenerate ID if content changed (but not if setting 'id' itself)
-        if name != "id":
-            object.__setattr__(self, "id", self._generate_content_based_id())
+
+    def with_framework_metadata_from(self, source: "Event") -> Self:
```
joeyutong: The Java and Python helpers currently have different contracts: Python returns a copy and includes the Event ID, while Java mutates the current object and does not copy the ID. Also, “framework metadata” describes ownership rather than the Event-domain meaning. Could the names and contracts be aligned around preserving/reconstructing the same Event occurrence?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -454,6 +456,13 @@ private void processActionTaskForKey(Object key) throws Exception {
         }
     }
 
+    private static void setOutputEventLineage(ActionTask actionTask, List<Event> outputEvents) {
```
joeyutong: This rule describes how a concrete Action invocation finalizes its output Events, rather than Operator scheduling. Could it be owned by `ActionTask`, which already holds the triggering Event and Action? That would keep the Operator independent of lineage field details and provide one place to reject `outputEvent.id == triggerEvent.id`; currently `ctx.sendEvent(event)` can create a self-loop and mutate the triggering Event shared by sibling ActionTasks.

python/flink_agents/api/events/event.py
```diff
@@ -96,35 +115,44 @@ def model_dump_json(self, **kwargs: Any) -> str:
         # Set fallback if not provided in kwargs
         if "fallback" not in kwargs:
             kwargs["fallback"] = self.__serialize_unknown
+        if "by_alias" not in kwargs:
```
joeyutong: Defaulting `by_alias` to `True` here changes serialization for every alias declared by a custom Event subclass, not only the two lineage fields. Could the camelCase mapping be scoped to the lineage fields, or should this broader compatibility change be documented and covered by a custom-alias test?

docs/content/docs/operations/monitoring.md
```diff
@@ -184,14 +184,44 @@ Each record contains a top-level `timestamp`, the resolved `logLevel`, and a top
...
+    "upstreamActionName": "action1"
   }
 }
 ```
 
+`upstreamEventId` identifies the Event consumed by the Action that emitted the current Event, and `upstreamActionName` identifies that Action. The framework maintains both fields directly on the `event` object, outside business `attributes`. A root `InputEvent` omits both fields.
+
+### Trace Tree Reconstruction
+
+The local reader rebuilds InputEvent-rooted Trace Trees from a saved File Event Log. From the repository root, pass either one log file for text output or a log directory for Trace Tree JSON:
```
weiqingy: This is an operations page, so the audience is people running jobs from an installed wheel, but `tools/reconstruct_trace_tree.py` only exists in a git checkout. `python/pyproject.toml` packages `flink_agents` only and has no `[project.scripts]` block, so nothing installs the reader or puts it on `PATH`. Everything else in `tools/` is a build or dev script (`build.sh`, `lint.sh`, `ut.sh`, and so on), and this is the first user-facing program to land there.

The placement has two side effects worth weighing at the same time. `tools/lint.sh` points ruff at `python/` only (lines 89-113), so the new 203-line file gets no ruff rules, no 88-column limit and no docstring checks, which are the Python standards `AGENTS.md` states for this repo. And `python/flink_agents/api/tests/test_trace_tree_reader.py:19-20` has to walk up to `parents[4]` and `subprocess.run` a repo-root script from inside the `api` module.

Should the reader live under `python/flink_agents/` with a `[project.scripts]` entry (something like `flink-agents-trace-tree`), or would you rather drop the docs pointer until it is packaged?

tools/reconstruct_trace_tree.py
```diff
@@ -0,0 +1,203 @@
...
+    """Read consecutive JSON objects separated by whitespace."""
+    content = path.read_text(encoding="utf-8")
+    decoder = json.JSONDecoder()
+    position = 0
+    while position < len(content):
+        while position < len(content) and content[position].isspace():
+            position += 1
+        if position == len(content):
+            return
+        record, position = decoder.raw_decode(content, position)
```
weiqingy: The reader has a five-code warning framework and `monitoring.md:223` promises that valid InputEvent-rooted trees are retained while warnings go to stderr, but there is no `try`/`except` anywhere in the file. One malformed record makes `raw_decode` raise `json.JSONDecodeError` and the whole run aborts, so every valid tree in every file is lost, not just the bad record. That is reachable while a job is still running: `FileEventLogger.append` writes with `writer.println(json)` and exposes `flush()` as a separate method, so a log read mid-run can genuinely end mid-record.

There is a second path at line 59. `event = record["event"]` raises a bare `KeyError` on any JSON object that is not an `EventLogRecord`, and the fallback glob at line 33 (`return log_files or sorted(path.glob("*.log"))`) sweeps in any `.log` file in the directory when no `events-*.log` matches.

Could the per-file and per-record decoding be wrapped and routed through the existing `warning()` helper, as something like `MALFORMED_RECORD` and `UNREADABLE_FILE`?

runtime/src/test/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperatorTest.java
```diff
@@ -748,6 +781,16 @@ agentPlanWithStateStore, true, new InMemoryActionStateStore(false)),
             testHarness.processElement(new StreamRecord<>(inputValue));
             operator.waitInFlightEventsFinished();
         }
+
+        for (Map<String, ActionState> states : actionStateStore.getKeyedActionStates().values()) {
```
weiqingy: This pre-null loop manufactures a state the runtime can no longer produce. `setOutputEventLineage` now runs before `maybePersistTaskResult` (which is why `outputEvents = actionTaskResult.getOutputEvents();` moved up), and `testCompletedActionStatePersistsOutputEventLineage` asserts that persisted states always carry lineage. So the only way to reach null lineage in a stored state is to write it by hand here.

The assertion at line 827 is `isNotNull()`. What the re-stamp on the reuse branch actually guards is that the replayed trigger event is a different object with a different UUID, so a value check is what would catch a regression: as written, the test still passes if the re-stamp writes the wrong id. Asserting that the re-stamped `upstreamEventId` equals the id of the InputEvent processed in the second harness would pin the behavior that matters.

This is also riding on `testActionStateStoreReplayIncurNoFunctionCall`, whose stated concern is that functions are not re-executed. Would a separate test for the replay re-stamp, without the pre-nulling, read more clearly?

api/src/main/java/org/apache/flink/agents/api/Event.java
```diff
@@ -81,6 +87,30 @@ public Map<String, Object> getAttributes() {
...
 
+    /** Returns the ID of the Event consumed by the Action that emitted this Event. */
+    @Nullable
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public UUID getUpstreamEventId() {
+        return upstreamEventId;
+    }
+
+    /** Sets the ID of the Event consumed by the Action that emitted this Event. */
+    public void setUpstreamEventId(@Nullable UUID upstreamEventId) {
```
weiqingy, line 124: Python declares both fields with `AliasChoices` (`event.py:89-98`), so they are ordinary constructor kwargs, and the PR's own `test_typed_from_event_preserves_lineage` builds a fully-formed Event in a single `Event(type=..., attributes=..., upstreamEventId=..., upstreamActionName=...)` call.

On the Java side the `@JsonCreator` constructor stays `(id, type, attributes)`, so these setters are the only way in, which is why `EventTest` has to do `new Event("ChildEvent")` followed by two `set...` calls. `AGENTS.md` asks that public API changes keep Java and Python semantically aligned. This is the construction surface specifically, separate from the reconstruction-helper contract already under discussion on this PR.

Which way do you see this settling: a lineage-carrying constructor on the Java side, or keeping both sides framework-write-only and treating the Python kwargs as incidental to pydantic?

rosemarYuan: Thanks @joeyutong and @weiqingy for the careful reviews. I’ve addressed the feedback locally in three latest commits. The replay behavior is now documented and tested. A reused output Event keeps its ID, but its lineage is updated to the current trigger. For Event reconstruction, custom Events in Java now use` Event.reconstructFrom(source, factory)`, while custom Events in Python use`reconstruct_from(source)`. These helpers preserve the Event ID and lineage.

The Trace Tree reader accepts the same Event ID with different valid parent links when its type and content remain unchanged. A separate test checks the exact Event IDs and lineage before and after replay. I also made the Trace Tree reader available from the installed Python package, improved malformed log handling, and added the related tests. 

Thanks again for the reviews.

python/flink_agents/cli/trace_tree.py
```diff
@@ -0,0 +1,333 @@
...
+    if column_number is not None:
+        item["columnNumber"] = column_number
+    return item
+
+
+def render_text(trace_forest: dict[str, Any]) -> str:
+    """Render valid Trace Trees without assigning meaning to sibling order."""
+    lines: list[str] = []
+
+    def render_event(event_id: str, indent: str) -> None:
```
joeyutong, line 410: The reader intentionally accepts the same Event ID with multiple lineage edges, but that also means the resulting graph is not guaranteed to be acyclic. For example, observations `root -> A`, `A -> B`, and `B -> A` pass the current validation. `render_event` has no active-path or edge guard, so text rendering keeps pushing A and B indefinitely. This is broader than the direct self-loop case raised earlier. Could the reader detect a cyclic edge (or track the active render path), emit a warning, and stop only that branch while retaining valid DAG reuse?

python/flink_agents/api/events/event.py
```diff
@@ -66,17 +72,30 @@ class Event(BaseModel, extra="allow"):
...
     attributes : Dict[str, Any]
         Key-value properties for the event data.
+    upstream_event_id : UUID | None
+        The ID of the direct upstream Event, or None.
+    upstream_action_name : str | None
+        The name of the emitting Action, or None.
     """
 
-    id: UUID = Field(default=None)
+    id: UUID = Field(default_factory=uuid4, frozen=True)
```
joeyutong, line 86: `frozen=True` makes `Event.id` immutable, so direct reassignment now raises a `ValidationError`. On current `main`, the built-in `from_event` methods and workflow documentation use `result.id = event.id`; this PR migrates those call sites, but external custom Event subclasses following the existing guidance will break. Since immutability is a user-visible compatibility change separate from switching ID generation to UUIDv4, could it be called out explicitly in the PR description or a migration note, with `reconstruct_from()` as the replacement?

python/flink_agents/cli/trace_tree.py
```diff
@@ -0,0 +1,333 @@
...
+        ):
+            warnings.append(
+                warning(
+                    "EVENT_ID_CONFLICT",
+                    event_id,
+                    f"Event ID {event_id} has inconsistent Event type or content "
+                    f"across {len(matching_records)} records.",
+                )
+            )
+            continue
```
weiqingy: `continue` skips building a node for the conflicting id, and every descendant leaves the rendered trees with it, not just the conflicting Event.

Six records in (a root, `K` logged twice with different content, then `C <- K`, `D <- C`, `E <- D`):

```
warnings: [('EVENT_ID_CONFLICT', 'k'), ('MISSING_PARENT', 'c')]
nodes:    ['c', 'd', 'e', 'r']
text:
Trace Tree 1
  _input_event (r)
```

`C` gets a `MISSING_PARENT`, but `D` and `E` get nothing, since they link cleanly to nodes that are themselves unreachable. Four Events drop out of the rendered tree. They stay in `nodes`, but nothing links them to a root, and two warnings name two of them.

The path I can see reaching this: `FileEventLogger` truncates at write time from `event-log.standard.max-string-length` and friends, so a reused Event ID logged before and after a threshold change differs in content and reads as a conflict. That is the cross-restart case the new hint block describes.

Could the conflicting id keep its first observation as the node, so the branch survives and the inconsistency still surfaces as a warning?

rosemarYuan: Thanks @joeyutong and @weiqingy for the review. I have pushed the fixes.

- The Trace Tree reader can now find multi-node cycles. It reports a `CYCLE_DETECTED` warning and removes only the edge that creates the cycle. Other valid branches are kept.
- The document now explains that Python Event IDs cannot be changed. The old `result.id = event.id` code raises a `ValidationError`. Please use `reconstruct_from(event)` instead.
- For `EVENT_ID_CONFLICT`, the reader now keeps the first record. Later conflicting records are counted, but their lineage links are not used. Child Events are no longer lost.

I also added tests for these cases. Please take another look when convenient. Thanks!

weiqingy: Confirmed on the log from this thread: `k` logged twice with different content, then `c <- k`, `d <- c`, `e <- d`. One `EVENT_ID_CONFLICT`, and the full tree renders.

One case I hit while poking at it. When the discarded observation is the only record linking a parent to the conflicting Event, that parent loses its child edge and nothing names it:

```
records: rootA, rootB, k<-rootA(actA, content "full"), k<-rootB(actB, content "trunc"), c<-k

warnings:      [EVENT_ID_CONFLICT k "...across 2 records"]
rootB.actions: []

Trace Tree 2
  _input_event (rootB)
```

`rootB` reads as a root that produced nothing. Reachable via the cross-restart case from earlier: a reused Event ID is re-stamped with the current trigger's lineage, so a record logged either side of an `event-log.standard.max-string-length` change differs in content, and the replay parent is the one that drops.

`CYCLE_DETECTED` names its omitted edge with `upstreamEventId` and `upstreamActionName`, and `warning()` takes both as of this commit. Carry the discarded edge the same way, or keep the warning at the Event level and let `observationCount` versus `upstreamEdges` tell the story?

rosemarYuan: Thanks for the suggestion. I kept the warning at the Event level, but included the excluded observation’s `upstreamEventId` and `upstreamActionName` in the warning.
I also reproduced the six-record case you described locally: a root, `k` logged twice with different content, followed by `c <- k`, `d <- c`, and` e <- d`. With the fix, the result is:

```text
[EVENT_ID_CONFLICT] Event ID k has inconsistent Event type or content across 2 records. Conflicting observation records lineage from r to k through Action create_k; that observation does not contribute lineage to the canonical Event node.

Trace Tree 1
  _input_event (r)
    [Action: create_k]
      KEvent (k)
        [Action: create_c]
          CEvent (c)
            [Action: create_d]
              DEvent (d)
                [Action: create_e]
                  EEvent (e)
```
  
The conflicting` rootB → k `edge is not added to the canonical DAG, but the warning still identifies the excluded edge and its Action.

joeyutong (approved):
LGTM. Thanks for addressing the previous review comments.

python/flink_agents/cli/trace_tree.py
weiqingy: Thanks for addressing the comments. Nothing further from me.

Issue #928, opened 2026-07-26
[Feature] Add VLLM_CONNECTION to the ChatModel out-of-the-box


piosystems:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

[VLLM ](https://vllm.ai/) is a superior choice for local model deployment for production. I believe that support for VLLM ChatModel out-of-the-box will drive adoption of flink-agents.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

Zhuoxi2000:
Thanks for raising this. I’d like to pick it up.

Since vLLM is OpenAI-compatible and our existing OpenAI connections already support a custom `api_base_url`, I think the simplest path could be either:

1. Add a `vllm` YAML alias pointing to the existing OpenAI-compatible setup, or
2. Add a thin dedicated `VLLM_CONNECTION` with vLLM-friendly defaults, like a local default `api_base_url` and optional API key.

I’d lean toward the dedicated connection for discoverability, but happy to go with whichever direction the maintainers prefer. Once the scope is confirmed, I can open a PR.

piosystems:
> Thanks for raising this. I’d like to pick it up.
> 
> Since vLLM is OpenAI-compatible and our existing OpenAI connections already support a custom `api_base_url`, I think the simplest path could be either:
> 
> 1. Add a `vllm` YAML alias pointing to the existing OpenAI-compatible setup, or
> 2. Add a thin dedicated `VLLM_CONNECTION` with vLLM-friendly defaults, like a local default `api_base_url` and optional API key.
> 
> I’d lean toward the dedicated connection for discoverability, but happy to go with whichever direction the maintainers prefer. Once the scope is confirmed, I can open a PR.

I would lean as well toward the dedicated VLLM connection, for discoverability.

Zhuoxi2000:
Thanks @piosystems — agreed on going with a dedicated connection.

The shape I’m planning is:

* **Java**: add `VLLMChatModelConnection` / `VLLMChatModelSetup` under the existing `integrations/chat-models/openai` module, similar to how Azure OpenAI is packaged today. This would reuse the OpenAI-compatible client, default `api_base_url` to `http://localhost:8000/v1`, and make `api_key` optional.
* **Python**: add `flink_agents/integrations/chat_models/vllm/vllm_chat_model.py`, mirroring the existing `chat_models/azure` layout.
[...]

PR #945, opened 2026-08-01
[integration] Add out of the box vLLM chat model support for Java and Python

Zhuoxi2000:
Linked issue: #928

### Purpose of change

Add a dedicated vLLM chat model connection out of the box, in both Java and Python, per the direction agreed on #928 (dedicated connection over alias-only, for discoverability).

vLLM exposes an OpenAI-compatible API, so both connections reuse the existing OpenAI integrations with vLLM-friendly defaults:

- `api_base_url` defaults to `http://localhost:8000/v1` (the default address of `vllm serve`).
- `api_key` defaults to a placeholder: vLLM servers started without `--api-key` require no credential, while both OpenAI SDKs require a non-empty key — an alias-only route would have forced users to pass a dummy key by hand.

One deliberate difference from the OpenAI setups: `model` is **required** and has no default, because a vLLM server only serves the model(s) it was started with.

Implementation notes:

- **Java**: `VLLMChatModelConnection` / `VLLMChatModelSetup` live in the existing `integrations/chat-models/openai` module (same packaging as `AzureOpenAIChatModelConnection`, so no new Maven module or distribution change). The connection injects defaults by rewriting the `ResourceDescriptor` before delegating to `OpenAICompletionsConnection`; the caller's descriptor is not mutated. The setup delegates to `OpenAICompletionsSetup` and only adds the model-required validation.
- **Python**: `flink_agents/integrations/chat_models/vllm/vllm_chat_model.py`, subclassing the OpenAI connection/setup the same way, mirroring the `chat_models/azure` package layout.
- Registered in `ResourceName.ChatModel` (Java + `Python` inner class) and in the Python `ResourceName`, plus the `vllm` YAML alias for both languages in `Aliases.java` and `aliases.py`.
- Failure behavior: a missing/blank `model` on the setup raises `IllegalArgumentException` / `ValueError` at construction with a message pointing at `vllm serve <model>` / `GET /v1/models`. Everything else (timeouts, retries, error wrapping) is inherited unchanged.
- Compatibility: additive only; no existing behavior changes.

### Tests

- Java: new `VLLMChatModelConnectionTest` (defaults injected — construction succeeds with no arguments where the parent would throw on missing `api_key`; blank args defaulted; explicit args honored; caller descriptor not mutated) and `VLLMChatModelSetupTest` (missing/blank model rejected; `getParameters` carries model + inherited fields). Existing Azure/OpenAI module tests pass unchanged.
- Python: new `test_vllm_chat_model.py` (connection defaults + overrides, setup model-required, `model_kwargs` passthrough).

### API

New public classes and constants only (`VLLMChatModelConnection` / `VLLMChatModelSetup`, `ResourceName.ChatModel.VLLM_CONNECTION` / `VLLM_SETUP` and Python variants, `vllm` YAML alias); no changes to existing APIs.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`


python/flink_agents/integrations/chat_models/vllm/vllm_chat_model.py
```diff
@@ -0,0 +1,83 @@
...
+    Behaves like :class:`OpenAIChatModelSetup` with one difference: ``model``
+    is required and has no default, because a vLLM server only serves the
+    model(s) it was started with — there is no meaningful universal default.
+    The value must match the model name announced by the server (see
+    ``vllm serve <model>``, or query ``GET /v1/models``).
+    """
+
+    def __init__(self, *, model: str | None = None, **kwargs: Any) -> None:
+        """Init method."""
+        if not model:
```
weiqingy: Java guards with `isBlank()`, Python with plain falsiness, so this line accepts a whitespace-only `model` while `VLLMChatModelSetup.java:52` rejects it. Same split for `api_key` / `api_base_url` at lines 59-60. `AGENTS.md` asks that "Public API changes must keep Java, Python, and YAML APIs semantically aligned", and both classes are new public API. The Java suite covers `" "` (`VLLMChatModelSetupTest.java:51-58`, `VLLMChatModelConnectionTest.java:51-59`); Python covers only `""`. Worth aligning Python on `not model or not model.strip()`?

weiqingy (commented):

integrations/chat-models/openai/src/test/java/org/apache/flink/agents/integrations/chatmodels/openai/VLLMChatModelConnectionTest.java
```diff
@@ -0,0 +1,81 @@
...
+    @DisplayName("Constructor honors explicit api_key and api_base_url")
+    void testConstructorExplicitArguments() {
+        ResourceDescriptor desc =
+                connectionDescriptor()
+                        .addInitialArgument("api_key", "secret-key")
+                        .addInitialArgument("api_base_url", "http://vllm-host:8000/v1")
+                        .addInitialArgument("timeout", 30)
+                        .addInitialArgument("max_retries", 1)
+                        .build();
+        assertThatCode(() -> new VLLMChatModelConnection(desc, NOOP)).doesNotThrowAnyException();
```
weiqingy, line 83: None of the four tests here asserts the values the connection ends up with. If `withVLLMDefaults` overwrote `api_key` unconditionally, all four would still pass, because `"EMPTY"` never throws in the parent constructor. A real key silently discarded looks like the case this test is aiming at. `OpenAIClient` is private, so there's no seam today, but dropping `private` on `withVLLMDefaults` (`VLLMChatModelConnection.java:72`) would let a same-package test assert on the descriptor it returns. Something like this, if useful:

```java
assertThat(VLLMChatModelConnection.withVLLMDefaults(desc).getInitialArguments())
        .containsEntry("api_key", "secret-key")
        .containsEntry("api_base_url", "http://vllm-host:8000/v1");
```

Is there a reason to keep it `private`?

python/flink_agents/integrations/chat_models/vllm/vllm_chat_model.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(
+        self,
+        *,
+        api_key: str | None = None,
+        api_base_url: str | None = None,
+        **kwargs: Any,
+    ) -> None:
+        """Init method."""
+        super().__init__(
+            api_key=api_key or DEFAULT_VLLM_API_KEY,
```
weiqingy: Both arguments are non-`None` by the time they reach the parent, so `resolve_openai_credentials` (`openai_chat_model.py:100`) never gets past `_get_from_param_or_env`'s first branch, and the documented param → env → `openai` module → default order collapses to param-only here. Someone running `vllm serve --api-key <token>` with `OPENAI_API_KEY` exported gets `"EMPTY"` sent and an unexplained 401. There's a real upside too: the defaults stop depending on the developer's environment, and Java has no env fallback either. Was pre-empting the parent's resolution deliberate? If so, would the docstring and the Python table in `chat_models.md` be the place to note that the env vars aren't consulted?

docs/content/docs/development/chat_models.md
```diff
@@ -1207,6 +1207,113 @@ Some popular options include:
 Model availability and specifications may change. Always check the official DashScope documentation for the latest information before implementing in production.
 {{< /hint >}}
 
+### vLLM
```
weiqingy, line 1210: There's a second place that lists providers: the "Native Integration Support Matrix" in `docs/content/docs/faq/faq.md:98-107`, a three-column table of provider, Python support, Java support. It currently names all eight existing chat models and not vLLM, so a reader who lands there would come away thinking vLLM isn't supported in either language. The rows are alphabetical, so a vLLM row linking to this new section would sit after Tongyi, with both columns ticked. Does that fit in this PR, or would you rather keep the faq change separate?

docs/content/docs/development/chat_models.md
```diff
@@ -1207,6 +1207,113 @@ Some popular options include:
...
+
+{{< /tab >}}
+
+{{< tab "Java" >}}
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `api_base_url` | String | `"http://localhost:8000/v1"` | vLLM server URL |
+| `api_key` | String | `"EMPTY"` | Only needed when the server is started with `--api-key`; the placeholder default works for unauthenticated servers |
+| `timeout` | Number | SDK default | Seconds before an API call times out |
```
weiqingy: The inherited Java constructor casts both of these to `Integer` (`OpenAICompletionsConnection.java:101,106`) through an unchecked `getArgument`, so a YAML `timeout: 30.5` throws `ClassCastException` at construction. The other Java tables on this page say `int` (523-524, 926-927); want to match them here?

weiqingy (commented):
Nothing to change here. Just a heads-up for whenever you rebase: main has since added native structured output to the OpenAI connections, so these classes will inherit it. Worth a look at that point.

docs/content/docs/development/chat_models.md
Zhuoxi2000: Will match the other tables

docs/content/docs/development/chat_models.md
Zhuoxi2000: Will add the row in this PR

python/flink_agents/integrations/chat_models/vllm/vllm_chat_model.py
Zhuoxi2000: Deliberate, for the two reasons you note. Will document in the docstring + the Python table.

integrations/chat-models/openai/src/test/java/org/apache/flink/agents/integrations/chatmodels/openai/VLLMChatModelConnectionTest.java
Zhuoxi2000: yeah you're right  no reason to keep private will adopt exactly that assertion

python/flink_agents/integrations/chat_models/vllm/vllm_chat_model.py
Zhuoxi2000: Good catch, will align on strip()-based checks + tests.

Zhuoxi2000:
Addressed all five comments (plus a merge of main to pick up #962 — the `vllm` alias now has `resolveClazz` coverage in both `AliasesTest` and `test_aliases.py`, following that PR's pattern):

- Python treats whitespace-only values as absent (strip-based), with `" "` tests mirroring the Java suite
- `withVLLMDefaults` is package-visible and the suggested assertion is adopted verbatim
- The env-var pre-emption is documented as deliberate, in the docstring and the Python parameter table
- The FAQ support matrix gained the vLLM row
- The Java table types now say `int`, matching the other tables

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/VLLMChatModelConnection.java
```diff
@@ -0,0 +1,86 @@
...
+ *   @ChatModelConnection
+ *   public static ResourceDescriptor vllm() {
+ *     return ResourceDescriptor.Builder.newBuilder(VLLMChatModelConnection.class.getName())
+ *             .addInitialArgument("api_base_url", "http://my-vllm-host:8000/v1")
+ *             .build();
+ *   }
+ * }
+ * }</pre>
+ */
+public class VLLMChatModelConnection extends OpenAICompletionsConnection {
```
wenjin272, line 57: [P1] Please use a vLLM-specific structured-output capability check. Because this class inherits `supportsNativeStructuredOutput()` unchanged, capability is still decided by the OpenAI model-name allowlist. Consequently, the documented `Qwen/Qwen2.5-7B-Instruct` model (and ordinary Llama names) returns `false`, so `chat(..., outputSchema)` silently omits `response_format`, while a model named `gpt-4o` returns `true`. vLLM supports the OpenAI `json_schema` response format for served models independently of OpenAI model names: https://docs.vllm.ai/en/stable/examples/features/structured_outputs/. Could we override this in both Java and Python (or introduce a vLLM-specific capability setting) and add request-building tests with a Qwen model?

docs/content/docs/development/chat_models.md
```diff
@@ -1207,6 +1207,113 @@ Some popular options include:
 Model availability and specifications may change. Always check the official DashScope documentation for the latest information before implementing in production.
 {{< /hint >}}
 
+### vLLM
+
+[vLLM](https://docs.vllm.ai) serves open-weight models behind an OpenAI-compatible API and is a popular choice for self-hosted production deployments. Flink Agents provides a dedicated connection that reuses the OpenAI integration with vLLM-friendly defaults, in both Java and Python.
+
+#### Prerequisites
+
+1. Install vLLM and start a server: `vllm serve Qwen/Qwen2.5-7B-Instruct`
```
wenjin272: [P2] Please document the server flags required for tool calling. The command shown starts basic chat serving, but Flink Agents sends tools without a named `tool_choice` and relies on vLLM automatic tool calling. vLLM requires `--enable-auto-tool-choice` and a model-specific `--tool-call-parser` for that path; for Qwen2.5 it recommends the `hermes` parser: https://docs.vllm.ai/en/stable/features/tool_calling/. Without those flags, tool calls are not parsed into the OpenAI `tool_calls` field. Could the prerequisite show a tool-enabled command and note that the parser is model-specific, so users do not get a server that can chat but cannot drive an agent’s tools?

wenjin272 (commented):
Thanks for contributing out-of-the-box vLLM support for both Java and Python — this is a valuable addition. The defaults, aliases, Java/Python parity, and documentation are thoughtfully covered. I found two follow-up issues around capabilities inherited from the OpenAI connection and the vLLM server prerequisites for tool calling.

integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/VLLMChatModelConnection.java
Zhuoxi2000: Good catch! the inherited allowlist wasn’t the right fit here. I’ve overridden the capability check in both Java and Python so it follows the served model instead.

vLLM’s guided decoding supports `json_schema` independently of OpenAI model names, while null/blank models still return false. I also added capability and request-building tests with `Qwen/Qwen2.5-7B-Instruct` in both languages, following the existing patterns.

docs/content/docs/development/chat_models.md
Zhuoxi2000: Done the docs now show the tool-enabled command for Qwen2.5:

`--enable-auto-tool-choice --tool-call-parser hermes`

I also noted that the parser is model-specific and linked to the vLLM tool-calling docs. Without those flags, normal chat still works, but tool calls won’t show up in `tool_calls`.

integrations/chat-models/openai/src/test/java/org/apache/flink/agents/integrations/chatmodels/openai/VLLMChatModelConnectionTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Test
+    @DisplayName("Constructor defaults blank api_key and api_base_url")
+    void testConstructorBlankArgumentsAreDefaulted() {
+        ResourceDescriptor desc =
+                connectionDescriptor()
+                        .addInitialArgument("api_key", "")
+                        .addInitialArgument("api_base_url", " ")
+                        .build();
+        assertThatCode(() -> new VLLMChatModelConnection(desc, NOOP)).doesNotThrowAnyException();
```
weiqingy, line 70: Now that `withVLLMDefaults` is package-visible, this one could pin values too. Drop just the `isBlank()` clause for `api_base_url` (`VLLMChatModelConnection.java:89-92`) and the whole Java suite stays green: a blank `api_base_url` makes the parent skip `builder.baseUrl(...)` (`OpenAICompletionsConnection.java:100`) rather than throw, so `doesNotThrowAnyException()` still holds. That leaves `api_base_url: ""` silently building against the SDK's default endpoint instead of the vLLM server. Python pins both values (`test_vllm_chat_model.py:62-66`). Worth extending this one the same way?

In case it helps:

```java
assertThat(VLLMChatModelConnection.withVLLMDefaults(desc).getInitialArguments())
        .containsEntry("api_key", VLLMChatModelConnection.DEFAULT_VLLM_API_KEY)
        .containsEntry("api_base_url", VLLMChatModelConnection.DEFAULT_VLLM_API_BASE_URL);
```

python/flink_agents/api/yaml/tests/test_aliases.py
```diff
@@ -127,6 +127,21 @@ def test_resolve_clazz_default_language_is_python() -> None:
...
 
 
+def test_resolve_clazz_covers_chat_model_vllm_in_both_languages() -> None:
+    assert resolve_clazz("vllm", ResourceType.CHAT_MODEL_CONNECTION).endswith(
+        "vllm_chat_model.VLLMChatModelConnection"
+    )
+    assert resolve_clazz("vllm", ResourceType.CHAT_MODEL).endswith(
+        "vllm_chat_model.VLLMChatModelSetup"
+    )
+    assert resolve_clazz("vllm", ResourceType.CHAT_MODEL_CONNECTION, "java").startswith(
```
weiqingy: `vllm` is a second alias whose Java and Python simple names collide, and these two Java assertions each pin half of what lines 158-162 pin for `azure_openai`: this one the package, line 140 the class. A Java entry that lost its `.Java` suffix and resolved to the Python class would still pass here, which is the case the comment at line 152 says to guard against. That comment also calls `azure_openai` the *only* such alias, which this PR makes false. Should this one pin both halves, and the comment pick up `vllm`?

weiqingy (commented):
Thanks for addressing the comments. I re-verified each at the current head and they're closed from my side. Two small test-strength points inline, both on tests this PR adds.

python/flink_agents/api/yaml/tests/test_aliases.py
Zhuoxi2000: Done both Java assertions now check the package and class, and I updated the azure_openai comment to call out vllm as the other colliding alias.

integrations/chat-models/openai/src/test/java/org/apache/flink/agents/integrations/chatmodels/openai/VLLMChatModelConnectionTest.java
Zhuoxi2000: Done! the blank-arguments test now checks both values after withVLLMDefaults (I used your suggested assertion). That way, if the isBlank() handling regresses, the test will catch it instead of silently falling back to the SDK default endpoint.

wenjin272 (approved):

Issue #936, opened 2026-07-28
[Tech Debt] Non-happy-path response handling diverges between Java and Python in OpenAI-family connections


weiqingy:
### Background

The OpenAI-family chat model connections (OpenAI and Azure OpenAI, both languages) converge on a shared message conversion helper. Comparing the two implementations, the non-happy-path branches diverge in three places. All three predate any current work, and all three are invisible on the happy path, which is why they have survived.

Native structured output makes the first two materially more reachable. With `strict: true` a model refuses rather than emitting non-conforming JSON, so the refusal branch stops being an edge case and becomes the primary non-happy path.

### 1. A refusal is preserved in Java and dropped in Python

Java, `OpenAIChatCompletionsUtils.convertFromOpenAIMessage` (`integrations/chat-models/openai/src/main/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIChatCompletionsUtils.java:113`):

```java
String content = message.content().orElse("");
ChatMessage response = ChatMessage.assistant(content);

message.refusal().ifPresent(refusal -> response.getExtraArgs().put("refusal", refusal));
```

Python, `convert_from_openai_message` (`python/flink_agents/integrations/chat_models/openai/openai_utils.py:217`):

```python
return ChatMessage(
    role=MessageRole(message.role),
    content=message.content or "",
    tool_calls=tool_calls,
    extra_args=extra_args,
)
```

There is no refusal handling on the Python side. `grep -rn "refusal" python/flink_agents/` returns zero hits across the whole tree, and `git log -S"refusal" -- python/flink_agents/` returns no commit, so the field has never been read there.

When a provider refuses, `content` is empty on both sides. A Java caller can recover the reason from `extraArgs["refusal"]`. A Python caller receives an empty assistant message with the reason discarded, and has no way to distinguish a refusal from a genuinely empty completion.

### 2. `finish_reason` is never inspected in either language

`grep -rni "finishReason\|finish_reason"` returns zero hits in both `integrations/chat-models/openai/src/main/java/` and `python/flink_agents/integrations/chat_models/`.

A completion truncated by a token limit therefore returns as ordinary partial content, with nothing indicating it is incomplete. Under structured output this is the more damaging case of the two, because truncated JSON fails to parse downstream with an error that points at the parser rather than at the truncation.

Unlike case 1 this one is symmetric, so it is a behavior question rather than a parity question: should a truncated completion raise, or surface the reason to the caller and let them decide?

[...]

PR #952, opened 2026-08-03
[integrations][openai] Preserve provider refusals in the Python message converter

weiqingy:
Linked issue: #936

Addresses case 1 only. Cases 2 and 3 are design questions and are left as follow-ups, as recorded on the issue.

### Purpose of change

Java and Python disagree on what a caller receives when a provider refuses a request. Java treats the refusal as recoverable information and Python discards it, so this is a parity fix with Java as the specification.

On a refusal the provider returns an assistant message with no content and the reason in `refusal`. `OpenAIChatCompletionsUtils.convertFromOpenAIMessage` records it as `extraArgs["refusal"]`. The Python converter never read the field, so the reason was lost and `content` became `""`. A Python caller could not tell a refusal apart from a genuinely empty completion.

The gap predates the native structured-output work, but `strict: true` makes the refusal path materially more reachable, because a model refuses rather than emitting non-conforming JSON.

#### Runtime flow

Unchanged except for one step. A connection calls `chat.completions.create`, builds a local `extra_args` holding `model_name` and token counts when usage is present, and passes it with `response.choices[0].message` to `convert_from_openai_message`. The converter builds the tool-call list, then, when `message.refusal is not None`, rebinds `extra_args` to a new dict carrying the reason before constructing the `ChatMessage`. Pydantic copies that dict during validation, so the write must precede construction to be visible.

Both the OpenAI and the Azure OpenAI connections call this helper, so one change covers both.

#### Key decisions

The guard is `is not None` rather than a truthiness test, so an empty refusal reason is still recorded. Java uses `Optional.ifPresent`, which keeps an empty string, and a truthiness test would have left a new divergence in the function meant to remove one.

There is no `isinstance` check. The inbound Java path has none. Java's `instanceof String` guards the outbound direction, where `extra_args` holds arbitrary caller data, which is a different situation.

The write rebinds rather than mutating the argument. Both callers pass a fresh local dict and pydantic copies it, so this is hygiene and not a guarantee. No test pins it, deliberately.

Java gets tests but no production change. Its behavior had no coverage, so a refactor could have removed it and reopened the gap from the other side.

#### Behavioral contracts

1. A refusal reason present on the SDK message appears unchanged at `extra_args["refusal"]` on the returned `ChatMessage`.
2. An empty-string refusal is recorded, not skipped.
3. When the SDK message carries no refusal, no `refusal` key is added.
4. Keys already present in the `extra_args` passed by the caller survive, including the token metrics both connections put there.
5. A refusal leaves `content` as `""`. The reason is never written into `content`.
6. The Java converter's behavior is unchanged.

#### Failure behavior

Nothing in this change raises, falls back, or retries. It adds one conditional dict write with no error path. Invalid configuration does not apply, since the converter takes none.

An error from the provider is unaffected and still propagates out of the Python connection unwrapped, which is case 3 on the linked issue and is deliberately not touched.

A response violating the expected shape is rejected before the converter sees it, since `refusal` is declared `Optional[str]` on the SDK's pydantic model. The converter does no type checking of its own, so a caller bypassing that model and supplying another type would have it stored as-is.

A message lacking the attribute would raise `AttributeError`. The field exists at the floor of the `openai>=1.66.3` pin, so no version guard is used.

### Tests

There was no non-integration coverage of the response-conversion path before this, in either language.

| Contract | Tests |
|---|---|
| 1, refusal preserved | `test_refusal_is_preserved_in_extra_args`, Python. `testRefusalPreservedInExtraArgs`, Java |
| 2, empty refusal preserved | `test_refusal_is_preserved_in_extra_args`, the `""` parameter |
| 3, no key when absent | `test_no_refusal_key_when_refusal_absent`, Python. `testNoRefusalKeyWhenAbsent`, Java |
| 4, caller keys survive | `test_refusal_is_preserved_in_extra_args`, which passes a non-empty `extra_args` and asserts it survives |
| 5, content not overwritten | `test_refusal_is_preserved_in_extra_args` |
| 6, Java unchanged | both Java tests, which lock existing behavior |

Each test was run against a deliberately broken implementation to confirm it fails when the behavior it covers breaks. A truthiness guard is caught only by the empty-string parameter, and replacing the merge with a plain assignment only by the surviving-keys assertion.

One existing fixture in `test_openai_native_structured_output.py` mocked the SDK message with only `role`, `content` and `tool_calls`. It now sets `refusal` too, so it does not hand back an auto-generated attribute.

Java runs 22 tests in the openai integration module, up from 20. The Python unit suite runs 657.

Azure is covered through the shared helper rather than by a test of its own, because every Azure test in the repo is integration-marked.

### API

No signature change, no new class, no new dependency.

`extra_args` is an existing field on `ChatMessage`, and `refusal` is a key Java already writes, so the cross-language contract is unchanged. Python starts honoring it.

For a caller who does nothing differently, one thing changes: on a refused response, `extra_args` now carries an extra key. Responses that were not refused are unaffected and gain no key.

There is one effect beyond the changed files. `convert_to_openai_message` merges `extra_args` into the outbound assistant message, so a `ChatMessage` that carries a refusal and is later sent back to the provider will now include `refusal` in that request. It is a declared field there, so this is valid and matches Java.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/integrations/chat-models/openai/src/test/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIChatCompletionsUtilsTest.java
+++ b/integrations/chat-models/openai/src/test/java/org/apache/flink/agents/integrations/chatmodels/openai/OpenAIChatCompletionsUtilsTest.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.integrations.chatmodels.openai;
+
+import com.openai.models.chat.completions.ChatCompletionMessage;
+import org.apache.flink.agents.api.chat.messages.ChatMessage;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Unit tests for how {@link OpenAIChatCompletionsUtils} carries a provider refusal from a Chat
+ * Completions message onto the returned {@link ChatMessage}. The converter is shared by every Chat
+ * Completions connection in this module.
+ */
+class OpenAIChatCompletionsUtilsTest {
+
+    @Test
+    @DisplayName("A refusal reason is carried into extraArgs")
+    void testRefusalPreservedInExtraArgs() {
+        // A refused response has no content, so the reason is the only thing separating it
+        // from a genuinely empty completion.
+        ChatCompletionMessage message =
+                ChatCompletionMessage.builder()
+                        .content(Optional.empty())
+                        .refusal("I cannot help with that")
+                        .build();
+
+        ChatMessage result = OpenAIChatCompletionsUtils.convertFromOpenAIMessage(message);
+
+        assertThat(result.getExtraArgs()).containsEntry("refusal", "I cannot help with that");
+    }
+
+    @Test
+    @DisplayName("No refusal key is added when the provider did not refuse")
+    void testNoRefusalKeyWhenAbsent() {
+        ChatCompletionMessage message =
+                ChatCompletionMessage.builder().content("hello").refusal(Optional.empty()).build();
+
+        ChatMessage result = OpenAIChatCompletionsUtils.convertFromOpenAIMessage(message);
+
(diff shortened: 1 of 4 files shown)


weiqingy:
Hi @wenjin272 , could you take a look at this PR when you get a chance?  A second sample for the Implementation Description experiment on #894, the Python refusal parity fix that came out of writing the #930 description.

It is much smaller than #930, 123 lines across 4 files, so it may be an easier one to start with if #930 is a lot to get through.

Two things changed in how it is written, both because #930's description turned out to be unreadable at 24k characters.

The description is the PR body now rather than a separate comment. Four of the six fields you listed already have homes in the current template, so only behavioral contracts and failure behavior needed a new heading. The whole body is about 6k characters.

The tests table maps one row per contract instead of one row per test, which is what made the #930 version long.

I have not touched the PR template itself. That seemed worth holding until you have actually reviewed one of these and can say whether it makes the review easier or just adds reading.

wenjin272:
Thanks for trying this format. It did reduce my review burden. By reading the PR description, I could quickly understand the intent, runtime flow, key decisions, test coverage, and edge cases such as preserving an empty refusal and existing `extra_args`.

That gave me enough confidence to focus my review on the behavioral contract and delegate the remaining implementation-level checks to a coding agent, including verifying that the implementation matches the description and running targeted tests. Compared with #930, keeping the description concise and mapping tests to contracts made it much easier to consume. Based on this PR, I think this format is genuinely helpful.

wenjin272 (approved):

weiqingy:
Thanks for the detailed feedback, and for running your agent against the description.

Good to know the shorter version worked better as a reviewer-facing document. I will keep the same shape on the next ones.

Issue #953, opened 2026-08-03
[Bug] Shaded distributions include Log4j dependencies from Plan


rosemarYuan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The current logging dependency setup is inconsistent across the Plan, Runtime, Dist, and E2E modules. This issue was discovered while working on #821. 

In the current dependency setup, the Plan module declares `slf4j-api`, `log4j-core`, and `log4j-slf4j-impl` with compile scope.

Because the Dist modules depend on Plan and package their dependencies with Maven Shade, these logging dependencies are copied into the generated Dist JARs.

However, the default logging setup of the supported Flink distributions already provides Log4j2 and the SLF4J binding for the JobManager and TaskManager processes. After a Flink Agents Dist JAR is installed, the same process may therefore contain:

- the Log4j API and implementation provided by Flink; and
- another Log4j implementation bundled inside Flink Agents.

If the two sets use different versions, the JVM may load incompatible Log4j API and Core classes together. This can cause runtime linkage errors such as:

```text
NoSuchMethodError
```

Expected behavior:

- Dist JARs do not include Log4j or SLF4J classes inherited from Plan.
- Flink Agents uses the logging implementation provided by the selected Flink distribution.
- `META-INF/NOTICE` matches the actual contents of the generated JARs.

### How to reproduce

1. Build the Dist artifacts:

```bash
mvn -DskipTests \
  -pl dist/common,dist/flink-1.20,dist/flink-2.0,dist/flink-2.1,dist/flink-2.2,dist/flink-2.3 \
  -am clean package
```

2. Inspect the generated JARs:

```bash
for artifact in \
  dist/common/target/flink-agents-dist-common-*.jar \
  dist/flink-*/target/flink-agents-dist-flink-*.jar; do
    echo "===== ${artifact} ====="
    jar tf "${artifact}" | grep -E \
      '(^|/)org/apache/logging/log4j/|(^|/)org/slf4j/|log4j2'
done
```

Expected:

```text
No Log4j or SLF4J classes are found.
```

Actual:

```text
Some Dist artifacts contain Log4j and SLF4J classes inherited from Plan.
```

### Version and environment

- Flink Agents version: `0.3-SNAPSHOT`

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #954, opened 2026-08-03
[dist][runtime] Use Flink-provided Log4j dependencies

rosemarYuan:
Linked issue: #953 

### Purpose of change

The shaded Dist artifacts currently receive Log4j and SLF4J dependencies transitively from Plan, even though the default logging setup of the supported Flink distributions already provides Log4j2 and the SLF4J binding for JobManager and TaskManager processes.

This can place different Log4j API and Core versions on the same runtime classpath and cause linkage errors such as `NoSuchMethodError`.

This PR makes the logging dependency and packaging boundary explicit:

- Keep `slf4j-api` as a normal Plan dependency so standalone Plan tools remain usable, but remove `log4j-core` and `log4j-slf4j-impl` from Plan.
- Keep Runtime's Log4j dependencies as `provided` and align the compile-time Log4j version with 2.24.3, the version used by the oldest supported Flink distributions.
- Exclude Log4j and SLF4J artifacts, classes, service metadata, and `log4j2.*` configuration files from all shaded Dist artifacts.
- Add test-scoped Log4j dependencies for the E2E test JVM and align them with the selected Flink version: 2.24.3 for Flink 1.20 through 2.2 and 2.25.3 for Flink 2.3.
- Update `META-INF/NOTICE` so it no longer lists Log4j or SLF4J as bundled dependencies.

After this change, Flink Agents uses the Log4j2 implementation provided by the default logging setup of the selected supported Flink distribution and no longer packages a second logging stack.

### Tests

- [x] Built all common, full, and thin Dist artifacts for the supported Flink versions.
- [x] Verified that the generated Dist JARs contain no Log4j or SLF4J classes, Log4j service metadata, or `log4j2.*` configuration files.
- [x] Verified that the generated NOTICE files match the updated artifact contents.

Full E2E and clean Flink deployment tests were not run locally.

### API

No public API changes.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/plan/pom.xml
+++ b/plan/pom.xml
@@ -85,16 +85,6 @@ under the License.
             <artifactId>slf4j-api</artifactId>
             <version>${slf4j.version}</version>
         </dependency>
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-core</artifactId>
-            <version>${log4j2.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-slf4j-impl</artifactId>
-            <version>${log4j2.version}</version>
-        </dependency>
     </dependencies>
 
     <profiles>
--- a/dist/common/pom.xml
+++ b/dist/common/pom.xml
@@ -79,15 +79,6 @@ under the License.
                             <shadeTestJar>false</shadeTestJar>
                             <shadedArtifactAttached>false</shadedArtifactAttached>
                             <filters>
-                                <!-- Exclude signature files -->
-                                <filter>
-                                    <artifact>*:*</artifact>
-                                    <excludes>
-                                        <exclude>META-INF/*.SF</exclude>
-                                        <exclude>META-INF/*.DSA</exclude>
-                                        <exclude>META-INF/*.RSA</exclude>
-                                    </excludes>
-                                </filter>
                                 <!-- Exclude flink-agents own code from common JAR -->
                                 <filter>
                                     <artifact>org.apache.flink:flink-agents-*</artifact>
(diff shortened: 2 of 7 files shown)


wenjin272:
Could you help review this pr @GreatEugenius?

plan/pom.xml
```diff
@@ -85,16 +85,6 @@ under the License.
             <artifactId>slf4j-api</artifactId>
             <version>${slf4j.version}</version>
```
GreatEugenius, line 86:
```suggestion
            <version>${slf4j.version}</version>
            <scope>provided</scope>
```

GreatEugenius (approved):
Hi, @rosemarYuan. Thanks for the PR. Overall, it looks good to me. I have just one minor suggestion.

plan/pom.xml
rosemarYuan: Thanks for you comment. Fix it.

rosemarYuan:
> Hi, @rosemarYuan. Thanks for the PR. Overall, it looks good to me. I have just one minor suggestion.

Thanks for the suggestion! I applied it, but it turns out this change breaks the cross-language E2E tests — the CI failure is NoClassDefFoundError: org/slf4j/LoggerFactory.The reason is that in cross-language tests, Python spawns a Java subprocess running the Plan tool directly (not inside a Flink cluster). When slf4j-api is marked as provided, it's excluded from the Plan module's transitive dependencies, so the subprocess's classpath doesn't include it. Unlike IT tests (where Flink's lib/ provides SLF4J) or unit tests (where Maven automatically adds provided deps to the test classpath), the cross-language test environment has no external provider for this dependency.This is why I originally kept slf4j-api as a compile-scope dependency in Plan — standalone Plan tool usage (including being invoked from Python) requires it to be bundled. The Dist shade plugin already excludes slf4j-api from the fat jar, so there's no duplication risk in the Flink deployment scenario.

I'll revert this commit to fix the CI.

wenjin272 (approved):
Thanks for picking this up @rosemarYuan.

PR #911, opened 2026-07-16
[infra] Add change-type review guides with runtime/ template

weiqingy:
Linked issue: #894

### Purpose of change

Item 2 of #894, in the shape agreed on the issue: split the worked examples into a per-change-type structure rather than a single appendix, so `code_review.md` stays a thin index and the detailed guidance for a change type loads only when an agent touches that kind of change.

This first PR establishes the template:

- `code_review.md` gains a **Change-Type Review Guides** index — a small matrix mapping each change type to its focused guide, used as an on-ramp before the full passes.
- `review-guides/runtime-state-recovery.md` is the first fully-worked guide: a focused checklist for runtime state, serde, and recovery reviews, plus two examples drawn from real review threads (#874 serde compatibility flag, #880 deterministic failure-path fixture).

The remaining change types (`api/`, Python-Java bridge, `dist`/dependency, docs-only) are listed as `planned` and will follow one small PR each once this shape is settled, so the template can be reviewed before the rest are filled in. The examples come from actual human reviews per the issue discussion, and each links the specific review comment.

### Tests

Not applicable — documentation only, no logic.

Ran `./tools/check-license.sh` with the change: RAT passes, confirming `review-guides/` is correctly excluded (a review-guide file has no place for a license header, matching the existing `AGENTS.md` and `code_review.md` exemptions).

### API

No. No code or public API change.

### Documentation

- [x] `doc-included`

PR #957, opened 2026-08-03
[infra] Add a Python-Java bridge review guide

weiqingy:
Linked issue: #894

### Purpose of change

Continues item 2 of #894 by filling the `Python-Java bridge` row that #911 left as `planned`. #911 established the template and noted the remaining change types would follow one small PR each once the shape settled. This is the first of those.

`review-guides/python-java-bridge.md` follows the merged `runtime-state-recovery.md` shape: a focused checklist that narrows the full passes in `code_review.md`, and two examples drawn from real review threads on merged PRs, each linking the specific comment.

It adds one section the first guide does not have, a short `Validation` block naming both language lanes. The bridge is the change type where running one language's tests is most easily mistaken for running the tests, so the guide gives the narrowed Java and Python commands and says what they cover between them.

The checklist is six bullets, matching the sibling. Interpreter close ordering was left out deliberately. It is a real crash source, but the rarest trigger, since only PRs touching open and close reach it. Straightforward to add if you would like it covered.

The other three rows stay `planned` and will follow the same way.

### Tests

Not applicable, documentation only, no logic.

`./tools/check-license.sh` passes with the change. No license header is needed, since `tools/.rat-excludes` already covers `review-guides/`.

Every checklist claim was checked against source rather than written from memory, and both review permalinks were re-fetched with the cited PRs confirmed merged.

The Java command was run. `mvn --batch-mode -pl runtime -am` resolves and runs the upstream module tests. The Python lane was verified by collection rather than by running `uv` itself, and it needed a correction: `uv sync --extra test` on its own is not enough, because `python/flink_agents/api/events/event.py` imports `pyflink.common.Row` at module level while `apache-flink` is declared in neither the `test` extra nor the base dependencies, so collection fails across all three directories without it. The guide points at `tools/ut.sh` for the supported version list rather than naming a version, so it does not drift at the next bump.

### API

No. No code or public API change.

### Documentation

- [x] `doc-included`

### DIFF
--- a/code_review.md
+++ b/code_review.md
@@ -40,7 +40,7 @@ Guides load on demand, so the general passes here stay short.
 | Change type | Focus | Guide |
 |---|---|---|
 | `runtime/` state and recovery | serde and replay type fidelity, real failure-path tests | [review-guides/runtime-state-recovery.md](review-guides/runtime-state-recovery.md) |
-| Python-Java bridge | cross-language parity, type mapping across Pemja | planned |
+| Python-Java bridge | cross-language parity, type mapping across Pemja | [review-guides/python-java-bridge.md](review-guides/python-java-bridge.md) |
 | `api/` contract | API shape, compatibility policy, deprecation | planned |
 | `dist` and dependency | shading, LICENSE and NOTICE, dist registration | planned |
 | docs-only | facts match their source of truth | planned |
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):
Thanks for taking this on @weiqingy. LGTM.

Issue #946, opened 2026-08-01
[Bug] Anthropic chat model drops token usage metrics on non-tool_use responses


Zhuoxi2000:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

In the Python Anthropic connection, `extra_args` (carrying `model_name` / `promptTokens` / `completionTokens`) is built after every API call, but it is only attached to the returned `ChatMessage` on the `tool_use` branch. The plain-text branch (`stop_reason == "end_turn"`, the common case) returns `ChatMessage(role, content=text)` without `extra_args`, so token metrics are dropped for every normal response.

Where it happens (`python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py`): `extra_args` is populated right after `client.messages.create(...)`, the `tool_use` branch passes `extra_args=extra_args`, and the `else` branch does not.

**Expected**: `promptTokens` / `completionTokens` counters increase for every Anthropic response that carries `usage`, matching the other connectors — the consumer in `chat_model_action.py` reads exactly these keys before calling `_record_token_metrics`.

**Actual**: only tool-calling turns are counted. In a ReAct loop, the intermediate tool-use turns are recorded but the final answer's tokens are always lost, so Anthropic token metrics are systematically under-reported.

All sibling connectors (openai / azure / ollama / tongyi) pass `extra_args` on every path, so this looks like an omission rather than a design choice. Related in spirit to the non-happy-path parity audit in #936.

Fix is one line (pass `extra_args=extra_args` in the `else` branch) plus a regression test — the existing `test_plain_text_response` only asserts `content` and calls `chat()` without a `model=` kwarg, so this path was never covered.

### How to reproduce

Unit-level (no network):

1. Build an `AnthropicChatModelConnection` with a mocked client returning
   `Message(stop_reason="end_turn", content=[TextBlock("Hello!")], usage=Usage(input_tokens=7, output_tokens=3))`.
2. Call `connection.chat([ChatMessage(role=USER, content="hi")], model="claude-sonnet-4-5")`.
3. `response.extra_args` is `{}` — expected `{"model_name": "claude-sonnet-4-5", "promptTokens": 7, "completionTokens": 3}`.

Repeat the same call with a `tool_use` response: `extra_args` is populated. Same request shape, divergent metrics behavior.

[...]

PR #959, opened 2026-08-04
[integration][anthropic][python] Keep token usage on non-tool_use responses

Zhuoxi2000:
Linked issue: #946

### Purpose of change

`extra_args` (`model_name` / `promptTokens` / `completionTokens`) is built after every Anthropic API call, but was only attached to the returned `ChatMessage` on the `tool_use` branch. Plain-text responses (`stop_reason == "end_turn"`, the common case) returned without `extra_args`, so `_record_token_metrics` never fired for them: in a ReAct loop only the intermediate tool-calling turns were counted and the final answer's tokens were always lost. All sibling connectors (openai / azure / ollama / tongyi) already pass `extra_args` on every path.

Fix: pass `extra_args=extra_args` in the non-tool_use branch. One line; no behavior change on the tool_use path.

### Tests

Two new cases in `test_anthropic_response_parsing.py`: `test_plain_text_response_keeps_token_usage` (regression for the fixed branch) and `test_tool_use_response_keeps_token_usage` (guard for the already-working branch). Both are mock-based with explicit token counts and a `model=` kwarg so the metrics path is actually exercised — the pre-existing `test_plain_text_response` only asserted `content` and never covered this path.

### API

None.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
+++ b/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
@@ -239,6 +239,7 @@ class AnthropicChatModelConnection(BaseChatModelConnection):
             return ChatMessage(
                 role=MessageRole(message.role),
                 content=text,
+                extra_args=extra_args,
             )
 
     @override
--- a/python/flink_agents/integrations/chat_models/anthropic/tests/test_anthropic_response_parsing.py
+++ b/python/flink_agents/integrations/chat_models/anthropic/tests/test_anthropic_response_parsing.py
@@ -94,3 +94,46 @@ def test_plain_text_response() -> None:
         [ChatMessage(role=MessageRole.USER, content="hi")]
     )
     assert response.content == "Hello!"
+
+
+def test_plain_text_response_keeps_token_usage() -> None:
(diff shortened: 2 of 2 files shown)


weiqingy (commented):
Thanks for taking this on, and for covering the tool_use branch alongside the fixed one.

One thing that might be useful for the issue: the Java connector already attaches these three keys to every response, whatever the stop reason.

Fix and tests check out against main. Nothing further from me, and I'll leave the final call to the maintainers.

Zhuoxi2000:
> Thanks for taking this on, and for covering the tool_use branch alongside the fixed one.
> 
> One thing that might be useful for the issue: the Java connector already attaches these three keys to every response, whatever the stop reason.
> 
> Fix and tests check out against main. Nothing further from me, and I'll leave the final call to the maintainers.

Thanks for checking against main! Good catch on the Java side, confirmed that it preserves `model_name`, `promptTokens`, and `completionTokens` on every response path, so this fix brings Python back in line with Java. I’ve noted that on #946 as well.

wenjin272 (approved):
Thanks for fixing this. LGTM.

Issue #947, opened 2026-08-01
[Bug] Gemini tool calling fails on the second turn when functionCall carries no id


Zhuoxi2000:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The Gemini Developer API frequently returns `functionCall` parts without an `id`. In
`GeminiChatModelConnection.convertFunctionCall`, the id is read via `functionCall.id().orElse(null)`,
and when absent, neither `id` nor `original_id` is put into the tool-call map. Two things break downstream:

1. `ToolCallAction` reads the id via `String.valueOf(toolCall.get("id"))`, producing the literal
   string `"null"`. With two or more parallel id-less calls, all results collide under the `"null"`
   key in the `success` / `responses` / `error` maps, so only the last one survives.
2. `original_id` is what `ToolCallAction` propagates as the TOOL message's `externalId`, which is
   how the follow-up turn recovers the function name for Gemini's `functionResponse` part
   (`buildToolCallIdToNameMap` + `resolveToolFunctionName`). Without it, the second request throws
   `IllegalArgumentException("Tool message must carry the function name...")`.

**Expected**: a documented, supported flow — Gemini setup with tools — completes the tool round trip.

**Actual**: turn 1 executes the tool, turn 2 crashes. The class javadoc assumes a native id is always
present ("Tool calls are returned as functionCall parts carrying a native id"), but the Developer API
does not guarantee one.

Existing unit tests always build `FunctionCall.builder().id("call_1")...`, so the id-less path is
never exercised end-to-end (the one id-less case, `testConvertFunctionCallNoSignature`, only asserts
`thought_signature` absence).

Proposed fix: synthesize a UUID when the native id is absent and set both `id` and `original_id`,
marked (e.g. `synthetic_id`) so `convertToolCallToPart` never echoes a fabricated id back to the
Gemini API — on-the-wire behavior for id-less calls stays unchanged, while the runtime round trip
works.

### How to reproduce

Unit-level (no network):

1. `Map<String, Object> call = connection.convertFunctionCall(FunctionCall.builder().name("get_weather").args(Map.of()).build(), null);`
   → the map contains neither `id` nor `original_id`.
2. Feed that tool call through the runtime shape: the TOOL `ChatMessage` never gets an `externalId`,
   so `convertToContent(toolMessage, buildToolCallIdToNameMap(messages))` throws
[...]

PR #960, opened 2026-08-04
[integration][gemini] Synthesize a tool-call id when the API omits functionCall.id

Zhuoxi2000:
Linked issue: #947

### Purpose of change

The Gemini Developer API frequently returns `functionCall` parts without an id. `convertFunctionCall` previously emitted neither `id` nor `original_id` in that case, which broke the runtime round-trip twice:

1. `ToolCallAction` reads the id via `String.valueOf(toolCall.get("id"))`, so every id-less call keyed the `success`/`responses`/`error` maps on the literal string `"null"` — two parallel id-less calls collided and only the last result survived.
2. `original_id` is what `ToolCallAction` propagates as the TOOL message's `externalId`; without it the follow-up turn cannot resolve the function name for Gemini's `functionResponse` part, and `resolveToolFunctionName` throws `"Tool message must carry the function name"` — turn 1 executes the tool, turn 2 crashes.

Fix: when the native id is absent, synthesize a UUID and set both `id` and `original_id`, plus a `synthetic_id` marker. `convertToolCallToPart` skips echoing the id back to Gemini when the marker is present, so on-the-wire behavior for id-less calls is unchanged — the synthetic id exists purely for runtime correlation. Calls that carry a native id are untouched.

### Tests

Three new cases in `GeminiChatModelConnectionTest`: id-less calls get unique synthetic ids (`id == original_id`, `synthetic_id` set, two calls never collide); a synthetic id is never echoed back to the API (`Part.functionCall().id()` stays empty); and the full id-less round trip — assistant turn → `buildToolCallIdToNameMap` → TOOL turn with `externalId` — resolves the function name (this exact flow threw before the fix). Existing 16 tests pass unchanged (20/20 total).

### API

None (`synthetic_id` is an internal marker key inside the tool-call map).

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included`

### DIFF
--- a/integrations/chat-models/gemini/src/main/java/org/apache/flink/agents/integrations/chatmodels/gemini/GeminiChatModelConnection.java
+++ b/integrations/chat-models/gemini/src/main/java/org/apache/flink/agents/integrations/chatmodels/gemini/GeminiChatModelConnection.java
@@ -48,6 +48,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):
LGTM

Issue #958, opened 2026-08-03
[Bug] Six resource implementations have no YAML alias, and the `azure` alias documents the wrong provider


weiqingy:
### Background

`clazz:` values in a YAML agent definition are resolved through three hand-maintained copies of the same alias table, with nothing cross-checking them:

- Java: `api/src/main/java/org/apache/flink/agents/api/yaml/Aliases.java:82-142`
- Python: `python/flink_agents/api/yaml/aliases.py`
- Docs: `docs/content/docs/development/yaml.md:522-551`

Adding an integration and adding its alias are separate steps, and nothing fails when the second one is skipped. Both loaders pass an unrecognized name through unchanged, so the class stays reachable by fully-qualified name and no test notices. Java, `Aliases.java:183-187`:

```java
Map<String, String> bucket = byLang.get(language);
if (bucket == null) {
    return name;
}
return bucket.getOrDefault(name, name);
```

Python, `aliases.py`:

```python
bucket = CLAZZ_ALIASES.get(resource_type, {}).get(language, {})
return bucket.get(name, name)
```

`yaml.md:554` documents that pass-through as intended user-facing behavior, which is reasonable on its own but also means an omitted alias is invisible.

This has been caught after the fact before. #812 was a whole PR whose only substance was adding Bedrock's aliases to all three tables, well after the Bedrock integration itself shipped.

Comparing `ResourceName.java` against the alias tables at `main` turns up two problems.

### 1. Six implementations exist with a `ResourceName` constant and no alias

All six classes exist at `main`; only the alias entry is missing.

| Resource type | Implementation | Constant | In alias tables |
|---|---|---|---|
| `chat_model_connections` / `chat_model_setups` (java) | Gemini | `ResourceName.java:69-72` | no |
| `chat_model_connections` / `chat_model_setups` (java) | Azure OpenAI | `ResourceName.java:93-96` | no |
| `vector_stores` (java) | OpenSearch | `ResourceName.java:199-200` | no |
| `vector_stores` (java) | S3 Vectors | `ResourceName.java:203-204` | no |
| `vector_stores` (java) | Milvus | `ResourceName.java:207-208` | no |
| `vector_stores` (python) | Mem0 | `ResourceName.java:225-226` | no |

The vector-store table is the thinnest. `Aliases.java:137-142` registers exactly one alias per language:

```java
// VECTOR_STORE
Map<String, String> vsJava = new HashMap<>();
[...]

PR #962, opened 2026-08-04
[api][yaml] Add aliases for six unaliased resource implementations

hulincup:
Linked issue: #958

### Purpose of change

The YAML loader resolves short `clazz:` aliases through three hand-maintained copies of the same alias table (Java `Aliases.java`, Python `aliases.py`, docs `yaml.md`). Six resource implementations already have a `ResourceName` constant but **no alias entry** in any of the three tables, so users must reach them by fully-qualified class path. Per item 1 of #958, this adds the missing entries across all three tables, mirroring the shape of existing entries (e.g. the Bedrock aliases added in #812).

| Alias | Resource type | Language | Implementation |
| --- | --- | --- | --- |
| `gemini` | chat model connection + setup | java | Gemini |
| `azure_openai` | chat model connection + setup | java | Azure OpenAI |
| `opensearch` | vector store | java | OpenSearch |
| `s3_vectors` | vector store | java | S3 Vectors |
| `milvus` | vector store | java | Milvus |
| `mem0` | vector store | python | Mem0 |

All additions are **purely additive**. Java/Python/API-YAML alignment is preserved (the same six entries land in the Java table, the Python table, and the docs table).

Scope: **item 1 of #958 only**. **Item 2** — the `azure` alias resolving to the Azure AI Inference connection (`AzureAIChatModelConnection`) while `yaml.md` documents it as "Azure OpenAI (Java)" — is a behavior change for users who already have `clazz: azure` in their YAML and is intentionally left for a separate PR, as the issue suggests. The existing `azure` alias and its docs row are untouched here. Adding `azure_openai` as a *new* Java alias is independent of that decision and gives the Azure OpenAI Java connection a reachable short name either way.

Open question from the issue: *should any of these six be deliberately unaliased?* I defaulted to aliasing all six, since an alias is an optional short name (fully-qualified `clazz:` still works) and adding one is reversible. Happy to drop any a maintainer feels should stay first-class-only.

### Tests

- `python/flink_agents/api/yaml/tests/test_aliases.py`: added two cases asserting the six new aliases resolve to the expected FQNs. For `azure_openai` — the only alias whose Java/Python simple names collide (`AzureOpenAIChatModelConnection`) — the assertions pin the package prefix (`startswith("org.apache.flink.agents")`) in addition to the simple name, so a Java entry that lost its `.Java` suffix and resolved to the Python class would fail. Full file passes locally: `13 passed`.
- `api/src/test/java/org/apache/flink/agents/api/yaml/AliasesTest.java`: added two parallel cases mirroring the existing `clazzAliasJavaBucket` style.
- `ruff check` clean on the changed Python file.

Java build was not run locally (sparse checkout of a large repo; host JDK is 8 vs the project Java 11 target). The change is mechanical — each new entry mirrors the immediately adjacent existing `put`/dict lines — and the referenced `ResourceName` constants all exist at `main`. CI will confirm.

### API

No public API signature changes. Aliases are optional short names; fully-qualified `clazz:` values keep passing through unchanged.

### Documentation

- [x] `doc-included`

`docs/content/docs/development/yaml.md` is updated in this PR: the `azure_openai` row gains a `type: java` column entry, and `mem0` / `opensearch` / `s3_vectors` / `milvus` rows are added to the vector-store alias table.

Part of #958 (item 1).

Generated-by: Claude Code (Anthropic) — used to author this patch and PR description; reviewed and submitted by a human contributor.

### DIFF
--- a/python/flink_agents/api/yaml/tests/test_aliases.py
+++ b/python/flink_agents/api/yaml/tests/test_aliases.py
@@ -134,12 +134,17 @@ def test_resolve_clazz_covers_chat_model_java_gemini_and_azure_openai() -> None:
     assert resolve_clazz("gemini", ResourceType.CHAT_MODEL, "java").endswith(
         "GeminiChatModelSetup"
     )
-    assert resolve_clazz(
+    # `azure_openai` is the only alias whose Java/Python simple names collide
+    # (AzureOpenAIChatModelConnection); pin the package so a Java entry that lost
+    # its `.Java` suffix and resolved to the Python class would fail here.
+    azure_conn = resolve_clazz(
         "azure_openai", ResourceType.CHAT_MODEL_CONNECTION, "java"
-    ).endswith("AzureOpenAIChatModelConnection")
-    assert resolve_clazz("azure_openai", ResourceType.CHAT_MODEL, "java").endswith(
-        "AzureOpenAIChatModelSetup"
     )
+    assert azure_conn.startswith("org.apache.flink.agents")
+    assert azure_conn.endswith("AzureOpenAIChatModelConnection")
+    azure_setup = resolve_clazz("azure_openai", ResourceType.CHAT_MODEL, "java")
+    assert azure_setup.startswith("org.apache.flink.agents")
+    assert azure_setup.endswith("AzureOpenAIChatModelSetup")
 
 
 def test_resolve_clazz_covers_vector_store_java_and_python() -> None:
--- a/python/flink_agents/api/yaml/tests/test_aliases.py
+++ b/python/flink_agents/api/yaml/tests/test_aliases.py
@@ -127,6 +127,36 @@ def test_resolve_clazz_default_language_is_python() -> None:
     assert default == explicit
 
 
+def test_resolve_clazz_covers_chat_model_java_gemini_and_azure_openai() -> None:
+    assert resolve_clazz("gemini", ResourceType.CHAT_MODEL_CONNECTION, "java").endswith(
+        "GeminiChatModelConnection"
+    )
+    assert resolve_clazz("gemini", ResourceType.CHAT_MODEL, "java").endswith(
+        "GeminiChatModelSetup"
+    )
+    assert resolve_clazz(
+        "azure_openai", ResourceType.CHAT_MODEL_CONNECTION, "java"
+    ).endswith("AzureOpenAIChatModelConnection")
+    assert resolve_clazz("azure_openai", ResourceType.CHAT_MODEL, "java").endswith(
+        "AzureOpenAIChatModelSetup"
+    )
+
+
+def test_resolve_clazz_covers_vector_store_java_and_python() -> None:
+    assert resolve_clazz("opensearch", ResourceType.VECTOR_STORE, "java").endswith(
+        "OpenSearchVectorStore"
+    )
+    assert resolve_clazz("s3_vectors", ResourceType.VECTOR_STORE, "java").endswith(
+        "S3VectorsVectorStore"
+    )
+    assert resolve_clazz("milvus", ResourceType.VECTOR_STORE, "java").endswith(
+        "MilvusVectorStore"
+    )
+    assert resolve_clazz("mem0", ResourceType.VECTOR_STORE, "python").endswith(
+        "Mem0VectorStore"
+    )
+
+
 def test_java_wrapper_clazz_table_covers_supported_types() -> None:
     # The Python-side wrappers must exist for every cross-language type
     expected = {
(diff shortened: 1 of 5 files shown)


python/flink_agents/api/yaml/tests/test_aliases.py
```diff
@@ -127,6 +127,36 @@ def test_resolve_clazz_default_language_is_python() -> None:
...
+def test_resolve_clazz_covers_chat_model_java_gemini_and_azure_openai() -> None:
+    assert resolve_clazz("gemini", ResourceType.CHAT_MODEL_CONNECTION, "java").endswith(
+        "GeminiChatModelConnection"
+    )
+    assert resolve_clazz("gemini", ResourceType.CHAT_MODEL, "java").endswith(
+        "GeminiChatModelSetup"
+    )
+    assert resolve_clazz(
+        "azure_openai", ResourceType.CHAT_MODEL_CONNECTION, "java"
+    ).endswith("AzureOpenAIChatModelConnection")
```
weiqingy: `AzureOpenAIChatModelConnection` is the simple name on both sides of the bridge:

- Python: `...chat_models.azure.azure_openai_chat_model.AzureOpenAIChatModelConnection`
- Java: `...integrations.chatmodels.openai.AzureOpenAIChatModelConnection`

So this passes even if the java entry ever lost its `.Java` and resolved to the Python class, and the same goes for the `Setup` assertion two lines down. The Java mirror catches that case because `isEqualTo` compares the whole constant, so the two sides aren't guarding equally right now.

`test_resolve_clazz_dispatches_on_language` above already pins the package for this reason. Would it be worth borrowing that here?

```python
azure_conn = resolve_clazz("azure_openai", ResourceType.CHAT_MODEL_CONNECTION, "java")
assert azure_conn.startswith("org.apache.flink.agents")
assert azure_conn.endswith("AzureOpenAIChatModelConnection")
```

`azure_openai` is the only one of the six where the names collide, so it would just be these two.

weiqingy (commented):
Thanks for taking this on. Only one inline comment below.

FYI in case it's useful, since you deliberately scoped item 2 out: it has a doc-only PR up in #963, so that half is covered and this one gets to stay purely additive.

weiqingy:
One note on the body: GitHub's closing-keyword parser stops at the number, so `Closes #958 (item 1).` doesn't actually scope to item 1. Merging would close all of #958, taking item 2 and the still-open question about whether any of the six should stay deliberately unaliased along with it. Would `Part of #958 (item 1).` do the job here? Same link, minus the auto-close.

One other thing worth folding in: the repo keeps a PR template at `.github/PULL_REQUEST_TEMPLATE.md` asking for a `Linked issue:` line plus `### Purpose of change`, `### Tests`, `### API`, and `### Documentation`. Easy to miss, since GitHub only pre-fills it when you open the PR through the web form. Your body already covers most of it, so it's mostly remapping headings, and `doc-included` looks like the right box given the `yaml.md` change.

hulincup:
Thanks @weiqingy, both points addressed in ab251a7:

1. **Closing keyword** — switched `Closes #958 (item 1).` to `Part of #958 (item 1).` so merging doesn't auto-close #958 and drag item 2 (and the open "should any stay unaliased" question) along with it.

2. **PR template** — remapped the body to the `.github/PULL_REQUEST_TEMPLATE.md` headings (`Linked issue`, `### Purpose of change` / `### Tests` / `### API` / `### Documentation`) and ticked `doc-included`, since `yaml.md` is updated in-tree.

python/flink_agents/api/yaml/tests/test_aliases.py
hulincup: Good catch, thanks. ab251a7 pins the package on both the connection and setup assertions for `azure_openai` — `startswith("org.apache.flink.agents")` plus the existing `endswith`, mirroring `test_resolve_clazz_dispatches_on_language`. It's the only one of the six where the Java/Python simple names collide, so the other five keep the `endswith`-only form. `13 passed` locally on `test_aliases.py`.

wenjin272:
Thanks for taking this on @hulincup. LGTM. 

Sorry, I merged #963 first without noticing the overlap, so this PR now needs a conflict resolution.

hulincup:
Thanks @wenjin272 and @weiqingy for the review.

**Conflict resolved** — rebased onto `main` to absorb #963. The `yaml.md` overlap is gone: kept #963's `Azure AI (Java)` wording on the `azure` row and layered the new `gemini` row + the `azure_openai` Java column on top. Force-pushed; CI is rerunning (`mergeable` is back to `MERGEABLE`).

**@weiqingy's top-level points** — both addressed:
- Closing keyword switched to `Part of #958 (item 1).` so merging won't auto-close #958 (item 2 + the open question stay visible).
- Body remapped to the PR template sections (`Linked issue`, `### Purpose of change / Tests / API / Documentation`) with `doc-included` checked.

**@weiqingy's line comment on `test_aliases.py`** — addressed in 78b8328; replying in that thread.

python/flink_agents/api/yaml/tests/test_aliases.py
hulincup: Good catch — done in 78b8328. `azure_openai` now asserts both `startswith("org.apache.flink.agents")` and `endswith(...)` on the connection and the setup, mirroring `test_resolve_clazz_dispatches_on_language`. `13 passed` locally, including `test_resolve_clazz_covers_chat_model_java_gemini_and_azure_openai`.

wenjin272 (approved):

PR #964, opened 2026-08-04
[api][plan][runtime][examples] Add pluggable in-chat model routing (MODEL_ROUTER)

purushah:
Linked issue: #897 (design discussion). Supersedes #852.

### Purpose of change

Today an agent must name a concrete chat model in each `ChatRequestEvent`; any routing logic lives in user code and is not first-class in the runtime. Most requests are easy and a small model handles them fine, but a minority really need the strong model — so a fixed choice either overpays on the easy majority or under-serves the hard ones. This PR makes per-request routing a framework capability: an agent sends its request to a router, and the runtime picks one of several candidate models.

My first attempt at this (#852) made the router a special `ChatModelSetup` that called the chosen backend itself. Review rightly pointed out that this breaks the two things the framework is good at: token metrics collapse onto the router instead of the real backend, and the actual model call disappears from the EventLog. The redesign discussed in #897 fixes this by making the router **select instead of delegate**: the router only returns a model *name*, and `ChatModelAction` runs the normal chat path against that model. A routed request produces the same events and the same per-model metrics as if the agent had named the chosen model directly.

From the user side it stays simple:

```java
env.addResource("router", ResourceType.MODEL_ROUTER,
    ModelRouter.of("small", "big")
        .describe("small", "fast and cheap; chit-chat, simple facts")
        .describe("big", "strong; code, SQL, analysis")
        .strategy(Strategies.rules(Map.of("big", "\\b(code|sql|analyze)\\b")))
        .defaultModel("small")
        .fallback(true)
        .build());

// the agent just sends its ChatRequestEvent to "router"
```

A few behaviors were worth getting right, and they came out of the #897 discussion:

- **The decision is durable.** The strategy runs inside its own durable call (`"route:<router>"` — deterministic across recovery re-processing; per-request uniqueness comes from the store's key/sequence/event scoping), so on recovery the persisted decision is replayed instead of re-running the strategy. Decision latency is measured inside that call, so a replayed run reports the original `decision_ms`.
- **Route once per ReAct loop.** Tool-call rounds reuse the already-selected model and carry the routing metadata onto the final response — no re-routing mid-conversation.
- **Fallback sits on top of retries, not instead of them.** The selected model gets its full retry budget first; only then are the remaining candidates tried in declaration order, and a fallback emits a second `ModelRoutingEvent` so the event log shows which model actually answered.
- **Abstain is not an error, an invalid pick is.** A strategy that has no opinion abstains and the router's default model handles the request (`decision_source=default`); a strategy that returns a non-candidate name fails loudly, because that is a bug, not a runtime condition.
- **`ModelRoutingEvent` is observability-only.** It has no built-in consumer and does not drive dispatch — removing every listener does not change which model runs.

Since routers and chat models share the `ChatRequestEvent` namespace, one name must not be registered as both; this is validated at the `addResource` call site (with an `AgentPlan` backstop) so the failure points at the user's own line.

Per the discussion, LLM-as-judge routing is deliberately **not** in this PR. `RoutingContext` exposes no chat API, so a strategy cannot make hidden model calls; the framework-managed observable judge (the judge call running on the normal durable/metered chat path) is the agreed follow-up. YAML support is intentionally left out of v1 to keep the first Java API/runtime change small; it and Python support are follow-ups.

### Tests

- 20 API unit tests (`RoutingTest`, `RoutingResourceValidationTest`): rule matching and abstain over the latest user message, builder and registration validation, candidate descriptions reaching strategies, `RoutingDecision` invariants and snake_case JSON round-trips, `decision_ms` surviving replay deserialization, event attribute normalization.
- 9 `ChatModelActionRoutingTest` integration tests against a scripted fake chat model: routing to the matched candidate, abstain-to-default, invalid candidate failing clearly, non-router requests keeping the legacy `"chat"` durable id unchanged, fallback across candidates (including exhaustion), and route-once semantics across a tool round with metadata carried to the final response.
- A `ResourceCacheTest` case covering `hasResource` visibility of resources inserted directly into the cache (no provider).

Verified locally (JDK 17):

- `mvn spotless:check -pl api,plan,runtime,examples`
- `mvn test -pl api,plan` — 312 + 196 tests, 0 failures
- `mvn test -pl runtime -Dtest=ResourceCacheTest`
- `mvn compile -pl examples`

### API

New public surface, all additive:

- `ResourceType.MODEL_ROUTER`
- `org.apache.flink.agents.api.chat.model.routing`: `ModelRouter` (+ builder), `RoutingStrategy`, `RoutingContext`, `RoutingDecision`, `RoutingCandidate`, `RoutingStrategyDescriptor`, `Strategies`, `RuleBasedRoutingStrategy`
- `ModelRoutingEvent`
- `RunnerContext.hasResource(name, type)` (default method, returns false)

### Compatibility impact

A request naming a plain chat model takes the unchanged success path, including the legacy durable call id, and the default `error-handling-strategy` (`FAIL`) is unaffected. Three deliberate behavior changes on non-default failure paths:

- **`retry`:** retry metrics are recorded per attempt (previously once with cumulative totals on the final response), so requests that ultimately fail now contribute their retry counts and each attempt lands on the model that served it. Totals over a completed request are unchanged.
- **`retry`:** the retry WARN log now emits under `ChatModelInvoker`'s logger (was `ChatModelAction`) — external log alerting keyed to the old logger name needs updating.
- **`ignore`:** a request naming a missing chat-model resource is warned and dropped (consistent with other request failures under `ignore`) instead of failing the job with a raw lookup exception.

### Documentation

- [x] `doc-needed`
- [ ] `doc-not-needed`
- [ ] `doc-included`

The Java examples show usage, but this adds public API, so proper docs are warranted — happy to do a docs follow-up PR once the API settles in review.

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/EventType.java
+++ b/api/src/main/java/org/apache/flink/agents/api/EventType.java
@@ -41,6 +41,8 @@ public final class EventType {
             org.apache.flink.agents.api.event.ContextRetrievalRequestEvent.EVENT_TYPE;
     public static final String ContextRetrievalResponseEvent =
             org.apache.flink.agents.api.event.ContextRetrievalResponseEvent.EVENT_TYPE;
+    public static final String ModelRoutingEvent =
+            org.apache.flink.agents.api.event.ModelRoutingEvent.EVENT_TYPE;
 
     public static final String ShortTermWriteEvent =
             org.apache.flink.agents.api.event.ShortTermWriteEvent.EVENT_TYPE;
--- a/api/src/main/java/org/apache/flink/agents/api/EventType.java
+++ b/api/src/main/java/org/apache/flink/agents/api/EventType.java
@@ -72,6 +72,7 @@ public final class EventType {
                     Map.entry("ToolResponseEvent", ToolResponseEvent),
                     Map.entry("ContextRetrievalRequestEvent", ContextRetrievalRequestEvent),
                     Map.entry("ContextRetrievalResponseEvent", ContextRetrievalResponseEvent),
+                    Map.entry("ModelRoutingEvent", ModelRoutingEvent),
                     Map.entry("ShortTermWriteEvent", ShortTermWriteEvent),
                     Map.entry("ShortTermReadEvent", ShortTermReadEvent),
                     Map.entry("SensoryWriteEvent", SensoryWriteEvent),
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ModelRouter.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ModelRouter.java
@@ -205,6 +205,21 @@ public class ModelRouter extends Resource {
             if (strategy == null) {
                 throw new IllegalStateException("ModelRouter requires a strategy(...).");
             }
+            // Rule keys are candidate names; validate here, where both lists are in hand, so a
+            // typo fails at the registration call site instead of throwing per record at runtime.
+            if (RuleBasedRoutingStrategy.class.getName().equals(strategy.getClazz())) {
+                Object rules = strategy.getArguments().get("rules");
+                if (rules instanceof Map) {
+                    for (Object ruleKey : ((Map<?, ?>) rules).keySet()) {
+                        if (!candidates.contains(String.valueOf(ruleKey))) {
+                            throw new IllegalArgumentException(
+                                    String.format(
+                                            "Routing rule key '%s' is not one of the candidates %s.",
+                                            ruleKey, candidates));
+                        }
+                    }
+                }
+            }
             Map<String, Object> args = new HashMap<>();
             args.put("candidates", new ArrayList<>(candidates));
             if (!descriptions.isEmpty()) {
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ModelRouter.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ModelRouter.java
@@ -33,6 +33,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 /**
  * A framework resource that <b>selects</b> a concrete chat model for a request. It does not call
@@ -205,17 +207,37 @@ public class ModelRouter extends Resource {
             if (strategy == null) {
                 throw new IllegalStateException("ModelRouter requires a strategy(...).");
             }
-            // Rule keys are candidate names; validate here, where both lists are in hand, so a
-            // typo fails at the registration call site instead of throwing per record at runtime.
+            // Rule keys are candidate names and rule values are regex patterns; validate both
+            // here, where the full map is in hand, so a typo fails at the registration call site
+            // instead of throwing per record at runtime (an invalid pattern is never cached by
+            // the resource cache, so it would otherwise re-throw on every routed request).
             if (RuleBasedRoutingStrategy.class.getName().equals(strategy.getClazz())) {
                 Object rules = strategy.getArguments().get("rules");
                 if (rules instanceof Map) {
-                    for (Object ruleKey : ((Map<?, ?>) rules).keySet()) {
-                        if (!candidates.contains(String.valueOf(ruleKey))) {
+                    for (Map.Entry<?, ?> rule : ((Map<?, ?>) rules).entrySet()) {
+                        if (!candidates.contains(String.valueOf(rule.getKey()))) {
                             throw new IllegalArgumentException(
                                     String.format(
                                             "Routing rule key '%s' is not one of the candidates %s.",
-                                            ruleKey, candidates));
+                                            rule.getKey(), candidates));
+                        }
+                        if (!(rule.getValue() instanceof String)) {
+                            throw new IllegalArgumentException(
+                                    String.format(
+                                            "Routing rule pattern for candidate '%s' must be a non-null String, got %s.",
+                                            rule.getKey(),
+                                            rule.getValue() == null
+                                                    ? "null"
+                                                    : rule.getValue().getClass().getSimpleName()));
+                        }
+                        try {
+                            Pattern.compile((String) rule.getValue());
+                        } catch (PatternSyntaxException e) {
+                            throw new IllegalArgumentException(
+                                    String.format(
+                                            "Routing rule pattern '%s' for candidate '%s' is not a valid regex.",
+                                            rule.getValue(), rule.getKey()),
+                                    e);
                         }
                     }
                 }
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ModelRouter.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ModelRouter.java
@@ -1,0 +1,222 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.api.chat.model.routing;
+
+import org.apache.flink.agents.api.resource.Resource;
+import org.apache.flink.agents.api.resource.ResourceContext;
+import org.apache.flink.agents.api.resource.ResourceDescriptor;
+import org.apache.flink.agents.api.resource.ResourceType;
+
+import java.lang.reflect.Constructor;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * A framework resource that <b>selects</b> a concrete chat model for a request. It does not call
+ * the backend itself — {@code ChatModelAction} resolves the router, runs its {@link
+ * RoutingStrategy} to get a {@link RoutingDecision}, and then runs the normal chat path against the
+ * chosen model.
+ *
+ * <p>Built with the fluent {@link #of(String...)} builder, which produces a {@link
+ * ResourceDescriptor} the framework instantiates reflectively. The strategy is carried by class
+ * name + args (see {@link RoutingStrategyDescriptor}) so it is plan-serializable.
+ *
+ * <p>Abstain ({@link RoutingDecision#abstain()}) → {@link #getDefaultModel()}. A returned name that
+ * is not a candidate is an invalid decision and is failed clearly by the caller.
+ */
+public class ModelRouter extends Resource {
+
+    private final List<RoutingCandidate> candidates;
+    private final String defaultModel;
+    private final boolean fallbackEnabled;
+    private final RoutingStrategy strategy;
+
+    public ModelRouter(ResourceDescriptor descriptor, ResourceContext resourceContext)
+            throws Exception {
+        super(descriptor, resourceContext);
+        List<String> names = descriptor.getArgument("candidates");
+        if (names == null || names.isEmpty()) {
+            throw new IllegalArgumentException("ModelRouter requires at least one candidate.");
+        }
+        Map<String, String> descriptions =
+                descriptor.getArgument("candidate_descriptions", Collections.emptyMap());
+        List<RoutingCandidate> parsed = new ArrayList<>();
+        Set<String> uniqueNames = new LinkedHashSet<>();
+        for (String name : names) {
+            if (!uniqueNames.add(name)) {
+                throw new IllegalArgumentException(
+                        String.format("ModelRouter candidate '%s' is duplicated.", name));
+            }
+            parsed.add(new RoutingCandidate(name, descriptions.get(name)));
+        }
+        this.candidates = Collections.unmodifiableList(parsed);
+        this.defaultModel = descriptor.getArgument("default_model");
+        if (this.defaultModel != null && !isCandidate(this.defaultModel)) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "ModelRouter default model '%s' is not one of the candidates %s.",
+                            this.defaultModel, getCandidateNames()));
+        }
+        this.fallbackEnabled =
+                Boolean.TRUE.equals(descriptor.getArgument("fallback", Boolean.FALSE));
+        String strategyClazz = descriptor.getArgument("strategy_clazz");
+        Map<String, Object> strategyArgs =
+                descriptor.getArgument("strategy_args", Collections.emptyMap());
+        this.strategy = instantiateStrategy(strategyClazz, strategyArgs);
+    }
+
+    @SuppressWarnings("unchecked")
+    private static RoutingStrategy instantiateStrategy(String clazz, Map<String, Object> args)
+            throws Exception {
+        if (clazz == null || clazz.isEmpty()) {
+            throw new IllegalArgumentException("ModelRouter requires a routing strategy.");
+        }
+        Class<?> c = Class.forName(clazz, true, Thread.currentThread().getContextClassLoader());
+        try {
+            Constructor<?> ctor = c.getConstructor(Map.class);
+            return (RoutingStrategy) ctor.newInstance(args);
+        } catch (NoSuchMethodException noMapCtor) {
+            return (RoutingStrategy) c.getConstructor().newInstance();
+        }
+    }
+
+    /** Run the strategy for the given context. */
+    public RoutingDecision route(RoutingContext context) throws Exception {
+        return strategy.route(context);
+    }
+
+    public List<RoutingCandidate> getCandidates() {
+        return candidates;
+    }
+
+    public List<String> getCandidateNames() {
+        List<String> names = new ArrayList<>();
+        for (RoutingCandidate candidate : candidates) {
+            names.add(candidate.getName());
+        }
+        return names;
+    }
+
+    public Optional<String> getDefaultModel() {
+        return Optional.ofNullable(defaultModel);
+    }
+
+    public boolean isFallbackEnabled() {
+        return fallbackEnabled;
+    }
+
+    /** Whether the given model name is one of this router's candidates. */
+    public boolean isCandidate(String model) {
+        for (RoutingCandidate candidate : candidates) {
+            if (candidate.getName().equals(model)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public ResourceType getResourceType() {
+        return ResourceType.MODEL_ROUTER;
+    }
+
+    /**
+     * Start building a router over the given candidate model names (order matters for fallback).
+     */
+    public static Builder of(String... candidates) {
+        return new Builder(Arrays.asList(candidates));
+    }
+
+    /** Fluent builder that produces a {@link ResourceDescriptor} for a {@link ModelRouter}. */
+    public static final class Builder {
+        private final List<String> candidates;
+        private final Map<String, String> descriptions = new HashMap<>();
+        private RoutingStrategyDescriptor strategy;
+        private String defaultModel;
+        private boolean fallback = false;
+
+        private Builder(List<String> candidates) {
+            this.candidates = candidates;
+        }
+
+        public Builder strategy(RoutingStrategyDescriptor strategy) {
+            this.strategy = strategy;
+            return this;
+        }
+
+        /**
+         * Attach a human-readable description to a candidate, surfaced to strategies via {@link
+         * RoutingCandidate#getDescription()}. Descriptions are how semantic strategies — and future
+         * framework-managed LLM routing — learn what each candidate is for, so declare them here
+         * (once, on the router) rather than in per-strategy arguments.
+         */
+        public Builder describe(String candidate, String description) {
+            if (!candidates.contains(candidate)) {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "Cannot describe '%s': not one of the candidates %s.",
+                                candidate, candidates));
+            }
+            descriptions.put(candidate, description);
+            return this;
+        }
+
+        public Builder defaultModel(String defaultModel) {
+            this.defaultModel = defaultModel;
+            return this;
+        }
+
+        /**
+         * Whether to try remaining candidates (in declaration order) after the selected model has
+         * exhausted its own retry policy. Applies to the initial routed request only; tool-call
+         * rounds keep the already-selected model for conversation coherence. Fallback outcomes are
+         * recorded on the response ({@code model_routing} extra args) and as a second {@code
+         * ModelRoutingEvent} with source {@code fallback}.
+         */
+        public Builder fallback(boolean fallback) {
+            this.fallback = fallback;
+            return this;
+        }
+
+        public ResourceDescriptor build() {
+            if (strategy == null) {
+                throw new IllegalStateException("ModelRouter requires a strategy(...).");
+            }
+            Map<String, Object> args = new HashMap<>();
+            args.put("candidates", new ArrayList<>(candidates));
+            if (!descriptions.isEmpty()) {
+                args.put("candidate_descriptions", new HashMap<>(descriptions));
+            }
+            if (defaultModel != null) {
+                args.put("default_model", defaultModel);
+            }
+            args.put("fallback", fallback);
+            args.put("strategy_clazz", strategy.getClazz());
+            args.put("strategy_args", strategy.getArguments());
+            return new ResourceDescriptor(ModelRouter.class.getName(), args);
+        }
+    }
+}
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
@@ -52,10 +52,13 @@ public final class RoutingContext {
             List<RoutingCandidate> candidates) {
         this.requestId = requestId;
         this.router = router;
+        // Deep copy: the wrapping list is unmodifiable, but ChatMessage is mutable and the
+        // caller passes the same instances that go to the model — a strategy calling
+        // setContent(...) on a shallow copy would silently rewrite the prompt actually sent.
         this.messages =
                 messages == null
                         ? Collections.emptyList()
-                        : Collections.unmodifiableList(new ArrayList<>(messages));
+                        : Collections.unmodifiableList(deepCopy(messages));
         this.promptArgs =
                 promptArgs == null
                         ? Collections.emptyMap()
@@ -64,6 +67,24 @@ public final class RoutingContext {
                 candidates == null
                         ? Collections.emptyList()
                         : Collections.unmodifiableList(new ArrayList<>(candidates));
+    }
+
+    private static List<ChatMessage> deepCopy(List<ChatMessage> messages) {
+        List<ChatMessage> copy = new ArrayList<>(messages.size());
+        for (ChatMessage m : messages) {
+            if (m == null) {
+                continue;
+            }
+            // ChatMessage's constructor copies extraArgs but stores toolCalls by reference, so
+            // copy the list AND each tool-call map — otherwise a strategy could still mutate the
+            // tool calls of the message actually sent.
+            List<Map<String, Object>> toolCalls = new ArrayList<>(m.getToolCalls().size());
+            for (Map<String, Object> call : m.getToolCalls()) {
+                toolCalls.add(call == null ? null : new HashMap<>(call));
+            }
+            copy.add(new ChatMessage(m.getRole(), m.getContent(), toolCalls, m.getExtraArgs()));
+        }
+        return copy;
     }
 
     /**
--- a/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
+++ b/api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
@@ -1,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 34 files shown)


plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -400,56 +669,199 @@ public ChatMessage call() throws Exception {
...
+                    public RoutingDecision call() throws Exception {
+                        // Timed inside the durable call so the latency is persisted with the
+                        // decision: a replayed run reports the original strategy wall time.
+                        long start = System.nanoTime();
+                        RoutingDecision decision = router.route(routingContext);
+                        return decision.withDecisionMs((System.nanoTime() - start) / 1_000_000.0);
+                    }
+                };
+
+        RoutingDecision decision = ctx.durableExecute(routeCallable);
```
weiqingy, line 817: `resolveRouter` (781-863) has no try/catch and runs before `chat()`, so a strategy exception escapes `error-handling-strategy`: a job set to `ignore` still dies.

It's reachable from the built-in. `RuleBasedRoutingStrategy.java:61-66` throws on a rule key that isn't a candidate, and `Builder` validates names in `describe()` (`ModelRouter.java:176-185`) but not in `strategy()`, so `Strategies.rules(Map.of("bg", "..."))` builds fine and throws per record.

Could rule keys be checked in `build()`, where both lists are already in hand? And should a strategy failure honor `error-handling-strategy`, or is fail-fast the contract?

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -344,15 +505,129 @@ public static void chat(
...
+                                    initialRequestId,
+                                    result.response,
+                                    totalRetryCount,
+                                    totalRetryWaitSec));
+                }
+                return;
+            } catch (ChatAttemptFailed e) {
+                recordAttemptRetryStats(
+                        ctx, initialRequestId, e.chatModel, e.retryCount, e.totalRetryWaitSec);
+                lastError = e.error;
```
weiqingy, line 584: `lastError = e.error` overwrites and 598 throws only the last, so with `[A, B, C]` all failing you see C's. The debug line at 585 logs the model and messages but not the error, so A's and B's are gone, including the routed model's own failure. The second `ModelRoutingEvent` only fires on success (536-549), so exhaustion records nothing about what was tried.

Something like this before the overwrite, if it helps:

```java
if (lastError != null && lastError != e.error) {
    e.error.addSuppressed(lastError);
}
```

`fallbackExhaustedRethrows` can't tell today: both scripted errors contain "down" (`ChatModelActionRoutingTest.java:408-410`). Would distinct messages help?

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -344,15 +505,129 @@ public static void chat(
...
+            List<ChatMessage> messages,
+            Map<String, Object> promptArgs,
+            @Nullable Object outputSchema,
+            RunnerContext ctx,
+            Agent.ErrorHandlingStrategy strategy,
+            int numRetries,
+            int retryWaitIntervalSec)
+            throws ChatAttemptFailed, Exception {
+        BaseChatModelSetup chatModel =
+                (BaseChatModelSetup) ctx.getResource(model, ResourceType.CHAT_MODEL);
```
weiqingy, line 614: This lookup sits outside the try/catch that produces `ChatAttemptFailed` (644-676), and the call site catches only that (581). So an unresolvable candidate B propagates straight out of `chat()`: A's real failure is discarded, the caller sees "resource not found", and the candidate loop is bypassed, including under `IGNORE`.

Nothing catches the typo upstream either, since `ModelRouter`'s constructor (57-89) never touches the registry.

Could the lookup move inside the conversion, so an unresolvable candidate just counts as that candidate failing? And is there room to cross-validate candidate names at plan construction, the way `Agent.java:135-139` argues for the clash check?

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
```diff
@@ -0,0 +1,123 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * Read-only view a {@link RoutingStrategy} sees when deciding which model to route to.
```
weiqingy, line 32: The collections are unmodifiable (55-58, 59-62) but the copy is shallow, and `ChatMessage` is mutable on main. These are the same instances that go to the model: `event.getMessages()` is passed into `resolveRouter` (758) and into `chat(...)` (764). So a `setContent(...)` inside a strategy rewrites the prompt that's actually sent, with no event and no trace.

The strategy's reach to other resources is already closed by the type. This is the same thing one level down, and narrowing it is free before v1 ships and breaking after.

`firstUserMessage()` and `lastUserMessage()` (100, 114) are what a selection strategy needs. Deep-copy the list, or drop `getMessages()` until something needs more than the text?

api/src/main/java/org/apache/flink/agents/api/EventType.java
```diff
@@ -39,6 +39,8 @@ public final class EventType {
             org.apache.flink.agents.api.event.ContextRetrievalRequestEvent.EVENT_TYPE;
     public static final String ContextRetrievalResponseEvent =
             org.apache.flink.agents.api.event.ContextRetrievalResponseEvent.EVENT_TYPE;
+    public static final String ModelRoutingEvent =
```
weiqingy, line 42: After a rebase, `ModelRoutingEvent` won't be in `main`'s `ALL_CONSTANTS` (`EventType.java:45-54`), and `ConditionExpressionValidator.java:125` rejects any `EventType.X` that isn't a key of it. So `@Action("type == EventType.ModelRoutingEvent && ...")` compiles, then fails plan validation. Plain `@Action(EventType.ModelRoutingEvent)` is unaffected.

Worth adding the entry when you rebase?

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RuleBasedRoutingStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+    private final Map<String, Pattern> rules;
+
+    @SuppressWarnings("unchecked")
+    public RuleBasedRoutingStrategy(Map<String, Object> args) {
+        this.rules = new LinkedHashMap<>();
+        Object raw = args == null ? null : args.get("rules");
+        if (raw instanceof Map) {
+            for (Map.Entry<String, ?> entry : ((Map<String, ?>) raw).entrySet()) {
+                String candidate = entry.getKey();
+                String regex = String.valueOf(entry.getValue());
```
weiqingy, line 47: `String.valueOf(entry.getValue())` returns `"null"` for a null value, never `null`, so the `regex != null` guard below is dead and the entry compiles to `Pattern.compile("null", CASE_INSENSITIVE)`. Any message containing "null" then routes there. Non-String values coerce silently too, and `rules` arrives as `Map<String,Object>` off the plan JSON.

Should a null or non-String value be rejected instead?

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -51,14 +55,43 @@
...
+ *
+ * <h2>Model routing overview</h2>
+ *
+ * <p>When a {@link ChatRequestEvent} names a {@code MODEL_ROUTER} instead of a chat model, this
+ * action layers five jobs on top of the normal chat path; each is localized to one place:
+ *
+ * <ol>
+ *   <li><b>Decide</b> — {@code resolveRouter} runs the router's strategy and normalizes the result
+ *       (abstain → default model; non-candidate → fail).
+ *   <li><b>Durably</b> — the strategy runs inside a durable call ({@code "route:<requestId>:
```
weiqingy, line 69: The guarantee reads unconditional here (69-71) and on `RoutingDecision.java:93-95`, but the journal is off by default: `actionStateStoreBackend` defaults to null (`AgentConfigOptions.java:65-66`), and `setupDurableExecutionContext` returns early when the store is null (`DurableExecutionManager.java:289-292`).

The default path is still self-consistent, since the decision and the chat call re-run together. It's the javadoc that promises more than a default deployment gets. Worth a qualifying clause?

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -344,15 +505,129 @@ public static void chat(
...
+                    ctx.sendEvent(
+                            new ChatResponseEvent(
+                                    initialRequestId,
+                                    result.response,
+                                    totalRetryCount,
+                                    totalRetryWaitSec));
+                }
+                return;
+            } catch (ChatAttemptFailed e) {
+                recordAttemptRetryStats(
```
weiqingy, line 582: Retry metrics moved: `origin/main:433-434` recorded once on the final response with cumulative totals; this records per attempt (681-698), including on the failure path. Totals over a completed request are unchanged, but requests that ultimately throw now increment where they didn't.

`AGENTS.md` asks for compatibility impact in the description. Worth a line?

weiqingy (commented):

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
purushah: Great catch — and you found *two* holes in one comment (the per-record throw AND the policy bypass). Fixed both: `build()` validates rule keys so the typo fails at registration, and `resolveRouter` failures now honor `error-handling-strategy` — deliberately without retries, since v1 strategies are CPU-only. Tests: `builderRejectsRuleKeyThatIsNotACandidate`, `strategyFailureIsIgnoredUnderIgnorePolicy`, `strategyFailurePropagatesUnderDefaultPolicy`.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
purushah: Really appreciate you noticing the test itself couldn't detect this (both scripted errors shared "down") — that's a level of review rigor we rarely get. Applied your suggestion: errors chain via `addSuppressed`, exhaustion WARNs with the tried list, and the test now uses distinct markers so it fails against the old behavior.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
purushah: This was the sharpest catch of the review — the failure mode (real error discarded, confusing "resource not found" for a model the user never picked, both fallback and IGNORE bypassed) is exactly the worst debugging experience. Fixed: the lookup moved inside the attempt conversion, so it counts as a failed attempt and flows through fallback/IGNORE with prior errors preserved. (Side effect: a plain chat with a missing model now honors IGNORE too — we think that's more consistent.) Test: `unresolvableCandidateCountsAsFailedAttemptAndFallsBack`.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
purushah: Good catch — fixed with a true deep copy, including the toolCalls list and inner maps (`ChatMessage`'s ctor stores toolCalls by reference; re-verifying your comment caught that second hole too, so this one paid off twice). Tests: `routingContextMessagesAreDeepCopied`, `routingContextToolCallsAreDeepCopiedToo`.

api/src/main/java/org/apache/flink/agents/api/EventType.java
purushah: Right — will register `ModelRoutingEvent` (plus a condition-expression test) during the rebase onto current main.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RuleBasedRoutingStrategy.java
purushah: Nice spot on a classic `String.valueOf(null)` trap. Fixed: null/non-String rule values are rejected with a descriptive error. Tests: `ruleStrategyRejectsNullRuleValue`, `ruleStrategyRejectsNonStringRuleValue`.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
purushah: Fair — the guarantee is conditional and now reads that way: replay requires a configured action-state store (`actionStateStoreBackend`); without one the decision re-executes. Doc-only.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
purushah: Good eye for catching the semantic shift — it's intentional (per-attempt recording makes fallback attribution land on the model that served the attempt), and it's now documented in a compat note on `recordAttemptRetryStats` plus the PR description.

purushah:
Thanks @weiqingy for the sharp review — all eight comments are addressed in the follow-up commits with new tests. The `EventType` registration will come with the rebase onto main.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -503,7 +971,14 @@ private static void processToolResponse(ToolResponseEvent event, RunnerContext c
                         Collections.emptyList(),
                         toolResponseMessages);
 
-        chat(initialRequestId, model, messages, promptArgs, outputSchema, ctx);
+        // Tool rounds reuse the already-selected concrete model (no re-routing); if the initial
+        // request was routed, carry its routing metadata onto the eventual final response.
+        Map<String, Object> routingMetadata = (Map<String, Object>) context.get(ROUTING);
```
wenjin272: Do we need to carry `routingMetadata` through every tool-request context? The concrete model used for subsequent tool rounds is already stored separately in `MODEL`; `routingMetadata` does not participate in model selection and is only used to attach `model_routing` to the eventual `ChatResponseEvent`.

Since its lifetime is the whole ReAct loop, could we store it once in an initial-request context keyed by `initialRequestId`, then retrieve it only when producing the final response? This would avoid repeatedly copying it across tool rounds, remove the special `RoutingSelection.carried(...)` state, and keep observability metadata out of intermediate `ChatMessage.extraArgs`. The context could then be cleaned up when the loop completes.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -67,6 +105,121 @@ public class ChatModelAction {
 
     private static final ObjectMapper mapper = new ObjectMapper();
 
+    private static final class RoutingSelection {
```
wenjin272: `ChatModelAction` has grown to roughly 1,000 lines and now owns several distinct responsibilities: route resolution, concrete-model invocation/retry, fallback orchestration, routing metadata propagation, and tool-loop state management.

Could we extract the routing-specific pieces into a package-private component such as `ModelRoutingResolver` / `ResolvedModelRoute`? `RoutingSelection`, `resolveRouter`, candidate ordering, routed durable-call IDs, and routing metadata construction form a cohesive unit. Similarly, `chatWithRetries` together with `ChatAttemptResult` / `ChatAttemptFailed` could become a concrete-model invoker. This would leave `ChatModelAction` responsible primarily for event orchestration while keeping the routing implementation easier to review and extend.

wenjin272 (commented):
Thanks for pushing this design forward and for addressing the earlier feedback. I have two structural suggestions around tool-round state and the organization of `ChatModelAction`.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
purushah: Good suggestion — implemented. The metadata is now stored once, keyed by the initial request id, attached to the loop's final response, and cleaned up on every loop exit. `RoutingSelection.carried(...)` is gone and intermediate messages are no longer stamped; the tool-round tests assert both (including a mutation-verified cleanup test for the abandoned-loop path).

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
purushah: Done — extracted `ResolvedModelRoute`, `ModelRoutingResolver`, and `ChatModelInvoker` as you sketched. `ChatModelAction` is down to ~690 lines of event orchestration. No behavior change; full suites green. One heads-up: the retry warning now logs under `ChatModelInvoker`'s logger, so log alerting keyed to the old logger name would need updating.

purushah:
Thanks @wenjin272 — both suggestions implemented in the follow-up commits.

api/src/main/java/org/apache/flink/agents/api/resource/ResourceType.java
```diff
@@ -32,7 +32,8 @@ public enum ResourceType {
     PROMPT("prompt"),
     TOOL("tool"),
     MCP_SERVER("mcp_server"),
-    SKILLS("skills");
+    SKILLS("skills"),
+    MODEL_ROUTER("model_router");
```
weiqingy, line 36: Java gets `MODEL_ROUTER` here, but the Python enum doesn't have it (`python/flink_agents/api/resource.py:38-46`), and no Python files changed in this PR. The Java plan is serialized over to Python and parsed against `Dict[ResourceType, ...]` (`python/flink_agents/plan/agent_plan.py:77`), so a Java agent using a router plus any Python action would fail at operator open with a pydantic `ValidationError`. Full Python routing is a follow-up per the description. Would adding just this one enum member be enough to keep mixed jobs starting, or is there a reason to hold it for the follow-up?

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -352,106 +390,220 @@ public static void chat(
...
-                    messages,
-                    promptArgs,
-                    outputSchema,
-                    ctx);
-        } else {
-            Map<String, Long> retryStats = getRetryStats(ctx.getSensoryMemory(), initialRequestId);
-            int totalRetryCount = retryStats.get(TOTAL_RETRY_COUNT).intValue();
-            int totalRetryWaitSec = retryStats.get(TOTAL_RETRY_WAIT_SEC).intValue();
+    /**
+     * Compatibility note: retry metrics are recorded per attempt (including attempts on the failure
```
weiqingy, line 520: The description still says the opposite of this — line 58 reads "Existing agents are unaffected." Three things do change for a plain-chat agent, all only when `error-handling-strategy` isn't the default: with `retry`, retry metrics now count per attempt including failures, and the retry WARN moved to `ChatModelInvoker`'s logger; with `ignore`, a request naming a missing chat model is warned and dropped instead of failing the job. Default `FAIL` really is unaffected, so the blast radius is small. `AGENTS.md` asks for compatibility impact in the description — is line 58 still the message you want, or should it name those three?

plan/src/main/java/org/apache/flink/agents/plan/actions/ModelRoutingResolver.java
```diff
@@ -0,0 +1,143 @@
...
+        DurableCallable<RoutingDecision> routeCallable =
+                new DurableCallable<>() {
+                    @Override
+                    public String getId() {
+                        // Deterministic across recovery re-processing: the durable store already
+                        // scopes call results by (key, sequence number, event, action), so the id
+                        // must NOT embed the request id — event ids are regenerated when Flink
+                        // rolls back and re-processes, and a non-deterministic id turns every
+                        // replay lookup into a miss (measured: 0/138 decisions replayed).
+                        return "route:" + model;
```
weiqingy, line 79: nit: the description hasn't caught up with this line — body line 26 still describes the id as `"route:<requestId>:<router>"`, the form this commit removed. Nothing is broken, since this line and the javadoc at `:36` are already right. It's just that durability is the claim people read closest here, so worth syncing line 26 too?

plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelActionRoutingTest.java
```diff
@@ -0,0 +1,725 @@
...
+        assertThat(ctx.hasChatResponse()).isTrue();
+    }
+
+    @Test
+    void directModelKeepsLegacyDurableCallId() throws Exception {
+        FakeRunnerContext ctx = new FakeRunnerContext(null);
+        ChatModelAction.processChatRequestOrToolResponse(
+                new ChatRequestEvent("plain", List.of(ChatMessage.user("hi"))), ctx);
+        // a non-router request must keep the unchanged legacy durable chat-call id
+        assertThat(ctx.durableCallIds).containsExactly("chat");
```
weiqingy, line 417: This is the only place `durableCallIds` gets asserted, and it pins the legacy `"chat"` id. The routed ones, `route:<router>` and `chat:<router>:<candidate>`, go through the same list (`:207`, `:213`) but nothing checks them — including `fallsBackToRemainingCandidateWhenSelectedModelFails` (`:367`), where the two candidates' journal entries need to stay distinct. Since `6ef088fae` had to change that format once already, could the routed ids get an assertion here too?

plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelActionRoutingTest.java
```diff
@@ -0,0 +1,725 @@
...
+            return this;
+        }
+
+        /** Marks a chat-model name whose resource lookup fails (e.g. a typo'd candidate). */
+        FakeRunnerContext unresolvable(String name) {
+            unresolvable.add(name);
+            return this;
+        }
+
+        FakeRunnerContext withErrorHandling(Agent.ErrorHandlingStrategy strategy) {
```
weiqingy, line 135: `withErrorHandling` only ever gets `IGNORE` (`:511`, `:634`), and `MAX_RETRIES` isn't in this file at all, so every routing test runs with `numRetries == 0`. Nothing here would notice if retry and fallback swapped order, even though the class javadoc calls that ordering a guarantee — selected model with its full retry budget first, then the rest (`ChatModelAction.java:73-75`). `FakeChatModel` already takes a scripted sequence of outcomes. Is one `RETRY` case with two failures scripted enough to pin it?

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
```diff
@@ -617,11 +617,39 @@ private ResourceProvider createResourceProvider(
...
                 .put(provider.getName(), provider);
     }
 
+    /**
+     * A name must not be registered as both a {@link ResourceType#CHAT_MODEL} and a {@link
+     * ResourceType#MODEL_ROUTER}: an agent references either by putting it in {@code
+     * ChatRequestEvent.model}, so a name that is both would resolve ambiguously. Fail clearly at
+     * plan-construction time rather than at request time.
+     */
+    private void checkNoRouterModelNameClash(ResourceProvider provider) {
```
weiqingy, line 673: `checkNoRouterModelNameClash` doesn't have a test yet. `RoutingResourceValidationTest.java:44-79` exercises the two `addResource` call sites, but each of those only sees its own map. Registering the same name in both places — `env.addResource("smart", CHAT_MODEL, ...)` and `agent.addResource("smart", MODEL_ROUTER, ...)` — passes both checks, and `Agent.addResourcesIfAbsent` (`Agent.java:104`) merges them without one, so this is what actually catches it. That's the case `Agent.java:139` calls out as the backstop. Should that path get a case of its own?

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
```diff
@@ -0,0 +1,152 @@
...
+        this.requestId = requestId;
+        this.router = router;
+        // Deep copy: the wrapping list is unmodifiable, but ChatMessage is mutable and the
+        // caller passes the same instances that go to the model — a strategy calling
+        // setContent(...) on a shallow copy would silently rewrite the prompt actually sent.
+        this.messages =
+                messages == null
+                        ? Collections.emptyList()
+                        : Collections.unmodifiableList(deepCopy(messages));
+        this.promptArgs =
```
weiqingy, line 69: nit: the javadoc calls this a read-only view (`:32`), but each copy only goes one level — `promptArgs` here, and in `deepCopy` both `extraArgs` (via `ChatMessage`'s constructor, `ChatMessage.java:72`) and each tool-call map (`:90`). Nested values stay shared with what gets sent (`ChatModelAction.java:604-610`). No in-tree strategy mutates one, so this is about the shape the API locks in. Is the top-level copy deliberate, or would deep-copying these fit better?

weiqingy (commented):
Thanks for addressing the comments. While reviewing the reordered code, a few inline comments below.

api/src/main/java/org/apache/flink/agents/api/resource/ResourceType.java
purushah: Done — `MODEL_ROUTER` added to the Python enum with a cross-language round-trip test, so mixed jobs deserialize; full Python routing stays the follow-up (which should also make Python-side `add_resource(MODEL_ROUTER)` an explicit not-yet-supported error — it currently drops silently).

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
purushah: Fair — the description now has a "Compatibility impact" section naming exactly your three; default `FAIL` unchanged.

plan/src/main/java/org/apache/flink/agents/plan/actions/ModelRoutingResolver.java
purushah: Synced — description now shows `"route:<router>"`.

plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelActionRoutingTest.java
purushah: Added: `routedRequestUsesRoutedDurableCallIds` pins `route:<router>` + `chat:<router>:<candidate>`, and the fallback test asserts the two candidates' distinct ids in order.

plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelActionRoutingTest.java
purushah: Added `retryBudgetRunsBeforeFallback`: the selected model recovers on its retry, the fallback candidate is never resolved, no fallback event — the ordering is pinned.

plan/src/main/java/org/apache/flink/agents/plan/AgentPlan.java
purushah: Added `AgentPlanRoutingBackstopTest`: same name via `addResource` (router) + `addResourcesIfAbsent` (chat model) bypasses both per-call checks, and plan construction throws.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
purushah: Deliberate, and now documented in the javadoc: copies are one level deep, nested values shared, and the SPI forbids mutation — deeper copies would tax every decision to guard a forbidden case. Happy to revisit if a real strategy needs more.

purushah:
Thanks @weiqingy for another careful pass — all seven addressed: the Python enum member is in (mixed jobs deserialize; full Python routing stays the follow-up), the three test gaps are covered, the `RoutingContext` boundary is documented as deliberate, and the PR description now carries the compatibility-impact section and the corrected durable-id format.

wenjin272 (approved):
LGTM from my side. @weiqingy, do you have any remaining comments?

weiqingy (commented):
Thanks for addressing the comments. Two small things inline, neither blocking. Otherwise, LGTM.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ModelRouter.java
```diff
@@ -0,0 +1,237 @@
...
+            this.fallback = fallback;
+            return this;
+        }
+
+        public ResourceDescriptor build() {
+            if (strategy == null) {
+                throw new IllegalStateException("ModelRouter requires a strategy(...).");
+            }
+            // Rule keys are candidate names; validate here, where both lists are in hand, so a
+            // typo fails at the registration call site instead of throwing per record at runtime.
```
weiqingy: The comment here says rule keys are validated at build time `so a typo fails at the registration call site instead of throwing per record at runtime`. The rule values have the same shape and no check: `Pattern.compile` runs only in the strategy constructor (`RuleBasedRoutingStrategy.java:62`), reached from `ModelRouter:88` when the resource is provided, and `ResourceCache` caches only after `provide()` returns (`ResourceCache.java:144`). A `PatternSyntaxException` therefore never caches and re-throws per routed request: under the default `FAIL` the job dies on the first one, under `ignore` requests through that router are dropped with a WARN while the job stays RUNNING. Would compiling the patterns in this same block move that to `build()` too?

plan/src/main/java/org/apache/flink/agents/plan/actions/ResolvedModelRoute.java
```diff
@@ -0,0 +1,134 @@
...
+    Map<String, Object> buildResponseMetadata(String finalModel, List<String> triedModels) {
+        boolean fallbackAttempted = !finalModel.equals(this.selectedModel);
+        List<String> fallbackModelsTried = new ArrayList<>();
+        for (int i = 1; i < triedModels.size(); i++) {
+            fallbackModelsTried.add(triedModels.get(i));
+        }
+        Map<String, Object> routing = new LinkedHashMap<>();
+        routing.put("router", this.requestedModel);
+        routing.put("selected_model", this.selectedModel);
+        routing.put("initial_selected_model", this.selectedModel);
```
weiqingy: nit: this and `selected_model` above are both `this.selectedModel`, and `initial_selected_model` has no other reference in the PR. Nothing reads it and no test asserts it. The fallback winner is already `final_model` on the next line. Removing a key from a shipped payload is a compat change later. Is the second key meant to differ from the first on some path, or can it go?

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/ModelRouter.java
purushah: Good catch on the caching consequence — fixed: `build()` now compiles every rule pattern (and rejects null/non-String values) in the same block as the key check, so a malformed regex fails at the registration call site. Two new builder tests cover it.

plan/src/main/java/org/apache/flink/agents/plan/actions/ResolvedModelRoute.java
purushah: You are right — it duplicated `selected_model` on every path (fallback winners are `final_model`). Removed before it ships and becomes a compat constraint.

weiqingy (commented):
Both addressed at `e95f20bd`. No remaining comments from me.

wenjin272:
Thanks for taking this on @purushah. I noticed that this feature has the design discussion (#897) and implementation PR (#964), but no corresponding issue. Could you please create one summarizing the agreed scope, link both the discussion and this PR? This will make the feature easier to track in the project history and release notes.

purushah:
Thanks @wenjin272 and @weiqingy for the careful reviews and the merge — the feature came out much stronger.

A heads-up on what's next: the `Strategies.llm(...)` follow-up we agreed to defer in #897 — framework-managed LLM-as-judge, where the engine runs the judge call on its own durable/metered/observable chat path — has been in the works alongside this review cycle for a while now, and it's nearly ready. The shape is exactly what the #897 thread settled on: the judge executes under a durable `"judge:<router>"` call with engine retries and token attribution, verdicts are constrained to candidate names, judge failures honor the error-handling strategy, and a mistyped judge model fails at plan construction rather than at runtime. I'll open the PR soon.

Issue #968, opened 2026-08-05
[Tech Debt] The coding-agent skill's bundled YAML schema can drift from docs/yaml-schema.json unnoticed


weiqingy:
### Background

`docs/yaml-schema.json` is exported from the Pydantic specs (`python -m flink_agents.api.yaml.specs`), and two checks already keep it honest: `python/flink_agents/api/yaml/tests/test_specs.py:377` asserts it matches the Pydantic export, and `api/src/test/java/org/apache/flink/agents/api/yaml/spec/SchemaParityTest.java:52` asserts it matches the Java POJOs.

The coding-agent skill added in #918 ships a third copy so it can answer YAML questions offline, plus a record of where that copy came from:

- `dev/agent-skills/flink-agents-dev/assets/yaml-schema.json`, byte-identical to `docs/yaml-schema.json` today
- `dev/agent-skills/flink-agents-dev/assets/yaml-contracts.yaml`, whose `main` contract records `blob_sha: 183cc7ac800cf97d030b75b166ea8146368747d5`

At `bc55d97e` all three agree. `git rev-parse HEAD:docs/yaml-schema.json` returns that same SHA.

### The gap

Neither existing check extends past `docs/yaml-schema.json`, and nothing regenerates the bundled copy. Regenerating the schema and forgetting the skill assets leaves no failing test, so the drift is only found by someone reading both files.

Two distinct things go stale, with different consequences.

The copy itself goes stale, and an agent consulting it is then taught a schema the repository no longer has. That is the failure the skill exists to prevent.

The recorded `blob_sha` also goes stale, which is the quieter half. `SKILL.md:45` tells agents the bundled main schema "describes only the repository revision that published this skill", and `references/local-development.md:62` says a source checkout should fall back to the bundled schema "only when they describe the same revision". Both instructions are decided by that SHA, so a stale value does not just lag. It asserts a provenance that is false, and it does so in the exact place an agent goes to decide whether the bundled copy is safe to trust.

### Proposal

A dependency-free check under `tools/`, run from the existing `lint` job next to `Check AGENTS.md freshness`, asserting that the bundled copy and the recorded `blob_sha` both match `docs/yaml-schema.json`. This mirrors `tools/check-agents-md.py`, which guards the same class of hand-copied fact.

The failure message should print the two edits needed, since the fix is mechanical and the person hitting it is usually mid-way through an unrelated schema change.

### Out of scope

[...]

PR #969, opened 2026-08-05
[ci] Guard the coding-agent skill's bundled YAML schema against drift

weiqingy:
Linked issue: #968

### Purpose of change

The `flink-agents-dev` skill ships a copy of `docs/yaml-schema.json` at `dev/agent-skills/flink-agents-dev/assets/yaml-schema.json`, and records the blob SHA it was taken from in `assets/yaml-contracts.yaml`. Nothing regenerates either one, and the two existing schema checks (`test_specs.py`, `SchemaParityTest`) stop at `docs/yaml-schema.json`, so re-exporting the schema leaves both stale with no failing test.

The stale copy teaches agents a schema the repository no longer has. The stale `blob_sha` is the quieter half: `SKILL.md` and `references/local-development.md` both instruct agents to trust the bundled schema only when it describes the same revision as the checkout, and that comparison is decided by exactly this value.

`tools/check-skill-schema.py` asserts the bundled copy and the recorded SHA both match `docs/yaml-schema.json`. It runs from the existing `lint` job next to `Check AGENTS.md freshness`, scoped to `main` pushes and PRs targeting `main` since the bundled skill only exists on `main`. Its Python side is regex-only and stdlib-only for the same reason `check-agents-md.py` is: that job has no Python environment beyond the interpreter.

The SHA itself comes from `git hash-object` rather than from hashing the file's bytes, so it goes through git's clean filters. A worktree-content hash is only the blob SHA where no filter applies: under `core.autocrlf` these JSON files are checked out CRLF, and the check would then report a current pin as stale and prescribe a SHA that the Linux CI rejects.

The versioned schemas beside it (`assets/yaml-schemas/release-0.3.0.json`) pin released refs and are expected to differ, so only the unversioned `main` contract is checked.

`docs/content/docs/development/yaml.md` documents the regeneration command, so it now also says to refresh the bundled copy. Without that, following the documented steps literally leaves CI red.

### Tests

No unit test. The check is itself the test, so it was verified by mutation instead: each pin was desynced independently and the check confirmed red, then restored.

| Mutation | Expected | Result |
|---|---|---|
| unmodified tree | green | green |
| `docs/yaml-schema.json` edited, both copies left behind | red, 2 failures | red, both reported |
| bundled copy edited alone | red, 1 failure | red |
| `blob_sha` edited alone | red, 1 failure | red |
| `main` contract loses its `blob_sha`, `0.3.0` keeps one | red, must not read the sibling SHA | red, `'main' contract has no blob_sha` |
| bundled copy deleted | red | red, names the missing path |
| `release-0.3.0.json` edited | green, out of scope | green |

The fifth row is why the manifest is read with a bounded match rather than a document-wide search for `blob_sha`. An unbounded search falls through to the `0.3.0` contract's SHA and the guard passes on a manifest that has lost the value it is supposed to check.

The hashing path was checked separately in a scratch repository with `core.autocrlf=true`, committing LF content and checking it back out: the byte hash and the blob SHA diverge there, and only `git hash-object` returns the committed blob. All seven rows above were re-run afterwards and are unchanged. Two further cases were covered by hand, since the check now depends on git being present: run outside a git repository it still reports the correct SHA, and with git off `PATH` it exits with a plain message rather than a traceback.

Also run locally: `tools/check-license.sh` (RAT passed), `python3 tools/check-agents-md.py`.

### API

No public API change. New file is a repository tooling script.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,8 @@ jobs:
         run: ./tools/lint.sh -c
       - name: Check AGENTS.md freshness
         run: python3 tools/check-agents-md.py
+      - name: Check bundled YAML schema freshness
+        run: python3 tools/check-skill-schema.py
 
   install_sh_tests:
     name: install.sh tests (${{ matrix.os }})
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ jobs:
       - name: Check AGENTS.md freshness
         run: python3 tools/check-agents-md.py
       - name: Check bundled YAML schema freshness
+        if: github.ref == 'refs/heads/main' || github.base_ref == 'main'
         run: python3 tools/check-skill-schema.py
 
   install_sh_tests:
--- a/tools/check-skill-schema.py
+++ b/tools/check-skill-schema.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+#################################################################################
+"""Assert the coding-agent skill still bundles the current YAML schema.
+
+The skill ships a copy of docs/yaml-schema.json so it can answer YAML questions
+offline, and records in yaml-contracts.yaml the git blob SHA that copy was taken
+from. Nothing regenerates either one, so both drift silently the moment the
+schema is re-exported. A stale copy teaches agents a schema the repository no
+longer has, and a stale blob SHA misreports which revision the copy describes.
+
+Only the unversioned "main" contract is checked. The versioned schemas beside it
+pin released refs, so they are expected to differ from the working tree.
+
+Regex-only and dependency-free so it runs on any Python 3 without a build step.
+"""
+
+import hashlib
+import re
+import sys
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+DOCS_SCHEMA = REPO_ROOT / "docs" / "yaml-schema.json"
+ASSETS = REPO_ROOT / "dev" / "agent-skills" / "flink-agents-dev" / "assets"
+BUNDLED_SCHEMA = ASSETS / "yaml-schema.json"
+MANIFEST = ASSETS / "yaml-contracts.yaml"
+
+
+def blob_sha(path: Path) -> str:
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -49,6 +49,8 @@ jobs:
         run: ./tools/lint.sh -c
       - name: Check AGENTS.md freshness
         run: python3 tools/check-agents-md.py
+      - name: Check bundled YAML schema freshness
+        run: python3 tools/check-skill-schema.py
```
wenjin272, line 54: This workflow also runs for `release-*` branches, but the new step always validates the contract whose `source.ref` is `main`. If a schema change is backported to a release branch, the check either leaves that branch's CI red or asks us to record the release branch's blob SHA as `main` provenance. This is a realistic path: `release-0.3` has already received schema-changing backports. Could we restrict this step to `main` pushes and PRs targeting `main`, or make contract selection branch-aware?

tools/check-skill-schema.py
```diff
@@ -0,0 +1,101 @@
...
+DOCS_SCHEMA = REPO_ROOT / "docs" / "yaml-schema.json"
+ASSETS = REPO_ROOT / "dev" / "agent-skills" / "flink-agents-dev" / "assets"
+BUNDLED_SCHEMA = ASSETS / "yaml-schema.json"
+MANIFEST = ASSETS / "yaml-contracts.yaml"
+
+
+def blob_sha(path: Path) -> str:
+    """Return the git blob SHA of a file, matching `git hash-object <path>`."""
+    data = path.read_bytes()
+    return hashlib.sha1(b"blob %d\0" % len(data) + data).hexdigest()
```
wenjin272: Hashing `read_bytes()` gives a worktree-content hash, which is not always the Git blob SHA. With `core.autocrlf=true` and no `.gitattributes`, these JSON files are checked out as CRLF, so this reports the recorded LF blob SHA as stale and suggests a SHA that Linux CI will reject. Could we hash through Git's clean filters (for example, `git hash-object --stdin --path=docs/yaml-schema.json`) or enforce LF for these files?

.github/workflows/ci.yml
weiqingy: Good catch. Scoped it to main: `if: github.ref == 'refs/heads/main' || github.base_ref == 'main'`.

One thing I found while checking, in case it changes your read. Release branches carry their own `ci.yml`, and this step only exists on main, so nothing runs there today. `release-0.3` doesn't even have the older `Check AGENTS.md freshness` step. The skill directory isn't on `release-0.3` at all either, so a backported step would just fail with `does not exist`.

Your premise still holds though. Those backports are real (`0c8da869` and `0bb16280` both touch `docs/yaml-schema.json` on `release-0.3`), so this would bite as soon as someone syncs the workflow across branches. Seemed worth guarding either way.

I skipped the branch-aware option since the skill only ships one unversioned schema, so there's nothing to select between. Is that what you had in mind, or were you picturing the manifest carrying per-branch entries at some point?

tools/check-skill-schema.py
weiqingy: You're right, thanks. Reproduced it in a scratch repo with `core.autocrlf=true`: the byte hash comes out `4ca505...` while the real blob is `8d6b85...`, so the check calls a good pin stale and then hands you a SHA that CI rejects.

`blob_sha()` now runs `git hash-object -- <path>`, which applies whatever clean filter the path's attributes select. Nothing moves on an LF checkout, both files still hash to `183cc7ac...` and the mutation cases behave the same.

I went with hashing through git rather than a `.gitattributes` pin, mostly because it covers any filter and not just line endings, and it keeps line-ending policy out of this PR. Would you want the `.gitattributes` too, or is going through git enough on its own?

wenjin272 (approved):

PR #974, opened 2026-08-06
[hotfix][runtime] Aggregate close failures across the Exception/Error boundary

weiqingy:
Linked issue: none, this is a hotfix.

### Purpose of change

`FlussActionStateStore.close()` and `KafkaActionStateStore.close()` each close two resources and caught only `Exception` while doing it. That leaves the non-`Exception` cases incomplete, and it cost four distinct failures.

In `FlussActionStateStore.close()`, `table.close()` was wrapped in a `try` with `connection.close()` in the `finally`. The connection did get closed, so there was no leak, but a `finally` that throws discards the in-flight exception. When both closes threw, the caller saw only the connection failure and had no signal that the table close failed at all. When `table.close()` threw an `Error`, the `finally` recorded a connection `Exception` that was then discarded as the pending `Error` resumed, so that connection failure vanished. Conversely, a connection `Error` thrown from the `finally` bypassed the inner catch and replaced an already-recorded table failure outright.

`KafkaActionStateStore.close()` had the same boundary problem in a worse form. A producer `Error` left the method before `consumer.close()` ran at all, which is a live resource leak rather than only exception loss. A consumer `Error` replaced an earlier producer failure.

All four paths were reproduced against the previous code before anything changed. The Kafka leak was confirmed by invocation check: with the producer close throwing an `Error`, `consumer.close()` was never invoked.

Both methods now use the same shape. Each close is wrapped in its own `catch (Throwable t)`, the first failure is kept primary via `ExceptionUtils.firstOrSuppressed`, and the method ends with `ExceptionUtils.rethrowException`, so an `Error` still reaches the caller as an `Error` and a checked exception as itself. `rethrowException` is required rather than a bare `throw` because the accumulator is `Throwable`-typed and would not compile otherwise.

Two notes on the shape, since both are easy to ask about.

Widening to `Throwable` is what removes the need for the `finally` in the Fluss method. The reason to hold a `finally` was that a `catch (Exception)` does not run for an `Error`, so a flat ladder would skip the connection close and leak it. A catch that runs for any `Throwable` already keeps the second close reachable on every path, so the `finally` no longer buys anything and the two methods are now structurally identical. `SkillManager.java` uses the same pattern in its constructor cleanup.

`IOUtils.closeAll` was considered and does not fit. Its first-wins suppression and null-skipping match this contract, but with the default `Exception.class` it rethrows a non-`Exception` `Throwable` immediately without closing the remaining resources, which is the same leak being fixed here. Passing `Throwable.class` avoids that but wraps the throwable, so an `OutOfMemoryError` would reach the caller as an ordinary `Exception`.

The Kafka method is fixed here rather than deferred, because its `Error` path is a resource leak and leaving it would keep the two stores divergent.

### Tests

Each test class now covers the same six close outcomes, one per path: both closes succeed, the first close fails, both fail, only the second fails, and the two `Exception`/`Error` orderings.

Four of those are new or strengthened in this change. The Fluss error-path test now asserts that the connection failure survives as a suppressed exception rather than only that the `Error` stays primary. A new Fluss test pins a checked table exception surviving a connection `Error`. A new Kafka test pins the leak directly, asserting `consumer.close()` still runs when the producer close throws an `Error`. A fourth pins a producer exception surviving a consumer `Error`.

Each of the four kills a distinct mutant. Narrowing any single one of the four `catch (Throwable)` arms back to `catch (Exception)` fails exactly one of them and no others. Swapping the two `firstOrSuppressed` arguments fails seven tests, so the argument order is pinned as well.

The stubs deliberately span checked (`IOException`), unchecked (`RuntimeException`) and `Error` failures. With unchecked stubs only, narrowing a `catch` to `catch (RuntimeException)` would pass the whole suite while reinstating the defect for checked exceptions, which is the likely case in practice given that `FlussConnection.close()` calls `RemoteFileDownloader.close() throws IOException`.

The full `runtime` module suite passes: 419 tests, 0 failures, 0 errors.

### API

No API change. Both methods keep their `@Override public void close() throws Exception` signature, and the caller-visible failure type is unchanged on every path that previously threw.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

`Generated-by: Claude Code (claude-opus-5)`, also present in the commit messages.

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
@@ -21,6 +21,7 @@ import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.plan.AgentConfiguration;
 import org.apache.flink.agents.plan.actions.Action;
 import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.Preconditions;
 import org.apache.fluss.client.Connection;
 import org.apache.fluss.client.ConnectionFactory;
@@ -490,31 +491,27 @@ public class FlussActionStateStore implements ActionStateStore {
 
     @Override
     public void close() throws Exception {
-        Exception firstException = null;
-        try {
-            if (table != null) {
+        // Catching Throwable rather than Exception is what keeps the connection close reachable
+        // when the table close fails with an Error, and what keeps that first failure the one the
+        // caller sees — a later connection failure rides along as suppressed instead of replacing
+        // it.
+        Throwable firstException = null;
+        if (table != null) {
+            try {
                 table.close();
+            } catch (Throwable t) {
+                firstException = t;
             }
-        } catch (Exception e) {
-            firstException = e;
-        } finally {
-            // Must stay a finally: a Throwable that is not an Exception skips the catch above, and
-            // only a finally still closes the connection on the way out. The rethrow below stays
-            // outside this block so a connection failure cannot displace such a Throwable.
-            if (connection != null) {
-                try {
-                    connection.close();
-                } catch (Exception e) {
-                    if (firstException == null) {
-                        firstException = e;
-                    } else {
-                        firstException.addSuppressed(e);
-                    }
-                }
+        }
+        if (connection != null) {
+            try {
+                connection.close();
+            } catch (Throwable t) {
+                firstException = ExceptionUtils.firstOrSuppressed(t, firstException);
             }
         }
         if (firstException != null) {
-            throw firstException;
+            ExceptionUtils.rethrowException(firstException);
         }
     }
 
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
@@ -490,14 +490,31 @@ public class FlussActionStateStore implements ActionStateStore {
 
     @Override
     public void close() throws Exception {
+        Exception firstException = null;
         try {
             if (table != null) {
                 table.close();
             }
+        } catch (Exception e) {
+            firstException = e;
         } finally {
+            // Must stay a finally: a Throwable that is not an Exception skips the catch above, and
+            // only a finally still closes the connection on the way out. The rethrow below stays
+            // outside this block so a connection failure cannot displace such a Throwable.
             if (connection != null) {
-                connection.close();
+                try {
+                    connection.close();
+                } catch (Exception e) {
+                    if (firstException == null) {
+                        firstException = e;
+                    } else {
+                        firstException.addSuppressed(e);
+                    }
+                }
             }
+        }
+        if (firstException != null) {
+            throw firstException;
         }
     }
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -490,15 +490,32 @@ public void pruneState(Object key, long seqNum) {
 
     @Override
     public void close() throws Exception {
+        Exception firstException = null;
         try {
             if (table != null) {
                 table.close();
             }
+        } catch (Exception e) {
+            firstException = e;
```
wenjin272: Catching only `Exception` leaves the non-`Exception` cases incomplete. If `table.close()` throws an `Error` and `connection.close()` throws an `Exception`, the connection failure is assigned to `firstException` but is then discarded when the pending table `Error` resumes after the `finally`. Conversely, if `connection.close()` throws an `Error`, it bypasses the inner catch and still replaces the earlier table failure. Could we aggregate `Throwable` from both closes, keep the first as primary, add the second as suppressed, and rethrow the original type? The error-path test should also assert that the connection failure is preserved. `KafkaActionStateStore.close()` has the same pre-existing boundary problem: a producer `Error` skips `consumer.close()`, while a consumer `Error` can replace an earlier producer exception. That can be fixed here for consistency or tracked as a follow-up if it is outside this PR’s scope.

weiqingy: You're right on all of it. I reproduced each path against the previous code before changing anything:

- table `Error` plus connection `Exception`: the `Error` reaches the caller with `getSuppressed()` empty, so the connection failure is gone.
- table `Exception` plus connection `Error`: the connection `Error` is thrown and the table exception is lost.
- Kafka producer `Error`: `consumer.close()` is never invoked at all, so that one is a live resource leak and not only exception loss.
- Kafka producer `Exception` plus consumer `Error`: the producer failure is lost.

Both methods now use the same shape. Each close gets its own `catch (Throwable t)`, the first failure stays primary via `ExceptionUtils.firstOrSuppressed`, and the method ends with `ExceptionUtils.rethrowException` so an `Error` still arrives as an `Error` and a checked exception as itself. `rethrowException` is needed rather than a bare `throw` because the accumulator is `Throwable`-typed and would not otherwise compile.

I took the Kafka fix here rather than as a follow-up, since its `Error` path leaks the consumer and deferring it would leave the two stores divergent.

Widening to `Throwable` also removes the reason the `finally` existed, so it is gone and the two methods are now structurally identical. That supersedes the design note in the PR description arguing the `finally` had to stay, and I have rewritten the description accordingly. That argument only ever held for `catch (Exception)` not running on an `Error`, and it stops holding once both arms catch `Throwable`.

On the test: `testCloseKeepsTableErrorWhenConnectionCloseAlsoFails` now asserts the connection failure survives as a suppressed exception, rather than only that the `Error` stays primary. Each test class covers the same six close outcomes, and each of the four new or strengthened tests kills a distinct mutant: narrowing any single one of the four `catch (Throwable)` arms back to `catch (Exception)` fails exactly one of them and no others.

wenjin272 (approved):
Thanks for addressing my comment. LGTM.

PR #821, opened 2026-06-09
[runtime][java][python] Add CEL Action condition filtering

rosemarYuan:
Linked issue: #754 

### Purpose of change

This is the second PR in the three-PR stack tracked by #754:

1. #756 — API changes (under review)
2. **This PR** — CEL runtime (Java and Python)
3. #977 Documentation (planned)

It adds the Java-side CEL runtime for `@Action` trigger conditions, enabling actions to filter events by field-level CEL expressions (e.g. `event.tokenCount > 100`).

It adds `trigger_conditions` to the Python `@action` decorator, `Agent.add_action`, and YAML. Python keeps the conditions in their original order as raw strings in `AgentPlan` JSON.

### Tests

- Java Plan, Runtime, routing, and serialization tests
- Python API, YAML, plan construction, and preflight tests
- Java ↔ Python E2E tests
- Shared CEL conformance and macro fixtures

### API

- Existing `trigger_conditions` now accepts event types or CEL Boolean expressions in Java, Python, and YAML. Multiple entries use OR semantics.
- Adds the condition evaluation failure strategy: `WARN_AND_SKIP` (default) or `FAIL`.
- Plans serialize only `trigger_conditions`; legacy `listen_event_types`-only JSON is rejected.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed`  (documentation will be covered in PR4)
- [ ] `doc-included`

PR #977, opened 2026-08-07
[docs] Document Action trigger conditions

rosemarYuan:
Linked issue: #754 

### Purpose of change

This is the third PR in the three-PR stack tracked by #754:
1. #756 — API changes (under review)
2. #821 — CEL runtime (Java and Python)
3.  **This PR** —Documentation

PR #821 introduced Action trigger conditions, but the existing documentation still described Actions mainly as event-type listeners. This PR documents the user-visible trigger-condition contract consistently across Java, Python, and YAML.

A trigger condition can be either an exact event type or a condition expression, which uses Common Expression Language (CEL). This change:

- explains OR semantics across trigger-condition entries, fan-out across matching Actions, exact event-type precedence, and per-Action short-circuit behavior;
- documents the event data available to condition expressions, including nested values, name collisions, missing values, numeric conversion, and supported macros;
- explains YAML validation and quoting rules for condition expressions, custom event types, and event aliases;
- clarifies that event aliases replace only complete event-type entries;
- provides decorated, annotated, and programmatic Action examples, including a complete Java registration example;
- documents `action.trigger-condition.evaluate-failure-strategy`; and

### Tests

This is a documentation-only change. No Tests.

### API

No public API or runtime behavior is changed. This PR documents the Action trigger-condition APIs and configuration introduced by #821.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included`


docs/content/docs/development/workflow_agent.md
```diff
@@ -264,9 +278,142 @@ public class ReviewAnalysisAgent extends Agent {
...
+attributes with the same names. Use the `attributes` namespace, such as `attributes["type"]` or
+`attributes["id"]`, to access colliding attribute keys.
+
+Nested values are not flattened. For an input event whose attributes are
+`{input: {status: "ok"}}`, use `input.status` or `attributes.input.status`; bare `status` is not
+available. Other event payloads keep their top-level envelope, for example `response.content`.
+
+Use a literal index for top-level keys containing dots, such as `attributes["a.b.c"]`, and test
+static membership with `"a.b.c" in attributes`. Dynamic access at the root, such as
+`attributes[key]`, and expressions over the whole `attributes` map are not supported. Dynamic access
```
weiqingy, line 394: "Not supported" sits in the same register as line 398's "remains absent", but this one is a hard failure: `ConditionExpressionCompiler.classifyIdent` throws from the `ActionMatcher` constructor, so the job fails at operator initialization. The fixture `dynamic_whole_attributes_index` captures the split, `plan_validation: pass` with `runtime_compilation: fail`.

Would naming the outcome help a reader tell the two apart? Maybe something like "rejected when the runtime compiles the condition, so the job fails to start", if that reads right to you.

docs/content/docs/development/yaml.md
```diff
@@ -278,25 +278,78 @@ Inline action (map) fields:
...
+Different matching actions all run (fan-out), while one action runs at most once for an event.
+Separate entries are OR branches. To combine a type restriction with an attribute predicate using
+AND, write both checks in one Boolean expression, as shown by `action2` above. Exact type matches run
+before matches from condition expressions.
+
+Expressions expose the event type as `type`, built-in constants under `EventType`, and the event
+attribute envelope under `attributes`. Referenced top-level entries from `Event.attributes` are also
+available directly. Nested values stay under their top-level attribute path and are never flattened.
+For an input event whose attributes are `{input: {status: "ok"}}`, both `input.status` and
+`attributes.input.status` are valid, while bare `status` is not. Other event payloads follow the same
```
weiqingy, line 342: "is not" reads as rejected, though nothing rejects bare `status`. It clears plan validation and the type check (`ConditionExpressionCompiler.java:124` declares non-framework idents `SimpleType.DYN`), then throws unbound at evaluation. Under the default `WARN_AND_SKIP` that is logged and treated as false, so the action just never fires.

That default lives only in the `configuration.md` row this PR adds, and neither page links to it. Would a sentence on the failure semantics plus a link fit here, the way `monitoring.md:335` points at `event-log.level`? Same two gaps on `workflow_agent.md`, if so.

docs/content/docs/development/yaml.md
```diff
@@ -498,7 +551,7 @@ When `type:` resolves to the **opposite** language of the loader, the loader bui
 
 ### Provider aliases
 
-For `clazz:` on resource descriptors and for event names in `trigger_conditions:`, you can use a short alias instead of a fully-qualified class path.
+For `clazz:` on resource descriptors and for complete event-type entries in `trigger_conditions:`, you can use a short alias instead of a fully-qualified class path. Event alias replacement is an exact complete-entry lookup: `input` is replaced, while `type == input`, `attributes.kind == 'input'`, and the quoted event type `'input'` remain unchanged.
```
weiqingy, line 554: This sentence now frames the alias table below as the menu of legal `trigger_conditions` values, and that table offers `output`. An action triggered on `output` never runs: `ActionExecutionOperator.java:247` sends OutputEvents downstream in the `if` branch, so the `getActionsTriggeredBy` call in the `else` at `:269` is never reached. Nothing fails and nothing logs.

You document exactly this at `workflow_agent.md:472-474`. Should the caveat follow the alias table here, or hang off the pointer at 350-351?

weiqingy (commented):

wenjin272:
Thanks for the detailed comments, @weiqingy. I refactored the two pages to keep the trigger-condition contract in `workflow_agent.md` and only YAML-specific guidance in `yaml.md`. The three issues you raised are now covered.

Could you please take another look and let me know if you have any further suggestions?

docs/content/docs/development/workflow_agent.md
```diff
@@ -320,14 +391,67 @@ public static void emitOutput(Event event, RunnerContext ctx) {
...
+### Trigger Condition Reference
+
+#### Custom Event-Type Names
+
+A custom event type may be a bare name such as `order.created` or `order-created`. Each dot-separated
+segment must start with an ASCII letter or underscore and may then contain ASCII letters, digits,
+underscores, or hyphens. Quote a name that contains other punctuation or would otherwise be parsed as
+an expression, for example `'order:created'`, `'true'`, or `'EventType.custom'`. A quoted name matches
+the literal event-type string; it does not reference a built-in `EventType` constant. Quoted names
+must be non-empty and cannot contain whitespace, quotes, backslashes, or control characters.
```
weiqingy: In Java and Python you pass the condition as a string, so an event type that needs quotes ends up with two layers: `@Action("'order:created'")`. This paragraph shows `'order:created'` but never says the inner quotes belong in the string, and the example that made that clear came out in this commit. Drop a layer and `@Action("order:created")` looks right, but it gets read as an expression rather than an event type (`TriggerCondition.java:32` leaves `:` out of the bare-name pattern), so the plan fails to build. The YAML page still says it outright at `yaml.md:292`. Since this page owns the rule now, would one line here help? Maybe: "In Java and Python the quotes are part of the condition string, for example `@Action("'order:created'")`."

weiqingy (commented):
Thanks for addressing the comments. All three are closed, and moving the trigger-condition rules onto one page with `yaml.md` pointing at it reads much better.

One thing inline that doesn't need to hold this up, otherwise LGTM.

Issue #983, opened 2026-08-08
[Tech Debt] De-flake the long-term memory e2e test, which asserts an exact LLM consolidation count


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Summary

`test_long_term_memory_async_execution_in_action` asserts an exact count of long-term memory items, but that count is decided by an LLM and is not guaranteed. The test has turned CI red on an unrelated change on `main`.

The assertion is `assert len(items) == 1` in `check_result` (`python/flink_agents/e2e_tests/e2e_tests_integration/long_term_memory_test.py`). It checks how many items mem0 stores after alice's two inputs, "My favorite fruit is watermelon." followed by "My favorite fruit is bananas.". Whether those collapse into one item or stay separate is the extraction model's judgment call.

The comment directly above the assertion already concedes this: "LLMs may treat different review comments as updates to the same fact or as distinct facts."

## Evidence

On `main`, this assertion has failed once:

| Date | Result |
|---|---|
| 2026-08-08 | `assert 0 == 1` ([run 31266039736](https://github.com/apache/flink-agents/actions/runs/31266039736)) |

That failure landed on a commit that changed three markdown files under `docs/content` with one line each, so it was unrelated to the change under test. All three attempts failed, exhausting the `--reruns 2` budget added in #717.

Two nearby data points, neither of which is a count failure on this assertion:

- The run immediately before it ([31262378878](https://github.com/apache/flink-agents/actions/runs/31262378878)) passed this test only on retry, and it was the only rerun in that job. The first attempt's cause is not recoverable, because the e2e arm runs with `log_cli_level=OFF` and pytest-rerunfailures reports only the final attempt.
- An earlier `main` run on 2026-07-25 ([30150040643](https://github.com/apache/flink-agents/actions/runs/30150040643)) failed the same test with a `Py4JJavaError` on `execute`, which is a job execution failure rather than an assertion failure.

The 0.2 release line has failures in a file of the same name, but that version is a different test: it covers compaction and asserts on `ChromaVectorStore` documents rather than on memory set items. Those are not occurrences of this assertion.

## Why the count is unstable

The fact-extraction model is a remote hosted endpoint (`qwen3.6-plus` over `coding.dashscope.aliyuncs.com`), configured at `long_term_memory_test.py:145-165`. Only the embedding model is local Ollama, despite the resource being named `ollama_qwen3`.

[...]

PR #984, opened 2026-08-08
[test][python] Assert long-term memory content instead of an exact LLM item count

weiqingy:
Linked issue: #983

### Purpose of change

`test_long_term_memory_async_execution_in_action` asserts `len(items) == 1`, which pins how many items mem0 stores after alice's two inputs. Whether those two facts collapse into one item or stay separate is the extraction model's decision, so the count is not guaranteed. The comment above the assertion already conceded the variability.

On 2026-08-08 the test failed on `main` with `assert 0 == 1`, on a commit that changed three markdown files under `docs/content` with one line each. All three attempts failed, exhausting the `--reruns 2` budget. The run before it passed this test only on retry, and it was the only rerun in that job.

This asserts on retrievable content instead of an exact count.

Two things this does not claim. Counts above `1` have not been observed on this test, so loosening them is preventive rather than a response to a recorded failure. And the one failure that has been recorded, the empty set, still fails after this change, deliberately. The immediate value here is the diagnostic and the restored coverage rather than a greener CI.

The empty case is deliberately kept as a failure rather than tolerated. Zero stored items means neither add landed, which may be a real defect in the add path rather than model variance, and it is the only one of the four observed outcomes that carries information. To make the next occurrence diagnosable, bob's memory set is reported alongside the failure: if bob has items and alice does not, the miss is per-key; if both are empty, extraction or the store failed wholesale. That matters because the e2e arm runs with `log_cli_level=OFF` and pytest-rerunfailures reports only the final attempt, so today an occurrence leaves no trace.

Two related changes come with it.

The `created_at < updated_at` assertion is dropped. On a plain add mem0 sets `updated_at` equal to `created_at`, so the strict comparison holds only when the model picked the update branch. That made it a second instance of the same flake, and it can also fail on a legitimate delete-then-add even when exactly one item survives.

An explicit cross-key leak check is added. The exact count was incidentally covering this: alice and bob share a job id and a memory set name, so `agent_id` is the only isolation boundary between them, and a scoping break would have surfaced as an inflated count. Asserting that bob's facts never appear in alice's set restores that coverage without depending on a count.

The content match is case-insensitive because the stored text is the model's paraphrase rather than the input, so a title-cased "Bananas" would otherwise fail for the same reason the count does.

Timestamp coverage is kept, in a form that does not depend on which branch the model picked. The dropped ordering assertion was incidentally the only thing in the repo asserting that `created_at` and `updated_at` parse at all: `_to_memory_set_item` suppresses `ValueError`/`TypeError` and yields `None`, so an unrecognized timestamp format would otherwise go unnoticed. Asserting both fields are populated covers that and holds on both the create and the update path, since mem0 sets `updated_at` equal to `created_at` on a plain add.

### Tests

Test-only change. The modified assertions were exercised against the observed and adjacent outcomes:

| Case | Result |
|---|---|
| 0 items | fails, with bob's set in the message |
| 1 item, consolidated | passes |
| 2 items, kept separate | passes |
| 4 items, kept separate | passes |
| single item, title-cased "Bananas" | passes |
| single item added without an update, timestamps equal | passes |
| stale read, only the watermelon fact | fails |
| bob's facts present in alice's set | fails |
| timestamps parsed as `None` | fails |

The first row is the failure seen on `main`, which still fails. Rows two through four are the counts the model may legitimately produce. Rows five and six are cases the previous assertions failed on for reasons unrelated to the product. The last three are regressions that must still fail.

`ruff check` and `ruff format --check` pass on the file.

The test is gated on `ACTION_API_KEY` and needs a Flink cluster plus a local Ollama embedding model, so it does not run locally without those. Verification of the end-to-end path comes from CI.

### API

No public API change.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/python/flink_agents/e2e_tests/e2e_tests_integration/long_term_memory_test.py
+++ b/python/flink_agents/e2e_tests/e2e_tests_integration/long_term_memory_test.py
@@ -303,13 +303,35 @@ def check_result(*, result_dir: Path) -> None:
     for record in actual_result:
         records[f"{record.name}.{record.count}"] = record
 
+    assert "alice.2" in records, f"missing alice.2; got {sorted(records)}"
     items = records["alice.2"].items
-    # LLMs may treat different review comments as updates to the same
-    # fact or as distinct facts.
-    assert len(items) == 1
-    item: MemorySetItem = items[-1]
-    assert item.created_at < item.updated_at
-    assert "bananas" in item.value
+    # The extraction model decides whether alice's two facts collapse into a
+    # single item or stay separate, so the item count is not fixed. bob's set
+    # is reported alongside a failure to tell a per-key miss from a store-wide
+    # one.
+    bob_items = records["bob.2"].items if "bob.2" in records else None
+    bob_values = None if bob_items is None else [item.value for item in bob_items]
+    assert items, f"alice's memory set is empty (bob's set: {bob_values})"
+
+    values = [item.value for item in items]
+    # The stored text is the model's paraphrase of the input, so match loosely.
+    assert any("bananas" in value.lower() for value in values), (
+        f"no stored item carries the updated fact: {values}"
+    )
+    # Each partition key is scoped to its own memories, so bob's facts must
+    # never surface in alice's set.
+    leaked = [
+        value
+        for value in values
+        if "swimming" in value.lower() or "vegetarian" in value.lower()
+    ]
+    assert not leaked, f"alice's set contains bob's facts: {values}"
+    # Both timestamps are populated whether an item was created or updated, and
+    # the parser turns an unrecognized format into None rather than raising.
+    assert all(item.created_at and item.updated_at for item in items), (
+        f"stored items are missing timestamps: "
+        f"{[(item.created_at, item.updated_at) for item in items]}"
+    )
 
     # verify async add doesn't block process other key
     assert datetime.fromisoformat(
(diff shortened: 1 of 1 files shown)


python/flink_agents/e2e_tests/e2e_tests_integration/long_term_memory_test.py
```diff
@@ -303,13 +303,45 @@ def check_result(*, result_dir: Path) -> None:
...
-    # fact or as distinct facts.
-    assert len(items) == 1
-    item: MemorySetItem = items[-1]
-    assert item.created_at < item.updated_at
-    assert "bananas" in item.value
+    # The extraction model decides whether alice's two facts collapse into a
+    # single item or stay separate, so the item count is not fixed. bob's set
+    # is reported alongside a failure to tell a per-key miss from a store-wide
+    # one.
+    bob_items = records["bob.2"].items if "bob.2" in records else None
```
wenjin272: `bob.2` is optional here, so a missing output record, `items is None`, and an empty item list are all normalized into “no leaked values.” The new reverse-direction check can therefore pass without ever observing the Bob memory set. I verified that the current `check_result` passes both when `bob.2` is absent and when `bob.2.items == []`, provided the Alice set is healthy. Since this PR is intended to check isolation in both directions—and zero items are deliberately treated as a failure for Alice—could we assert that `bob.2` exists and that its item list is non-empty before scanning it for Alice facts?

weiqingy: Good catch, the `or []` made "no leak" and "never looked" the same result. Fixed in 396efd96: `bob.2` now has to exist and be non-empty before the scan, same as alice's set. The three cases you named pass on `d1aff9e0` and fail now, and alice's fact landing under bob's `agent_id` still surfaces as the leak rather than as her empty set.

One knock-on I'd like your read on. Checking bob's emptiness ahead of alice's means a both-empty run now fails on bob, so the store-wide vs per-key signal would have dropped out of the message. I kept it by reporting each key's set in the other's failure message. Does that read right to you?

wenjin272 (approved):

Issue #982, opened 2026-08-08
[Tech Debt] tools/e2e.sh breaks without uv on PATH, and the tools/ut.sh pip fallback passes a pytest log level that errors out


weiqingy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Two independent defects in the build scripts, both found while working on #980. They are unrelated in mechanism but both sit in the no-`uv` / fallback paths that CI never exercises, which is why neither has surfaced before. Filing them together since one change can cover both.

## 1. `tools/e2e.sh` breaks without `uv` on PATH

`tools/e2e.sh` invokes `uv` by bare name seven times, at `:48`, `:57`, `:69`, `:75`, `:84`, `:93` and `:156`, and it has no `command -v uv` guard anywhere. Its only `command -v` is for `mktemp` at `:125`. So on a machine where `uv` is not on PATH, every one of those fails with `uv: command not found`.

This is unlike the sibling scripts. `tools/lint.sh:67` and `tools/ut.sh:222` both guard their `uv` usage and fall back to pip, and `tools/build.sh` installs `uv` itself before using it. `tools/e2e.sh` does neither. It only delegates to `tools/build.sh` at `:101` and `:114`, and both of those are conditional:

```bash
if [[ ! -d "e2e-test/target" ]]; then
  bash tools/build.sh
fi
...
if [[ ! -f "uv.lock" ]]; then
  bash tools/build.sh
fi
```

So once a build has happened, the delegation is skipped. `tools/build.sh -j` also never runs the Python half at all, so it never installs `uv`.

Worth noting what the fix is *not*: rewriting these to `python3 -m uv` looks like the obvious parallel to #979, but it would make things worse here. That form requires an importable `uv` module, which only exists if `uv` was pip installed. Someone using a standalone `uv` (the curl installer or homebrew) has a working `uv` on PATH and no module, so the rewrite would turn a working run into a failing one. #979 deliberately leaves this file alone for that reason.

A guard is the right shape: probe once, prefer the PATH binary, fall back to the interpreter form only when the module is actually importable, and fail with a clear message when neither is available. That also fixes the current failure mode, where the script reports a bare `command not found` rather than telling you what to install.

## 2. `tools/ut.sh` pip fallback passes a log level pytest rejects

[...]

PR #986, opened 2026-08-08
[infra] Guard uv in e2e.sh and fix the pytest log level in ut.sh

weiqingy:
Linked issue: #982

### Purpose of change

Two defects in fallback paths that CI never exercises, which is why neither has surfaced before. Both were found while working on #979.

#### `tools/e2e.sh` breaks without `uv` on PATH

It invokes `uv` by bare name seven times and has no `command -v uv` guard anywhere, so a machine without uv on PATH gets seven `uv: command not found` lines and no hint about what to install.

It is the odd one out among the build scripts. `tools/lint.sh` and `tools/ut.sh` both guard their uv usage and fall back to pip, and `tools/build.sh` installs uv before using it. `tools/e2e.sh` does neither. It only delegates to `tools/build.sh` conditionally, and that delegation is skipped once `e2e-test/target` and `python/uv.lock` exist.

So it now probes once, after the build delegation so that a build performed in the same run counts:

```bash
if command -v uv >/dev/null 2>&1; then
  UV=(uv)
elif python3 -m uv --version >/dev/null 2>&1; then
  UV=(python3 -m uv)
else
  echo "Error: uv not found. Run tools/build.sh first, or install uv." >&2
  exit 1
fi
```

The binary on PATH has to win over the module. A standalone uv, from the curl installer or homebrew, gives a working binary and no importable `uv` module, so preferring the module would break exactly those installs.

The module branch tests uv by *running* it rather than by importing it. `uv/__init__.py` imports `find_uv_bin` without calling it, and the five-directory search for the binary happens at call time inside `uv/__main__.py`. So `import uv` succeeds even when the binary is missing, and a probe built on it would select a module that then fails at all seven sites, relocating the very failure this change removes.

The `uv pip install` site also names its target environment. `uv pip` honors the `VIRTUAL_ENV` that uv's module entry point derives from the calling interpreter, so without `--python .venv` the fallback could install `apache-flink` outside the venv the tests actually run in. `tools/build.sh` already does the same at its own wheel install.

#### `tools/ut.sh` passes a log level pytest rejects

The pip fallback passed `-o log_cli_level=${LOG_LEVEL:-OFF}` while the uv branch in the same function passed `${LOG_LEVEL:-CRITICAL}`. `OFF` is not a Python logging level name, so pytest rejects it during startup:

```
ERROR: 'OFF' is not recognized as a logging level name for 'log_cli_level'. Please consider passing the logging level num instead.
```

That is exit 4, raised before any test is collected, which `ut.sh` then reports through its "unknown error" arm. Aligning the fallback with the uv branch lets the path run.

### Tests

CI behavior is unchanged. `LOG_LEVEL: INFO` is set at four workflow sites, and the two Python jobs that fall back to the default take the uv branch, which already used `CRITICAL`. The runners also provide uv on PATH, so the probe selects the same binary the script used before.

The probe was exercised across every combination of binary and module availability:

| uv on PATH | module importable | binary findable | Selected | Exit |
| --- | --- | --- | --- | --- |
| yes | yes | yes | `uv` | 0 |
| yes | no | n/a | `uv` | 0 |
| no | yes | yes | `python3 -m uv` | 0 |
| no | no | n/a | error | 1 |
| no | yes | **no** | error | 1 |

The last row is the case that motivated testing the module by running it. With an import-based probe it selected `python3 -m uv` and exited 0, then failed at the first real call site.

Argument passing was checked at all seven sites against the previous version with a stub on PATH, including the one inside a command substitution and the ones in `cd ... &&` chains. Byte-identical under both probe branches.

For `tools/ut.sh`, the fallback path previously exited 4 with nothing collected. It now collects 375 tests and 293 pass locally. The rest are `apache-flink~=2.3.0` being uninstallable on macOS arm64, since no `apache-beam` wheel at the required version exists for that platform. That is environmental and unrelated to this change, so this fixes the first blocker on that path rather than making it green everywhere.

`bash -n` passes on both scripts, and shellcheck reports the same finding set before and after, with nothing landing on the new block.

### API

No API change.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

### DIFF
--- a/tools/e2e.sh
+++ b/tools/e2e.sh
@@ -45,7 +45,7 @@ function run_agent_plan_compatibility_test {
     return 1
   fi
 
-  cd "$python_dir" && uv run --no-sync bash ../e2e-test/test-scripts/test_agent_plan_compatibility.sh "$tempdir" "$jar_path"
+  cd "$python_dir" && "${UV[@]}" run --no-sync bash ../e2e-test/test-scripts/test_agent_plan_compatibility.sh "$tempdir" "$jar_path"
 }
 
 function run_cross_language_config_test {
@@ -54,7 +54,7 @@ function run_cross_language_config_test {
     return 1
   fi
 
-  cd "$python_dir" && uv run --no-sync bash ../e2e-test/test-scripts/test_java_config_in_python.sh
+  cd "$python_dir" && "${UV[@]}" run --no-sync bash ../e2e-test/test-scripts/test_java_config_in_python.sh
 }
 
 function run_resource_cross_language_test_in_java {
@@ -66,13 +66,13 @@ function run_resource_cross_language_test_in_java {
   cd "$python_dir"
 
   # Get Python version from uv's virtual environment and set PYTHONPATH
-  local python_version=$(uv run --no-sync python -c "import sys; print(f'python{sys.version_info.major}.{sys.version_info.minor}')" 2>/dev/null)
+  local python_version=$("${UV[@]}" run --no-sync python -c "import sys; print(f'python{sys.version_info.major}.{sys.version_info.minor}')" 2>/dev/null)
   if [[ -n "$python_version" ]]; then
     export PYTHONPATH="$python_dir/.venv/lib/$python_version/site-packages:$PYTHONPATH"
     echo "PYTHONPATH set to: $PYTHONPATH"
   fi
 
-  uv run --no-sync bash ../e2e-test/test-scripts/test_resource_cross_language.sh
+  "${UV[@]}" run --no-sync bash ../e2e-test/test-scripts/test_resource_cross_language.sh
 }
 
 function run_resource_cross_language_test_in_python {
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):
Thanks for taking this on. LGTM. Merging after resolving conflicts.

weiqingy:
@wenjin272 Thanks for the review. Rebased onto main now that #979 is in.

The conflict was the squash: #979's two commits landed as one commit with a new SHA, so this branch and main were changing the same three files from different ancestries. I dropped those two commits with `git rebase --onto origin/main 16f01719` and confirmed first that the squashed commit is byte-identical to them, so nothing was lost in the drop.

The PR is down to the single new commit, touching only `tools/e2e.sh` and `tools/ut.sh`. No source change from what you reviewed. I also dropped the "stacked on #979" note from the description since it no longer applies.

PR #987, opened 2026-08-08
[hotfix][runtime][python] Close every component when an earlier close fails

emecii:
Linked issue: none (hotfix)

### Purpose of change

Six Java `close()` methods and one Python one in the operator shutdown chain closed
their components as
sequential statements with no per-call guard, so the first failure skipped
everything behind it.

| Site | Stranded when an earlier close fails |
|---|---|
| `ActionExecutionOperator.close()` | `contextManager`, `pythonBridge`, `eventRouter`, `durableExecManager`, `super.close()` |
| `PythonBridgeManager.close()` | the Pemja `PythonInterpreter` and the `PythonEnvironmentManager` |
| `ActionTaskContextManager.close()` | the `ContinuationActionExecutor` thread pool |
| `ResourceCache.close()` | the remaining cached resources, `cache.clear()`, and `resourceContext.close()` — for a non-`Exception` `Throwable` only |
| `PythonActionExecutor.close()` | the Python runner-context cleanup, which releases that context's long-term memory and resource cache |
| `SkillManager.closeRepos()` | the remaining skill repositories and their materialized temp directories — for a non-`Exception` `Throwable` only |
| Python `ResourceCache.close()` | the remaining resources, `_cache.clear()`, and `_resource_context.close()` |

The last three were found in review by @wenjin272, each a place the guarantee the
first four establish did not actually reach. `PythonActionExecutor` sits directly
under `PythonBridgeManager`, which closes the interpreter on the next rung, so its
skipped cleanup never runs at all. `SkillManager` is the nested branch below
`ResourceCache` → `ResourceContextImpl`, which clears its manager reference in a
`finally`, so a skipped repository can never be retried. The Python cache is the
cross-language counterpart of the Java one and had diverged from it.

#### Runtime flow

`ActionExecutionOperator.close()` is the entry point and the sharpest case. It
closes `resourceCache` first, and `ResourceCache.close()` aggregates its own
component failures and rethrows them by design. So the exception `ResourceCache`
propagates is precisely the one that skipped the rest of the chain, including
`pythonBridge.close()` — the call that releases the embedded Python interpreter
and its environment manager. A single resource failing to close could leak native
Python state for the lifetime of the TaskManager JVM.

`PythonBridgeManager.close()` and `ActionTaskContextManager.close()` are reached
from that same chain and had the same shape.

`ResourceCache.close()` was added to this list during review. It already
aggregated `Exception`s, but caught only `Exception`, so a non-`Exception`
`Throwable` out of a cached `Resource.close()` propagated immediately. That gap
matters *more* after the rest of this patch, not less: the operator ladder now
continues past the failure and closes the Python interpreter, while cached
resources that may hold Python references are still open — which inverts the
ordering the ladder exists to preserve. Thanks to @weiqingy for catching it.

#### Key decisions

Capture and rethrow, rather than closing later components in a `finally`. A
`finally` that completes abruptly discards the in-flight exception (JLS 14.20.2),
which is the defect #974 is fixing in `FlussActionStateStore`. The shape here
matches `KafkaActionStateStore.close()` (#948), and `ResourceCache.close()`
already aggregated this way before this patch widened it.

The ladders catch `Throwable`, not `Exception`. A `catch (Exception)` ladder
stops at a non-`Exception` `Throwable` and skips the remaining closes, which is
the same leak with a narrower trigger. `ExceptionUtils.rethrowException` then
rethrows `Error` and `Exception` unchanged, so the caller sees the original type
and instance rather than a wrapper.

`IOUtils.closeAll` was considered and rejected, for the reason already set out in
#974: with the default `Exception.class` it rethrows a non-`Exception`
`Throwable` immediately without closing the remaining resources. I verified this
against `flink-core-2.3.0` rather than assuming — with an `Error` thrown from the
first closeable, the second is never closed, while a plain `Exception` closes all
three. An earlier revision of this patch used `closeAll` and the `Error` test
below is what caught it. The same finding is recorded on #944, which takes the
`closeAll` path over these same files.

`ActionTaskContextManager` spells the aggregation out rather than delegating:
neither `RunnerContextImpl` nor `ContinuationActionExecutor` implements
`AutoCloseable`, and `ContinuationActionExecutor` has separate `java/` and
`java21/` implementations, so making it closeable would touch a source set the
JDK 11 profile does not compile.

#### Behavioral contracts

1. Every component close is attempted on every call, in the existing order.
2. A null component is skipped rather than raising.
3. When one close fails, its exception reaches the caller unchanged in type and
   identity, with nothing suppressed. Both halves are asserted, on every
   single-failure path rather than only the identity half.
4. When several fail, the first is thrown and the later ones are attached to it
   via `addSuppressed`.
5. A non-`Exception` `Throwable` does not prevent the remaining closes, and
   reaches the caller as itself rather than wrapped.
6. When nothing fails, `close()` returns normally.
7. `ActionExecutionOperator` still calls `super.close()`, and a `super.close()`
   failure aggregates with component failures rather than replacing them.
8. `ResourceCache.close()` holds contracts 1, 4, and 5 for a non-`Exception`
   `Throwable` too: the remaining cached resources, the cache clear, and
   `resourceContext.close()` all still run.

Contract 1's ordering clause is pinned with `InOrder`, not left implicit. Order
is load-bearing: `resourceCache` must close before `pythonBridge` because cached
resources may hold Python references, `PythonActionExecutor.close()` calls back
into the interpreter, and `super.close()` must come last.

### Tests

| Test | Contract |
|---|---|
| `PythonBridgeManagerTest.closeReleasesInterpreterAndEnvironmentWhenActionExecutorFails` | 1, 3 |
| `PythonBridgeManagerTest.closeReportsFirstFailureWithLaterOnesSuppressed` | 4 |
| `PythonBridgeManagerTest.closeReleasesInterpreterAndEnvironmentWhenActionExecutorThrowsError` | 5 |
| `ActionTaskContextManagerTest.closeClosesContinuationExecutorWhenRunnerContextFails` | 1, 3 |
| `ActionExecutionOperatorTest.closeClosesEveryComponentWhenAnEarlierCloseFails` | 1, 7 |
| `ActionExecutionOperatorTest.closeAggregatesSuperCloseFailureWithComponentFailure` | 7 |
| `ResourceCacheTest.closeClosesEveryResourceWhenAnEarlierResourceThrowsError` | 8 |
| `ResourceCacheTest.closeReportsFirstResourceFailureWithLaterOnesSuppressed` | 4 |
| `ActionTaskContextManagerTest.closeReportsFirstFailureWithLaterOneSuppressed` | 4 |
| `PythonActionExecutorTest.closeCleansRunnerContextWhenThreadPoolShutdownFails` | 1, 3 |
| `PythonActionExecutorTest.closeReportsFirstFailureWithLaterOneSuppressed` | 4 |
| `PythonActionExecutorTest.closeCleansRunnerContextWhenThreadPoolShutdownThrowsError` | 5 |
| `SkillManagerTest.closeAttemptsEveryRepoWhenAnEarlierRepoThrowsError` | 1, 5 |
| `ResourceCacheTest.closeClosesEverySkillRepositoryWhenAnEarlierRepoThrowsError` | 1, 5 (through the production path) |
| Python `test_resource_cache_close.py` — four cases | 1, 3, 4, 6 |

Each assertion was checked against a mutation that should break it, rather than
merely observed green. All of these fail the listed tests:

| Mutation | Fails |
|---|---|
| the original `close()` restored in place | the first five tests, four with Mockito's `Wanted but not invoked` |
| `IOUtils.closeAll` substituted into `PythonBridgeManager.close()` | only the `Error` test — the discriminating case for the `Throwable` decision |
| `ResourceCache` catch narrowed back to `Exception` | `closeClosesEveryResourceWhenAnEarlierResourceThrowsError` |
| the `super.close()` try/catch deleted | both `ActionExecutionOperatorTest` close tests |
| `if (firstFailure == null) super.close();` — a partial revert of the fix | both `ActionExecutionOperatorTest` close tests |
| operator close order swapped (`resourceCache` ↔ `pythonBridge`) | both `ActionExecutionOperatorTest` close tests |
| `PythonBridgeManager` close order swapped | `closeReleasesInterpreterAndEnvironmentWhenActionExecutorFails` |
| `ActionTaskContextManager`'s second rung overwrites instead of suppressing | `closeReportsFirstFailureWithLaterOneSuppressed` |
| `PythonActionExecutor.close()` restored to its sequential shape | all three new `PythonActionExecutorTest` cases |
| `SkillManager.closeRepos()` catch narrowed back to `Exception` | the `SkillManager` and `ResourceCache` repo tests |
| Python `close()` restored to its sequential shape | both Python failure-path tests |

One mutation check found a flaw in the new tests rather than in the code. The
`SkillManager` and `ResourceCache` repository tests first used one failing and one
healthy repo, and the `SkillManager` one *passed* against the narrowed catch: the
de-dup set is an `IdentityHashMap`, so when the healthy repo happened to iterate
first, both were closed even with the bug present. Both now use two failing repos
and fail deterministically whichever runs first, confirmed over three runs.

Three things are deliberately not mutation-covered, flagged rather than left
implied:

- `closeReportsFirstResourceFailureWithLaterOnesSuppressed` — `ResourceCache`
  already aggregated `Exception`s correctly, so this characterizes that behavior
  through the rewrite rather than guarding a live defect.
- the `getSuppressed()).isEmpty()` assertions added for contract 3's "nothing
  suppressed" clause. No mutation of the current ladders produces a spurious
  suppressed entry on a single-failure path.
- the `firstOrSuppressed` call at `ActionTaskContextManager`'s first rung.
  Reverting it leaves all ten tests in that file green; it is a fail-open guard
  for a close inserted above it later, not a fix.

All three align the tests with what the contracts already claimed.

Every aggregation point routes through `ExceptionUtils.firstOrSuppressed`,
including the one rung where a previous failure is not yet possible, so that
inserting a close above it cannot silently overwrite.

`./tools/ut.sh -j` passes: 1374 tests, 0 failures, 0 errors (38 skipped, all
pre-existing integration tests that need external services). Python unit tests:
816 passed, 13 skipped. Ruff, `spotless:check` and RAT are clean.

### API

No public API change. All six Java `close()` methods keep their
`@Override public void close() throws Exception` signature.

`PythonActionExecutor` gains `implements AutoCloseable`; it already declared a
matching `close() throws Exception`, so this is additive and no call site
changes.

Caller-visible behavior changes:

- When several closes fail, the exception received is now the first failure
  rather than the last, and `getSuppressed()` is non-empty.
- `ResourceCache.close()` now finishes the remaining closes before propagating a
  non-`Exception` `Throwable`, instead of propagating it immediately. The
  `Throwable` still reaches the caller as itself.

- Python `ResourceCache.close()` likewise finishes the remaining closes before
  re-raising the first failure. Later failures are logged rather than attached,
  since `ExceptionGroup` requires 3.11 and this package supports 3.10; the catch
  is `Exception` rather than the literal analogue of `Throwable`, because
  Python's `Exception` already covers what Java calls `Error` while
  `BaseException` would swallow `KeyboardInterrupt` and `SystemExit`.

No code in the repo catches these by type, unwraps a cause, or reads
`getSuppressed()`. `ResourceCache.close()`'s only production caller is the
`ActionExecutionOperator` ladder, which catches `Throwable`.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

`Generated-by: Claude Code (claude-opus-5)`, also present in the commit messages.

---

Overlap note: #944 rewrites three of these same `close()` methods with
`IOUtils.closeAll` and touches a superset of these files. The `Error` finding is
recorded there so the two do not land opposite decisions on that case;
sequencing is under discussion on that PR.

Same defect class, left out to keep this to one module and one call path, happy
to follow up separately:

- `OpenSearchVectorStore.close()` — `httpClient.close()` failing strands
  `credentialsProvider`
- `BedrockEmbeddingModelConnection.close()` — `embedPool.shutdown()` failing
  strands `client`

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/ResourceCache.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/ResourceCache.java
@@ -25,6 +25,7 @@ import org.apache.flink.agents.plan.resourceprovider.PythonResourceProvider;
 import org.apache.flink.agents.plan.resourceprovider.ResourceProvider;
 import org.apache.flink.agents.plan.tools.FunctionTool;
 import org.apache.flink.agents.runtime.resource.ResourceContextImpl;
+import org.apache.flink.util.ExceptionUtils;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -158,32 +159,32 @@ public class ResourceCache implements AutoCloseable {
 
     @Override
     public void close() throws Exception {
-        Exception firstException = null;
+        // Close every cached resource, then the resource context, even when an earlier close
+        // fails. The first failure is rethrown with the later ones suppressed.
+        //
+        // The ladders catch Throwable, not Exception: ActionExecutionOperator.close() closes this
+        // cache before the Python interpreter because cached resources may hold Python references,
+        // so a non-Exception Throwable escaping here would leave the remaining resources open
+        // while the interpreter behind them is torn down anyway. ExceptionUtils.rethrowException
+        // passes Error and Exception through unchanged, so the caller still sees the original.
+        Throwable firstFailure = null;
         for (Map<String, Resource> resources : cache.values()) {
             for (Resource resource : resources.values()) {
                 try {
                     resource.close();
-                } catch (Exception e) {
-                    if (firstException == null) {
-                        firstException = e;
-                    } else {
-                        firstException.addSuppressed(e);
-                    }
+                } catch (Throwable t) {
+                    firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
                 }
             }
         }
         cache.clear();
         try {
             resourceContext.close();
-        } catch (Exception e) {
-            if (firstException == null) {
-                firstException = e;
-            } else {
-                firstException.addSuppressed(e);
-            }
+        } catch (Throwable t) {
+            firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
         }
-        if (firstException != null) {
-            throw firstException;
+        if (firstFailure != null) {
+            ExceptionUtils.rethrowException(firstFailure);
         }
     }
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
@@ -591,24 +591,38 @@ public class ActionExecutionOperator<IN, OUT> extends AbstractStreamOperator<OUT
 
     @Override
     public void close() throws Exception {
-        // Must close before pythonInterpreter since cached resources may hold Python references.
-        if (resourceCache != null) {
-            resourceCache.close();
-        }
-        if (contextManager != null) {
-            contextManager.close();
-        }
-        if (pythonBridge != null) {
-            pythonBridge.close();
-        }
-        if (eventLogWriter != null) {
-            eventLogWriter.close();
-        }
-        if (durableExecManager != null) {
-            durableExecManager.close();
+        // Close every component even when an earlier one fails, so a failing close cannot leak
+        // the components behind it or skip super.close(). The first failure is rethrown with
+        // the later ones suppressed. Order is preserved: the resource cache must close before
+        // pythonInterpreter since cached resources may hold Python references.
+        //
+        // The ladder catches Throwable, not Exception, and IOUtils.closeAll is deliberately not
+        // used: both stop at the first non-Exception Throwable without closing what follows,
+        // which is the very leak this method has to avoid.
+        Throwable firstFailure = null;
+        for (AutoCloseable closeable :
+                new AutoCloseable[] {
+                    resourceCache, contextManager, pythonBridge, eventLogWriter, durableExecManager
+                }) {
+            if (closeable == null) {
+                continue;
+            }
+            try {
+                closeable.close();
+            } catch (Throwable t) {
+                firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
+            }
         }
 
-        super.close();
+        try {
+            super.close();
+        } catch (Throwable t) {
+            firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
+        }
+
+        if (firstFailure != null) {
+            ExceptionUtils.rethrowException(firstFailure);
+        }
     }
 
     @Override
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
@@ -35,6 +35,7 @@ import org.apache.flink.agents.runtime.python.context.PythonRunnerContextImpl;
 import org.apache.flink.agents.runtime.trace.ExecutionEventSink;
 import org.apache.flink.agents.runtime.trace.ReportedExecutionKey;
 import org.apache.flink.api.common.state.MapState;
+import org.apache.flink.util.ExceptionUtils;
 
 import javax.annotation.Nullable;
 
@@ -350,15 +351,31 @@ class ActionTaskContextManager implements AutoCloseable {
 
     @Override
     public void close() throws Exception {
+        // Close the continuation executor even when the runner context fails to close. The first
+        // failure is rethrown with the later one suppressed.
+        //
+        // The ladder catches Throwable, not Exception, so a non-Exception Throwable from the
+        // runner context cannot strand the executor's thread pool. Neither type implements
+        // AutoCloseable, so the aggregation is spelled out rather than delegated.
+        Throwable firstFailure = null;
         if (runnerContext != null) {
             try {
                 runnerContext.close();
+            } catch (Throwable t) {
+                firstFailure = t;
             } finally {
                 runnerContext = null;
             }
         }
         if (continuationActionExecutor != null) {
-            continuationActionExecutor.close();
+            try {
+                continuationActionExecutor.close();
+            } catch (Throwable t) {
+                firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
+            }
+        }
+        if (firstFailure != null) {
+            ExceptionUtils.rethrowException(firstFailure);
         }
     }
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
@@ -356,13 +356,15 @@ class ActionTaskContextManager implements AutoCloseable {
         //
         // The ladder catches Throwable, not Exception, so a non-Exception Throwable from the
         // runner context cannot strand the executor's thread pool. Neither type implements
-        // AutoCloseable, so the aggregation is spelled out rather than delegated.
+        // AutoCloseable, so the aggregation is spelled out rather than delegated. Both rungs go
+        // through firstOrSuppressed even though the first one cannot yet have a previous failure,
+        // so that a close inserted above it later suppresses rather than overwrites.
         Throwable firstFailure = null;
         if (runnerContext != null) {
             try {
                 runnerContext.close();
             } catch (Throwable t) {
-                firstFailure = t;
+                firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
             } finally {
                 runnerContext = null;
             }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/operator/PythonBridgeManager.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/operator/PythonBridgeManager.java
@@ -36,6 +36,7 @@ import org.apache.flink.agents.runtime.python.utils.PythonResourceAdapterImpl;
 import org.apache.flink.api.common.ExecutionConfig;
 import org.apache.flink.api.common.JobID;
 import org.apache.flink.python.env.PythonDependencyInfo;
+import org.apache.flink.util.ExceptionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import pemja.core.PythonInterpreter;
@@ -315,14 +316,29 @@ class PythonBridgeManager implements AutoCloseable {
 
     @Override
     public void close() throws Exception {
-        if (pythonActionExecutor != null) {
-            pythonActionExecutor.close();
+        // Close every component even when an earlier one fails, so a failing action executor
+        // cannot leak the interpreter or the environment manager. The first failure is
+        // rethrown with the later ones suppressed.
+        //
+        // The ladder catches Throwable, not Exception, and IOUtils.closeAll is deliberately not
+        // used: both stop at the first non-Exception Throwable without closing what follows, and
+        // what follows here is the native Python state.
+        Throwable firstFailure = null;
+        for (AutoCloseable closeable :
+                new AutoCloseable[] {
+                    pythonActionExecutor, pythonInterpreter, pythonEnvironmentManager
+                }) {
+            if (closeable == null) {
+                continue;
+            }
+            try {
+                closeable.close();
+            } catch (Throwable t) {
+                firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
+            }
         }
-        if (pythonInterpreter != null) {
-            pythonInterpreter.close();
-        }
-        if (pythonEnvironmentManager != null) {
-            pythonEnvironmentManager.close();
+        if (firstFailure != null) {
+            ExceptionUtils.rethrowException(firstFailure);
         }
     }
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
@@ -34,7 +34,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import static org.apache.flink.util.Preconditions.checkState;
 
 /** Execute the corresponding Python action in the agent. */
-public class PythonActionExecutor {
+public class PythonActionExecutor implements AutoCloseable {
 
     private static final String PYTHON_IMPORTS =
             "from flink_agents.plan import function\n"
@@ -201,6 +201,7 @@ public class PythonActionExecutor {
         return (boolean) ((Object[]) invokeResult)[0];
     }
 
+    @Override
     public void close() throws Exception {
         if (interpreter != null) {
             if (pythonAsyncThreadPool != null) {
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
@@ -25,6 +25,7 @@ import org.apache.flink.agents.plan.AgentPlan;
 import org.apache.flink.agents.plan.PythonFunction;
 import org.apache.flink.agents.runtime.python.context.PythonRunnerContextImpl;
 import org.apache.flink.types.Row;
+import org.apache.flink.util.ExceptionUtils;
 import pemja.core.PythonInterpreter;
 import pemja.core.object.PyObject;
 
@@ -203,18 +204,33 @@ public class PythonActionExecutor implements AutoCloseable {
 
     @Override
     public void close() throws Exception {
-        if (interpreter != null) {
-            if (pythonAsyncThreadPool != null) {
+        // The two Python-side cleanups are independent, so attempt both even when the first
+        // fails. Skipping the runner-context cleanup leaves that context's long-term memory and
+        // resource cache unreleased, and PythonBridgeManager closes the interpreter right behind
+        // us, so there is no later chance to run it. The first failure is rethrown with the later
+        // one suppressed, matching the ladders in the managers above.
+        if (interpreter == null) {
+            return;
+        }
+        Throwable firstFailure = null;
+        if (pythonAsyncThreadPool != null) {
+            try {
                 interpreter.invoke(CLOSE_ASYNC_THREAD_POOL, pythonAsyncThreadPool);
+            } catch (Throwable t) {
+                firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
             }
-
-            if (pythonRunnerContext != null) {
-                try {
-                    interpreter.invoke(CLOSE_FLINK_RUNNER_CONTEXT, pythonRunnerContext);
-                } finally {
-                    pythonRunnerContext = null;
-                }
+        }
+        if (pythonRunnerContext != null) {
+            try {
+                interpreter.invoke(CLOSE_FLINK_RUNNER_CONTEXT, pythonRunnerContext);
+            } catch (Throwable t) {
+                firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
+            } finally {
+                pythonRunnerContext = null;
             }
+        }
+        if (firstFailure != null) {
+            ExceptionUtils.rethrowException(firstFailure);
         }
     }
 
--- a/runtime/src/test/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperatorTest.java
+++ b/runtime/src/test/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperatorTest.java
@@ -62,6 +62,8 @@ import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.java.functions.KeySelector;
 import org.apache.flink.runtime.checkpoint.OperatorSubtaskState;
 import org.apache.flink.runtime.state.KeyGroupRangeAssignment;
+import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
+import org.apache.flink.streaming.api.operators.StreamOperatorStateHandler;
 import org.apache.flink.streaming.api.watermark.Watermark;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.tasks.mailbox.TaskMailbox;
@@ -71,6 +73,7 @@ import org.apache.flink.util.ExceptionUtils;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.mockito.InOrder;
 
 import java.io.IOException;
 import java.io.Serializable;
@@ -93,8 +96,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assertions.catchThrowable;
 import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
 
 /** Tests for {@link ActionExecutionOperator}. */
 public class ActionExecutionOperatorTest {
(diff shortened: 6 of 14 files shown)


weiqingy (commented):

runtime/src/main/java/org/apache/flink/agents/runtime/operator/PythonBridgeManager.java
```diff
@@ -299,14 +300,29 @@ boolean isInitialized() {
...
-        }
-        if (pythonInterpreter != null) {
-            pythonInterpreter.close();
+        // Close every component even when an earlier one fails, so a failing action executor
+        // cannot leak the interpreter or the environment manager. The first failure is
+        // rethrown with the later ones suppressed.
+        //
+        // The ladder catches Throwable, not Exception, and IOUtils.closeAll is deliberately not
+        // used: both stop at the first non-Exception Throwable without closing what follows, and
+        // what follows here is the native Python state.
```
weiqingy, line 309: Worth knowing about #944, which is open and takes the other path here: it rewrites these same three `close()` methods with `IOUtils.closeAll(...)`, and it touches all seven files this PR does.

Your reasoning in this comment holds up as far as I can tell. `closeAll` defaults `suppressedException` to `Exception.class` and rethrows anything that is not one before closing the rest (flink-core 2.3.0), so `closeReleasesInterpreterAndEnvironmentWhenActionExecutorThrowsError` would fail against #944's version of this method. The component lists differ too, since that call also passes `longTermMemory` and `pythonResourceAdapter`.

So the two disagree on the mechanism, not just on the same seven files, and whichever lands second inherits the other's decision. Either way the `Error` case looks like the thing to settle first. How would you like to sequence them?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -553,24 +553,38 @@ public void waitInFlightEventsFinished() throws Exception {
...
+        // the later ones suppressed. Order is preserved: the resource cache must close before
+        // pythonInterpreter since cached resources may hold Python references.
+        //
+        // The ladder catches Throwable, not Exception, and IOUtils.closeAll is deliberately not
+        // used: both stop at the first non-Exception Throwable without closing what follows,
+        // which is the very leak this method has to avoid.
+        Throwable firstFailure = null;
+        for (AutoCloseable closeable :
+                new AutoCloseable[] {
+                    resourceCache, contextManager, pythonBridge, eventRouter, durableExecManager
```
weiqingy, line 567: Something I noticed while reading the order comment above. `resourceCache` closes first because cached resources may hold Python references, so they need to go before the interpreter.

`ResourceCache.close()` itself still catches only `Exception` (`ResourceCache.java:148` and `:160`), which is the shape this comment argues against. If a cached `Resource.close()` throws something that is not an `Exception`, it propagates straight out: the remaining cached resources are skipped, `cache.clear()` never runs, and `resourceContext.close()` never runs. This ladder then catches it and closes `pythonBridge` anyway, so the interpreter goes down while those resources still point at it.

Not a regression, and an `Error` out of `Resource.close()` is unlikely in practice. What made me look was the PR body citing `ResourceCache.close()` as prior art for the aggregation shape. Is it intentionally left as-is, or would widening those two catches be in scope here?

runtime/src/test/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperatorTest.java
```diff
@@ -525,6 +529,65 @@ private static void replaceOperatorLtm(
...
+        try {
+            assertThatThrownBy(operator::close)
+                    .isInstanceOf(IllegalStateException.class)
+                    .hasMessage("resource cache close failed");
+
+            // The components behind the failing one are still released.
+            verify(contextManager).close();
+            verify(pythonBridge).close();
+            verify(eventRouter).close();
+            verify(durableExecManager).close();
```
weiqingy: The test table maps this one to contracts 1 and 7. Contract 1 is well covered by the four `verify`s, but I cannot find anything here that looks at `super.close()`, and there is no case where it fails, so the aggregation half of contract 7 is not exercised.

I have not run this, so treat what follows as reasoning rather than a result. Two changes both look like they would leave this test green: deleting the try/catch at `ActionExecutionOperator.java:579-583`, or restoring just the old shape for the super call (`if (firstFailure == null) super.close();`). Neither the throw nor the four `verify`s depend on it, and the `finally` empties the ladder before teardown.

That second one is the interesting case, since it is a partial revert of what this PR fixes. `AbstractStreamOperator.close()` is `stateHandler.dispose()`, so skipping it strands the state backends.

The "it still runs" half looks cheap to pin, because a side effect of `dispose()` is visible once the throwing `close()` returns. Making it actually fail is harder, since `super.close()` binds statically and a subclass cannot intercept it. Would an assertion for the first half be worth it, or would you rather drop contract 7 from the table?

runtime/src/test/java/org/apache/flink/agents/runtime/operator/PythonBridgeManagerTest.java
```diff
@@ -59,4 +67,92 @@ void openIsNoOpWhenPlanHasNeitherPythonActionsNorResources() throws Exception {
...
+        setField(bridge, "pythonActionExecutor", actionExecutor);
+        setField(bridge, "pythonInterpreter", interpreter);
+        setField(bridge, "pythonEnvironmentManager", environmentManager);
+
+        assertThatThrownBy(bridge::close)
+                .isInstanceOf(IllegalStateException.class)
+                .hasMessage("action executor close failed");
+
+        verify(interpreter).close();
+        verify(environmentManager).close();
```
weiqingy: Contract 1 mentions the existing order, and I am curious how much weight that clause is meant to carry. `verify()` does not check order, and none of the new tests use `InOrder`, so swapping the array at `PythonBridgeManager.java:311-314` to `{pythonInterpreter, pythonActionExecutor, pythonEnvironmentManager}` leaves all three tests in this file green. The action executor still throws, and it is still the first failure even though it is no longer the first close.

Order does look load-bearing rather than incidental: the class javadoc at `:70-71` documents reverse-of-creation order, and `PythonActionExecutor.close()` calls into the interpreter twice (`PythonActionExecutor.java:205-219`), which throws once the interpreter is already closed.

Something like this in one of the three, if useful:

```java
InOrder inOrder = inOrder(actionExecutor, interpreter, environmentManager);
inOrder.verify(actionExecutor).close();
inOrder.verify(interpreter).close();
inOrder.verify(environmentManager).close();
```

One test would be enough to pin it. Does that seem worth it?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/PythonBridgeManager.java
emecii: Thanks — I had not connected the two beyond the file overlap, and you are right that the disagreement is on mechanism.

I settled the `Error` case first, as you suggested, and checked it empirically against `flink-core-2.3.0` rather than reading the source. Three closeables that record whether `close()` ran, first one throwing:

```
Error     -> a1=true a2=false a3=false     // remaining two never closed
Exception -> b1=true b2=true  b3=true
```

The chain is `closeAll(AutoCloseable...)` -> `closeAll(Iterable)` -> `closeAll(Iterable, Exception.class)`, and the last one rethrows anything not assignable to `suppressedException` before closing the rest. So `closeAll` continues past an `Exception` but stops dead on any non-`Exception` `Throwable`, exactly as you read it.

That is now recorded on #944 (apache/flink-agents#944 (comment)), including the consequence at each of its three call sites — its `ActionExecutionOperator.close()` passes `super::close` as the trailing closeable, so an `Error` from `resourceCache` skips `stateHandler.dispose()` too — with a request to adopt the `Throwable` ladder.

On sequencing, my proposal is this PR first, since it is the smaller change and only rewrites the `close()` methods: #944's rebase then becomes additive, with `longTermMemory` and `pythonResourceAdapter` slotting into ladders that already exist, rather than a mechanism swap in one direction or the other. I have said on #944 that I am equally happy to invert it and rework this as a follow-up on top of theirs. Deferring to whatever you and @joeyutong prefer — the part I care about is that the two do not land opposite decisions on the `Error` case.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
emecii: Good catch — taken in scope, fixed in 9b5688d0.

Not intentional on my part. And I think it is slightly worse than "not a regression": this patch makes that path more consequential rather than less. Before, an `Error` out of a cached `Resource.close()` skipped `pythonBridge.close()` entirely, so nothing was torn down out of order. Now the ladder continues and closes the interpreter while those cached resources are still open — which inverts the very ordering the comment at `:556` exists to preserve. Widening `ResourceCache` restores it, so the two changes belong in the same patch.

`ResourceCache.close()` now uses the same shape as the other three: `Throwable` ladders, `ExceptionUtils.firstOrSuppressed`, `ExceptionUtils.rethrowException`. Its only production caller is the operator ladder, which already catches `Throwable`, so nothing downstream changes.

Two tests in `ResourceCacheTest`. The `Error` one pins all three consequences you listed, including the two I would otherwise have been assuming from position in the method rather than observing: `cache.clear()` is checked by reflecting the cache map, and `resourceContext.close()` by standing a mock `SkillManager` into the context, since `ResourceContextImpl.close()` closes it. Narrowing the catch back to `Exception` fails that test.

I also updated the PR body, which cited `ResourceCache.close()` as untouched prior art for the aggregation shape — no longer accurate now that this patch modifies it.

runtime/src/test/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperatorTest.java
emecii: You are right on both counts, and I confirmed your reasoning by running it: against the code as it stood, deleting the try/catch at `:579-583` and replacing it with `if (firstFailure == null) super.close();` both left this test green. The second one especially should not have — it is a partial revert of the fix. Contract 7 was not being tested. Fixed in 9b5688d0.

Rather than drop it from the table, I think the aggregation half is reachable — you stopped one step short. `super.close()` does bind statically, but its *effect* does not: `AbstractStreamOperator.close()` compiles to `stateHandler.dispose()` (verified in `flink-runtime-2.3.0` bytecode), `stateHandler` is a `protected` field, and `StreamOperatorStateHandler` is a public non-final class with a public non-final `dispose()`. Swapping the inherited handler therefore makes the super call both observable and failable — no subclass interception needed.

So:

- `closeClosesEveryComponentWhenAnEarlierCloseFails` now asserts `dispose()` ran.
- A new `closeAggregatesSuperCloseFailureWithComponentFailure` makes `dispose()` throw and asserts the `resourceCache` failure still reaches the caller with the super failure attached as suppressed — which also pins the `firstOrSuppressed` argument order, the easy thing to get backwards.

The real handler is restored in the `finally` before teardown, so the harness still disposes it for real.

Both of your mutations now fail both tests. While in here I also applied your `InOrder` point from the other thread to these two tests: order is documented even more explicitly on this path (`resourceCache` before `pythonBridge`, per the comment at `:556`), and swapping those two left the tests green until I did. The chain now covers all five components plus `stateHandler.dispose()` last.

runtime/src/test/java/org/apache/flink/agents/runtime/operator/PythonBridgeManagerTest.java
emecii: Worth it — added in 9b5688d0, using your snippet in `closeReleasesInterpreterAndEnvironmentWhenActionExecutorFails`, with a javadoc note on why the order is load-bearing rather than incidental (the class javadoc at `:70-71`, and `PythonActionExecutor.close()` calling back into the interpreter at `:205-219`). Your swap to `{pythonInterpreter, pythonActionExecutor, pythonEnvironmentManager}` now fails that test; it passed before.

The clause in contract 1 was meant to carry that weight, so this closes the gap between what it claims and what was checked.

Reviewing the rest of the diff against the same question, `ActionExecutionOperator.close()` had the identical hole and a more explicitly documented constraint — `resourceCache` must close before `pythonBridge` because cached resources may hold Python references (the comment at `:556`). Swapping those two left both operator close tests green. They now use one `InOrder` chain across all five components plus `stateHandler.dispose()`, so the same swap fails there too, and `super.close()` is pinned last in the same assertion.

runtime/src/test/java/org/apache/flink/agents/runtime/operator/PythonBridgeManagerTest.java
```diff
@@ -59,4 +69,98 @@ void openIsNoOpWhenPlanHasNeitherPythonActionsNorResources() throws Exception {
...
+                .when(actionExecutor)
+                .close();
+
+        setField(bridge, "pythonActionExecutor", actionExecutor);
+        setField(bridge, "pythonInterpreter", interpreter);
+        setField(bridge, "pythonEnvironmentManager", environmentManager);
+
+        assertThatThrownBy(bridge::close)
+                .isInstanceOf(IllegalStateException.class)
+                .hasMessage("action executor close failed");
```
weiqingy, line 97: Both this test and `ActionTaskContextManagerTest.closeClosesContinuationExecutorWhenRunnerContextFails` map to contract 3, which reads "unchanged in type and identity, with nothing suppressed".

The `nothing suppressed` half isn't asserted anywhere in the PR. Every `getSuppressed()` assertion in it (`:128`, `ResourceCacheTest.java:308`, `ActionExecutionOperatorTest.java:688`) is a contract-4 non-empty check, so nothing pins the array as empty on a single-failure path. The identity half I raised earlier is covered now: `:154` uses `isSameAs`, and both paths rethrow through the one site at `PythonBridgeManager.java:325`.

The code is fine either way, it's the table claiming a bit more than the tests check. Either remedy costs a line, so whichever you prefer: a `.satisfies(t -> assertThat(t.getSuppressed()).isEmpty())` here, or dropping `with nothing suppressed` from contract 3?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
```diff
@@ -322,15 +323,31 @@ void removePythonAwaitableRef(ActionTask actionTask) {
...
+        //
+        // The ladder catches Throwable, not Exception, so a non-Exception Throwable from the
+        // runner context cannot strand the executor's thread pool. Neither type implements
+        // AutoCloseable, so the aggregation is spelled out rather than delegated.
+        Throwable firstFailure = null;
         if (runnerContext != null) {
             try {
                 runnerContext.close();
+            } catch (Throwable t) {
+                firstFailure = t;
```
weiqingy, line 337: nit: this is the one aggregation point in the PR that doesn't go through `ExceptionUtils.firstOrSuppressed`. It's equivalent today, since `firstFailure` is still null from `:332`, but it fails open: a close added above it later would have its failure overwritten here rather than suppressed. Any objection to `firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);` for consistency with the other six?

runtime/src/test/java/org/apache/flink/agents/runtime/operator/PythonBridgeManagerTest.java
emecii: You're right, and I'd rather pin the claim than trim it — fixed in 0c4df632.

`.satisfies(t -> assertThat(t.getSuppressed()).isEmpty())` is now on every single-failure path, not just the two contract-3 tests: this one, `ActionTaskContextManagerTest.closeClosesContinuationExecutorWhenRunnerContextFails`, both `Error` tests, and `ActionExecutionOperatorTest.closeClosesEveryComponentWhenAnEarlierCloseFails` — the last being the case where the assertion says something slightly different, namely that a healthy `super.close()` contributes nothing.

One caveat I should state rather than let the table imply otherwise: these are fidelity guards, not regression tests. I tried to construct a mutation of the current ladders that produces a spurious suppressed entry on a single-failure path and could not — `firstOrSuppressed` guards self-suppression, and a duplicated rung rethrows the same instance, so the guard absorbs it. They pin what contract 3 claims; they do not catch a defect that exists today. I've noted that in the PR body alongside the same caveat for `closeReportsFirstResourceFailureWithLaterOnesSuppressed`.

Thanks for reading the table against the tests this carefully — that's twice now it was claiming more than it checked.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionTaskContextManager.java
emecii: No objection — done in 0c4df632, and I think it's more than a nit.

Your reasoning is exactly right on both halves, and I checked the second half rather than assuming it: reverting the change leaves all ten tests in `ActionTaskContextManagerTest` green, so it is provably a no-op today and no test can distinguish it. It's a fail-open guard, so I added a comment recording why the redundant call is deliberate — otherwise it reads like something to simplify away.

That was the one aggregation point of seven not going through `firstOrSuppressed`, so they're now uniform.

Looking at that site for the overwrite case you describe, I noticed it had no two-failure test at all — the aggregation there was entirely unpinned, at either rung. Added `closeReportsFirstFailureWithLaterOneSuppressed`, which fails if the second rung overwrites instead of suppressing. That doesn't cover the rung you flagged, since the failure it would drop cannot exist yet, but it does mean the site is no longer relying on the other ladders' tests for its aggregation behavior.

emecii:
Hi @weiqingy, thanks for reviewing and I have addressed the comments. PTAL again. Ty!

weiqingy (commented):
Thanks for addressing the comments. Nothing further from me. I’ll leave the final call to the maintainers.

wenjin272 (commented):
Moved the findings to inline review comments: https://github.com/apache/flink-agents/pull/987#pullrequestreview-4948248510

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
```diff
@@ -201,6 +201,7 @@ public boolean callPythonAwaitable(String pythonAwaitableRef) {
         return (boolean) ((Object[]) invokeResult)[0];
     }
 
+    @Override
```
wenjin272, line 205: [P1] `PythonActionExecutor.close()` can still skip the second cleanup step. If `interpreter.invoke(CLOSE_ASYNC_THREAD_POOL)` throws, `CLOSE_FLINK_RUNNER_CONTEXT` is never invoked. `PythonBridgeManager` then continues closing the interpreter, so the runner context's LTM and resource cache are not explicitly closed first. I reproduced this with a mocked interpreter: after the first `invoke()` throws, the second invocation is not made. Could this method use the same first-failure/suppressed-failure ladder and add a regression test?

runtime/src/main/java/org/apache/flink/agents/runtime/ResourceCache.java
```diff
@@ -140,32 +141,32 @@ public void put(String name, ResourceType type, Resource resource) {
...
         try {
             resourceContext.close();
-        } catch (Exception e) {
-            if (firstException == null) {
-                firstException = e;
-            } else {
-                firstException.addSuppressed(e);
-            }
+        } catch (Throwable t) {
+            firstFailure = ExceptionUtils.firstOrSuppressed(t, firstFailure);
```
wenjin272, line 184: [P2] Catching `Throwable` here only aggregates failures after `ResourceContextImpl.close()` returns or throws, but its nested [`SkillManager.closeRepos()`](https://github.com/apache/flink-agents/blob/0c4df632101d1a059bb0cb6e1b35cf434cfb5181/runtime/src/main/java/org/apache/flink/agents/runtime/skill/SkillManager.java#L221-L242) still catches only `Exception`. If a repository's `close()` throws an `Error`, the remaining repositories are skipped; `ResourceContextImpl` then clears the manager reference in `finally`, so that cleanup cannot be retried. The new close-all guarantee therefore does not reach this nested branch. Could `SkillManager` use the same `Throwable` aggregation and get a regression test through the `ResourceCache`/`ResourceContext` path?

runtime/src/main/java/org/apache/flink/agents/runtime/ResourceCache.java
```diff
@@ -140,32 +141,32 @@ public void put(String name, ResourceType type, Resource resource) {
...
+        // Close every cached resource, then the resource context, even when an earlier close
+        // fails. The first failure is rethrown with the later ones suppressed.
+        //
+        // The ladders catch Throwable, not Exception: ActionExecutionOperator.close() closes this
+        // cache before the Python interpreter because cached resources may hold Python references,
+        // so a non-Exception Throwable escaping here would leave the remaining resources open
+        // while the interpreter behind them is torn down anyway. ExceptionUtils.rethrowException
+        // passes Error and Exception through unchanged, so the caller still sees the original.
+        Throwable firstFailure = null;
         for (Map<String, Resource> resources : cache.values()) {
```
wenjin272, line 171: [P2] The Java path now closes every resource after an earlier failure, but the equivalent [Python `ResourceCache.close()`](https://github.com/apache/flink-agents/blob/0c4df632101d1a059bb0cb6e1b35cf434cfb5181/python/flink_agents/runtime/resource_cache.py#L96-L108) is still sequential: the first resource failure skips the remaining resources, `_cache.clear()`, and `_resource_context.close()`. I reproduced `first_closed=True, second_closed=False, cache_size=1`. Could we align the Python lifecycle behavior and add the corresponding regression test?

wenjin272 (commented):
Thanks @emecii for the careful work on this shutdown path. I left three inline comments on the remaining cleanup gaps.

runtime/src/main/java/org/apache/flink/agents/runtime/python/utils/PythonActionExecutor.java
emecii: Agreed, and fixed in 23312147. Your reproduction matches what I see: the two cleanups were sequential statements, so a throwing `CLOSE_ASYNC_THREAD_POOL` skipped `CLOSE_FLINK_RUNNER_CONTEXT` entirely.

It's the worst place in the chain for that, for the reason you give — `PythonBridgeManager` closes the interpreter on the very next rung, so the skipped cleanup has no later chance to run. `close()` now uses the same first-failure/suppressed ladder as the managers above it, and still nulls `pythonRunnerContext` in the `finally` so a repeated close can't double-free.

Three tests in `PythonActionExecutorTest`: the runner-context cleanup still runs when the shutdown fails, the first failure is rethrown with the later one suppressed, and the `Error` case. Restoring the original method fails all three.

One note on scope, since this overlaps #944: @joeyutong had offered to handle this method's `Error` path as part of their rebase, which is why I'd left it alone. Doing it here instead means that's one less thing for the rebase, and their `IOUtils.closeAll` inside this method lands on top of a ladder rather than replacing one. Flagging so the two of you aren't both holding it.

runtime/src/main/java/org/apache/flink/agents/runtime/ResourceCache.java
emecii: Good catch — this is the branch the guarantee genuinely did not reach, and the `finally` you point at is what makes it unrecoverable rather than merely delayed. Fixed in 23312147.

`closeRepos()` now uses the `Throwable` ladder with `firstOrSuppressed`/`rethrowException`. Two tests, including the one through the path you asked for: `ResourceCacheTest.closeClosesEverySkillRepositoryWhenAnEarlierRepoThrowsError` drives real `SkillRepository` instances through `ResourceCache.close()` → `ResourceContextImpl.close()` → `SkillManager.close()`, rather than asserting at the `SkillManager` level only.

Worth reporting how the first attempt went wrong, since it affects how much the result is worth. I initially wrote both tests with one failing repo and one healthy one — and the mutation check showed the `SkillManager` test *passing* against the narrowed catch. The de-dup set is an `IdentityHashMap`, so iteration order is unspecified: when the healthy repo happens to run first, both end up closed even with the bug present. Both tests now use two failing repos, which fails deterministically whichever runs first; I confirmed that over three runs before pushing.

Also updated a comment in `errorDuringRegistrationStillClosesRepoAndSuppressesCloseError` that documented the old limitation ("closeRepos() catches only Exception per repo, so an Error ends the iteration") as intended behavior — no longer true.

runtime/src/main/java/org/apache/flink/agents/runtime/ResourceCache.java
emecii: Agreed — aligned in 23312147, with the same `first_closed=True, second_closed=False` behavior as your reproduction. Python now attempts every resource, always reaches `_cache.clear()` and `_resource_context.close()`, and re-raises the first failure.

Two deliberate differences from the Java side, documented in the docstring rather than left for a reader to notice:

- **`Exception`, not the literal analogue of `Throwable`.** The Java argument does not carry over: Python's `Exception` already covers what Java calls `Error` (`MemoryError`, `RecursionError`), while `BaseException` would additionally swallow `KeyboardInterrupt` and `SystemExit`, which cleanup must not do. So the parity is in the contract, not the catch clause.
- **Later failures are logged, not attached.** `ExceptionGroup` needs 3.11 and `requires-python` is `>=3.10,<3.13`, so there is no real equivalent of `addSuppressed` available.

Four tests in `python/flink_agents/runtime/tests/test_resource_cache_close.py` covering the failing-resource path, first-failure-wins, a resource-context failure, and the healthy path. Restoring the sequential version fails the first two.

Two related things I looked at and deliberately did not change, so they are visible rather than silently skipped:

- `SkillManager.close()` on the Python side uses `contextlib.suppress(Exception)` per repo. It already closes every repo, so the close-all property holds, but it reports nothing at all — the mirror image of the Java gap. Changing swallow → raise is a behavior change for its callers, so it seemed wrong to fold into this PR uninvited. Happy to do it here or file it, whichever you prefer.
- `FlinkRunnerContext.close()` above it already sequences its two components correctly.

wenjin272 (approved):

One remaining issue is the conflict with the latest main.

emecii:
Rebased onto main (`fabc5f14`) — conflicts resolved, `mergeable` is clean again and CI is 28/28. Thanks for the review.

Flagging one part of the resolution, because it was semantic rather than textual and is worth a second pair of eyes.

Three files conflicted, two of them only on import blocks. The real one was `ActionExecutionOperator.close()`: #924 replaced `eventRouter.close()` with `eventLogWriter.close()` in the close chain, and `EventRouter` no longer has a `close()` method at all. Taking either side wholesale would have been wrong — "ours" drops `eventLogWriter` from the close path and does not compile, "theirs" drops the aggregation. The ladder now closes:

```java
resourceCache, contextManager, pythonBridge, eventLogWriter, durableExecManager
```

then `super.close()`, which matches main's sequence element for element. `ActionExecutionOperatorTest` mocks `EventLogWriter` instead of `EventRouter` accordingly, so the `InOrder` chain still pins the full set.

Verified on the rebased branch rather than assuming the pre-rebase runs still held: `./tools/ut.sh -j` 1661 tests / 0 failures / 0 errors / 40 skipped, Python 936 passed / 13 skipped, Ruff, Spotless and RAT clean. Note this needed a full `./tools/build.sh` first, since main moved to `0.4-SNAPSHOT`.

Also updated the `Generated-by` trailers to the format AGENTS.md adopted while this was open (`Generated-by: Claude Code 2.1.227 (Claude Opus 5)`). The old line had a blank line before `Co-Authored-By`, so git was not parsing it as a trailer at all; both are now in the same block and `git log --format=%(trailers)` picks them up.

emecii:
@wenjin272 Thanks for the review. The conflicts have been resolved. PTAL again. Ty!

Issue #990, opened 2026-08-09
[Bug] The bundled NOTICE omits 18 third-party artifacts shaded into the distribution jar


weiqingy:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The `META-INF/NOTICE` bundled inside the distribution jar declares 177 third-party artifacts. The jar actually shades **195**. The 18 missing ones are distributed with no attribution entry at all.

By entry count, the largest omissions are `com.google.genai:google-genai` (3,363 classes), `io.opencensus:opencensus-api` and `opencensus-contrib-http-util` (413), `com.google.auth:google-auth-library-*` (188), and `org.java-websocket:Java-WebSocket` (100).

All 18 trace to a single cause: the Gemini integration's transitive tree, plus five platform classifiers of `netty-tcnative-boringssl-static` whose bare coordinate was already listed. This looks like one missed NOTICE update rather than accumulated drift, and there is **no over-listing** on the other side: all 177 existing rows do correspond to bundled artifacts.

Four of the omitted dependencies are under licenses whose text the distribution does not carry at all: `Java-WebSocket` (MIT), `api-common` and `google-auth-library` (BSD-3-Clause), and `javax.annotation-api` (CDDL 1.1 with the GPLv2 Classpath Exception). For the BSD-3 and MIT ones the retained copyright notice is the substance of the obligation, so the omission is not merely cosmetic.

This is worth fixing before the next release vote, since LICENSE and NOTICE accuracy is checked there.

### How to reproduce

Build the distribution and compare what the shaded jar contains against what the NOTICE declares:

```bash
./tools/build.sh
# rows declared in the bundled NOTICE
unzip -p dist/common/target/flink-agents-dist-common-*.jar META-INF/NOTICE | grep -c '^- '
# third-party artifacts actually shaded in
mvn dependency:list -pl dist/common -DincludeScope=runtime
```

The dependency list resolves 195 third-party artifacts (excluding the project's own modules and `slf4j-api`, which `artifactSet` excludes) against 177 declared rows.

A direct spot-check of a single omission:

```bash
unzip -l dist/common/target/flink-agents-dist-common-*.jar | grep -c 'com/google/genai'   # 3363
unzip -p dist/common/target/flink-agents-dist-common-*.jar META-INF/NOTICE | grep -c genai  # 0
```

### Version and environment

[...]

PR #991, opened 2026-08-09
[dist] Add missing NOTICE entries for bundled dependencies

weiqingy:
Linked issue: #990

### Purpose of change

The bundled `META-INF/NOTICE` declared 177 third-party artifacts while the shaded distribution jar contains 195, so the distribution shipped several thousand classes of third-party code with no attribution entry.

This adds the 18 missing rows and the four license texts the distribution did not yet carry. The change is additive: no existing entry is removed, reworded or reordered.

All 18 trace to one cause rather than to accumulated drift. `com.google.genai:google-genai` and its transitive tree account for 13 of them, and the other five are platform classifiers of `netty-tcnative-boringssl-static`. The bare coordinate was already listed, but the file's invariant is one row per resolved coordinate including classifier, which is why `netty-transport-native-epoll` already appears only in its classified form. Both the bare and the classified tcnative jars are bundled, so tcnative needs all six rows.

Four dependencies are under licenses whose text was not present under `META-INF/licenses/`, so their texts are added: `Java-WebSocket` under MIT, `api-common` and `google-auth-library` under BSD-3-Clause, and `javax.annotation-api` under CDDL 1.1 with the GPLv2 Classpath Exception. The last needed a new section, since the file had no CDDL section before. For the MIT and BSD-3 dependencies the retained copyright notice is the substance of the obligation, so these are the entries where accuracy matters most.

Each license text was taken from the dependency's own release, and the license classification for every added artifact was read from that artifact's POM or parent POM rather than from general knowledge of what those projects usually use. None of the added dependencies ships its own `META-INF/NOTICE`, so there was nothing to propagate.

There is no over-listing to correct here. All 177 pre-existing rows do correspond to artifacts actually bundled.

### Tests

No new tests. This changes a resource file, and there is no existing test that reads NOTICE.

Verified on the branch: `tools/build.sh` clean, `tools/check-license.sh` clean (relevant, since the change adds files to a licensing directory). The jar inside the built wheel was checked to carry the updated NOTICE with 195 rows, matching the 195 bundled third-party artifacts one to one in both directions, and the four new license files.

The omission set was derived twice by different routes and reconciled: once from the shade plugin's own `Including <coordinate>` log, and once from `mvn dependency:list -DincludeScope=runtime` across the `dist` modules. Both give the same 195 and the same 18. As a completeness check, every entry in the built jar was reconciled against the contents of the 211 bundled artifact jars: of 83,265 class entries, none were unaccounted for.

### API

No. This changes only bundled license metadata.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

`Generated-by: Claude Code 2.1.226`, included in the commit message.

### DIFF
--- a/dist/src/main/resources/META-INF/NOTICE
+++ b/dist/src/main/resources/META-INF/NOTICE
@@ -66,6 +66,11 @@ This project bundles the following dependencies under the Apache Software Licens
 - io.netty:netty-transport-native-kqueue:osx-x86_64:4.1.118.Final
 - io.netty:netty-transport-classes-kqueue:4.1.118.Final
 - io.netty:netty-tcnative-boringssl-static:2.0.70.Final
+- io.netty:netty-tcnative-boringssl-static:linux-x86_64:2.0.70.Final
+- io.netty:netty-tcnative-boringssl-static:linux-aarch_64:2.0.70.Final
+- io.netty:netty-tcnative-boringssl-static:osx-x86_64:2.0.70.Final
+- io.netty:netty-tcnative-boringssl-static:osx-aarch_64:2.0.70.Final
+- io.netty:netty-tcnative-boringssl-static:windows-x86_64:2.0.70.Final
 - io.netty:netty-tcnative-classes:2.0.70.Final
 - io.projectreactor.netty:reactor-netty-http:1.0.48
 - io.netty:netty-resolver-dns:4.1.112.Final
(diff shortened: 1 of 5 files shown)


wenjin272 (approved):
Thanks for taking this on. LGTM.

PR #992, opened 2026-08-09
[e2e] Add a checkpoint-recovery test for restored Python state

weiqingy:
Linked issue: #836

### Purpose of change

No test covers Python agent memory surviving the loss of a TaskManager process. Existing e2e suites run on MiniCluster, which never recreates the embedded Python interpreter, so a value Pemja cannot materialize into a JVM type fails only on a real restore. This test hard-kills a real TaskManager and asserts the tool-call context and a raw `bytes` value both survived.

#### Runtime flow

1. The harness configures checkpointing and a restart strategy on a standalone cluster, stages this checkout's artifacts, and submits the job detached.
2. The agent writes the blob to memory, bumps a process-local counter, and requests a chat. The mock model emits one tool call; the tool writes `tool-entered` and blocks on `release`.
3. On `tool-entered` it waits for two further completed checkpoints, records the pid, resource id and restore count, SIGKILLs the TaskManager, starts a replacement, waits for a restore, then writes `release`.
4. The re-executed tool returns its sentinel. The agent checks the restored blob and transcript against the counters, writes `verdict.json`, and raises on mismatch.

#### Key decisions

The source must be unbounded. A bounded one raises end-of-input right after the record, and `ActionExecutionOperator.endInput()` then holds the mailbox thread while the run is parked, so no checkpoint completes. Measured on the pinned stack: unbounded 16 completions in 15s, bounded 0.

Provenance rests on module-level counters outside Flink state, since anything checkpointed is restored with the payload and cannot tell a restored read from the reader's own write. Pemja's singleton interpreter keeps globals across an in-place task restart, so zero counters prove a different OS process.

`counts.restored` rather than `numRestarts`, which counts a restart that restored nothing.

#### Behavioral contracts

1. A pass requires a published `verdict: "pass"`; a run recording no assertion exits non-zero.
2. The interval comes from the job's own coordinator config, not the harness's file.
3. The payload sits in a checkpoint taken while parked: two completions past a park-time baseline.
4. `counts.restored` rises above its pre-kill baseline.
5. `latest.restored` is non-null. If 4 and 5 disagree the run fails rather than picking one.
6. The process was replaced: pid and resource id both differ from pre-kill.
7. The values came from another process: both counters are 0 where they are asserted.
8. The blob is exactly `bytes` with the exact content; a `bytearray`, a `bytes` subclass, or a list of the same ints all fail.
9. The kill/restart window fits the tool's deadline: a pre-flight check plus a per-wait clamp.
10. The TaskManager ran this checkout: version `0.3.dev0`, `flink_agents.api` under its path.
11. Config edits are idempotent and reverted, leaving no duplicate top-level YAML key.
12. The new workflow job never runs on a pull request.
13. The output schema the request carries is rebuilt from restored state and applied to the round-two response, and a round that never applied it cannot pass.

#### Failure behavior

A missing `handshake_dir` or `verdict_dir` raises `ValueError`. Checkpointing off or a wrong interval records a FAIL and exits non-zero, and an infeasible budget exits 1.

An unreachable, slow, or wrongly shaped REST response never reads as a false condition. Requests are bounded by `CURL_MAX_TIME`; an empty body, bad JSON, a missing path, or a wrongly typed leaf all fail the read.

One path degrades deliberately. The framework absorbs a tool exception into a `... execute failed.` response rather than failing the action, so the tool returns a sentinel the assertion requires. The agent publishes the verdict before raising `AssertionError`, so a failed run leaves a readable file.

### Tests

| Contract | Tests |
|---|---|
| 1 verdict only | 7 `assert_verdict`, 3 `print_summary` |
| 2 checkpointing | 4 `assert_checkpointing_enabled` |
| 3 two checkpoints | none; see negative control |
| 4 restored count | none; see provenance control |
| 5 restored non-null | none; see provenance control |
| 6 process replaced | none |
| 7 counters | none; exercised via 8 |
| 8 blob predicate | `_blob_matches`, 10 cases |
| 9 budget | 4 `assert_handshake_budget`, 3 `charged_timeout`, 1 `handshake_budget_left` |
| 10 identity | none |
| 11 config idempotent | 6 `configure_flink` / `delete_config_key` |
| 12 not on a PR | none |
| 13 output schema | `_structured_transcript`, 7 cases |
| handshake, for 3 and 7 | 3 `await_release` |

About 45 of the 82 bats tests pin harness plumbing the contracts above deliberately do not cover: REST reads, file waits, cluster and venv setup, and EXIT-trap cleanup.

The 82 bats and 19 Python tests run pre-merge; the cluster test does not, since its job is `schedule` and `workflow_dispatch` only. Fork runs are the only evidence, and the controls are what make the pass meaningful.

Happy path: 9/9 assertions, verdict `pass`, pid 4567 to 5892 with the resource id changed, `counts.completed` 3 while parked, `counts.restored` 0 to 1, both counters 0, blob back as `bytes`, and the restored output schema applied (`structured_ok`, observed as a `dict`).

Negative control, release marker pre-created: failed at `checkpoint-containment`, `counts.completed` stuck at 1 against a required 3.

Provenance control, kill skipped: failed at `restore`, `counts.restored` stayed 0 and the pid was never replaced. The failure lands at the restore wait rather than at the verdict.

It does not reproduce the SIGSEGV that motivated the issue: the memory value validator now raises `TypeError` at `set()` before Pemja is reached, so that control no longer exists. Unit tests pin which types the validator admits; only this one shows an admitted type survives a restore.

Known limitations: timeouts are sized from observed runs, not proven bounds; and 11 pre-existing bats assertions are vacuous under bash 3.2 bodies, since `errexit` does not apply to `[[ ]]`.

### API

No public API change, and no product code touched. Six of the seven files are new; the seventh adds a job to the nightly workflow, leaving its existing job, triggers and paths filter untouched. Nothing changes for a caller who does nothing differently.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Claude Code 2.1.224

### DIFF
--- a/e2e-test/test-scripts/test_checkpoint_recovery.sh
+++ b/e2e-test/test-scripts/test_checkpoint_recovery.sh
@@ -123,11 +123,58 @@ CHECKPOINT_DIR=""
 INPUT_DIR=""
 INPUT_FILE=""
 FLINK_CONF=""
+FLINK_CONF_BACKUP=""
 TM_PID_BEFORE=""
 TM_RESOURCE_ID_BEFORE=""
 RESTORED_BEFORE=""
 RELEASE_DEADLINE_S=""
 HANDSHAKE_DEADLINE_AT=""
+
+# ---------------------------------------------------------------------------
+# A reused Flink home carries someone's own config.yaml. set_config_key deletes any
+# line already present for a key before appending ours, so reverting key by key
+# cannot put back a value the installation already had: the whole file is copied
+# aside before the first edit and copied back at exit. config.yaml is the only
+# thing copied aside — the jars this run stages into lib/ are not.
+#
+# The copy carries mode as well as content. delete_config_key rewrites through
+# mktemp and mv, which replaces the file with a 0600 temp file, so restoring the
+# bytes alone would leave the installation's config readable only by this user.
+#
+# The copy lives under TMPDIR rather than in WORK_DIR or in the installation:
+# WORK_DIR does not exist yet when the copy is taken and is removed on a clean
+# exit, and writing into conf/ mutates the directory being protected.
+# ---------------------------------------------------------------------------
+backup_flink_conf() {
+    FLINK_CONF_BACKUP="$(mktemp "${TMPDIR:-/tmp}/flink-agents-config.yaml.XXXXXX")" || {
+        log_error "Could not create a temporary file to copy $FLINK_CONF aside"
+        exit 1
+    }
+    # `if !` rather than a bare cp: set -e would abort before the diagnostic prints.
+    if ! cp -p "$FLINK_CONF" "$FLINK_CONF_BACKUP"; then
+        rm -f "$FLINK_CONF_BACKUP"
+        FLINK_CONF_BACKUP=""
+        log_error "Could not copy $FLINK_CONF aside. Refusing to edit an installation that cannot be put back."
+        exit 1
+    fi
+    log_info "Copied $FLINK_CONF aside for restore at exit"
+}
+
+restore_flink_conf() {
+    # Guarded on the copy, not on the destination: a config.yaml that went missing
+    # mid-run is exactly when the restore is needed, and cp recreates it. The
+    # explicit `|| return 0` is load-bearing under bash 3, where a bare [[ ]] as a
+    # non-final command does not trigger errexit.
+    [[ -n "$FLINK_CONF" && -n "$FLINK_CONF_BACKUP" && -f "$FLINK_CONF_BACKUP" ]] || return 0
+    if cp -p "$FLINK_CONF_BACKUP" "$FLINK_CONF"; then
+        log_info "Restored $FLINK_CONF from the pre-run copy"
+        rm -f "$FLINK_CONF_BACKUP"
+    else
+        # Keep the only surviving original and say where it is.
+        log_warn "Could not restore $FLINK_CONF; the pre-run copy is kept at $FLINK_CONF_BACKUP"
+        return 1
+    fi
+}
 
 cleanup() {
     local exit_code=$?
@@ -159,16 +206,18 @@ cleanup() {
         fi
     fi
 
-    # Strip the keys we appended. A second run against a reused Flink home would
-    # otherwise append them twice, and a duplicate key is a hard YAML parse failure
-    # that stops the cluster from starting at all.
-    if [[ -n "$FLINK_CONF" && -f "$FLINK_CONF" ]]; then
-        local key
-        for key in "${CONFIG_KEYS_SET[@]:-}"; do
-            [[ -n "$key" ]] || continue
-            delete_config_key "$key"
-        done
-        log_info "Reverted ${#CONFIG_KEYS_SET[@]} appended config key(s)"
+    # Put the installation's config.yaml back exactly as it was, mode included.
+    # Restoring the whole file also drops the keys this run appended, so a second
+    # run against a reused Flink home finds no duplicate top-level key — which
+    # would be a hard YAML parse failure that stops the cluster from starting at
+    # all.
+    #
+    # Tested inside the `if` so a failure cannot trip errexit here: aborting the
+    # trap would skip print_summary, which is what reports the results and settles
+    # the exit status. A run that left the installation mutated is not a success,
+    # so a restore failure with nothing else to report becomes the exit code.
+    if ! restore_flink_conf && (( exit_code == 0 )); then
+        exit_code=1
     fi
 
     # The killed TaskManager's entry stays in /tmp/flink-*-taskexecutor.pid, which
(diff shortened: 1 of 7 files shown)


e2e-test/test-scripts/test_checkpoint_recovery.sh
```diff
@@ -0,0 +1,1307 @@
...
+    tmp="$(mktemp)"
+    # grep -v exits 1 when it filters everything out, which set -e would treat as
+    # fatal. The dots in the key are regex "any char"; harmless for these keys.
+    grep -v "^${config_key}: " "$FLINK_CONF" > "$tmp" || true
+    mv "$tmp" "$FLINK_CONF"
+}
+
+set_config_key() {
+    local config_key="$1" value="$2"
+    delete_config_key "$config_key"
```
wenjin272, line 718: Could we preserve and restore the original `config.yaml`? When reusing `FLINK_HOME`, `set_config_key` removes pre-existing values, while cleanup only removes the test values, so the original configuration is lost. Backing up the file before modification and restoring it on `EXIT` would avoid mutating the reused installation.

python/flink_agents/e2e_tests/e2e_tests_integration/checkpoint_recovery_agent.py
```diff
@@ -0,0 +1,447 @@
...
+        ctx.short_term_memory.set(_BLOB_MEMORY_KEY, _KNOWN_BLOB)
+        _mark_blob_written()
+
+        ctx.send_event(
+            ChatRequestEvent(
+                model="recovery_chat_model",
+                messages=[
+                    ChatMessage(role=MessageRole.USER, content=input_data.content)
+                ],
+                prompt_args={"task": input_data.content},
```
wenjin272, line 449: This request leaves `output_schema` unset, so the E2E only exercises the `None` branch. Since #828 specifically normalized `OutputSchema` for checkpoint safety, a regression in its serialization or reconstruction would remain undetected. Could we include a non-null `OutputSchema` in this recovery flow?

python/flink_agents/e2e_tests/e2e_tests_integration/checkpoint_recovery_agent.py
```diff
@@ -0,0 +1,447 @@
...
+
+    ``bytes`` is admitted because it is the only one of the two the memory value
+    validator accepts at write time. ``bytearray`` is admitted because the value
+    comes back across the bridge from a Java ``byte[]`` and may materialize as
+    either — a read-side allowance this test makes, not one the write-side contract
+    grants. Nothing else is admitted on the chance it might appear; anything else
+    fails here rather than aborting the action, so the run still publishes a verdict
+    and ``blob_observed_type`` records what did come back.
+    """
+    if not isinstance(raw, bytes | bytearray):
```
wenjin272: Could we require exact `bytes` here? The memory contract deliberately accepts `bytes` but rejects `bytearray`. With the current check, a restored `bytearray` with the same content would still pass. Using `type(raw) is bytes and raw == _KNOWN_BLOB` would make this a true `bytes` round-trip assertion.

wenjin272 (commented):
Thanks for putting together this comprehensive recovery harness. The checkpoint provenance checks and failure diagnostics are particularly thoughtful. I left four comments on cases that could mutate a reused installation or leave important regressions undetected.

e2e-test/test-scripts/test_checkpoint_recovery.sh
```diff
@@ -0,0 +1,1307 @@
...
+}
+
+start_cluster() {
+    log_section "Start Flink standalone cluster"
+    "$FLINK_HOME/bin/start-cluster.sh"
+
+    log_info "Waiting for JobManager REST API at $REST_URL ..."
+    local deadline=$((SECONDS + CLUSTER_TIMEOUT))
+    while (( SECONDS < deadline )); do
+        if curl -fsS "$REST_URL/overview" >/dev/null 2>&1; then
```
wenjin272: Could this poll use `rest_get` or add `curl --max-time`? The current request is unbounded, so a single stalled connection can exceed `CLUSTER_TIMEOUT` and potentially hang until the workflow-level timeout.

wenjin272 (approved):

weiqingy:

While working through your comments I ran into something separate and filed it as #1035. The bash 4+ guard in tools/test/run.sh only binds the launcher. bats-core's bats-exec-test re-execs through /usr/bin/env bash, so test bodies still run under /bin/bash 3.2 on macOS. Under 3.2 errexit does not apply to [[ ]], so a non-final [[ ]] assertion that is false reports ok and execution carries on.

There are 42 such assertions across the bats suites, and 33 of them are in the file this PR added, so most of the cleanup is mine.

I will send a separate PR for it.

PR #995, opened 2026-08-11
[hotfix] Add help handling to build script

kaiwangleo:
Linked issue: N/A (hotfix)

### Purpose of change

Add `-h`/`--help` handling to `tools/build.sh` and show usage for unknown options instead of failing because `show_help` is undefined.

### Tests

- Added `tools/test/integration/build_help.bats` covering `--help`, `-h`, and unknown options.
- Manually verified `bash tools/build.sh --help`, `bash tools/build.sh -h`, and `bash tools/build.sh --no-such-option` (Bats is not installed locally).

### API

No public API changes.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

### DIFF
--- a/tools/test/integration/build_help.bats
+++ b/tools/test/integration/build_help.bats
@@ -1,0 +1,47 @@
+#!/usr/bin/env bats
+
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+BUILD_SCRIPT="${BATS_TEST_DIRNAME}/../../build.sh"
+
+@test "build --help prints usage and exits 0" {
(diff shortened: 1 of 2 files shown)


tools/test/integration/build_help.bats
```diff
@@ -0,0 +1,47 @@
...
+    run bash "$BUILD_SCRIPT" -h
+
+    [ "$status" -eq 0 ]
+    [[ "$output" == *"Usage:"* ]]
+}
+
+@test "build rejects an unknown option with usage" {
+    run bash "$BUILD_SCRIPT" --no-such-option
+
+    [ "$status" -ne 0 ]
```
weiqingy: nit: `-ne 0` passes whether the script is broken or fixed. On `main` this path exits 127, and with your fix it exits 1, so both satisfy the check. I tried changing the `*)` branch to `exit 2` locally and all three tests still passed, so nothing here locks in the exit code your fix actually produces. The test does still go red against `main`, just through the `Usage:` check below it, so it is doing its job.

Would `[ "$status" -eq 1 ]` be worth using instead? That would say an unknown option is a usage error rather than a crash. Curious whether you left it loose on purpose.

weiqingy (commented):
Thanks for taking this on. One question inline, nothing blocking.

tools/test/integration/build_help.bats
kaiwangleo: Updated in commit 3605821f: the unknown-option regression test now asserts the expected usage-error exit code with `[ $status -eq 1 ]`. The focused Bats test passes.

kaiwangleo: Thanks for pointing this out. I left the assertion broad initially because the test only needed to verify that an unknown option failed, regardless of the specific non-zero status. However, since `tools/build.sh` treats unknown options as usage errors and explicitly exits with status `1`, asserting `[ $status -eq 1 ]` makes the expected behavior clearer and prevents unrelated failures such as exit code `127` from satisfying the test. I’ve updated the test in commit `3605821f`; the focused Bats tests pass.

wenjin272 (approved):

Issue #999, opened 2026-08-12
[Bug] Elasticsearch queryEmbedding applies filters as a post-filter, returning fewer than k documents


weiqingy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

`ElasticsearchVectorStore.queryEmbedding` attaches its filter with `SearchRequest.Builder.postFilter(...)`, while the KNN clause itself carries no `filter` option.

Elasticsearch applies `post_filter` after the kNN phase has already selected its `k` nearest hits, so the filter can only remove documents from that set. It never pulls in matching documents that fell outside the top `k`. A caller asking for 5 documents matching `user_id=alice` can get zero back, even with hundreds of alice's documents indexed, if the 5 nearest vectors happen to belong to other users.

`get` and `delete` build the same filter map into a real query clause, so they are exhaustive. The three methods diverge, and the divergence is not visible from the API.

Both filter forms are affected, since the unified `filters` map and a raw `filter_query` are merged into a single query by `combineQueryJson` before it is attached.

Expected: `queryEmbedding` returns up to `k` documents that match the filter.
Actual: it returns whichever of the unfiltered `k` nearest happen to match, which can be none.

The Elasticsearch Java client exposes `KnnSearch.Builder.filter(...)`, which applies the filter during the kNN search rather than after it.

This would be a user-visible behavior change. Callers who pass a filter to `queryEmbedding` today receive at most the matching subset of the top `k`, and would afterwards receive up to `k` matching documents.

### How to reproduce

1. Start a single-node Elasticsearch with `xpack.security.enabled=false`.
2. Index documents belonging to several users, each carrying metadata such as `user_id`.
3. Call `queryEmbedding` with `k` of 5, a filter for one user, and an embedding nearest to another user's documents.
4. Fewer than 5 documents come back, possibly none, even though more than 5 documents match the filter.

Note that `ElasticsearchVectorStoreTest` is annotated `@Disabled("Should setup Elasticsearch server.")`, so the existing `testFiltersDsl` coverage does not run in CI and would not catch this.

### Version and environment

`main`, elasticsearch-java 8.19.0. Noticed while reviewing #997.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

kaiwangleo:
I reviewed the current implementation and the proposed fix against the Elasticsearch Java client API.

The issue is reproducible from the current code path:

[...]

PR #1001, opened 2026-08-12
[integrations][vector-store][java] Apply Elasticsearch KNN filters before selecting neighbors

weiqingy:
Linked issue: #999

### Purpose of change

`queryEmbedding` attached its filter with `post_filter`, which Elasticsearch applies after the KNN phase has already selected its `k` nearest hits. The filter could then only discard documents from that set, and never reach matching documents that fell outside the top `k`. A caller asking for five documents matching `user_id=alice` could get none back while hundreds were indexed, whenever the nearest vectors belonged to other users.

This moves the filter into the KNN clause, so the nearest neighbors are selected from among the documents that match. `get` and `delete` already build the same filter map into a real query clause, so this also removes the divergence between the three methods.

Both filter forms are affected, since the unified `filters` map and a raw `filter_query` are merged into a single query by `combineQueryJson` before it is attached.

### Tests

`ElasticsearchVectorStoreTest#testQueryEmbeddingFiltersBeforeSelectingNeighbors` indexes six documents pointing along the query vector and three orthogonal documents belonging to another user, then queries with `k` of 5 and a filter selecting the three. It fails with `expected: <3> but was: <0>` before the change and passes after, run against Elasticsearch 8.19.0.

The test uses `addEmbedding` with explicit vectors so the result depends only on where the filter is applied.

Note that `ElasticsearchVectorStoreTest` is annotated `@Disabled("Should setup Elasticsearch server.")`, so this test does not run in CI and needs a local Elasticsearch, as with the rest of the class.

### API

No signature changes. This does change observable behavior: callers passing a filter to `queryEmbedding` previously received at most the matching subset of the top `k`, and now receive up to `k` matching documents.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

- [x] Yes
- [ ] No

### DIFF
--- a/docs/content/docs/development/vector_stores.md
+++ b/docs/content/docs/development/vector_stores.md
@@ -800,7 +800,7 @@ Elasticsearch is currently supported in the Java API only. To use Elasticsearch 
 | `dims`            | int  | `768`                     | Vector dimensionality                                              |
 | `k`               | int  | None                      | Number of nearest neighbors to return; can be overridden per query |
 | `num_candidates`  | int  | None                      | Candidate set size for ANN search; can be overridden per query     |
-| `filter_query`    | str  | None                      | Raw JSON Elasticsearch filter query (DSL) applied as a post-filter |
+| `filter_query`    | str  | None                      | Raw JSON Elasticsearch filter query (DSL) restricting which documents a KNN query can match |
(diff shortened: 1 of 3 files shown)


wenjin272 (approved):
Thanks for fixing this. LGTM

Issue #998, opened 2026-08-12
[Bug] Long-term memory operations read the partition key from shared mutable state on the async worker thread


weiqingy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

`Mem0LongTermMemory` is shared across partition keys and holds the current key in a mutable field. `switch_context` sets it on the mailbox thread before each action (`mem0_long_term_memory.py:310`), while `add`, `get`, `search`, `delete` and `delete_memory_set` can run on a pool thread when an action passes them to `durable_execute_async`. Those operations read the field themselves, so the key they see is whatever the mailbox thread last wrote, not the key of the action that submitted the work.

The key is the isolation boundary, not just observability metadata: it is passed to Mem0 as `agent_id` (`mem0_long_term_memory.py:504` for `add`, `:457` for `delete_memory_set`). Two keys sharing a job id and a memory set name are separated by `agent_id` alone, so reading the wrong one means one key's items land in or are read from another key's set.

The same applies to `_observation_id` and `_observation_suppressed`, which are read from the same shared state in all five operations (`:450-451, :488-489, :553-554, :608-609, :665-666`), so observations can be misattributed independently of the memory itself.

Several keys really can be in flight at once. An action that awaits suspends without completing (`plan/function.py:436,451`), the operator re-queues it as a fresh mail (`ActionExecutionOperator.java:502-508,537`), and processing keys are tracked in a `ListState` rather than as a single key (`OperatorStateManager.java:83`). Nothing in `Mem0LongTermMemory` holds a lock or thread-local.

Two things worth stating plainly:

- **The window is narrow.** Each operation copies the field into a local on its first lines, so the exposure is from `executor.submit` (`flink_runner_context.py:159`) to that first statement, not the whole Mem0 round trip.
- **No failure has been attributed to this.** It is a structural defect found by reading the code. #983 recorded one unexplained `assert 0 == 1` in the long-term memory e2e test, and this race would produce that symptom, but nothing links the two. #984 adds a cross-key leak check in both directions so a future occurrence is diagnosable.

[...]

PR #1002, opened 2026-08-12
[runtime][api] Scope long-term memory operations to the action that obtained the memory set

weiqingy:
Linked issue: #998

### Purpose of change

`Mem0LongTermMemory` is shared across partition keys and holds the current key, observation id and suppression flag in mutable fields. `switch_context` writes them on the mailbox thread before each action, while `add`, `get`, `search` and `delete` can run on a worker thread when an action passes them to `durable_execute_async`, and those operations read the fields themselves. The key they see is whatever the mailbox thread wrote most recently, not the key of the action that submitted the work.

The key is the isolation boundary rather than observability metadata: it reaches Mem0 as `agent_id`, and two keys sharing a job id and a memory set name are separated by `agent_id` alone. Reading the wrong one means one key's items land in, or are read from, another key's set. The observation id and suppression flag come from the same fields, so observations can be misattributed independently.

This binds the partition key, observation id and suppression flag onto the `MemorySet` when it is created, and has the four operations take them from the set. The four already receive `memory_set` as their first parameter, so nothing new is threaded through. A set is consequently scoped to one action and must not be held across actions.

Java is covered in the same change rather than separately. Its operations all delegate into the same Python object, and the bridge rebuilt the Python set from its name alone, so once the Python side sources the key from the set, a Java-originated call would arrive without one. The Java wrapper now records the context when it switches and carries it across the bridge, and `to_python_memory_set` requires the key rather than defaulting it.

#### Refusing a key Mem0 cannot isolate

Mem0 tests `agent_id` for truthiness instead of matching on it, so both an unbound key and an empty one would drop the filter entirely and widen an operation to every key sharing the job id and set name, which for a delete would remove another key's items. An `add` under such a key would be worse than mis-filtered: Mem0 omits `agent_id` from the stored metadata, leaving the item unattributed to any key.

The empty string is a legal Flink key, reachable by keying a stream on a string field that can be empty, so both states are now refused at the point of use. They carry separate messages because the remedies differ. Encoding the empty key instead was rejected: keeping already-stored memories readable requires the mapping to be the identity on non-empty keys, which leaves no unoccupied non-empty value for the empty key to take. Reserving a sentinel only moves which key fails, since a Flink key arrives as `String.valueOf(key)` and can be any string. A collision-free encoding would therefore have to transform every key and orphan every memory already written.

The unbound state is now `None` and `null` in both languages. Both previously defaulted the memory's own key to the empty string, so neither could distinguish an absent context from the empty key, and the Java wrapper additionally stamped that value onto a `MemorySet` whose own default was null, so an unbound set originating in Java arrived in Python looking bound and passed both guards.

Validation deliberately stays out of `switch_context`, so a job carrying an empty key still runs unless it actually uses long-term memory.

#### Scoping the management operations to the mailbox thread

Binding context onto the set makes the four set-scoped operations safe on a worker thread, but obtaining or deleting a set still reads the key currently in scope. Submitting either to `durable_execute_async` therefore preserved the same cross-key race one call earlier. `getMemorySet` and `deleteMemorySet` now run the runner context's mailbox-thread check, the same one `getResource` uses. The four set-scoped operations stay unchecked, since being usable from a worker thread is the point of binding the context to the set.

The check fires in Java as well as Python. Java's `getMemorySet` is a pure factory that stamps the Java-side fields and never crosses into Python, so a Python-only check would leave Java agents unguarded. Python receives the checker through the constructor rather than through a new method on the `RunnerContext` ABC, because Java's `checkMailboxThread` lives on the runtime class rather than on the API interface, and adding one to the Python ABC would put a user-facing API method there with no Java counterpart.

`delete_memory_set` keeps its name-based signature. It has no bound context and applies to the key currently in scope, which the mailbox-thread check now makes correct, so closing the gap no longer requires a public signature change. Its documentation records that it can target a different key than `MemorySet.delete` on a same-named set.

#### CI

The cross-language `Mem0LongTermMemoryTest` and `SkillsIntegrationTest` both skip themselves unless `ACTION_API_KEY` is set. The python-it job passes the secret; the java-it and cross-language jobs set only `LOG_LEVEL`, so both Java suites went green with every case skipped. Both jobs now receive it. Repository secrets are still withheld from `pull_request` runs on forks, so this restores the coverage for push builds and same-repository pull requests, matching what python-it already gets.

### Tests

| Test | What it pins |
|---|---|
| `test_memory_set_stays_on_its_own_key_after_the_owner_switches` | all four operations keep the set's key after the owner switches |
| `test_observations_stay_with_the_action_that_obtained_the_set` | observation ownership follows the set for add, get, search and delete |
| `test_suppression_follows_the_set_not_the_current_context` | the set's own suppression flag decides, in both directions |
| `test_unbound_memory_set_is_refused_rather_than_widened` | all four operations raise, and none reaches Mem0 |
| `test_empty_partition_key_is_refused_by_set_operations` | an empty key is refused on all four, and none reaches Mem0 |
| `test_empty_partition_key_is_refused_by_memory_set_management` | obtaining and deleting a set are refused under an empty key |
| `test_memory_set_management_before_any_context_switch_is_refused` | the unbound state is refused rather than treated as a key |
| `test_memory_set_management_is_refused_off_the_mailbox_thread` | both management operations run the check |
| `test_set_scoped_operations_run_without_the_mailbox_thread` | the four set-scoped operations stay usable off the mailbox thread |
| `testForwardedSetCarriesTheContextItWasObtainedIn` | the Java bridge forwards the context bound at creation |
| `testUnboundSetIsRefusedRatherThanWidened` | the Java guard throws before forwarding |
| `testMemorySetIsRefusedBeforeAnyContextSwitch` | the unbound state is refused in Java rather than treated as a key |
| `test_to_python_memory_set_carries_the_action_context` | the bridge helper carries all four fields, not just the name |
| `testEmptyKeyedSetIsRefusedRatherThanWidened` | the Java guard refuses an empty key before marshalling the set |
| `testMemorySetManagementIsRefusedForAnEmptyKey` | obtaining and deleting are refused under an empty key in Java |
| `testMemorySetManagementIsRefusedOffTheMailboxThread` | both Java management operations run the check |
| `testSetScopedOperationsRunWithoutTheMailboxThread` | the four Java set-scoped operations stay usable off the mailbox thread |

`test_switch_context` and `test_context_switch_changes_observation_owner_and_current_suppression` reused a single set across a context switch and expected the current key to apply. They now assert that a set keeps its own key, and obtain one per context. `test_empty_context_key_is_used_consistently_for_mem0_operations` asserted that an empty `agent_id` is forwarded to Mem0, recording the widening as intended. It is replaced by `test_empty_partition_key_is_refused_by_set_operations`, which asserts refusal and that Mem0 is never reached.

Every assertion was checked against a deliberately reverted implementation. Reverting the suppression binding, the observation id binding, any of the refusal guards, the key binding itself, either mailbox check, or the order in which the thread check precedes the key check each leaves at least one test failing.

126 Python tests pass across `api/memory`, `runtime/memory` and `runtime/tests`. 80 Java tests pass across `MemorySetTest`, the runtime `Mem0LongTermMemoryTest`, `MemoryRefTest`, `TestMemoryObservationFlush`, `ActionTaskContextManagerTest` and `ActionExecutionOperatorTest`. `ruff check` and `spotless:check` are clean, and the touched Python files are `ruff format` clean.

The long-term memory e2e test is gated on `ACTION_API_KEY` and needs a Flink cluster plus a local Ollama embedding model, so end-to-end coverage comes from CI on push builds.

### API

`MemorySet` gains three fields in both languages, excluded from serialization like the existing long term memory back-reference. Java adds `setActionContext` and three getters. `BaseLongTermMemory.get_memory_set` / `getMemorySet` and `delete_memory_set` / `deleteMemorySet` keep their signatures and gain a documented per-action, mailbox-thread scope. The runtime bridge helper `to_python_memory_set` now requires the partition key; its only caller is the Java wrapper, and a version mismatch fails loudly with a missing-argument error rather than silently.

`Mem0LongTermMemory` gains a mailbox-thread checker parameter in both languages. It is a runtime class rather than public API, and the public `RunnerContext` interface is unchanged in both languages.

Existing code that obtains a memory set inside the action using it is unaffected. Code that cached one across actions was already reading whichever key happened to be current, and now raises or stays on its original key instead. Code that obtains or deletes a memory set inside a `DurableCallable` or a `durable_execute_async` callable now raises on Python and on Java under JDK 21 and above; under JDK 11 the callable runs inline on the mailbox thread, so it does not raise there yet. A job keyed on a value that can be the empty string now raises when it uses long-term memory, rather than silently sharing one memory set across every key.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

- [x] Yes
- [ ] No

The first commit on the branch predates the model half of that convention and carries `Generated-by: Claude Code 2.1.228`.

### DIFF
--- a/python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
+++ b/python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
@@ -123,12 +123,30 @@ def _create_flink_agents_config_classes() -> tuple:
     return _FlinkAgentsLlmConfig, _FlinkAgentsEmbedderConfig
 
 
+def _reject_empty_partition_key(key: str) -> str:
+    """Return ``key`` unless it is empty, which cannot scope an operation.
+
+    Mem0 matches on ``agent_id`` only when it is truthy, so an empty key widens
+    every operation to all keys sharing the job id and set name, and stores added
+    items with no ``agent_id`` at all.
+    """
+    if not key:
+        msg = (
+            "Long-term memory cannot be scoped to an empty partition key. Mem0 "
+            "ignores an empty agent_id, so the operation would reach every key "
+            "sharing the job id and set name, and added items would be stored "
+            "unattributed. Key the stream by a non-empty value."
+        )
+        raise ValueError(msg)
+    return key
+
+
 def _bound_partition_key(memory_set: MemorySet) -> str:
     """Return the partition key the set is scoped to.
 
-    Mem0 ignores a falsy ``agent_id`` rather than matching on it, so an unbound set
-    would widen every operation to all keys sharing the job id and set name, which
-    for a delete means deleting another key's items. Refuse the operation instead.
+    An unbound set would widen every operation to all keys sharing the job id and
+    set name, which for a delete means deleting another key's items. Refuse the
+    operation instead.
     """
     if memory_set.partition_key is None:
         msg = (
@@ -137,7 +155,7 @@ def _bound_partition_key(memory_set: MemorySet) -> str:
             "than constructing it directly or reusing one across actions."
         )
         raise ValueError(msg)
-    return memory_set.partition_key
+    return _reject_empty_partition_key(memory_set.partition_key)
 
 
 class Mem0LongTermMemory(InternalBaseLongTermMemory):
@@ -155,8 +173,10 @@ class Mem0LongTermMemory(InternalBaseLongTermMemory):
 
     job_id: str = Field(description="Unique identifier for the job.")
 
-    key: str = Field(
-        default="", description="Unique identifier for the keyed partition."
+    key: str | None = Field(
+        default=None,
+        description="Keyed partition currently in scope, or None before the first "
+        "context switch.",
     )
 
     chat_model_name: str = Field(
@@ -442,6 +462,21 @@ class Mem0LongTermMemory(InternalBaseLongTermMemory):
 
         return json.dumps([record.to_wire() for record in records], ensure_ascii=False)
 
+    def _current_partition_key(self) -> str:
+        """Return the partition key currently in scope.
+
+        Whole-set management operations read the key from the memory rather than
+        from a set, so they are only correct once an action has switched context.
+        """
+        if self.key is None:
+            msg = (
+                "Long-term memory has no partition key in scope. Call this from "
+                "an action body, which always runs under a partition key, rather "
+                "than before the first action has run."
+            )
+            raise ValueError(msg)
+        return _reject_empty_partition_key(self.key)
+
     @override
     def get_memory_set(self, name: str) -> MemorySet:
         """Get the memory set by name.
@@ -455,11 +490,14 @@ class Mem0LongTermMemory(InternalBaseLongTermMemory):
 
         Returns:
             The memory set.
+
+        Raises:
+            ValueError: If no non-empty partition key is in scope.
         """
         return MemorySet(
             name=name,
             ltm=self,
-            partition_key=self.key,
+            partition_key=self._current_partition_key(),
             observation_id=self._observation_id,
             observation_suppressed=self._observation_suppressed,
         )
@@ -478,8 +516,11 @@ class Mem0LongTermMemory(InternalBaseLongTermMemory):
 
         Returns:
             True if the memory set was deleted.
+
+        Raises:
+            ValueError: If no non-empty partition key is in scope.
         """
-        observation_key = self.key
+        observation_key = self._current_partition_key()
         observation_id = self._observation_id
         observation_enabled = (
             self._update_observation_enabled and not self._observation_suppressed
--- a/python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
+++ b/python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
@@ -24,7 +24,7 @@ import queue
 from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
-from typing import Any, Dict, List
+from typing import Any, Callable, Dict, List
 
 from pydantic import ConfigDict, Field, PrivateAttr, field_validator
 from typing_extensions import override
@@ -171,6 +171,13 @@ class Mem0LongTermMemory(InternalBaseLongTermMemory):
         description="The runner context to retrieve resources.", exclude=True
     )
 
+    mailbox_thread_checker: Callable[[], None] = Field(
+        description="Raises when called off the Flink mailbox thread. Whole-memory-set "
+        "management reads the partition key currently in scope, which only the mailbox "
+        "thread can read consistently.",
+        exclude=True,
+    )
+
     job_id: str = Field(description="Unique identifier for the job.")
 
     key: str | None = Field(
@@ -221,6 +228,7 @@ class Mem0LongTermMemory(InternalBaseLongTermMemory):
         chat_model_name: str,
         embedding_model_name: str,
         vector_store_name: str,
+        mailbox_thread_checker: Callable[[], None],
     ) -> None:
         """Initialize the Mem0-based Long-Term Memory.
 
@@ -231,6 +239,8 @@ class Mem0LongTermMemory(InternalBaseLongTermMemory):
             embedding_model_name: Resource name of the embedding model.
             vector_store_name: Resource name of a
                 ``CollectionManageableVectorStore`` to back Mem0.
+            mailbox_thread_checker: Callable that raises when invoked off the
+                Flink mailbox thread.
         """
         # Resolve metric group upfront on the main thread so that it is
         # safe to use from any thread later.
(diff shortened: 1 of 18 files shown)


python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
```diff
@@ -429,18 +446,33 @@ def drain_ltm_observation_records(self, key: str, observation_id: str) -> str:
     def get_memory_set(self, name: str) -> MemorySet:
         """Get the memory set by name.
 
+        The current partition key and observation context are copied onto the set
+        so that operations submitted to a worker thread stay scoped to the action
+        that obtained it. Must be called on the mailbox thread.
```
wenjin272: **[P1] Enforce mailbox-thread access for memory-set management**

I think `get_memory_set` and `delete_memory_set`, as memory-set-level management operations, should only run on the mailbox thread. Otherwise, passing either to `durable_execute_async` can preserve the same cross-key race. Could we enforce this in both Java and Python, similar to `RunnerContextImpl#getResource`?

python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
```diff
@@ -123,6 +123,23 @@ def validate_config(cls, v: Any, values: Any) -> Any:
...
 
 
+def _bound_partition_key(memory_set: MemorySet) -> str:
+    """Return the partition key the set is scoped to.
+
+    Mem0 ignores a falsy ``agent_id`` rather than matching on it, so an unbound set
+    would widen every operation to all keys sharing the job id and set name, which
+    for a delete means deleting another key's items. Refuse the operation instead.
+    """
+    if memory_set.partition_key is None:
```
wenjin272, line 151: **[P1] Handle empty partition keys safely**

`""` is a valid Flink key, but Mem0 ignores a falsy `agent_id`, dropping partition isolation. Please reject empty keys or encode them into a non-empty, collision-free value. Also use `None`/`null`, rather than `""`, for the unbound state.

wenjin272 (commented):
Thanks for taking this on @weiqingy. 

BTW, while reviewing the change, I noticed that the cross-language job does not provide `ACTION_API_KEY` like python-it, which caused the Java LTM E2E test to be skipped. I think we can fix this in this PR or in a separate one.

python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
weiqingy: PR description updated.

On refusing rather than encoding: to keep already-stored memories readable the mapping has to leave non-empty keys alone, which leaves no free non-empty value for `""` to take. A reserved sentinel just moves which key breaks. Collision-free would mean rewriting every key's `agent_id` and orphaning what is already stored, so refusing looked like the better trade. Does that match how you would weigh it?

Both `None` and `""` are refused now in both languages, with different messages. The unbound default was `""` on both sides, so an unbound Java set already looked bound by the time it reached Python.

Two things worth knowing. On `add`, mem0 drops `agent_id` from the stored metadata, so items written under an empty key stay unrecoverable even after the key is fixed. And an existing test was asserting that `agent_id=""` gets forwarded, so this was pinned as intended behaviour.

python/flink_agents/runtime/memory/mem0/mem0_long_term_memory.py
weiqingy: Good point, done in both languages.

Both management methods now run the same mailbox check `getResource` uses. The four set-scoped operations stay unchecked on purpose, since being safe on a worker thread is what binding the context to the set buys.

Java needed its own check rather than relying on the bridge, since `getMemorySet` there never enters Python. On the Python side I passed the checker through the constructor rather than adding one to the `RunnerContext` ABC, since Java's equivalent sits on the runtime class rather than the API interface. Curious whether you would have put it on the interface instead.

`getLongTermMemory` is left alone, since capturing the handle on the mailbox thread and using it inside the callable is normal.

This settles the `delete_memory_set` question without touching its signature. One caveat: on JDK 11 the durable callable runs inline, so the same code will not raise there yet.

weiqingy:
@wenjin272 Thanks for the review! Good catch, and it turned out to be two jobs: the cross-language one you spotted, and java-it, which runs `SkillsIntegrationTest` behind the same gate. Both fixed here.

One limit worth knowing: secrets are withheld from `pull_request` runs on forks, so this restores coverage on push builds and same-repository PRs rather than on fork PRs.

weiqingy:
The `ut-java [ubuntu-latest] [java-11] [python-3.12]` failure looks unrelated to this PR. Maven Central returned `502 Bad Gateway` while resolving `com.fasterxml.jackson.core:jackson-databind:2.18.2`, so the build stopped during dependency resolution and no test ran. Evidence: [job log](https://github.com/apache/flink-agents/actions/runs/33294432135/job/99211555301).

That version is the existing `jackson.version` pin in the root `pom.xml`, this PR touches no `pom.xml` or dependency file, and the same job passed on the previous commit of this branch.

Re-triggering it once the current run finishes.

wenjin272 (approved):

There are a few conflicts with `main` following the merge of #944. Could you please rebase and resolve them?

weiqingy:
@wenjin272 Thanks for the review. Rebased onto main and pushed.

PR #1011, opened 2026-08-13
[ci][integrations] Enable Elasticsearch vector-store tests

### LINKED ISSUE #999: [Bug] Elasticsearch queryEmbedding applies filters as a post-filter, returning fewer than k documents

kaiwangleo:
Fixes #999

## Root cause

Filtered `queryEmbedding` requests applied the combined filter through `SearchRequest.postFilter(...)` after Elasticsearch had already selected the unfiltered KNN top-k results. This could return fewer than `k` matching documents.

## Fix

Apply the combined filter directly to the Elasticsearch KNN clause through `KnnSearch.Builder.filter(...)`.

Existing unified `filters`, raw `filter_query` handling, AND composition, `k`, `num_candidates`, and exception behavior are preserved.

## Tests

- WSL Maven compilation for the Elasticsearch module passed.
- Real Elasticsearch 8.19 integration tests passed: 6 tests, 0 failures, 0 errors, 0 skipped.
- Regression coverage includes:
  - unified `filters`;
  - raw `filter_query`;
  - combined filters;
  - matching documents outside the unfiltered top-k.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,23 +238,36 @@ jobs:
           distribution: 'temurin'
       - name: Install flink-agents Java
         run: bash tools/build.sh -j
-      - name: Start Elasticsearch
-        run: |
-          docker compose -f tools/docker/elasticsearch/docker-compose.yml down -v
-          docker compose -f tools/docker/elasticsearch/docker-compose.yml up -d
-          timeout 180 bash -c 'until curl -fsS http://localhost:9200/_cluster/health; do sleep 5; done'
-      - name: Run Elasticsearch integration tests
-        env:
-          ES_HOST: http://localhost:9200
-        run: |
-          mvn -B --no-transfer-progress -pl integrations/vector-stores/elasticsearch -am -Dspotless.skip=true -Drat.skip=true -Dtest=ElasticsearchVectorStoreTest -Dsurefire.failIfNoSpecifiedTests=false test
       - name: Install ollama
         run: bash tools/start_ollama_server.sh
       - name: Run Java IT
         env:
           LOG_LEVEL: INFO
-          ES_HOST: http://localhost:9200
         run: tools/ut.sh -j -e -f ${{ matrix.flink-version }}
+
+  elasticsearch_tests:
+    name: Elasticsearch vector-store tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    steps:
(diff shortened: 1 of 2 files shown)


wenjin272:
Thanks for taking this on. I have adjusted the PR title to reflect the current CI-focused scope.

wenjin272 (approved):

Issue #1017, opened 2026-08-16
[Tech Debt] Documentation index page has an empty column, a stale sync link, and unversioned Hugo build instructions


weiqingy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Three deficiencies in the documentation, all small and all under `docs/`.

**1. The index page left column is nearly empty.** The page splits into two columns through the `columns` shortcode. The left "Try Flink Agents" column holds a single sentence, while the right "Get Help with Flink Agents" column carries the Slack and Community Sync sections. The shortcode renders both as flex items, so they stretch to the height of the taller one and the closing "developed under the umbrella of Apache Flink" paragraph sits below both. The left column ends about 260px above that paragraph, leaving a large blank area in the middle of the landing page. The page is also thin as a navigation entry point, offering one link to a reader arriving at the docs root.

Proposal: add two short link lists to the left column, one for the three quickstart agents and one for the main development and operations pages. The Apache Flink documentation index does the same, listing its tutorials as a bulleted list with a one line description each.

**2. The community sync link points at the 2025 thread.** The index links to [discussion 66](https://github.com/apache/flink-agents/discussions/66), which covers the 2025 syncs and now opens with "This page is outdated. Please find information about the latest community syncs at: #419". The live schedule, agenda and minutes are in [discussion 419](https://github.com/apache/flink-agents/discussions/419). Both the English and Chinese index pages need the update.

**3. `docs/README.md` does not name the required Hugo version.** It links the generic Hugo install page and says to make sure Hugo is installed. `setup_hugo.sh` pins the extended Hugo 0.110.0, and a newer Hugo release removed the internal template that the pinned `hugo-book` theme calls in `layouts/404.html` and `layouts/partials/docs/html-head.html`. A contributor who installs a current Hugo gets `no such template "_internal/google_analytics_async.html"` with no indication of why. `setup_hugo.sh` also downloads the Linux binary only, which the README does not say.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #1018, opened 2026-08-16
[docs] Improve the documentation index page and build instructions

weiqingy:
Linked issue: #1017

### Purpose of change

Four small documentation fixes, all under `docs/`.

**1. The index page left column is nearly empty.** The `columns` shortcode renders both halves as flex items, so they stretch to the height of the taller one. The left "Try Flink Agents" column held a single sentence against the right column's Slack and Community Sync sections, leaving about 260px of blank space above the closing paragraph.

Two short link lists now fill it, one for the three quickstart agents and one for the main development and operations pages. The columns end within about 15px of each other, and the docs root gets real navigation instead of a single link. The Apache Flink documentation index lists its tutorials the same way.

**2. The community sync link points at the 2025 thread.** [Discussion 66](https://github.com/apache/flink-agents/discussions/66) now opens with "This page is outdated" and redirects to [discussion 419](https://github.com/apache/flink-agents/discussions/419), which carries the current schedule, agenda and minutes. Updated on both the English and Chinese index pages.

**3. `docs/README.md` does not name the required Hugo version.** `setup_hugo.sh` pins extended Hugo 0.110.0, but the README only links the generic install page. A newer Hugo removed an internal template the pinned `hugo-book` theme calls, so a contributor who installs a current Hugo gets `no such template "_internal/google_analytics_async.html"` with no hint why. The README now names the version, that failure, and where to get the binary on platforms the script does not cover.

**4. `setup_hugo.sh` installs an x86_64 binary on every Linux host.** It hard-coded `hugo_extended_0.110.0_Linux-64bit.tar.gz`, the legacy name for the amd64 build, so Linux ARM64 got a binary that cannot run. The script now selects `Linux-64bit` or `linux-arm64` from `uname -m`, and exits naming the architecture when neither matches.

### Tests

Built the site locally with extended Hugo 0.110.0.

- Build clean: 48 pages, no errors or warnings.
- All 26 docs links on the rendered index resolve to a built page, including the nine new ones.
- Compared against a render of the unmodified page, the blank area is gone.
- No `discussions/66` reference remains under `docs/content` or `docs/content.zh`.
- A current Hugo reproduces the `no such template` failure named in the README.
- `setup_hugo.sh` passes `bash -n` and `shellcheck`. `x86_64` selects `Linux-64bit`, `aarch64` selects `linux-arm64`, anything else exits 1. Both artifact URLs return 200 at the sizes the release lists.

The ARM64 binary itself was not executed, as no ARM64 Linux host was available.

### API

No public API change. Documentation only.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [X] `doc-included`

- [X] Yes
- [ ] No

Generated-by: Claude Code 2.1.233 (Claude Opus 5)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,11 +22,15 @@ $ docker run -v $(pwd):/src -p 1313:1313 jakejarvis/hugo-extended:latest server 
 
 #### Local Hugo installation:
 
-Make sure you have installed [Hugo](https://gohugo.io/getting-started/installing/) on your system.
+`setup_hugo.sh` installs Hugo 0.110.0, the version the docs build is pinned to. A newer Hugo release removed an internal template the pinned `hugo-book` theme calls, so building with a current Hugo fails with `no such template "_internal/google_analytics_async.html"`.
+
+On Linux, install it with:
 
 ```sh
 $ ./setup_hugo.sh
 ```
+
+`setup_hugo.sh` downloads the Linux binary only. On other platforms, install the extended Hugo 0.110.0 from the [Hugo releases page](https://github.com/gohugoio/hugo/releases/tag/v0.110.0) yourself.
 
 Then build the docs from source:
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,7 +30,7 @@ On Linux, install it with:
 $ ./setup_hugo.sh
 ```
 
-`setup_hugo.sh` downloads the Linux binary only. On other platforms, install the extended Hugo 0.110.0 from the [Hugo releases page](https://github.com/gohugoio/hugo/releases/tag/v0.110.0) yourself.
+`setup_hugo.sh` downloads a Linux binary, x86_64 or ARM64. On other platforms, install the extended Hugo 0.110.0 from the [Hugo releases page](https://github.com/gohugoio/hugo/releases/tag/v0.110.0) yourself.
 
 Then build the docs from source:
 
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -30,6 +30,18 @@ under the License.
 If you’re interested in playing around with Flink Agents, check out our [get started documentation]({{< ref "docs/get-started/overview" >}}). 
 It provides a step by step introduction on how to install Flink Agents and build agents with it.
 
+### Quickstart
+
+* [Workflow Agent]({{< ref "docs/get-started/quickstart/workflow_agent" >}}) - Define agent logic as an explicit event-driven workflow
+* [ReAct Agent]({{< ref "docs/get-started/quickstart/react_agent" >}}) - Let the model plan and call tools in a reason-act loop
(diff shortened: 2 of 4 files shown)


docs/README.md
```diff
@@ -22,12 +22,16 @@ $ docker run -v $(pwd):/src -p 1313:1313 jakejarvis/hugo-extended:latest server
 
 #### Local Hugo installation:
 
-Make sure you have installed [Hugo](https://gohugo.io/getting-started/installing/) on your system.
+`setup_hugo.sh` installs Hugo 0.110.0, the version the docs build is pinned to. A newer Hugo release removed an internal template the pinned `hugo-book` theme calls, so building with a current Hugo fails with `no such template "_internal/google_analytics_async.html"`.
+
+On Linux, install it with:
```
wenjin272, line 27: Could we narrow this to Linux x86_64? `setup_hugo.sh` currently hard-codes the `Linux-64bit` artifact, while Hugo provides a separate `linux-arm64` package. As written, Linux ARM64 users would install an incompatible binary. Alternatively, the script could select the artifact based on `uname -m`.

docs/content/_index.md
```diff
@@ -30,6 +30,18 @@ under the License.
 If you’re interested in playing around with Flink Agents, check out our [get started documentation]({{< ref "docs/get-started/overview" >}}). 
 It provides a step by step introduction on how to install Flink Agents and build agents with it.
 
+### Quickstart
+
+* [Workflow Agent]({{< ref "docs/get-started/quickstart/workflow_agent" >}}) - Define agent logic as an explicit event-driven workflow
+* [ReAct Agent]({{< ref "docs/get-started/quickstart/react_agent" >}}) - Let the model plan and call tools in a reason-act loop
+* [YAML Agent]({{< ref "docs/get-started/quickstart/yaml_agent" >}}) - Declare an agent without writing code
```
wenjin272: Could we avoid saying “without writing code” here? The linked quickstart explains that action and tool implementations still need to live in Python or Java, while YAML only describes how those pieces are wired together. “Configure an agent declaratively in YAML” may be more accurate.

wenjin272 (commented):
Thanks for improving the documentation! I left two comments about the accuracy of the new instructions and descriptions.

docs/README.md
weiqingy: Good catch. Went with the `uname -m` route since it turned out to be about six lines, so the script now picks `Linux-64bit` or `linux-arm64` and fails with the architecture name rather than quietly installing the wrong binary. README note updated to match.

docs/content/_index.md
weiqingy: Agreed, that's more accurate. Taken your wording.

wenjin272 (approved):

Issue #1010, opened 2026-08-13
[Bug] Recovered Durable ActionState for keys owned by other subtasks is never pruned


GreatEugenius:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

When the Kafka or Fluss Durable ActionState backend is enabled, restoring a job from a checkpoint can leave duplicated recovery state permanently retained in each subtask's in-memory cache.

During recovery:

1. Recovery markers are stored in `UnionListState`, so every restored subtask receives all markers.
2. `KafkaActionStateStore.rebuildState()` merges the markers and scans every referenced Kafka partition.
3. `FlussActionStateStore.rebuildState()` similarly merges the markers and scans every referenced Fluss bucket.
4. As a result, every subtask's `actionStates` map contains the complete recovery tail, including states for keys owned by other subtasks.

For example, with parallelism 2:

```text
After restore:

subtask 0 cache: {A, B}
subtask 1 cache: {A, B}

Key ownership:

subtask 0 owns A
subtask 1 owns B
```

When a subsequent checkpoint completes, snapshotLastCompletedSequenceNumbers() scans the current subtask's keyed state only. Therefore, notifyCheckpointComplete() invokes:

```text
subtask 0: pruneState(A, ...)
subtask 1: pruneState(B, ...)
```

The resulting caches are:
```text
subtask 0 cache: {B}
subtask 1 cache: {A}
```

The entries belonging to other subtasks are never passed to pruneState() and remain reachable until the current operator attempt is closed or fails again.

### How to reproduce

1. Enable the Kafka or Fluss Durable ActionState backend.
2. Set operator parallelism to 2.
3. Choose keys A and B such that they are assigned to different subtasks.
4. Complete checkpoint C0.
5. After C0, write ActionState records for both A and B.
6. Trigger failover before the next checkpoint completes.
7. Restore from C0.
8. Inspect the `actionStates` map in both subtasks.

   Both maps contain A and B:

   ```text
   subtask 0: {A, B}
   subtask 1: {A, B}
   ```
9. Allow the replayed actions for A and B to complete.
10. Complete checkpoint C1.
11. Inspect both caches again.

The current implementation leaves:
```text
subtask 0: {B}
subtask 1: {A}
```

### Version and environment

Apache Flink Agents 0.2+

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

da-daken:
[...]

PR #1024, opened 2026-08-17
[Bug] Recovered Durable ActionState for keys owned by other subtasks is never pruned

da-daken:
Linked issue: https://github.com/apache/flink-agents/issues/1010

### Purpose of change

I've confirmed the root cause: `rebuildState` replays all recovery markers (UnionListState broadcasts every subtask's marker) into each subtask's `actionStates` cache, while `notifyCheckpointComplete` only prunes keys present in the current subtask's own keyed state — so keys owned by other subtasks are never pruned and stay resident for the whole operator attempt.

**Adopted fix**: keep the bulk replay, but skip records not owned by the current subtask during rebuild, so foreign keys never enter the cache.
#### Changes
- **`ActionStateStore`**: add default method `setOwnershipFilter(Predicate<String>)` (`null` means no filter, safe for in-memory/test backends).
- **`KafkaActionStateStore` / `FlussActionStateStore`**: hold the predicate; in `rebuildState`'s replay loop, skip records whose business key is not owned by the current subtask, reusing the existing `OperatorStateManager.isKeyOwnedByCurrentSubtask` (Flink key-group semantics).
- **`DurableExecutionManager.handleRecovery`**: compute the current subtask's `KeyGroupRange` and set it as the filter before `rebuildState`.
- **`ActionExecutionOperator.initializeState`**: compute `maxParallelism` + `KeyGroupRange` (already available here) and pass them into `handleRecovery`.

### Tests

- **Unit — `ActionStateUtilTest`**: backend-agnostic coverage of `isKeyRetained` (owned key retained, foreign key dropped, `null` filter keeps all, unparseable key retained). Both stores route through this function, so this is the shared filtering contract.
- **Unit — `KafkaActionStateStoreTest`** (MockConsumer): seed records for keys A/B, set the filter to "owner of A only", assert the cache keeps only A; plus no-filter-keeps-all and unparseable-key-retained cases.
- **Integration — `FlussActionStateStoreIT.testRebuildStateFiltersForeignKeys`** (real Fluss cluster via `FlussClusterExtension`): write A and B, capture a recovery marker, then rebuild into a fresh store with the filter "owner of A only" and assert A is recovered while B is filtered out. This is the only test that drives Fluss's real log-scan → deserialize → filter → cache replay path.

### API

no API

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included` 

- [x] Yes
- [ ] No

Generated-by: Cursor 2.4.22 (Claude Opus 4.8), WorkBuddy 5.3.11 (hy3)

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
@@ -23,16 +23,22 @@ import com.fasterxml.jackson.databind.json.JsonMapper;
 import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.plan.actions.Action;
 import org.apache.flink.util.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.UUID;
+import java.util.function.Predicate;
 
 /** Utility class for action state related operations. */
 public class ActionStateUtil {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ActionStateUtil.class);
 
     private static final JsonMapper MAPPER =
             JsonMapper.builder()
@@ -60,6 +66,25 @@ public class ActionStateUtil {
         String[] parts = key.split(KEY_SEPARATOR);
         Preconditions.checkArgument(parts.length == 4, "Invalid key format.");
         return List.of(parts);
+    }
+
+    /**
+     * Returns {@code true} if the composite {@code stateKey}'s business key should be retained in a
+     * subtask's in-memory cache under the given ownership filter. A {@code null} filter retains
+     * every key (the default for in-memory and test backends). If the key cannot be parsed, it is
+     * retained as a fail-safe: prefer keeping a valid key over dropping it on a parse error.
+     */
+    public static boolean isKeyRetained(
+            @Nullable Predicate<String> ownershipFilter, String stateKey) {
+        if (ownershipFilter == null) {
+            return true;
+        }
+        try {
+            return ownershipFilter.test(parseKey(stateKey).get(0));
+        } catch (Exception e) {
+            LOG.warn("Failed to parse state key for ownership filtering: {}", stateKey, e);
+            return true;
+        }
     }
 
     private static String generateUUIDForEvent(Event event) throws IOException {
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
@@ -49,11 +49,19 @@ public class ActionStateUtil {
                     .build();
     private static final String KEY_SEPARATOR = "_";
 
-    // Composite key layout: keyGroup_businessKey_seqNum_eventUUID_actionUUID.
+    // Composite key layout: keyGroup_seqNum_eventUUID_actionUUID_businessKey.
+    //
+    // Every fixed field before the business key (key-group, seq-num, and the two UUIDs) is
+    // guaranteed to be free of KEY_SEPARATOR, and the business key — the only caller-supplied,
+    // variable-length field — is placed LAST. Parsing therefore splits with a fixed limit so the
+    // final segment keeps the business key intact even when it contains the separator, e.g.
+    // "tenant_user". No escaping is required and the segment count is always exact.
     private static final int KEY_GROUP_SEGMENT = 0;
-    private static final int BUSINESS_KEY_SEGMENT = 1;
-    private static final int SEQ_NUM_SEGMENT = 2;
-    private static final int KEY_SEGMENT_COUNT = 5;
+    private static final int SEQ_NUM_SEGMENT = 1;
+    private static final int EVENT_UUID_SEGMENT = 2;
+    private static final int ACTION_UUID_SEGMENT = 3;
+    private static final int BUSINESS_KEY_SEGMENT = 4;
+    static final int KEY_SEGMENT_COUNT = 5;
 
     public static String generateKey(
             @Nonnull Object key,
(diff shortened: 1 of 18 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
```diff
@@ -62,6 +68,25 @@ public static List<String> parseKey(String key) {
...
+     * every key (the default for in-memory and test backends). If the key cannot be parsed, it is
+     * retained as a fail-safe: prefer keeping a valid key over dropping it on a parse error.
+     */
+    public static boolean isKeyRetained(
+            @Nullable Predicate<String> ownershipFilter, String stateKey) {
+        if (ownershipFilter == null) {
+            return true;
+        }
+        try {
+            return ownershipFilter.test(parseKey(stateKey).get(0));
```
GreatEugenius: The ownership check hashes the String parsed from the durable state key, but Flink assigned keyed-state ownership using the original typed key. KeyGroupRangeAssignment hashes different object types differently; for example, with max parallelism 128, Long(1) maps to key-group 86 while String("1") maps to 54. The true owner can therefore discard its recovered ActionState while another subtask retains it, allowing replay to execute an already completed action again. Please persist the key-group computed from the original typed key in the WAL record and compare that value with currentSubtaskKeyGroupRange, instead of reconstructing ownership from the string form.

GreatEugenius (commented):
Hi @da-daken, thank you for the PR. I left one comment, which I believe is important.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
da-daken: Good catch! Since the string key recovered during rebuild carries no type
information, we can't reconstruct the correct key-group from it (e.g. both
`Long(1)` and `String("1")` toString to `"1"`, but hash to different key-groups).
The fix is to store the key-group directly at write time, when the typed key is
still available, so recovery can read it back without relying on the string form.

There are two options for where to store the key-group:

| Dimension | A (embed in key) | C (store in ActionState value) |
|---|---|---|
| **Correctness** | Correct. Key-group computed once from typed key at write time. | Correct. |
| **Format change** | Key format (4 → 5 segments) | Value format (new field) |
| **Upgrade behavior** | Old 4-segment keys are deterministically dropped + warn log. At most one re-execution within the retention window. | Old values without keyGroup field must be dropped, or fall back to the buggy string-hash — which preserves the original bug. |
| **Upgrade cost** | One-time: old keys evicted after retention window. | Same one-time cost. |
| **Self-describing** | Yes. The key alone tells you which key-group it belongs to. | No. Must read the value to determine ownership. |
| **Identity vs. payload** | Key-group is identity information (like business key, seqNum), belongs in the key. | Key-group is stored as payload, mixing identity metadata with business data. |
| **Perf (Fluss rebuild)** | Filter by key before deserializing value. | Must deserialize all values to read keyGroup. |

Both A and C have the same upgrade cost — old records lack the key-group in
either location and must be dropped. The difference is semantic: key-group is
derived from the typed key at write time, just like the business key and
sequence number. It belongs in the key, not in the value payload.

**I prefer A** because it keeps identity information in the key where it
belongs, avoids mixing identity metadata with business payload, and in Fluss
avoids deserializing values for non-owned keys during rebuild.

GreatEugenius: Thanks for the detailed analysis. Persisting the key-group at write time makes sense.

I am concerned about dropping all legacy 4-segment records during upgrade. A missing key-group does not have to mean either dropping the record or falling back to the incorrect string hash. We could treat it as UNKNOWN ownership:

Records with a key-group: filter normally using the current subtask’s KeyGroupRange.

Legacy records without a key-group: temporarily retain them in every subtask.

This preserves the old memory amplification only for the legacy recovery tail during the first upgraded attempt, while avoiding the loss of durable state. Once a new checkpoint marker advances past those records, they will no longer be included in the next recovery tail.

Dropping legacy records may re-execute every affected action or durable call in the recovery tail, potentially repeating external side effects. For durable execution, I think the bounded one-time memory overhead is preferable to weakening recovery correctness.

da-daken: I think it can be kept. To avoid re‑running, the subsequent GET path will also need to be compatible with the 4‑segment key. I'm not sure if adding a fallback logic in the code is acceptable.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
```diff
@@ -58,10 +83,110 @@ public static String generateKey(
...
+     * state across a key-group upgrade at the cost of a bounded, one-time memory amplification for
+     * the legacy recovery tail, which ages out once a new checkpoint marker advances past those
+     * records. Lookups still find such records via {@link #legacyKeyOf}. A 5-segment key whose
+     * key-group segment fails to parse is likewise retained as a fail-safe.
+     */
+    public static boolean isKeyRetained(@Nullable IntPredicate ownershipFilter, String stateKey) {
+        if (ownershipFilter == null) {
+            return true;
+        }
+        String[] parts = stateKey.split(KEY_SEPARATOR);
```
wenjin272: Could we avoid using the number of underscore-separated segments to parse the state key? `businessKey` is inserted without escaping, so a key such as `tenant_user` produces more than five segments and is treated as UNKNOWN, causing every subtask to retain it and leaving the orphan-state leak unfixed. Since the project is still beta and previous ActionState data does not need to be preserved, I suggest defining a single new, unambiguous key format and removing the legacy compatibility logic entirely. For example, the business key could be length-prefixed, or the fixed fields could be parsed from both ends.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
```diff
@@ -105,12 +106,20 @@ public class FlussActionStateStore implements ActionStateStore {
     /** In-memory cache for O(1) state lookups; rebuilt from Fluss log on recovery. */
     private final Map<String, ActionState> actionStates;
 
+    // When set, only records whose key-group is accepted by this predicate are kept in the
+    // in-memory cache during rebuildState; null means retain all keys (default).
+    private IntPredicate ownershipFilter;
+
+    // The operator's maximum parallelism, used to compute key-groups consistently with Flink.
+    private int maxParallelism;
```
wenjin272: `maxParallelism` remains `0` when the public `FlussActionStateStore(config)` constructor is used, and the Kafka constructor has the same issue. Any direct `put()` or `get()` then fails in `generateKey()` because it requires a positive value. I reproduced this by running `FlussActionStateStoreIT` explicitly: 10 of its 11 tests fail with `maxParallelism must be positive but was 0`. Since API compatibility is not required during beta, could we make `maxParallelism` a required constructor argument and final, rather than relying on a later mutable setter?

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStoreIntegrationTest.java
```diff
@@ -206,6 +206,44 @@ void testRebuildStateWithRecoveryMarkers() throws Exception {
...
     }
 
+    /**
+     * Reproduces the orphan-state leak fix: after recovery, a subtask must keep only the keys it
+     * owns and drop keys owned by other subtasks. Here "A" is owned and "B" is foreign, so the
+     * rebuilt cache must contain "A" but not "B".
+     */
+    @Test
+    @SuppressWarnings("unchecked")
+    void testRebuildStateFiltersForeignKeys() throws Exception {
```
wenjin272, line 217: This regression test is not executed by the current CI. The class name ends with `IT`, while the repository runs ordinary `mvn test` with Surefire default discovery patterns, which do not include `*IT`. Could we either rename this class to something ending in `Test`, such as `FlussActionStateStoreIntegrationTest`, or explicitly wire `*IT` tests into Surefire/Failsafe and CI? Once enabled, the `maxParallelism=0` failure above will also need to be fixed.

wenjin272 (commented):
Thanks for working on this fix, @da-daken. I left three comments on the state-key format, max-parallelism initialization, and Fluss integration-test coverage.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
da-daken: Good catch! I have set the new stateKey to: keyGroup_seqNum_eventUUID_actionUUID_businessKey. The first four parts are in fixed formats and will not be affected by the underscore (_) delimiter.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStore.java
da-daken: Thank you for pointing that out. The fact that FlussActionStateStoreIT didn't run successfully caused me to miss this point. It has been fixed in the latest commit.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/FlussActionStateStoreIntegrationTest.java
da-daken: Thanks! renamed.

wenjin272 (approved):

Issue #1014, opened 2026-08-14
[Bug] Ollama tool conversion throws NPE when the tool schema has no "required" key


Zhuoxi2000:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

`OllamaChatModelConnection.convertToOllamaTools` assumes the tool schema always has a `required` list:

final List<String> required = (List<String>) schema.get("required");
...
.required(required.contains(paramName))

But `required` is optional in JSON Schema, and `SchemaUtils.generateSchema` only emits it when at least one parameter is required.

That means a Java `@Tool` whose parameters are all optional produces a valid schema without `required`, and the Ollama connector throws an NPE while building the request.

Expected: all-optional tools are converted normally, with each property marked `required(false)`.

Actual: `required` is `null`, so `required.contains(...)` fails before the request reaches Ollama. Tools with at least one required parameter are unaffected.

The fix should be small: treat a missing `required` field as an empty list in `convertToOllamaTools`, and add a regression test for a schema with no `required` key.

I think hardening the consumer is preferable to changing `SchemaUtils`, since omitting an empty `required` list is valid JSON Schema, but happy to follow whichever direction maintainers prefer.

### How to reproduce

1. Define a Java tool with at least one optional parameter:

   public class DemoTools {
   @Tool(description = "List recent orders")
   public static String listOrders(
   @ToolParam(name = "limit", required = false) Integer limit) {
   return "[]";
   }
   }

An unannotated parameter reproduces the same issue, since it is also omitted from the schema’s `required` list.

2. Register the tool with an Ollama chat model and send any request that includes it.

3. `convertToOllamaTools` fails with:

   java.lang.NullPointerException: Cannot invoke "java.util.List.contains(Object)" because "required" is null

A zero-parameter tool does not trigger this because there are no properties to iterate over. The failing case is a tool with one or more parameters where none are required.

### Version and environment

Current main (3070ee2). Java 11+, any OS; independent of the Ollama server version — the
NPE is thrown client-side during request construction.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wenjin272:
[...]

PR #1026, opened 2026-08-18
[integration][ollama] Tolerate tool schemas without a 'required' key

Zhuoxi2000:
Linked issue: #1014

### Purpose of change

`OllamaChatModelConnection.convertToOllamaTools` assumed the tool schema always carries a `required` list, but `required` is optional in JSON Schema and `SchemaUtils.generateSchema` only emits it when at least one parameter is required. A Java `@Tool` whose parameters are all optional therefore produced a valid schema that crashed the Ollama connector with an NPE before the request reached the server. The same shape can come from external tools (e.g. an MCP tool with only optional parameters), so this hardens the consumer rather than changing `SchemaUtils`: a missing `required` list is treated as empty, and all-optional tools convert normally with every property marked `required(false)`.

### Tests

New `OllamaChatModelConnectionTest` (the module previously had no connection test): a schema without a `required` key converts with every property optional, and a schema with a `required` key still marks the listed parameters required. Both are pure unit tests of the conversion — no Ollama server needed. `convertToOllamaTools` was narrowed from private to package-private for test access.

### API

None.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included` 

- [x] Yes
- [ ] No

Generated-by: Claude Fable 5

### DIFF
--- a/integrations/chat-models/ollama/src/main/java/org/apache/flink/agents/integrations/chatmodels/ollama/OllamaChatModelConnection.java
+++ b/integrations/chat-models/ollama/src/main/java/org/apache/flink/agents/integrations/chatmodels/ollama/OllamaChatModelConnection.java
@@ -106,8 +106,9 @@ public class OllamaChatModelConnection extends BaseChatModelConnection {
      * @return List of Ollama compatible tool specifications
      * @throws RuntimeException if schema parsing or conversion fails
      */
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):

Issue #1015, opened 2026-08-14
[Bug] Tool parameter schema maps long/float (and all other non-int/double numerics) to "object"


Zhuoxi2000:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Description

`SchemaUtils.getParamSchema` maps Java parameter types to JSON Schema with a small if/else chain. It currently handles `String`, `int`/`Integer`, `double`/`Double`, and `boolean`/`Boolean`; everything else falls back to `"object"`.

As a result, common numeric types such as `long`, `Long`, `float`, `Float`, `short`, and `byte` are advertised to the model as objects. For example, a tool like `getOrder(long orderId)` generates `"type": "object"` for `orderId`.

This can cause models to produce the wrong argument shape or avoid the tool call entirely, while the failure looks like a model/tool-calling issue rather than a schema-generation bug.

Expected:

* `long` / `Long` / `short` / `Short` / `byte` / `Byte` → `"integer"`
* `float` / `Float` → `"number"`

Suggested fix: extend the existing numeric mappings in `getParamSchema` and add `SchemaUtilsTest` coverage for these types. Arrays, collections, and POJOs can continue falling back to `"object"` for now.

### How to reproduce

1. Define a tool with a numeric type that is not currently recognized:

   @Tool(description = "Fetch an order")
   public static String getOrder(@ToolParam(name = "orderId") long orderId) { ... }

2. Call `SchemaUtils.generateSchema(method, null)`, or inspect the schema sent by any chat model connector.

3. The generated schema contains:

   "orderId": {"type": "object"}

Expected:

```
"orderId": {"type": "integer"}
```

### Version and environment

Current main (3070ee2). Pure schema-generation issue in `plan`'s SchemaUtils; affects every
chat-model integration that forwards the generated tool schema.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wenjin272:
Thanks for reporting this. I confirmed the issue on the latest `main`: `long`/`Long`, `float`/`Float`, `short`/`Short`, and `byte`/`Byte` are all currently mapped to `"object"`. Extending `SchemaUtils.getParamSchema` with the appropriate integer/number mappings and adding primitive/wrapper regression tests looks like the right fix.

PR #1027, opened 2026-08-18
[plan] Map long/float and other numeric tool parameters to proper JSON Schema types

Zhuoxi2000:
Linked issue: #1015

### Purpose of change

`SchemaUtils.getParamSchema` maps Java parameter types to JSON Schema with a small if/else chain that only handled `String`, `int`/`Integer`, `double`/`Double`, and `boolean`/`Boolean` — everything else fell back to `"object"`. Common numeric types such as `long`, `Long`, `float`, `Float`, `short`, and `byte` were therefore advertised to the model as objects (e.g. `getOrder(long orderId)` generated `"type": "object"` for `orderId`), causing models to produce the wrong argument shape or avoid the tool call entirely, with the failure looking like a model/tool-calling issue rather than a schema-generation bug.

This extends the numeric mappings: `long`/`Long`/`short`/`Short`/`byte`/`Byte` → `"integer"`, `float`/`Float` → `"number"`. Arrays, collections, and POJOs keep the `"object"` fallback for now, per the issue.

### Tests

New `SchemaUtilsTest#testGenerateSchemaWithWideNumericTypes` covering all six types (primitive and boxed), asserting `"integer"` for the integral types and `"number"` for the floating-point types. Existing SchemaUtils tests pass unchanged.

### API

None. Generated schemas for previously-"object" numeric parameters change to the correct JSON Schema types, which is the fix.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included` 

- [x] Yes
- [ ] No

Generated-by Claude Fable 5

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/tools/SchemaUtils.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/tools/SchemaUtils.java
@@ -102,9 +102,19 @@ public class SchemaUtils {
 
         if (paramType == String.class) {
             paramSchema.put("type", "string");
-        } else if (paramType == int.class || paramType == Integer.class) {
+        } else if (paramType == int.class
+                || paramType == Integer.class
+                || paramType == long.class
+                || paramType == Long.class
+                || paramType == short.class
+                || paramType == Short.class
+                || paramType == byte.class
+                || paramType == Byte.class) {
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):

Issue #1013, opened 2026-08-14
[Feature] Use descriptive thread names for Java and Python async executors


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Flink Agents currently relies on the default thread names for both Java and Python asynchronous execution:

- The JDK 21 Java implementation creates its executor with `Executors.newFixedThreadPool(numAsyncThreads)`, so worker threads have generic names such as `pool-N-thread-M`.
- The Python implementation creates a `ThreadPoolExecutor` without `thread_name_prefix`, so worker threads have generic names such as `ThreadPoolExecutor-N_M`.

These names make it difficult to identify Flink Agents async workers in TaskManager thread dumps, profiler output, and monitoring tools. A TaskManager process may contain many unrelated executor pools, and the current names do not indicate whether a thread belongs to the Flink Agents Java async executor, the Python async executor, or another component. Idle workers are especially difficult to attribute because their stacks contain only generic `ThreadPoolExecutor` frames.

Both executors should use descriptive, collision-resistant thread names. The naming convention should:

- clearly identify the thread as a Flink Agents async worker;
- distinguish Java and Python executor threads;
- distinguish different executor instances and workers within the same process; and
- remain useful in thread dumps and profiler output without changing execution behavior.

For example, names could follow conventions similar to:

- Java: `flink-agents-java-async-<pool-id>-thread-<worker-id>`
- Python: `flink-agents-python-async-<pool-id>_<worker-id>`

The exact convention can be decided during implementation. Java can provide a custom `ThreadFactory` when constructing the fixed thread pool, while Python can provide a unique `thread_name_prefix` to `ThreadPoolExecutor`.

Tests should verify that tasks submitted to each executor run on threads with the expected descriptive prefix and that names remain distinct across multiple executor instances.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

Zhuoxi2000:
I'd like to take this one. My plan is to name the async worker threads in both runtimes so they’re easier to identify in thread dumps and profiler output.

For Java, I’ll add a `ThreadFactory` and use names like `flink-agents-async-<n>`, keeping the current daemon behavior and pool sizing unchanged. For Python, I’ll set `thread_name_prefix="flink-agents-async"` on the `ThreadPoolExecutor`.

Happy to use a different naming scheme if you have a preference.

wenjin272:
[...]

PR #1028, opened 2026-08-18
[runtime][python] Use descriptive thread names for the async executors

Zhuoxi2000:
Linked issue: #1013

### Purpose of change

Both async executors relied on default thread names (`pool-N-thread-M` from the JDK 21 `Executors.newFixedThreadPool`, `ThreadPoolExecutor-N_M` from Python), which make Flink Agents async workers hard to attribute in TaskManager thread dumps, profiler output, and monitoring tools — especially idle workers, whose stacks contain only generic frames.

Following the naming convention suggested in the issue:

- **Java**: a new `AsyncExecutorThreadFactory` names workers `flink-agents-java-async-<pool-id>-thread-<worker-id>`, with a process-unique pool id so multiple executor instances stay distinguishable. The factory lives in the JDK-agnostic `src/main/java` source set (the JDK 21 `ContinuationActionExecutor` consumes it), so it is unit-testable on every JDK. Only the name changes — daemon status and priority follow the default factory behavior.
- **Python**: `create_async_thread_pool` passes a process-unique `thread_name_prefix`, yielding `flink-agents-python-async-<pool-id>_<worker-id>` via `ThreadPoolExecutor`'s standard prefix mechanism.

### Tests

- Java: `AsyncExecutorThreadFactoryTest` — tasks submitted to an executor built with the factory run on threads matching `flink-agents-java-async-\d+-thread-\d+`; names are distinct across workers and across factory instances; daemon status and priority are unchanged from the default factory.
- Python: `test_async_thread_pool.py` — submitted tasks observe names matching `flink-agents-python-async-\d+_\d+`, and two pools produce distinct prefixes.

### API

None; thread naming only, no execution-behavior change.

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included` 

- [x] Yes
- [ ] No

Generated-by:  Claude Fable 5


runtime/src/main/java/org/apache/flink/agents/runtime/async/AsyncExecutorThreadFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final AtomicInteger POOL_ID = new AtomicInteger();
+
+    private final String namePrefix;
+    private final AtomicInteger workerId = new AtomicInteger();
+
+    public AsyncExecutorThreadFactory() {
+        this.namePrefix = "flink-agents-java-async-" + POOL_ID.incrementAndGet() + "-thread-";
+    }
+
+    @Override
```
wenjin272, line 41: The PR states that only the thread name changes, but `new Thread(...)` inherits the daemon status and priority of the calling thread, unlike the previous default thread factory, which normalizes them to non-daemon and `NORM_PRIORITY`. Could we delegate thread creation to `Executors.defaultThreadFactory()` to preserve the existing execution semantics?

runtime/src/main/java/org/apache/flink/agents/runtime/async/AsyncExecutorThreadFactory.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * <p>Default executor names such as {@code pool-N-thread-M} make Flink Agents async workers hard to
+ * attribute in TaskManager thread dumps and profiler output, where many unrelated pools coexist.
+ * The pool id is process-unique so multiple executor instances in one TaskManager remain
+ * distinguishable; only the name changes — thread priority and daemon status follow the default
+ * factory behavior.
+ */
+public final class AsyncExecutorThreadFactory implements ThreadFactory {
+
+    private static final AtomicInteger POOL_ID = new AtomicInteger();
```
wenjin272: If we use `Executors.defaultThreadFactory()`, the generated thread name already contains the pool and worker IDs (`pool-N-thread-M`). Could we simply prepend `flink-agents-java-async-` to that name instead of maintaining separate `POOL_ID` and `workerId` counters? The resulting name would be `flink-agents-java-async-pool-N-thread-M`.

Zhuoxi2000: Done  thread names now look like `flink-agents-java-async-pool-N-thread-M`, and the extra counters are gone.

runtime/src/main/java/org/apache/flink/agents/runtime/async/AsyncExecutorThreadFactory.java
Zhuoxi2000: Good catch switched to `Executors.defaultThreadFactory()` so daemon status and priority keep the JDK defaults, and only the name is changed. Added a regression test for that as well.

wenjin272 (approved):

PR #1030, opened 2026-08-18
[api][runtime] Replay newObject memory updates as object creations during durable recovery

purushah:
Linked issue: #1029

### Purpose of change

With durable execution enabled, `MemoryObjectImpl.newObject()` records itself in the `ActionState` as `MemoryUpdate(path, null)` — indistinguishable from a user calling `set(path, null)`. The recovery replay path applies every recorded update via `set()`, so replaying a completed action that created a nested object either throws `IllegalArgumentException: Cannot overwrite object with value` against an object restored from the checkpoint (a **permanent recovery crash loop** — reproduced on a local mini-cluster job with a Kafka action-state store, where the job burned all restart attempts and died), or materializes the object as a `null` value leaf, after which replaying any child write throws `UnsupportedOperationException` (and the memory tree is silently corrupted).

This PR adds an `objectCreation` discriminator to `MemoryUpdate`, records it from `newObject()`, and replays object-creation updates via `newObject(path, true)` in a new shared `MemoryUpdateReplayer` used by the operator's short-term and sensory replay loops.

### Tests

- New `MemoryUpdateReplayerTest` (4 cases): replay into empty state restores a real nested object with children; replay over an already existing object (the crash-loop case) succeeds; `newObject(path, overwrite=true)` over a value leaf replays correctly; a user's genuine `set(path, null)` still replays as a null **value**, not an object.
- Two operator-level recovery tests in `ActionExecutionOperatorTest` (`testReplayReappliesNewObjectMemoryUpdatesIntoEmptyState`, `testReplayReappliesNewObjectMemoryUpdatesOverRestoredState`) run both scenarios end to end through `KeyedOneInputStreamOperatorTestHarness` with an `ActionStateStore`, including checkpoint snapshot/restore.
- `ActionStateSerdeTest` extended with a round-trip assertion for the new field.
- All of the above verified **red before the fix** (exact exceptions above) and **green after**; the affected runtime suites (130 tests, including all of `ActionExecutionOperatorTest` and `ShortTermMemoryTTLIntegrationTest`) pass.
- Additionally verified with a real local Flink job (mini-cluster, checkpointing, Kafka action-state store, injected one-time failure): on current `main` the job crash-loops and dies with `Cannot overwrite object with value: user`; with this patch the identical job recovers and completes with correct memory state. Logs attached to the linked issue.

### API

`MemoryUpdate` gains a three-arg constructor and `isObjectCreation()`; the existing two-arg constructor is kept (source- and binary-compatible). Compatibility notes:

- **Old durable journals read by new code:** the missing `objectCreation` field defaults to `false`, so pre-existing records replay exactly as before.
- **New journals read by old code:** rejected — consistent with `ActionStateSerde`'s documented recovery-compat contract ("the durable-execution journal ... never survives a code or Flink upgrade").
- **Flink checkpoints/savepoints:** untouched (no change to `MemoryItem` or state schemas).
- **Behavior change:** `MemoryUpdate.equals`/`hashCode`/`toString` now include the flag, so a recorded `newObject` update no longer compares equal to `set(path, null)` — which is the semantic distinction this fix introduces.
- Memory-event JSON schemas and the Python surface are unchanged.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Claude Code 2.1.234 (Claude Fable 5)

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/context/MemoryUpdate.java
+++ b/api/src/main/java/org/apache/flink/agents/api/context/MemoryUpdate.java
@@ -32,17 +32,36 @@ public class MemoryUpdate implements Serializable {
 
     private final String path;
     private final Object value;
+    private final boolean objectCreation;
+
+    /**
+     * Creates a new MemoryUpdate instance describing a value write.
+     *
+     * @param path the absolute path of the data in Short-Term Memory.
+     * @param value the new value to set at the specified path.
+     */
+    public MemoryUpdate(String path, Object value) {
+        this(path, value, false);
+    }
 
     /**
      * Creates a new MemoryUpdate instance.
      *
      * @param path the absolute path of the data in Short-Term Memory.
-     * @param value the new value to set at the specified path.
+     * @param value the new value to set at the specified path; always null when {@code
+     *     objectCreation} is true.
+     * @param objectCreation true if this update records the creation of a nested object rather than
+     *     a value write. Absent in records written before this field existed, in which case Jackson
+     *     defaults it to false, preserving their original replay behavior.
      */
     @JsonCreator
-    public MemoryUpdate(@JsonProperty("path") String path, @JsonProperty("value") Object value) {
+    public MemoryUpdate(
+            @JsonProperty("path") String path,
+            @JsonProperty("value") Object value,
+            @JsonProperty("objectCreation") boolean objectCreation) {
         this.path = path;
         this.value = value;
+        this.objectCreation = objectCreation;
     }
 
     /**
@@ -63,21 +82,41 @@ public class MemoryUpdate implements Serializable {
         return value;
     }
 
+    /**
+     * Whether this update records the creation of a nested object (via {@code newObject}) rather
+     * than a value write. Replay must re-create the object instead of setting a null value.
+     *
+     * @return true if this update is a nested-object creation.
+     */
+    public boolean isObjectCreation() {
+        return objectCreation;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
         if (!(o instanceof MemoryUpdate)) return false;
         MemoryUpdate that = (MemoryUpdate) o;
-        return Objects.equals(path, that.path) && Objects.equals(value, that.value);
+        return objectCreation == that.objectCreation
+                && Objects.equals(path, that.path)
+                && Objects.equals(value, that.value);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(path, value);
+        return Objects.hash(path, value, objectCreation);
     }
 
     @Override
     public String toString() {
-        return "MemoryUpdate{" + "path='" + path + '\'' + ", value=" + value + '}';
+        return "MemoryUpdate{"
+                + "path='"
+                + path
+                + '\''
+                + ", value="
+                + value
+                + ", objectCreation="
+                + objectCreation
+                + '}';
     }
 }
--- a/api/src/main/java/org/apache/flink/agents/api/context/MemoryUpdate.java
+++ b/api/src/main/java/org/apache/flink/agents/api/context/MemoryUpdate.java
@@ -59,6 +59,13 @@ public class MemoryUpdate implements Serializable {
             @JsonProperty("path") String path,
             @JsonProperty("value") Object value,
             @JsonProperty("objectCreation") boolean objectCreation) {
+        if (objectCreation && value != null) {
+            throw new IllegalArgumentException(
+                    "An object-creation update cannot carry a value, but got one for path '"
+                            + path
+                            + "': "
+                            + value);
+        }
         this.path = path;
         this.value = value;
         this.objectCreation = objectCreation;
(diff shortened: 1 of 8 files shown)


wenjin272:
Hi @joeyutong, could you help review this pr?

api/src/main/java/org/apache/flink/agents/api/context/MemoryUpdate.java
```diff
@@ -32,17 +32,36 @@ public class MemoryUpdate implements Serializable {
...
+     * @param objectCreation true if this update records the creation of a nested object rather than
+     *     a value write. Absent in records written before this field existed, in which case Jackson
+     *     defaults it to false, preserving their original replay behavior.
+     */
     @JsonCreator
-    public MemoryUpdate(@JsonProperty("path") String path, @JsonProperty("value") Object value) {
+    public MemoryUpdate(
+            @JsonProperty("path") String path,
+            @JsonProperty("value") Object value,
+            @JsonProperty("objectCreation") boolean objectCreation) {
```
joeyutong, line 61: Could we reject objectCreation=true with a non-null value in this constructor? MemoryUpdateReplayer ignores the value for object creation, so accepting this invalid combination can silently hide malformed updates.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerdeTest.java
```diff
@@ -61,11 +61,13 @@ public void testActionStateSerializationDeserialization() throws Exception {
 
         MemoryUpdate sensoryMemoryUpdate = new MemoryUpdate("sm.test.path", "sm test value");
         MemoryUpdate shortTermMemoryUpdate = new MemoryUpdate("stm.test.path", "stm test value");
+        MemoryUpdate objectCreationUpdate = new MemoryUpdate("stm.test.obj", null, true);
```
joeyutong, line 67: Could we add a legacy JSON case with objectCreation omitted and assert that it defaults to false? The current round trip only covers records written by the new serializer.

runtime/src/test/java/org/apache/flink/agents/runtime/memory/MemoryUpdateReplayerTest.java
```diff
@@ -0,0 +1,134 @@
...
+        return updates;
+    }
+
+    @FunctionalInterface
+    private interface ThrowingConsumer<T> {
+        void accept(T t) throws Exception;
+    }
+
+    @Test
+    void testReplayNewObjectWithChildWritesIntoEmptyState() throws Exception {
```
joeyutong, line 60: Could we add a case containing only newObject("empty") and assert that replay preserves an empty nested object rather than a null leaf?

api/src/main/java/org/apache/flink/agents/api/context/MemoryUpdate.java
purushah: Good catch — done in 4edd5f6. The creator now throws `IllegalArgumentException` when `objectCreation=true` carries a non-null value, so a malformed update fails loudly at construction/deserialization time instead of the replayer silently dropping the value.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateSerdeTest.java
purushah: Added in 4edd5f6: `testLegacyRecordWithoutObjectCreationFieldDefaultsToFalse` serializes a state, strips the `objectCreation` field from the JSON (simulating a record written by the old serializer), deserializes through `ActionStateSerde`, and asserts the update comes back as a value write with `isObjectCreation() == false`.

runtime/src/test/java/org/apache/flink/agents/runtime/memory/MemoryUpdateReplayerTest.java
purushah: Added in 4edd5f6: `testReplayLoneNewObjectPreservesEmptyNestedObject` replays an action that only called `newObject("empty")` and asserts the result is an empty nested object (`isNestedObject()` true, no field names, null value) rather than a null leaf.

joeyutong:
Thanks for addressing all three comments. I re-reviewed 4edd5f6 and don’t see any further issues with the change itself. The only remaining item is to rebase and resolve the conflicts with current main. When resolving ActionExecutionOperator, please keep notifyActionReused(actionTask) after both memory replays. Once the conflicts are resolved, this looks good to me.

purushah:
Rebased onto current main (0e36efd). The `ActionExecutionOperator` conflict was resolved exactly as requested — `notifyActionReused(actionTask)` sits after both `MemoryUpdateReplayer.replay(...)` calls (short-term, then sensory). The test-file conflict kept both your new `testActionStateStoreReplayRecordsReusedExecutions` and the two recovery tests from this PR. Verified locally on the rebased tree: spotless clean, 79/79 affected tests green including the full `ActionExecutionOperatorTest` suite (47 tests, trace-replay test included).

PR #1032, opened 2026-08-21
Fix #985: Handle PydanticInvalidForJsonSchema for unserializable BaseModel

### LINKED ISSUE #985: [Bug] An output schema pydantic cannot render raises instead of taking the prompt fallback
> **Decision, 2026-08-17: option 2.** Raise a clear, consistent error rather than falling back, and treat `RowTypeInfo` as a separate documented fallback. See [this comment](https://github.com/apache/flink-agents/issues/985#issuecomment-5312409060).
>
> The two options below are the original framing and are kept for the record.

### Search before asking

### Description

`OllamaChatModelConnection.chat` translates a `BaseModel` output schema by calling `model_json_schema()`. For some field types pydantic cannot produce a JSON Schema and raises `PydanticInvalidForJsonSchema`, so the chat call fails instead of falling back to the prompt-engineering path.

That is inconsistent with how the same connection treats the other schema form it cannot translate natively. A `RowTypeInfo` is skipped silently and the caller keeps the prompt fallback, while a `BaseModel` pydantic cannot render propagates an exception out of `chat`. Both cases are "this connection cannot express the schema natively", so it is not obvious they should behave differently.

There are two defensible directions, and I do not think the choice is clear cut:

1. Treat an unrenderable `BaseModel` like any other untranslatable schema and fall back to the prompt path. This makes the two cases consistent and keeps a schema from breaking a call that would otherwise succeed.
2. Keep raising, on the grounds that an unrenderable schema is a caller mistake and a silent fallback would hide it, and instead make the `RowTypeInfo` case louder.

The choice also affects what "capable" means. The connection reports native support for every model, so the caller has no way to ask in advance whether a particular schema will translate.

Exposure today is limited. No framework path passes an `output_schema` down to a connection, so this is reachable only from a direct call to the 4-arg `chat`.

The OpenAI, Azure OpenAI and Anthropic Python connections hand the class to their provider SDK rather than calling `model_json_schema()` themselves, so they raise from inside the SDK on the same input. Whatever is decided here likely applies to them as well, which is why this is filed against the behavior rather than against one connection.

### How to reproduce

```python
from typing import Callable

from pydantic import BaseModel

from flink_agents.api.agents.types import OutputSchema


class Bad(BaseModel):
    cb: Callable[[int], int]


# raises pydantic.errors.PydanticInvalidForJsonSchema
connection.chat(
    messages, model="qwen3", output_schema=OutputSchema(output_schema=Bad)
)
```

`Bad.model_json_schema()` on its own raises the same error, which is the call the connection makes.

### Version and environment

Flink Agents `main` (0.3-SNAPSHOT), pydantic 2.11.4, Python 3.10 to 3.12. Not platform specific.

hulincup:
## Description

Fixes #985

When a BaseModel contains fields that cannot be serialized to JSON schema (e.g., `Callable` fields), the `to_strict_json_schema()` function raises `PydanticInvalidForJsonSchema`. Currently, this exception propagates up and breaks the chat call.

The issue description notes that `RowTypeInfo` schemas gracefully fall back to prompt engineering when the connection can't translate them natively, but `BaseModel` schemas that can't be serialized raise an exception instead. This inconsistency should be fixed.

## Changes

This PR catches `PydanticInvalidForJsonSchema` in the three connection types that use `to_strict_json_schema()` or `transform_schema()`:

1. **OpenAI** (`openai_chat_model.py`):
   - Catches exception in `_native_response_format()`
   - Falls back to prompt engineering (returns `None`)
   - Logs a warning message
   - Added `logger` import

2. **Azure OpenAI** (`azure_openai_chat_model.py`):
   - Catches exception in `_native_response_format()`
   - Falls back to prompt engineering (returns `None`)
   - Logs a warning message
   - Updated docstring to document the fallback behavior

3. **Anthropic** (`anthropic_chat_model.py`):
   - Catches exception in `_native_output_config()`
   - Falls back to prompt engineering (returns `None`)
   - Logs a warning message
   - Updated docstring to document the fallback behavior

## Testing

Added comprehensive test file `test_unserializable_output_schema.py` with three test cases:

- `test_openai_handles_unserializable_schema()`: Verifies OpenAI falls back to prompt when schema cannot be serialized
- `test_azure_handles_unserializable_schema()`: Verifies Azure falls back to prompt when schema cannot be serialized
- `test_anthropic_handles_unserializable_schema()`: Verifies Anthropic falls back to prompt when schema cannot be serialized

Each test uses a `BadModel` with a `Callable` field that cannot be converted to JSON schema, and verifies that:
1. No exception is raised
2. The native structured output format is not included in the request (fallback to prompt engineering)

## Verification

The fix ensures that:
- Connections handle unserializable `BaseModel` schemas gracefully
- Behavior is consistent with `RowTypeInfo` fallback
- Users receive a warning log when fallback occurs
- Existing functionality for serializable schemas is unchanged

## Related Issues

- Fixes #985

### DIFF
--- a/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
+++ b/python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
@@ -15,13 +15,14 @@
 #  See the License for the specific language governing permissions and
 # limitations under the License.
 #################################################################################
+import logging
 import uuid
 from typing import Any, Dict, List, Sequence
 
 from anthropic import Anthropic, transform_schema
 from anthropic._types import NOT_GIVEN
 from anthropic.types import MessageParam, TextBlockParam, ToolParam
-from pydantic import BaseModel, Field, PrivateAttr
+from pydantic import BaseModel, Field, PrivateAttr, PydanticInvalidForJsonSchema
 from typing_extensions import override
 
 from flink_agents.api.agents.types import OutputSchema
@@ -31,6 +32,8 @@ from flink_agents.api.chat_models.chat_model import (
     BaseChatModelSetup,
 )
 from flink_agents.api.tools.tool import Tool, ToolMetadata
+
+logger = logging.getLogger(__name__)
 
 
 def to_anthropic_tool(
@@ -202,7 +205,8 @@ def _native_output_config(output_schema: Any) -> Dict[str, Any] | None:
 
     Returns ``None`` (leaving the request unchanged) unless the schema is a
     ``BaseModel`` subclass. A ``RowTypeInfo`` schema is skipped so it keeps the
-    prompt-engineering fallback.
+    prompt-engineering fallback. If the BaseModel cannot be converted to JSON schema
+    (e.g., contains Callable fields), returns ``None`` to fall back to prompt engineering.
 
     Anthropic's format object carries only the schema and its type, so it shares no
     shape with the providers that nest the schema under a named, strict
@@ -215,7 +219,15 @@ def _native_output_config(output_schema: Any) -> Dict[str, Any] | None:
     )
     if not (isinstance(model, type) and issubclass(model, BaseModel)):
         return None
-    return {"format": {"type": "json_schema", "schema": transform_schema(model)}}
+    try:
+        return {"format": {"type": "json_schema", "schema": transform_schema(model)}}
+    except PydanticInvalidForJsonSchema as e:
+        logger.warning(
+            "Pydantic model %s cannot be converted to JSON schema, falling back to prompt engineering: %s",
+            model.__name__,
+            str(e),
+        )
+        return None
 
 
 class AnthropicChatModelConnection(BaseChatModelConnection):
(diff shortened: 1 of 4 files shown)


wenjin272:
Thanks for working on this. We have decided on option 2 in #985: an unrenderable schema should raise a clear error, because the prompt fallback also needs to render the schema and therefore cannot handle this case. #1046 implements that approach across the relevant call paths. To avoid duplicate and conflicting fixes, could we close this PR in favor of #1046?

hulincup:
Hi @wenjin272, thanks for the heads up.

You're right — #1046 implements option 2 as decided on #985, which is the correct direction. My PR took the fallback approach (option 1), which the discussion on the issue explicitly moved away from on 2026-08-17, so it shouldn't land. I also missed @weiqingy's "I will open a PR" note when I opened this, so the overlap is on me.

Closing #1032 in favor of #1046. Thanks @weiqingy for the thorough implementation across the Python and Java paths.

PR #1033, opened 2026-08-21
[runtime] Resolve relative file: URLs when loading classpath skills (#966)

### LINKED ISSUE #966: [Bug] ClasspathSkillRepository fails to load skills from relative file URLs
### Search before asking

- [x] I searched the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

`ClasspathSkillRepository` fails to load skills when a classloader exposes an application JAR through a relative local URL, for example:

```text
jar:file:../../flink/usrlib/job.jar!/diagnosis-skills
```

This URL form can be produced when a Flink deployment adds user-code JARs relative to the TaskManager working directory.

`SkillMaterializer.copyJarEntries()` extracts the inner URL (`file:../../flink/usrlib/job.jar`), converts it to a `URI`, and passes it to `new File(URI)`. The relative `file:` URI is opaque rather than hierarchical, so `File(URI)` throws `IllegalArgumentException: URI is not hierarchical`. The exception is wrapped as an invalid JAR URL and skill initialization fails.

A representative stack trace is:

```text
java.lang.IllegalStateException: Failed to load skills from classpath:{resource=diagnosis-skills}
  at org.apache.flink.agents.runtime.skill.SkillManager.loadAll(SkillManager.java:191)
Caused by: java.io.IOException: Invalid JAR URL:
  jar:file:../../flink/usrlib/job.jar!/diagnosis-skills
  at org.apache.flink.agents.runtime.skill.repository.SkillMaterializer.copyJarEntries(SkillMaterializer.java:212)
  at org.apache.flink.agents.runtime.skill.repository.SkillMaterializer.extractClasspathFromJars(SkillMaterializer.java:183)
  at org.apache.flink.agents.runtime.skill.repository.ClasspathSkillRepository.materialize(ClasspathSkillRepository.java:132)
Caused by: java.lang.IllegalArgumentException: URI is not hierarchical
  at java.io.File.<init>(File.java:420)
  at org.apache.flink.agents.runtime.skill.repository.SkillMaterializer.copyJarEntries(SkillMaterializer.java:207)
```

The same assumption also exists in the `URLClassLoader#getURLs()` fallback scan in `ClasspathSkillRepository.findAllMatches()`, so a relative `file:` JAR URL can be skipped or fail before the resource is materialized. Direct relative `file:` directory resources are affected by the corresponding `Paths.get(url.toURI())` conversion as well.

Expected behavior: local relative `file:` URLs should be resolved relative to the process working directory and classpath skills should load normally. Absolute hierarchical `file:` URLs should keep their current behavior, while unsupported non-file URLs should be rejected or skipped explicitly.

One possible implementation direction is to centralize local URL conversion:

- accept only the `file` protocol;
- use `new File(uri)` for hierarchical file URIs;
- for an opaque relative file URI, use its decoded scheme-specific part as the relative path;
- reuse the conversion in direct resource materialization, JAR extraction, and fallback classpath scanning.

### How to reproduce

1. Create a JAR containing `diagnosis-skills/example/SKILL.md`.
2. Add the JAR to a `URLClassLoader` using a relative URL such as `file:../../flink/usrlib/job.jar`, or make `getResources("diagnosis-skills")` return `jar:file:../../flink/usrlib/job.jar!/diagnosis-skills`.
3. Construct `ClasspathSkillRepository` for the `diagnosis-skills` resource, directly or through `SkillManager`.
4. Observe that repository initialization fails with `IllegalArgumentException: URI is not hierarchical`.

The failure can also be reproduced in a unit test by creating a temporary skill JAR, relativizing its path against the current working directory, and constructing the URL with `new URL("file:" + relativePath)`.

### Version and environment

- Apache Flink Agents: current `main` (the affected conversions are present in `SkillMaterializer` and `ClasspathSkillRepository`)
- Apache Flink: 1.20-based distribution
- Deployment mode: application/user-code JAR loaded from Flink's `usrlib` directory through a relative classpath URL
- Java: 11

Olawoyin007:

Fixes #966.

A Flink deployment can add user-code JARs relative to the TaskManager working directory, so a classloader may expose a skill resource through a **relative** `file:` URL such as `file:../../flink/usrlib/job.jar` (or `jar:file:../../flink/usrlib/job.jar!/diagnosis-skills`). Such a URL parses to an *opaque* `URI`, which `new File(URI)` / `Paths.get(URI)` reject with `IllegalArgumentException: URI is not hierarchical`, so `ClasspathSkillRepository` initialization fails and skills never load.

## Brief change log

- Add `LocalUrls.toLocalFile(URL)`, a single place that converts a local `file:` URL to a `File`:
  - absolute hierarchical `file:` URLs keep their existing `new File(uri)` behavior;
  - opaque relative `file:` URLs are resolved against the process working directory via the decoded scheme-specific part;
  - non-`file` URLs are rejected explicitly.
- Reuse it in the three affected sites so their behavior can't drift:
  - `ClasspathSkillRepository.materializeFileUrl` (direct file/dir/zip resource),
  - `SkillMaterializer.copyJarEntries` (JAR entry extraction),
  - `ClasspathSkillRepository.findAllMatches` (`URLClassLoader` fallback scan).
- Remove now-unused imports.

## Verifying this change

- New `LocalUrlsTest` covers absolute resolution, relative/opaque resolution against the working directory, and non-`file` rejection.
- New `ClasspathSkillRepositoryTest#loadFromRelativeJarUrl` loads skills end-to-end through a relative jar URL (exercises both the fallback scan and jar extraction). Verified that this test fails against the pre-fix code with the reported `Classpath resource not found` / `URI is not hierarchical` error.
- Full `runtime` module test suite passes (668 tests, 0 failures).

- Dependencies (does it add or upgrade a dependency): **no**
- The public API: **no**
- The serializers: **no**
- The runtime per-record code paths: **no**
- Anything that affects deployment or recovery: **no**
- The threading model: **no**

## Documentation

- Does this pull request introduce a new feature? **no**

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/LocalUrls.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/LocalUrls.java
@@ -47,24 +47,28 @@ final class LocalUrls {
      * <p>Absolute hierarchical {@code file:} URLs keep their existing {@code new File(uri)}
      * behavior. Non-{@code file} URLs are rejected explicitly.
      *
-     * @throws IOException if the URL is not a {@code file:} URL, or is a malformed URI.
+     * @throws IOException if the URL is not a {@code file:} URL, is a malformed URI, or cannot be
+     *     represented as a local {@link File}.
      */
     static File toLocalFile(URL url) throws IOException {
         if (!"file".equals(url.getProtocol())) {
             throw new IOException("Not a local file URL: " + url);
         }
-        final URI uri;
         try {
-            uri = url.toURI();
-        } catch (URISyntaxException e) {
+            URI uri = url.toURI();
+            if (uri.isOpaque()) {
+                // Relative file: URL (e.g. file:../../flink/usrlib/job.jar). new File(URI)
+                // rejects opaque URIs, so resolve the decoded scheme-specific part relative to
+                // the working directory, which is exactly how a relative File is interpreted.
+                return new File(uri.getSchemeSpecificPart());
+            }
+            return new File(uri);
+        } catch (URISyntaxException | IllegalArgumentException e) {
+            // URISyntaxException: url.toURI() rejected the URL. IllegalArgumentException:
+            // new File(URI) cannot represent this file: URL as a local path, e.g. one carrying an
+            // authority component like file://host/share/job.jar. Callers only catch IOException,
+            // so wrap both so such a URL is skipped rather than escaping unchecked.
             throw new IOException("Malformed file URL: " + url, e);
         }
-        if (uri.isOpaque()) {
-            // Relative file: URL (e.g. file:../../flink/usrlib/job.jar). new File(URI) rejects
-            // opaque URIs, so resolve the decoded scheme-specific part relative to the working
-            // directory, which is exactly how a relative File is interpreted.
-            return new File(uri.getSchemeSpecificPart());
-        }
-        return new File(uri);
     }
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/LocalUrls.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/LocalUrls.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.agents.runtime.skill.repository;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+/**
+ * Centralizes the conversion of local {@code file:} URLs to {@link File} for the classpath skill
+ * loading path (direct resource materialization, JAR extraction, and the {@code URLClassLoader}
+ * fallback scan). Keeping this in one place ensures the three call sites agree on how relative
+ * {@code file:} URLs are resolved.
+ */
+final class LocalUrls {
+
+    private LocalUrls() {}
+
+    /**
+     * Resolve a local {@code file:} URL to a {@link File}.
+     *
+     * <p>A Flink deployment may add user-code JARs relative to the TaskManager working directory,
+     * producing relative {@code file:} URLs such as {@code file:../../flink/usrlib/job.jar}. Such a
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/LocalUrls.java
```diff
@@ -0,0 +1,70 @@
...
+        } catch (URISyntaxException e) {
+            throw new IOException("Malformed file URL: " + url, e);
+        }
+        if (uri.isOpaque()) {
+            // Relative file: URL (e.g. file:../../flink/usrlib/job.jar). new File(URI) rejects
+            // opaque URIs, so resolve the decoded scheme-specific part relative to the working
+            // directory, which is exactly how a relative File is interpreted.
+            return new File(uri.getSchemeSpecificPart());
+        }
+        return new File(uri);
```
wenjin272: Could we also wrap the `IllegalArgumentException` thrown by `new File(uri)`? For example, `file://host/share/job.jar` is hierarchical, but `new File(uri)` throws `IllegalArgumentException: URI has an authority component`. Since the callers only catch `IOException`, this escapes the fallback scan instead of being skipped and also changes the previous `copyJarEntries()` exception contract. Shall we keep the conversion inside the `try` and wrap both `URISyntaxException` and `IllegalArgumentException` as `IOException`?

runtime/src/test/java/org/apache/flink/agents/runtime/skill/ClasspathSkillRepositoryTest.java
```diff
@@ -89,6 +90,29 @@ void loadFromJarResource(@TempDir Path tempDir) throws IOException {
                         .collect(Collectors.toList()));
     }
 
+    @Test
+    void loadFromRelativeJarUrl(@TempDir Path tempDir) throws IOException {
```
wenjin272, line 96: This test covers the `URLClassLoader` fallback scan and JAR extraction paths, but the PR also changes direct resource materialization. Could we add regression cases where `getResources()` directly returns a relative `jar:file:` URL and a relative `file:` directory URL? That would exercise all three affected call paths described in the PR.

wenjin272 (commented):
Thanks for working on this fix! The overall approach looks good. I left two inline comments about an exception-handling edge case and regression-test coverage.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/LocalUrls.java
Olawoyin007: Good catch, fixed in 05b4a463. `new File(uri)` is now inside the `try`, and the `catch` handles both `URISyntaxException` and `IllegalArgumentException`, wrapping them as `IOException`. So a hierarchical-but-non-representable URL like `file://host/share/job.jar` now surfaces as `IOException` and is skipped by the fallback scan, restoring the previous `copyJarEntries` contract instead of escaping unchecked. Added a `LocalUrlsTest` case asserting the authority-component URL comes back as `IOException`.

runtime/src/test/java/org/apache/flink/agents/runtime/skill/ClasspathSkillRepositoryTest.java
Olawoyin007: Agreed, added in 05b4a463. Two new regression tests feed the URLs through `getResources()` directly (via a non-`URLClassLoader` loader, so the fallback scan is bypassed and each path is exercised in isolation): `loadFromDirectRelativeJarUrl` returns a relative `jar:file:` URL to drive `SkillMaterializer.copyJarEntries`, and `loadFromDirectRelativeDirectoryUrl` returns a relative `file:` directory URL to drive `ClasspathSkillRepository.materializeFileUrl`. Together with the existing fallback-scan test, all three affected call paths are now covered.

wenjin272 (approved):

Issue #1036, opened 2026-08-22
[Bug] [integrations][anthropic] temperature defaults to 0.1 and is always sent, returning HTTP 400 on Claude 4.7 and later


weiqingy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The Anthropic integration gives `temperature` a default of `0.1` and puts it on every request, with no check against the model. Claude 4.7 and later reject a non-default sampling parameter with an HTTP 400, so selecting one of those models turns every chat call into a provider error. Omitting the parameter, or sending the provider default, is still accepted, so the fix is to stop sending it rather than to change its value.

Both languages have the defect and neither gates on the model.

Java sets `DEFAULT_TEMPERATURE = 0.1d` in `AnthropicChatModelSetup` and puts the value on the parameter map unconditionally, and `AnthropicChatModelConnection` copies it onto the request builder whenever it is a `Number`. Python does the same through `DEFAULT_TEMPERATURE = 0.1` and the `model_kwargs` property, which always includes `temperature`.

The affected names are the 4.7 generation onward: `claude-opus-4-7`, `claude-opus-4-8`, `claude-opus-5`, `claude-sonnet-5`, `claude-fable-5`, `claude-mythos-5` and `claude-mythos-preview`. Sampling parameters are still accepted on `claude-opus-4-6` and `claude-sonnet-4-6`.

That boundary is worth stating explicitly, because it is not the boundary either existing list in this connection encodes. Native structured output starts at the 4.5 generation, prefill rejection starts at 4.6, and sampling rejection starts at 4.7. Claude 4.6 rejects a prefill while still accepting a temperature, so `PREFILL_UNSUPPORTED_MODELS` cannot be reused here and a third list is needed. The comment already sitting on `PREFILL_UNSUPPORTED_MODELS` anticipates exactly this case, noting that sharing one list would hold only until a model moved one boundary without moving the other.

An untouched configuration does not hit this. The default model is `claude-sonnet-4-20250514`, which accepts sampling parameters, so reaching the failure requires selecting a newer model. This is the same shape as the prefill defect in #1009.

Not in scope here: `top_p` and `top_k` reach the request through `additional_kwargs` in Java and `**kwargs` in Python, and the same models reject them. Those are values a caller passes explicitly rather than a default the integration injects, so they are a separate concern.

### How to reproduce

[...]

PR #1037, opened 2026-08-22
[integrations][anthropic] Stop sending rejected sampling parameters

weiqingy:
Linked issue: #1036

### Purpose of change

Claude Opus 4.7+ and the Sonnet, Fable and Mythos 5 lines return HTTP 400 for a non-default `temperature`, `top_p` or `top_k`. The Anthropic integration defaults `temperature` to `0.1` and sent it on every request without checking the model, so picking one of those models turned every chat call into a provider error. Both languages had the defect.

All three parameters are now dropped on models that reject them, in both languages. Dropped rather than clamped: omitting the parameter is accepted everywhere, while substituting the provider default would quietly change sampling behaviour. Each drop is logged once per model and parameter.

Three paths could reach the wire, and all three are gated:

| Path | Was |
|---|---|
| Java top-level `temperature` | sent unconditionally |
| Java `additional_kwargs` → `top_p` / `top_k` | sent unconditionally |
| Java `additional_kwargs` → `temperature` | no `case` label, so forwarded as a raw body property |

Python gates the merged kwargs immediately above its single `create` call.

The model list is deliberately separate from the prefill and structured-output lists in this connection: the three boundaries do not coincide. Structured output starts at 4.5, prefill rejection at 4.6, sampling rejection at 4.7, so Claude 4.6 rejects a prefill while still accepting a temperature. A test pins that.

**One behaviour change.** A caller setting `temperature` both at the top level and in `additional_kwargs` now has the `additional_kwargs` value win. Previously both were emitted under the same key, producing `{"temperature":0.1,"temperature":0.5}`, which is undefined. Typed and raw routes serialize identically for a number, so accepting models see no change on the wire. The dropped-parameter warning names the value that was actually withheld: the two routes are resolved before the gate rather than reporting whichever was read first.

**One bug fix beyond the 400.** The warn-once bookkeeping was keyed on the model alone, so a second parameter dropped for an already-warned model was discarded silently. It is now keyed on model and parameter.

### Tests

Java 90 in `AnthropicChatModelConnectionTest`, 92 in the module. Python 84 passed, 2 skipped. Spotless, Ruff and RAT clean.

Coverage was driven by mutation testing. Every mutant in scope dies in both languages: the guard removed per parameter, each parameter's routing removed, and the warn key reverted to model-only.

Worth knowing for future edits: an unrecognised key is forwarded as a raw body property, so asserting the typed accessor is empty does **not** prove a parameter was left off. The drop tests assert the raw body too.

Which value the warning names is pinned by passing both temperatures as a `Number` that records whether it was rendered, since the message is assembled with `%s`. `stop_sequences` is covered too, the last arm of the `additional_kwargs` switch that had no test.

Not covered: the `LOG.warn` call itself, and a rewrite that keeps the dedup helper but rebuilds the message at the call site and logs unconditionally. Both need a logging binding in a module that has none.

### API

No public API change. `temperature`, `top_p` and `top_k` keep their names, types and defaults; only whether they reach the provider depends on the model. No change on models that accept sampling parameters, including the default `claude-sonnet-4-20250514`.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

`chat_models.md`: both `temperature` rows note the parameter is not sent on Claude 4.7+; the Java `additional_kwargs` row adds `temperature` and states the precedence. The Python `additional_kwargs` row was removed because that parameter does not exist in the Python setup.

- [x] Yes
- [ ] No

Generated-by: Claude Code 2.1.234, 2.1.251 (Claude Opus 5)

### DIFF
--- a/integrations/chat-models/anthropic/src/main/java/org/apache/flink/agents/integrations/chatmodels/anthropic/AnthropicChatModelConnection.java
+++ b/integrations/chat-models/anthropic/src/main/java/org/apache/flink/agents/integrations/chatmodels/anthropic/AnthropicChatModelConnection.java
@@ -42,6 +42,8 @@ import org.apache.flink.agents.api.chat.model.BaseChatModelConnection;
 import org.apache.flink.agents.api.resource.ResourceContext;
 import org.apache.flink.agents.api.resource.ResourceDescriptor;
 import org.apache.flink.agents.api.tools.ToolMetadata;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.time.Duration;
 import java.util.ArrayList;
@@ -51,6 +53,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 /**
@@ -80,6 +83,8 @@ import java.util.stream.Collectors;
  * }</pre>
  */
 public class AnthropicChatModelConnection extends BaseChatModelConnection {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AnthropicChatModelConnection.class);
 
     private static final TypeReference<Map<String, Object>> MAP_TYPE = new TypeReference<>() {};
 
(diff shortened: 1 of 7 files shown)


integrations/chat-models/anthropic/src/main/java/org/apache/flink/agents/integrations/chatmodels/anthropic/AnthropicChatModelConnection.java
```diff
@@ -366,15 +471,16 @@ BuiltRequest buildRequest(
         }
 
         Object temperature = modelParams.remove("temperature");
-        if (temperature instanceof Number) {
+        if (temperature instanceof Number
+                && sendSamplingParam(modelName, "temperature", temperature)) {
```
wenjin272, line 502: Thanks for fixing this! When both the top-level `temperature` and `additional_kwargs.temperature` are set, the latter should take precedence. However, the warning is emitted for the top-level value first and may report the overridden value. Could we resolve the effective temperature before filtering/logging and add a test for this case?

weiqingy: Yes, good catch. The warning is owed once per model and parameter, so the top-level value claimed it first and named a temperature that was about to be overridden anyway.

Fixed in `daa13b34`. `additional_kwargs` is read before the gate now, and `effectiveTemperature` picks the winner, so the warning names the value the request would otherwise have carried. The `temperature` case in the switch stays, just empty: dropping the label would send the value through the default branch and onto the body raw.

The test was the awkward part. On a rejecting model neither value is sent, so the only place the choice shows up is the warning text, and the module has no logging binding. What I settled on was passing both temperatures as a `Number` that records whether `toString()` ran, since the warning is built with `%s`. The value that got rendered is the value that got named. Putting the old order back fails that test and nothing else.

There's also a parameterized test on `effectiveTemperature` for the cases where the top-level value still wins: no `temperature` key, or one holding a non-number.

wenjin272 (approved):
LGTM

weiqingy:
@wenjin272 Thanks for the review. Rebased onto main and pushed.

Issue #1038, opened 2026-08-22
[Tech Debt] Update the stale Anthropic default model claude-sonnet-4-20250514 to claude-sonnet-4-5


weiqingy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The Anthropic chat model defaults to `claude-sonnet-4-20250514` when the caller omits `model`. That model is deprecated, so a user who takes the default silently gets a deprecated model today and will hit a hard failure whenever it retires.

The stale value lives in both language runtimes and is restated in several more places, so a user meets it regardless of which API they use:

- Java, `AnthropicChatModelSetup.DEFAULT_MODEL` (`integrations/chat-models/anthropic/src/main/java/org/apache/flink/agents/integrations/chatmodels/anthropic/AnthropicChatModelSetup.java:78`), plus the Javadoc parameter list (`:40`) and the Javadoc usage example (`:63`)
- Python, `DEFAULT_ANTHROPIC_MODEL` (`python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py:453`), plus the docstring that repeats it (`:467`)
- Docs, the `AnthropicChatModelSetup` parameter tables and code examples for both languages (`docs/content/docs/development/chat_models.md:290`, `:305`, `:342`, `:367`)

Anthropic is also the only chat model integration whose default is a dated snapshot. Every other one uses an undated alias: `gpt-4o-mini`, `gpt-4o`, `gemini-3.1-pro-preview`, `qwen-plus`. The dated form is what let this default go stale unnoticed, since an alias tracks the current snapshot of its model version on its own.

Two related problems sit next to it in the same files. The Python docstring repeats the literal value rather than naming the constant, unlike the OpenAI (`openai_chat_model.py:318`) and Tongyi (`tongyi_chat_model.py:232`) integrations, which is why it drifted alongside the constant. And the docs list of available models (`chat_models.md:385-388`) recommends four models of which three are no longer current, including `claude-3-7-sonnet-20250219`, which is retired, and `claude-opus-4-1-20250805`, which is past its retirement date.

Proposed fix: move the default to `claude-sonnet-4-5` in both runtimes, update every restatement, make the Python docstring name its constant, and prune the docs model list. `claude-sonnet-4-5` is the same tier as the model it replaces, so callers who rely on the default see no change in cost or class of model, and it is already the name the connection tests use as their structured-output-capable example.

[...]

PR #1039, opened 2026-08-22
[integrations][anthropic] Update the stale default model to claude-sonnet-4-6

weiqingy:
Linked issue: #1038

### Purpose of change

The Anthropic chat model defaults to `claude-sonnet-4-20250514` when the caller omits `model`. That model is deprecated, so a user who takes the default silently gets a deprecated model today and will hit a hard failure whenever it retires. The same value is restated in a Javadoc parameter list, a Javadoc example, a Python docstring, two docs parameter tables and two docs examples, so it had drifted in nine places rather than one.

This change moves the default to `claude-sonnet-4-5` in both language runtimes and updates every restatement. The undated alias matches every other chat model default in the repo (`gpt-4o-mini`, `gpt-4o`, `gemini-3.1-pro-preview`, `qwen-plus`); Anthropic was the only one pinned to a dated snapshot, which is why it went stale. Unlike a dated snapshot, an alias cannot decay into a retired id again.

`claude-sonnet-4-5` keeps the same tier as the model it replaces, so a caller relying on the default sees no change in cost or class of model. It is also the name the connection tests introduced in #965 already use as their structured-output-capable example, on the grounds that it is the generation which is both capable and still accepts a JSON prefill.

Two smaller things ride along, both of the same defect class. The Python docstring now names `DEFAULT_ANTHROPIC_MODEL` instead of repeating its value, matching the OpenAI and Tongyi integrations, so it cannot drift from the constant a second time. The docs list of available models is pruned to three current entries, because three of its four entries named models that are deprecated or already retired, one of them past its retirement date.

Callers that pass `model` explicitly are unaffected. The constant is consulted only when the resource descriptor omits `model`.

### Tests

`AnthropicChatModelSetupTest.testGetParametersDefaults` asserts the documented defaults and is updated to the new value. The existing Python test asserts against the imported constant rather than a hard-coded string, so it continues to pin the default without change.

Two test constants in the suites added by #965 named the old model as the shipped default. They keep their value, because those tests need a model the provider does not document native structured-output support for and that is still what it is, but their comments no longer describe it as the default.

No new tests. The default is already pinned on both sides, and an assertion that a private constant equals its own literal would not catch a bug the compiler misses.

### API

No API signature change. Only the value of the default model changes.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [X] `doc-included`

- [X] Yes
- [ ] No

Generated-by: Claude Code 2.1.234 (Claude Opus 5)

### DIFF
--- a/docs/content/docs/development/chat_models.md
+++ b/docs/content/docs/development/chat_models.md
@@ -287,7 +287,7 @@ Anthropic provides cloud-based chat models featuring the Claude family, known fo
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
 | `connection` | str | Required | Reference to connection method name |
-| `model` | str | `"claude-sonnet-4-20250514"` | Name of the chat model to use |
+| `model` | str | `"claude-sonnet-4-5"` | Name of the chat model to use |
 | `prompt` | Prompt \| str | None | Prompt template or reference to prompt resource |
 | `tools` | List[str] | None | List of tool names available to the model |
 | `max_tokens` | int | `1024` | Maximum number of tokens to generate |
@@ -302,7 +302,7 @@ Anthropic provides cloud-based chat models featuring the Claude family, known fo
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
 | `connection` | String | Required | Reference to connection method name |
-| `model` | String | `"claude-sonnet-4-20250514"` | Name of the chat model to use |
+| `model` | String | `"claude-sonnet-4-5"` | Name of the chat model to use |
 | `prompt` | Prompt \| String | None | Prompt template or reference to prompt resource |
 | `tools` | List<String> | None | List of tool names available to the model |
 | `max_tokens` | long | `1024` | Maximum number of tokens to generate |
@@ -339,7 +339,7 @@ class MyAgent(Agent):
         return ResourceDescriptor(
             clazz=ResourceName.ChatModel.ANTHROPIC_SETUP,
             connection="anthropic_connection",
-            model="claude-sonnet-4-20250514",
+            model="claude-sonnet-4-5",
             max_tokens=2048,
             temperature=0.7
         )
@@ -364,7 +364,7 @@ public class MyAgent extends Agent {
     public static ResourceDescriptor anthropicChatModel() {
         return ResourceDescriptor.Builder.newBuilder(ResourceName.ChatModel.ANTHROPIC_SETUP)
                 .addInitialArgument("connection", "anthropicConnection")
-                .addInitialArgument("model", "claude-sonnet-4-20250514")
+                .addInitialArgument("model", "claude-sonnet-4-5")
                 .addInitialArgument("temperature", 0.7d)
                 .addInitialArgument("max_tokens", 2048)
                 .build();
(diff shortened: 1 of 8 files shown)


wenjin272:
Thanks for updating the stale default and keeping Java, Python, and the docs aligned. 

One concern: `claude-sonnet-4-5` only tracks snapshots within the 4.5 version and does not advance to 4.6 or 5; Anthropic states that every model ID has its own retirement schedule, and the current 4.5 snapshot may retire as early as September 29, 2026 ([model versioning](https://platform.claude.com/docs/en/about-claude/models/model-ids-and-versions), [deprecations](https://platform.claude.com/docs/en/about-claude/model-deprecations)). Could we use the recommended replacement `claude-sonnet-4-6`, or require an explicit model, so the stale-default problem is not merely postponed? The Python default test should also assert the chosen literal, since comparing with the same imported constant does not validate the public default. 

Also, [`ChatModelIntegrationAgent`](https://github.com/apache/flink-agents/blob/11e763a47589850c3187501156308a014ff2e8b0/e2e-test/flink-agents-end-to-end-tests-integration/src/test/java/org/apache/flink/agents/integration/test/ChatModelIntegrationAgent.java#L123-L127) still hard-codes the retired `claude-sonnet-4-20250514`, so the Anthropic E2E path will fail when run with a real API key.

weiqingy:
@wenjin272 Thanks for the review. All three are fixed in cb4453ab.

The default is now `claude-sonnet-4-6`. Your point about the alias not advancing past 4.5 is right, and the docs back it a step further: `claude-sonnet-4-6` is listed as the recommended replacement for `claude-sonnet-4-20250514`, which is the exact value this default used to carry.

The PR first picked 4.5 because of the JSON prefill, which 4.6 rejects. That turned out not to matter. Both runtimes already gate the prefill on the model (`AnthropicChatModelConnection.java:425`, `anthropic_chat_model.py:376`), so it is suppressed rather than sent. 4.6 also matches the native structured-output allowlist exactly, where the 4.5 alias only matched the prefix rule, so callers on the default now get the native path instead of the prompt-engineering fallback.

I stopped at 4.6 rather than going to the 5 line because `temperature` is always sent, and 4.7 and later reject a non-default sampling value with a 400. #1037 is adding that gating and would lift the ceiling. Does moving to 4.6 now and revisiting after #1037 seem like the right order, or would you rather hold and move once?

On requiring an explicit model instead, I lean toward keeping a default, since every other chat model integration here ships one (`gpt-4o-mini`, `gpt-4o`, `gemini-3.1-pro-preview`, `qwen-plus`). Is there something about Anthropic that you think warrants treating it differently?

You were right about the Python test. It compared against the constant it imported, so it checked the wiring but not the shipped value. It asserts the literal now, matching the Java test. `ChatModelIntegrationAgent` is fixed too, and it was worse than stale: `claude-sonnet-4-20250514` retired on June 15, 2026, so that path could not have run against a live key at all. I also refreshed the docs list of popular models, which named only 4.5-generation entries.

wenjin272 (approved):

Moving to claude-sonnet-4-6 now makes sense to me. I would also prefer to keep 4.6 as the default even after #1037 is merged.

PR #1040, opened 2026-08-23
[integrations][plan] Handle finish_reason in OpenAI-family connections

weiqingy:
Linked issue: #936

### Purpose of change

`finish_reason` was never read in either language, so a truncated completion came back as ordinary partial content. Under structured output that content went straight to a JSON parser, so the caller got an error naming the parser instead of the truncation. Worse, when the truncated content happened to still be valid JSON, nothing raised at all and the caller silently received wrong data.

Cases 1 and 3 of #936 shipped in #952 and #989. This is the last one.

The change has three parts:

1. Both languages record the provider's reason as `finish_reason` in message metadata. Unknown values are stored verbatim, and an absent reason writes no key.
2. The shared chat-response path checks it once per response, before that response is dispatched as text, structured output, or tool calls. `length` and `content_filter` fail with the cause named. `stop`, `tool_calls`, unknown values and an absent key proceed as before.
3. Python's outbound converter had to be fixed first. It merged all of `extra_args` into outbound messages, and since the chat action replays the assistant response on the next tool-call turn, `model_name`, `promptTokens`, `completionTokens` and `structured_output` were reaching the provider as message fields. Each role now sends only the fields OpenAI defines, matching Java.

Four things worth knowing when reading the diff:

- Java reads through `_finishReason().asKnown()` rather than `finishReason()`. The plain accessor throws `OpenAIInvalidDataException` when the member is absent or null, and `VLLMChatModelConnection` extends `OpenAICompletionsConnection`, so OpenAI-compatible servers that omit it would start failing calls that work today.
- The check runs on the common path, after the model execution is reported and its token metrics are recorded, and before the response is dispatched. A truncated response consumed its full token budget, so that metric is recorded before the rejection. Reporting the model execution as failed instead was considered and rejected: the call itself did not fail, and none of the existing problem categories describes a complete response that is unusable.
- Java's structured-output result now carries inbound metadata forward instead of starting from an empty map. It previously dropped `finish_reason` exactly when an output schema was in play, while Python kept it.
- Three visible behavior changes. A Python caller can no longer forward arbitrary `extra_args` to the provider, which Java never allowed and nothing in the repo relies on. A truncated or filtered response no longer reaches the caller under any error handling strategy: `ChatResponseEvent` was the only route that delivered it, and `FAIL` and `RETRY` raise while `IGNORE` drops the record without sending an event. And a deterministic `length` is now retried on every call rather than only on structured ones, a cost that repeats for each candidate in a model route.

Out of scope: the Responses API reports `status` and `incomplete_details.reason`, a different vocabulary. Providers that record no reason are unaffected and pick the behavior up if they later write the same key.

### Tests

Verified per commit against the modules each one touches: the `plan` module at 301 Java tests with 0 failures, and the full non-integration Python suite at 942 passed with 13 skipped. Spotless and Ruff are clean. CI covers the full build and the version matrix.

Coverage was driven by mutation testing, and every mutant in scope is killed in both languages: swapping in the throwing accessor, moving either capture inside the token-usage guard, dropping either terminal reason, reinstating the blanket merge, neutering the gate call, and widening the gate to reject an accepted reason. Acceptance and rejection are pinned by disjoint tests, so neither can mask the other.

Two test properties are load-bearing and easy to lose in a later refactor. The throwing tests use valid JSON, because truncated JSON raises on its own and a malformed fixture would pass with no gate at all. And one test asserts that a rejected response produces no parser execution report, which is what pins the gate ahead of the structured-output wrapper rather than inside it.

Both languages were also driven through identical inputs to confirm they agree, including that neither sends the metadata keys to a provider.

### API

No public API change. `finish_reason` is a metadata key rather than a field, and no new exception type or problem category is introduced. Failures reuse `ValueError` in Python and `IllegalStateException` in Java, both already used for this class of failure in their packages. Since `json.JSONDecodeError` subclasses `ValueError`, Python callers catching today's symptom keep working and simply get a clearer message.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Claude Code 2.1.234 (Claude Opus 5)

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
@@ -580,39 +580,35 @@ public class ChatModelAction {
     }
 
     /**
-     * Rejects a response the provider did not finish emitting. A finish reason reporting the
-     * content as cut off by the token budget or withheld by content filtering raises {@link
+     * Rejects a response the provider did not finish emitting. Evaluated once per chat response,
+     * before it is dispatched as text, structured output, or tool calls. A finish reason reporting
+     * the content as cut off by the token budget or withheld by content filtering raises {@link
      * IllegalStateException}; any other reason, and an absent one, are accepted.
      */
-    private static void rejectIncompleteResponse(ChatMessage response) {
+    static void rejectIncompleteResponse(ChatMessage response) {
         Object finishReason = response.getExtraArgs().get(FINISH_REASON);
         if (TRUNCATED_FINISH_REASON.equals(finishReason)) {
             throw new IllegalStateException(
                     String.format(
                             "ChatModel response is truncated (finish_reason='%s'): it"
                                     + " exhausted the completion token budget before the model"
-                                    + " finished, so the content is incomplete and cannot yield"
-                                    + " structured output. Raise the model's max output tokens,"
-                                    + " or ask for a smaller output.",
+                                    + " finished, so the content is incomplete. Raise the"
+                                    + " model's max output tokens, or ask for a smaller output.",
                             finishReason));
         }
         if (CONTENT_FILTERED_FINISH_REASON.equals(finishReason)) {
             throw new IllegalStateException(
                     String.format(
                             "ChatModel response was withheld by the provider's content"
-                                    + " filter (finish_reason='%s'), so the content is incomplete"
-                                    + " and cannot yield structured output. Adjust the prompt or"
-                                    + " the provider's content filtering configuration.",
+                                    + " filter (finish_reason='%s'), so the content is"
+                                    + " incomplete. Adjust the prompt or the provider's content"
+                                    + " filtering configuration.",
                             finishReason));
         }
     }
 
     static ChatMessage generateStructuredOutputWithReport(
             RunnerContext ctx, ChatMessage response, Object outputSchema) throws Exception {
-        // Precedes the start report: when this rejects, parsing is never attempted, so there is no
-        // parser execution to report.
-        rejectIncompleteResponse(response);
-
         ExecutionReporters.started(ctx, ExecutionReporter.EntityTypes.PARSER, STRUCTURED_OUTPUT);
         try {
             ChatMessage structuredResponse = generateStructuredOutput(response, outputSchema);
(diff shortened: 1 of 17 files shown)


plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
```diff
@@ -576,8 +579,40 @@ private static Map<String, Object> takeRoutingMetadata(
...
+                                    + " the provider's content filtering configuration.",
+                            finishReason));
+        }
+    }
+
     static ChatMessage generateStructuredOutputWithReport(
             RunnerContext ctx, ChatMessage response, Object outputSchema) throws Exception {
+        // Precedes the start report: when this rejects, parsing is never attempted, so there is no
+        // parser execution to report.
+        rejectIncompleteResponse(response);
```
wenjin272: Could we move this check to the common chat-response path rather than limiting it to structured output? Since Flink Agents currently only handles complete, non-streaming responses, we can validate every response before processing it as text, structured output, or tool calls. This would also prevent truncated responses such as finish_reason=length from being treated as successful.

wenjin272 (commented):
Thanks for taking this on. I have one question to discuss, but it’s not a blocker.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java
weiqingy: Yes, done in both languages.

Something I did not expect while tracing it: truncated tool call arguments mostly never reach the gate. They fail earlier in the connection's converter and come back as `MODEL_CALL_FAILED`. The one case the gate does catch is Java only, where `parseArguments` turns blank arguments into `Map.of()` and the tool runs with no arguments at all, while Python raises. That divergence looks pre-existing. Worth its own issue, or would you rather it rode along here?

The trade is that truncated content no longer reaches the caller at all, under any strategy, and a deterministic `length` now burns retries on every call and again per router candidate. That still seems better than returning a truncated answer that looks successful, but I would be glad to hear if you read the retry cost differently.

wenjin272 (approved):
LGTM, thank you for addressing the common response path. I think the tool-argument difference can be tracked separately. It may also be worth following up on the retry semantics for failures such as `length` and `content_filter`, since retrying the same request may not help in these cases. Thanks again for the contribution!

Issue #1035, opened 2026-08-22
[Tech Debt] bats test bodies run under bash 3.2, so non-final [[ ]] assertions never fail


weiqingy:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

`tools/test/run.sh` refuses to start under bash 3.2, and says why:

```
# Bash 4+ is required: bash 3.2 (macOS default) does not trigger `set -e`
# ... substring assertions in this suite would silently pass on bash 3.2.
```

The guard is aimed at something real, though the sentence above is broader than what holds: a *final* `[[ ]]` does fail the test, since it supplies the body's exit status. The narrow form is the one that bites, and it is measured below. Either way the guard only covers the launcher. `bats-core`'s helpers, including `bats-exec-test`, start with `#!/usr/bin/env bash`, so a test body runs under whatever `bash` resolves to on `PATH` rather than under the interpreter that satisfied the gate. On macOS that is `/bin/bash` 3.2, so the check passes and the bodies still run on 3.2.

Reproduced with the vendored bats, launched by a bash the gate accepts:

```
$ /opt/homebrew/bin/bash --version | head -1
GNU bash, version 5.3.15(1)-release (aarch64-apple-darwin24.6.0)

$ /opt/homebrew/bin/bash tools/test/.bats-cache/bats-core/bin/bats demo.bats
1..3
BODY BASH_VERSION=3.2.57(1)-release
BODY bash path=/bin/bash
ok 1 reports which bash runs this body
REACHED THE LINE AFTER A FALSE ASSERTION
ok 2 a bare non-final [[ ]] that is FALSE
not ok 3 the same check with || return 1
```

Test 2 is the problem. Under bash 3.2 `errexit` does not apply to `[[ ]]`, so a false `[[ ]]` that is not the last command in the body neither fails the test nor stops execution. It reports `ok`. The same condition written `[[ ... ]] || return 1` fails correctly, as test 3 shows, and `[ ... ]` fails correctly too.

A final `[[ ]]` is fine, since it supplies the body's exit status. Only non-final ones are affected. Counting those across the suites:

| File | Non-final bare `[[ ]]` |
|---|---|
| `tools/test/unit/checkpoint_recovery_harness.bats` | 33 |
| `tools/test/unit/verify_example_job.bats` | 4 |
| `tools/test/unit/revalidate_python_constraint.bats` | 2 |
[...]

PR #1043, opened 2026-08-24
[tools][test][e2e] Make the bats suite fail on assertions that stop holding

### LINKED ISSUE #1044: [Bug] edit_plan_quote produces invalid shell on bash 3.2, corrupting the plan state file
### Description

`tools/install.sh`'s `edit_plan_quote` wraps a value in single quotes and escapes any embedded quote as `'\''`, so the result can be sourced back by `edit_plan_dump_state`. It did that with an inline parameter expansion, `${v//\'/\'\\\'\'}`.

That form is not treated alike across bash versions. On 3.2.57 it emits `'it\'\\'\'s'` where the intended output is `'it'\''s'`. The result is not valid shell, so sourcing the dumped state file fails with `unexpected EOF while looking for matching '` and the value is lost.

This affects real users rather than only the test suite. `install.sh` declares macOS support, it is fetched and piped to bash, and macOS resolves `bash` to `/bin/bash` 3.2.57. So the plan-edit feature corrupts its own state file on a stock mac.

`normalize_path` hit the same parameter-expansion difference and works around it, with an implementation note recording why. `edit_plan_quote` did not carry that lesson.

### How to reproduce

Run the installer's plan edit on macOS with a value containing a single quote, for example a venv directory named `o'brien`. The dumped state file fails to source and the edited value is silently dropped.

Directly:

```
$ /bin/bash -c 'v="it'"'"'s"; printf "'"'"'%s'"'"'" "${v//\'"'"'/\'"'"'\\\'"'"'\'"'"'}"'
'it\'\\'\'s'      # 3.2.57, not valid shell
'it'\''s'         # 5.3.15, correct
```

### Version and environment

Reproduced against `8ad3a32b`. macOS, `/bin/bash` 3.2.57. Not reproducible on 4.4+ or 5.x.

Yes. Fixed in the PR for #1035, which is where it was found: pinning the interpreter the test suite runs test bodies under would otherwise have hidden it, since the unfixed code is correct on a newer bash.

weiqingy:
Linked issue: #1035

Also fixes #1044, a user-facing installer bug found while doing this. Follow-up work is tracked in #1045.

### Purpose of change

`tools/test/run.sh` requires bash 4+ because `errexit` does not apply to `[[ ]]` on older bash, so an assertion that stops holding still reports `ok`. But the check binds the shell running `run.sh`, while bats starts every test body through `#!/usr/bin/env bash`, which re-resolves the interpreter from `PATH`. On a stock mac that is `/bin/bash` 3.2. The check passed while the bodies ran under the interpreter it exists to reject.

| # | Change |
|---|---|
| 1 | Correct the floor to bash **4.1**. `[[ ]]` came under `set -e` in 4.1, not 4.0, so a major-version test admits a defective interpreter. |
| 2 | Append `\|\| false` to all 83 bare `[[ ]]` assertions, across 9 files. This is the form bats-core documents and the repo already uses. |
| 3 | Fix `edit_plan_quote` in `tools/install.sh`, which emits invalid shell on bash 3.2 and corrupts the plan state file for macOS users. Also fixes #1044. |
| 4 | Pin the interpreter bats resolves, with a `setup_suite` guard as a backstop if the pin ever stops binding. |
| 5 | Fix an empty-array expansion in `test_submit_examples_to_flink.sh` that errors below bash 4.4. Change 1 declares a 4.1 floor, so without this the branch would name a floor its own suite fails on. |

Change 5 is the only file outside `tools/` and `.github/`.

Three things worth knowing before reading the diff.

**The count in the issue is wrong.** It says 42 vacuous assertions across 5 files; it is 33 across 6. The original count never subtracted the 22 that are the last statement of their block, where a bare `[[ ]]` does supply the exit status and works fine. The issue also says the bulk of the cleanup is mine; it is 11 of 33.

**Nothing was silently failing.** No vacuous assertion is currently false, and CI was never exposed, since `ci.yml` runs `bash tools/test/run.sh` unqualified so a single `PATH` lookup feeds both the gate and the bodies. This closes a way for the suite to go quietly green on developer machines in future.

**All 83 are converted, not just the 33 vacuous today.** The other 50 work only because they happen to be last in their block, which a later edit can change with no signal. Converting all of them makes the rule checkable: a bare `[[ ]]` at statement position is now always a defect, with two documented exceptions.

One accepted cost: the pin re-interprets the scripts under test too, so the suite no longer exercises `install.sh` under the bash its macOS users have. That is how the `edit_plan_quote` bug surfaced. It is recorded in a comment at the pin and tracked in #1045.

### Tests

A passing suite proves nothing here, because it passed before this PR too. So each change is verified by making it fail on purpose, on real interpreters rather than by reasoning about versions.

| Change | Evidence it works |
|---|---|
| `\|\| false` conversion | Mutate an assertion to be false. With `\|\| false` it reports `not ok` on 3.2.57 and 5.3.15. Without it, the same mutation reports `ok` on 3.2.57. |
| bash 4.1 floor | Rejects a real 4.0.44, accepts a real 4.1.17. |
| `edit_plan_quote` fix | The new test fails against the unfixed function and passes against the fixed one. |
| empty-array fix | The test that exercises it fails on 3.2.57, 4.0.44, 4.1.17, 4.2.53 and 4.3.48, and passes from 4.4.23. After the fix it passes on all eight. |
| interpreter pin | A test body reports 5.3.15 from inside itself. The same `PATH` without the pin gives 3.2.57. |
| suite guard | Forced two ways, a dangling shim and an impossible threshold. Each aborts with zero test bodies run and exit 1. |

Also checked:

- Suite is 315/315, exit 0, from a stock-`PATH` shell.
- Conversion completeness: `grep -rnE '^[[:space:]]*\[\[' tools/test --include='*.bats' --exclude-dir=.bats-cache | grep -vE '\|\||&&'` returns exactly the two intended exceptions.
- Both CI legs simulated. With no bash 3.x present the new test skips cleanly; with one present it runs.
- Every shim failure mode: dangling link, shim path existing as a directory, leftover temp files, read-only shim, read-only cache, cold cache, unset interpreter. All are loud, self-healing, or caught by the guard.

Three things worth calling out.

**The `edit_plan_quote` test is built to catch wrong fixes, not just the original bug.** Its input carries two adjacent quotes, a later lone quote and a `$x`, so an escape handling only the first occurrence, only adjacent pairs, or wrapping in double quotes all fail it. The six existing tests for that function miss three such cases.

**The installer fix has independent evidence.** On a 3.2 body the suite used to report `Executed 313 instead of expected 314`, with that test missing from the TAP stream entirely. It now emits every test with no count warning.

**The obvious form of the empty-array fix is wrong**, and the code carries a comment saying so. `("${arr[@]:-}")` yields a one-element array holding the empty string rather than an empty array, which breaks the same test on every version including those that pass today. The alternate-value form is used instead.

### API

No public API change. The changes are to the test harness and to an internal function in the installer script.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Claude Code 2.1.240 (Claude Opus 5)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
         os: [ubuntu-latest, macos-latest]
     steps:
       - uses: actions/checkout@v4
-      - name: Install bash 4+ (macOS)
+      - name: Install bash 4.1+ (macOS)
         if: runner.os == 'macOS'
         run: brew install bash
       - name: Run tools script tests
--- a/e2e-test/test-scripts/test_submit_examples_to_flink.sh
+++ b/e2e-test/test-scripts/test_submit_examples_to_flink.sh
@@ -359,7 +359,11 @@ remove_submission_pid() {
             remaining_pids+=("$pid")
         fi
     done
-    SUBMISSION_PIDS=("${remaining_pids[@]}")
+    # Rebuild with the `+` form: on bash 4.3 and older "${arr[@]}" on an empty
+    # array is an unbound-variable error under `set -u`, and "${arr[@]:-}" would
+    # rebuild one empty-string element instead of nothing. The `:-` in the loop
+    # above is correct: its empty iteration is dropped by the `-n` guard.
+    SUBMISSION_PIDS=("${remaining_pids[@]+"${remaining_pids[@]}"}")
 }
 
 start_attached_java_submission() {
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -1022,9 +1022,20 @@ reconcile_plan_after_edit() {
 # Quote a value for safe re-sourcing by the parent shell via single-quoted
 # assignment. Any embedded single quotes are escaped by closing/reopening
 # the quoted string.
+#
+# Implementation note: the escape is supplied through a variable rather than
+# written inline. Written inline as ${v//\'/\'\\\'\'}, bash 3.2 (macOS
+# /bin/bash) emits 'it\'\\'\'s' where the intended output is 'it'\''s'. The
+# 3.2 form is not valid shell, so sourcing the dumped state file fails with
+# "unexpected EOF while looking for matching `'" and the value is lost.
+# Held in escaped_quote the escape reaches the output untouched: 3.2.57 and
+# 5.3.15 emit identical bytes. Keep escaped_quote at a single backslash;
+# with two, they stop matching.
+# tools/test/unit/edit_plan_quote.bats pins this under a real bash 3.x.
 edit_plan_quote() {
     local v="$1"
-    printf "'%s'" "${v//\'/\'\\\'\'}"
+    local escaped_quote="'\\''"
+    printf "'%s'" "${v//\'/$escaped_quote}"
 }
 
 # Write the subset of plan variables we may have modified to a sourceable
--- a/tools/test/helpers/bash_version_guard.bash
+++ b/tools/test/helpers/bash_version_guard.bash
@@ -1,0 +1,23 @@
+# Suite-level assertion that the interpreter running the test bodies is bash
+# 4.1 or newer. Wired in explicitly by run.sh via --setup-suite-file, because
+# auto-discovery probes the directory named by each path argument and would
+# resolve to unit/ instead.
+#
+# This backs up the PATH pin in run.sh rather than replacing it: bats starts
+# each test process through `#!/usr/bin/env bash`, so if the pin ever stops
+# resolving, the run would otherwise continue under whatever bash PATH finds.
+#
+# A failing setup_suite aborts the run before any test body executes. Use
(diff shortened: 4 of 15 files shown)


wenjin272 (approved):
Thanks for the thorough fix. The changes look good to me overall. 

My only concern is that pinning `bash` through `PATH` also makes `install.sh` run under the newer Bash used by Bats, so the regular suite no longer exercises the stock macOS Bash 3.2 environment that the installer supports. Since this trade-off is clearly documented and tracked in #1045, I don’t think it needs to block this PR, but we should ensure the follow-up restores separate Bash 3.2 coverage for the script under test.

weiqingy:
@wenjin272 thanks for the review. Agreed on the Bash 3.2 point, and I'll restore that coverage as part of #1045. I've linked your comment on that issue so it doesn't get lost.

PR #1046, opened 2026-08-24
[api][integrations][java][python] Raise a clear error for an output schema that cannot be rendered

weiqingy:
Linked issue: #985

### Purpose of change

An output schema that cannot be rendered as JSON Schema fails opaquely. Python leaks a raw `PydanticInvalidForJsonSchema` naming a pydantic-internal type, never the caller's model. Java throws past a catch clause too narrow to see it.

Per the decision on the issue, such a schema is now refused with a clear, cause-chained error naming the schema. `RowTypeInfo` stays a fallback and is documented as one on the public `chat` contract, which previously said nothing about how a connection treats each kind of schema.

A schema that renders but declares no fields is returned as rendered, in both languages. Whether such a document is usable belongs to whatever consumes it.

**Observable changes:**

| Where | Change |
|---|---|
| Python, all 4 sites | unrenderable schema raises `TypeError`, cause chained |
| Java ReAct | catch widened to what Jackson actually throws; rethrown as `IllegalArgumentException`, cause chained |
| Java ReAct | a self-referential POJO overflows the stack inside Jackson. That is an `Error`, so it escaped the catch entirely. Now reported as a schema that refers back to itself |
| Java ReAct | an unsupported output schema type now names the type received, matching the Python message |
| Anthropic (py) | a model pydantic renders but Anthropic refuses now raises `TypeError`, not `ValueError`. Only site where the new type is not a subtype of the old |
| Anthropic (py) | no render at all when the caller supplies `output_config`, so a discarded schema can no longer fail the call |
| Azure (py) | the caller `response_format` conflict is reported before the render |

**The Java connections are deliberately untouched.** Validating them with Jackson would refuse schemas the providers accept: a `@JsonTypeInfo` member renders empty under Jackson while the SDK ships a full `anyOf` union, and Jackson throws on field-less, colliding-property and self-referential POJOs that the SDKs send today. Validating the SDK's own schema is no better, since it renders a map member and an unrenderable member identically. Probed with 15 hostile classes, the SDKs never failed to render, so there is no failure there to report.

That leaves one asymmetry, imposed by the libraries rather than chosen: pydantic raises where the Java SDKs do not.

Not in scope: no change to `RowTypeInfo` handling, no new dependency, no payload change for a schema that renders, and `output_schema` still is not threaded from the runtime to a connection in either language.

### Tests

`ReActAgentTest` and `test_react_agent.py` are the first unit coverage of `ReActAgent` in either language.

Per site: an unrenderable schema raises with the cause chained, a `dict`/`Map` member is accepted and still sent whole, and a normal schema produces the payload it produced before.

Two assertions are load-bearing:

- The self-referential test asserts the chained cause, not only that something was thrown. Removing the catch clause fails it with a raw `StackOverflowError`, which is the regression it exists to catch.
- The Python self-reference test asserts a top-level `$ref`. A model that merely nests another model has `$defs` but no top-level `$ref`, so the weaker guards pass on a non-cyclic model and would miss a regression.

Both languages pin that an unsupported output schema type names the type it received, so neither message can silently stop reporting it.

Verified. Java: `mvn clean test -pl api` 380 passed, openai 104, anthropic 61, zero failures, `spotless:check` clean. Python: `pytest flink_agents --ignore=flink_agents/e2e_tests -m "not integration"` 940 passed, 13 skipped. `check-license.sh` clean. The e2e failures in this worktree are pre-existing, from jars never built here.

`tools/lint.sh` could not run locally (its Python step needs an editable install this pip is too old to perform). `ruff check`, which is what that step runs, is clean on every changed file.

### API

Yes. One helper in `flink_agents.api.agents.types`: `render_output_schema`, used by the three chat model connections and by the ReAct prompt path. It renders through a callable the caller supplies, and reports a render failure clearly.

No Java helper is added. `OutputSchema.java` is unchanged.

No existing signature changes. Exception types change as listed above, and two error messages change wording: the Java unsupported-type message now names the type, and the Python translation failure no longer describes its caller as a chat model, since the prompt path is also a caller.

### Documentation

- [x] `doc-included`

- [x] Yes

Generated-by: Claude Code v2.1.53 (claude-opus-5[1m])

### DIFF
--- a/api/src/main/java/org/apache/flink/agents/api/agents/ReActAgent.java
+++ b/api/src/main/java/org/apache/flink/agents/api/agents/ReActAgent.java
@@ -20,6 +20,7 @@ package org.apache.flink.agents.api.agents;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.lang3.ClassUtils;
 import org.apache.flink.agents.api.Event;
@@ -69,11 +70,28 @@ public class ReActAgent extends Agent {
                 jsonSchema = outputSchema.toString();
                 outputSchema = new OutputSchema((RowTypeInfo) outputSchema);
             } else if (outputSchema instanceof Class) {
+                Class<?> schemaClass = (Class<?>) outputSchema;
+                JsonNode schemaNode;
                 try {
-                    jsonSchema = mapper.generateJsonSchema((Class<?>) outputSchema).toString();
-                } catch (JsonMappingException e) {
-                    throw new RuntimeException(e);
+                    schemaNode = mapper.generateJsonSchema(schemaClass).getSchemaNode();
+                } catch (JsonMappingException | IllegalArgumentException e) {
+                    // Both are reachable: a class whose getters disagree on a property name fails
+                    // the mapping, and one that would not serialize as a JSON object at all is
+                    // refused by the generator with an IllegalArgumentException naming no remedy.
+                    throw new IllegalArgumentException(
+                            String.format(
+                                    "Output schema %s cannot be rendered as a JSON Schema, so it"
+                                            + " cannot constrain the response. Use a schema whose"
+                                            + " fields are all JSON-Schema-renderable, or pass no"
+                                            + " output schema. Rendering it reported: %s",
+                                    schemaClass.getName(), e.getMessage()),
+                            e);
                 }
+                // Deliberately outside the catch above: the check throws an
+                // IllegalArgumentException of its own naming the offending member's path, and
+                // re-wrapping it would bury that path in the cause.
+                OutputSchema.rejectUnconstrainedSchema(schemaNode, schemaClass.getName());
+                jsonSchema = schemaNode.toString();
             } else {
                 throw new IllegalArgumentException(
                         "Output schema must be RowTypeInfo or Pojo class.");
--- a/api/src/main/java/org/apache/flink/agents/api/agents/ReActAgent.java
+++ b/api/src/main/java/org/apache/flink/agents/api/agents/ReActAgent.java
@@ -20,7 +20,6 @@ package org.apache.flink.agents.api.agents;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonMappingException;
-import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.lang3.ClassUtils;
 import org.apache.flink.agents.api.Event;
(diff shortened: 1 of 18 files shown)


api/src/main/java/org/apache/flink/agents/api/agents/ReActAgent.java
```diff
@@ -69,11 +70,28 @@ public ReActAgent(
...
                 outputSchema = new OutputSchema((RowTypeInfo) outputSchema);
             } else if (outputSchema instanceof Class) {
+                Class<?> schemaClass = (Class<?>) outputSchema;
+                JsonNode schemaNode;
                 try {
-                    jsonSchema = mapper.generateJsonSchema((Class<?>) outputSchema).toString();
-                } catch (JsonMappingException e) {
-                    throw new RuntimeException(e);
+                    schemaNode = mapper.generateJsonSchema(schemaClass).getSchemaNode();
+                } catch (JsonMappingException | IllegalArgumentException e) {
```
wenjin272, line 75: Could we also handle self-referential POJOs here? Jackson may throw `StackOverflowError` while generating their schemas, which currently leaks as a raw error. Please catch `StackOverflowError` specifically and add a regression test.

python/flink_agents/api/agents/types.py
```diff
@@ -65,3 +65,177 @@ def __custom_deserialize(self) -> "OutputSchema":
...
+    ref = node.get("$ref")
+    if isinstance(ref, str):
+        prefix = "#/$defs/"
+        target = defs.get(ref[len(prefix) :]) if ref.startswith(prefix) else None
+        _reject_empty_objects(target, path, defs, visited, model)
+        return
+
+    properties = node.get("properties")
+    additional = node.get("additionalProperties")
+    if (
```
wenjin272: This check does not match JSON Schema semantics. For example, a field-less `BaseModel` with `ConfigDict(extra="forbid")` renders as `{"type": "object", "properties": {}, "additionalProperties": false}`, which accepts only `{}` and is therefore strongly constrained, but this branch rejects it. Conversely, omitting `additionalProperties` is semantically equivalent to allowing it, yet the omitted form is rejected while explicit `true` or `{}` is accepted. The check also misses a genuinely unconstrained `RootModel[Any]`, which renders without `type` or `properties`. Since `properties == {}` is not a reliable test, and #985 only concerns render failures, shall we remove the unconstrained-schema check and handle that policy separately?

wenjin272 (commented):
Thanks for working on this and for improving the error reporting! I left two comments below.

api/src/main/java/org/apache/flink/agents/api/agents/ReActAgent.java
weiqingy: Thanks for the review! 

Done. Turned out it isn't only direct self-reference: a mutual cycle, `A` holding a `B` that holds an `A`, blows the stack the same way. So the message talks about the schema referring back to itself rather than naming the direct case.

I gave it its own catch clause rather than adding to the existing one, since `StackOverflowError` carries no message and folding it in would have printed `reported: null`.

python/flink_agents/api/agents/types.py
weiqingy: Yes, removed, in both languages.

Your examples hold up. I ran them against pydantic 2.11.4: the `extra="forbid"` case was being rejected even though it only admits `{}`, and `RootModel[Any]` passed straight through because the check tested `type == "object"`. It was turning away the strictest schema and waving through the emptiest one.

Java had the same check and never looked at `additionalProperties` at all, so that went too. `OutputSchema.java` is back to its state on main.

Render failures are untouched, which is the part #985 was actually about.

Two knock-ons outside your comment, so they don't surprise you in the diff: with one renderer left the two long names collapsed into `render_output_schema`, which touches the imports in the three integrations, and the ReAct tests that asserted rejection now assert the opposite.

weiqingy:
The one failing check isn't related to this PR.

`it-python [ubuntu-latest] [java-17] [python-3.12] [flink-2.2]` hit `HTTP 429 Too Many Requests` from Maven Central while resolving the parent POM `org.apache:apache:pom:34`. Maven couldn't read the project at all, so the `Install flink-agents` step failed and `Run Python IT` never ran. [Job log](https://github.com/apache/flink-agents/actions/runs/33294759880/job/99212408988).

I'll re-run it once the rest of the run settles.

wenjin272 (approved):

Issue #1048, opened 2026-08-25
[Draft] [Bug] PythonActionExecutor leaks per-action Pemja PyObject references


joeyutong:
> [!IMPORTANT]
> **Draft.** Independent local validation is complete. The draft marker is kept while the scope and proposed fix are reviewed.

### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

`PythonActionExecutor` creates Java-side Pemja `PyObject` handles at several per-Action boundaries. Each handle owns a native Python reference and must be closed explicitly; dropping the Java local variable is not sufficient because Pemja's `PyObject` has no finalizer or cleaner.

The affected boundaries are:

1. converting the JSON input into a Python `Event`;
2. receiving the coroutine returned by an async Python Action;
3. retrieving that coroutine from interpreter globals for each poll;
4. keeping the completed coroutine in interpreter globals;
5. returning a yielded or returned Action value from Python to Java during a poll.

The fifth boundary is unnecessary: Java consumes only the coroutine completion state. Returning `(finished, output)` makes Pemja materialize non-built-in output values as additional `PyObject` handles, although the runtime ignores them. Nested containers can also contain additional Python objects, so closing only a top-level output wrapper would not provide a complete ownership rule.

Pemja's [`PyObject`](https://github.com/alibaba/pemja/blob/release-0.5.7/src/main/java/pemja/core/object/PyObject.java#L20-L54) implements `AutoCloseable`, and `close()` performs the native `decRef`. Converting a Python object to a Java `PyObject` [increments its Python reference count](https://github.com/alibaba/pemja/blob/release-0.5.7/src/main/c/pemja/core/pyutils.c#L2003-L2008).

The interpreter global and the temporary Java coroutine wrapper have independent ownership. `interpreter.set(...)` establishes the global reference, so the temporary wrapper can be closed immediately. The global should be deleted after the coroutine completes.

The result is native Python object retention proportional to the number of Python Action invocations. It can increase TaskManager RSS without a matching JVM heap or Flink managed-memory increase.

This is related to, but distinct from, #942 and #944. Those cover executor-lifetime handles released when a task attempt closes; this issue covers objects created repeatedly while the same executor remains open.

### Independent local A/B validation

[...]

PR #1049, opened 2026-08-25
[runtime][python] Release per-action Pemja objects

joeyutong:
Linked issue: #1048

### Purpose of change

`PythonActionExecutor` repeatedly creates Java-side Pemja `PyObject` handles while a Python Action is running. Those handles own native Python references and were previously dropped without calling `close()`.

This change:

- closes the temporary Python Event and coroutine wrappers after each Action invocation;
- keeps a pending coroutine alive through the independent interpreter-global reference established by `interpreter.set(...)`;
- closes the temporary wrapper returned by each `interpreter.get(...)` poll;
- deletes the interpreter-global coroutine reference after the coroutine completes;
- returns only the awaitable completion state from Python instead of materializing ignored yielded/returned values across JNI.

The last point avoids a recursive ownership problem: an Action output may contain arbitrary nested Python objects, while Java never consumes the output. Actions communicate with the runtime through emitted Events, so the narrow boundary is a boolean completion state.

This is complementary to #944: that PR handles executor-lifetime objects during attempt cleanup, while this PR handles objects created repeatedly during the lifetime of an open executor.

### Tests

- `PythonActionExecutorTest`: 11 passed, covering synchronous completion, async storage, Action failure, `set` failure, pending/completed polling, and polling failure.
- `test_function.py`: 27 passed, including yielded and returned values being reduced to completion state.
- `./tools/lint.sh -c`: passed for Python lint and all Java Spotless modules.
- Local Flink A/B: 12 fresh single-slot jobs covering six ownership boundaries in fixed/leak variants, 100 records with a 1 MiB payload each. Every fixed variant ended with zero tracked live objects and zero retained payload. Each leak variant retained exactly 100 affected objects or globals; payload-bearing variants retained 100 MiB. Full details are in #1048.

### API

No user-facing API changes. `PythonActionExecutor` and `call_python_awaitable` are internal runtime bridge helpers. The helper now returns `bool` rather than `(bool, output)` because the output was not consumed. `PyObject.close()` failures continue through the existing `ActionTask.invoke(...) throws Exception` path.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Codex (GPT-5)

### DIFF
--- a/python/flink_agents/plan/function.py
+++ b/python/flink_agents/plan/function.py
@@ -414,28 +414,21 @@ _ASYNCIO_ERROR_MESSAGE = (
 )
 
 
-def call_python_awaitable(awaitable: Any) -> Tuple[bool, Any]:
-    """Invokes the next step of a Python coroutine or generator and returns whether
-    it is done, along with the yielded or returned value.
+def call_python_awaitable(awaitable: Any) -> bool:
+    """Invoke the next step of a Python coroutine or generator.
 
     Args:
         awaitable: A Python coroutine or generator object that can be driven
         by the send() method.
 
     Returns:
-        Tuple[bool, Any]:
-            - The first element is a boolean flag indicating whether the awaitable
-            has finished:
-                * False: The awaitable has more values to yield.
-                * True: The awaitable has completed.
-            - The second element is either:
-                * The value yielded by the awaitable (when not exhausted), or
-                * The return value of the awaitable (when it has finished).
+        True if the awaitable has completed, otherwise False. Yielded and returned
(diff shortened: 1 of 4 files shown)


joeyutong:
The Action reference cleanup and regression tests from this PR are now included in #1058, rebased onto current main alongside the resource-bridge ownership changes. #1057 is the consolidated issue and retains the original evidence from #1048. Closing this PR in favor of #1058; please continue review there.

PR #1051, opened 2026-08-25
[python] Reject unknown constructor arguments on chat model connections/setups

### LINKED ISSUE #1050: [Tech Debt][python] Chat model connections/setups silently ignore unknown constructor arguments
### Search before asking

### Description

`BaseChatModelConnection` and `BaseChatModelSetup` (`python/flink_agents/api/chat_models/chat_model.py`) are pydantic models that inherit `Resource`'s `model_config = ConfigDict(arbitrary_types_allowed=True)`, which does not set `extra=`. Pydantic v2's default for that is `extra="ignore"`.

Every concrete chat-model connection/setup (OpenAI, Azure OpenAI, Anthropic, Watsonx, Ollama, Tongyi) accepts `**kwargs: Any` in `__init__` and forwards it straight into `super().__init__(**kwargs)`.

**Result:** any unrecognized or misspelled constructor argument is silently swallowed — no error, no warning, no log line. For example:

```python
from flink_agents.integrations.chat_models.watsonx.watsonx_chat_model import (
    WatsonxChatModelConnection,
)

# `api_version` only exists on the Java Watsonx connector; a user copying
# Java-side config, or simply mistyping a real field, gets no feedback at all
conn = WatsonxChatModelConnection(
    url="...", api_key="...", project_id="...", api_version="2024-05-01",
)
# constructs successfully; `api_version` is silently discarded
```

This also affects config supplied through YAML, since `PythonResourceProvider.provide()` (`python/flink_agents/plan/resource_provider.py:107-121`) forwards resolved YAML config keys as `**kwargs` into the same constructors — so a YAML typo in a chat-model setup/connection block fails exactly the same way: it loads without complaint and the intended setting simply never applies.

This runs counter to `code_review.md`'s explicit review lens: *"are invalid user inputs rejected with clear errors that name the invalid value?"* Right now they are not rejected at all.

### Proposed fix

Set `extra="forbid"` on `BaseChatModelConnection` and `BaseChatModelSetup`'s `model_config`, so an unknown constructor argument raises a clear pydantic `ValidationError` naming the bad key, instead of vanishing. (Scoped to these two base classes rather than the shared `Resource` base, since some other `Resource` subclasses rely on today's "ignore" behavior in ways that need separate review.)

Ashfaqbs:
### What is the purpose of the change

`BaseChatModelConnection` and `BaseChatModelSetup` inherit pydantic's default `extra="ignore"` behavior, and every concrete chat-model connection/setup forwards `**kwargs` straight into `super().__init__()`. An unknown or misspelled constructor argument -- whether passed directly or via YAML config through `PythonResourceProvider` -- is silently dropped instead of raising, so a typo'd or unsupported config key appears to apply but has no effect.

This PR sets `extra="forbid"` on both base classes so an unrecognized key raises a clear `ValidationError` naming it, per `code_review.md`'s guidance that invalid input should be rejected with a clear error.

Fixes #1050.

### Brief change log

- `python/flink_agents/api/chat_models/chat_model.py`: set `model_config = ConfigDict(arbitrary_types_allowed=True, extra="forbid")` on `BaseChatModelConnection` and `BaseChatModelSetup`; declare `name: str | None` (excluded from serialization) on both, since callers and the resource provider commonly pass a resource `name` through that isn't itself a connection/setup field.
- `python/flink_agents/integrations/chat_models/ollama_chat_model.py`: remove `OllamaChatModelSetup`'s `request_timeout` constructor parameter. It was forwarded to `super().__init__()` but never declared as a field on that class, so with `extra="ignore"` it silently did nothing even before this change (`request_timeout` is a real, used field only on `OllamaChatModelConnection`). Enabling `extra="forbid"` turned this pre-existing dead parameter into a hard failure in tests, which is what surfaced it.
- `python/flink_agents/api/chat_models/tests/test_chat_model_base.py`: regression tests asserting an unknown constructor argument raises `ValidationError` on both base classes, and that `name` is still accepted without being stored.

### Tests

- Added unit tests (above) verifying the new rejection behavior.
- Ran the full existing chat-model test suite (`api/chat_models`, `integrations/chat_models/**`, 227 tests) -- all pass with no regressions.
- Ran the broader `api/`, `plan/`, and `integrations/chat_models/` suites (722 tests) to check for any other silent-kwarg dependency this change might expose -- no new failures. The 4 pre-existing failures in that run (`test_bash_tool.py`, one YAML schema test) reproduce identically on `main` without this change and are Windows-environment-specific, unrelated to this PR.
- `./tools/lint.sh`-equivalent (`ruff check` / `ruff format --check`) passes on all changed files.

### Are these changes tested?

Yes, see above.

### Are there any user-facing changes?

Yes: constructing a chat-model connection or setup (via Python API or YAML) with an unrecognized keyword argument now raises `pydantic.ValidationError` instead of silently ignoring it. Any caller currently passing a stray/misspelled key will now see a clear error at construction time instead of the setting silently having no effect.

### Was this PR generated with the help of a generative AI tool?

Yes.


python/flink_agents/api/chat_models/chat_model.py
```diff
@@ -115,6 +115,17 @@ class BaseChatModelConnection(Resource, ABC):
     One connection can be shared in multiple chat model setup.
     """
 
+    # Reject unrecognized constructor arguments instead of silently ignoring them
+    # (pydantic's default extra="ignore"), so a misspelled or unsupported config
+    # key fails loudly at construction time instead of appearing to apply and
+    # then having no effect. `name` is declared below purely so the resource
+    # name callers and the resource provider commonly pass through survives
+    # this check; it is not otherwise used by this class.
+    model_config = ConfigDict(arbitrary_types_allowed=True, extra="forbid")
```
wenjin272, line 125: Applying `extra="forbid"` here also affects the Java chat-model wrappers. The current [cross-language CI](https://github.com/apache/flink-agents/actions/runs/32878446289/job/97901950082) fails because `JavaChatModelSetupImpl` receives `java_clazz` and `extract_reasoning`: `java_clazz` is bridge metadata, while `extract_reasoning` is a valid argument consumed by the Java Ollama setup, but both are rejected as unknown Python fields. Java descriptors intentionally carry implementation-specific arguments in an open map, so this also makes Java and Python validation semantics diverge.

Could we keep strict validation for Python-native chat models while filtering these arguments before constructing `JavaChatModel*Impl`, or explicitly retaining `extra="ignore"` on the Java wrapper classes?

python/flink_agents/api/chat_models/chat_model.py
```diff
@@ -115,6 +115,17 @@ class BaseChatModelConnection(Resource, ABC):
...
+    name: str | None = Field(
```
wenjin272: Could we avoid adding `name` as an accepted chat-model field? The in-tree `name=` call sites are all tests, while the resource name is owned by `ResourceProvider` and is not passed into the resource during construction. This therefore preserves a previously ignored, behaviorless argument solely to keep those tests passing, which works against the goal of rejecting unsupported inputs.

I think these tests should drop `name=` instead of adding a no-op exception to strict validation.

wenjin272 (commented):
Thanks for working on this! I left two inline comments about the cross-language wrapper behavior and the `name` field.

python/flink_agents/api/chat_models/chat_model.py
Ashfaqbs: Good catch, thanks. Fixed by overriding `model_config` back to `extra="ignore"` on `JavaChatModelConnection`/`JavaChatModelSetup` (the Java-backed subclasses), so strict validation stays scoped to Python-native chat models and Java descriptors keep passing through `java_clazz`, `extract_reasoning`, etc. as before. Pushed in `c971dee7`.

python/flink_agents/api/chat_models/chat_model.py
Ashfaqbs: Agreed, that's cleaner. Removed the `name` field from both base classes and dropped `name=` from the in-tree test call sites (confirmed `ResourceProvider` never forwards the resource name into the constructor -- it's kept on the provider itself). Pushed in `c971dee7`.

wenjin272 (approved):

PR #1054, opened 2026-08-27
[java][python][integrations][anthropic] Stop sending temperature on models that reject it

### LINKED ISSUE #1036: [Bug] [integrations][anthropic] temperature defaults to 0.1 and is always sent, returning HTTP 400 on Claude 4.7 and later

Ashfaqbs:
## Summary

Fixes #1036.

The Anthropic integration gives `temperature` a default of `0.1` and puts it on every request unconditionally, in both the Java and Python connections. Claude 4.7-generation models and later (`claude-opus-4-7`, `claude-opus-4-8`, `claude-opus-5`, `claude-sonnet-5`, `claude-fable-5`, `claude-mythos-5`, `claude-mythos-preview`) reject a non-default sampling parameter with an HTTP 400, so selecting one of those models turned every chat call into a provider error instead of a completion.

Claude 4.6-generation models (`claude-opus-4-6`, `claude-sonnet-4-6`) still accept `temperature`, so this boundary sits one generation narrower than the existing `PREFILL_UNSUPPORTED_MODELS` list and can't reuse it (the same reasoning already documented on that list applies here in reverse).

## Changes

- **Java** (`AnthropicChatModelConnection.java`): added a `SAMPLING_UNSUPPORTED_MODELS` allowlist and a `supportsSampling(String)` predicate, mirroring the existing `supportsJsonPrefill` gate. `buildRequest` now only forwards `temperature` onto the request builder when the effective model accepts it.
- **Python** (`anthropic_chat_model.py`): added the matching `_SAMPLING_UNSUPPORTED_MODELS` / `_supports_sampling` and popped `temperature` from `kwargs` before calling `messages.create` when the model doesn't accept it.
- Per the issue, `top_p` and `top_k` are out of scope — those only reach the request when a caller explicitly passes them via `additional_kwargs`/`kwargs`, not as a default the integration injects.

The fix omits the parameter rather than changing its value, since Anthropic's own default (or omitting the parameter) is accepted on every model.

## Test plan

- Added Java tests to `AnthropicChatModelConnectionTest`: parametrized coverage of every model in the new unsupported list reporting unsupported, every other model (including the 4.6-generation boundary models) reporting supported, and two `buildRequest` tests asserting `temperature` is present/absent on the built request.
- Added Python tests to `test_anthropic_response_parsing.py`: parametrized predicate coverage plus two request-level tests asserting `temperature` is omitted/forwarded in the kwargs sent to `messages.create`.
- Verified RED→GREEN: with the test files in place and only the source fix reverted (`git stash` on the source-only diff), the Java tests fail to compile against the missing `supportsSampling` symbol and the Python test module fails to import `_supports_sampling` — confirming the tests exercise the new behavior rather than passing vacuously. Restored the fix and reran: all new tests pass.
- `mvn -pl integrations/chat-models/anthropic -am test` — full module suite green, no regressions.
- `pytest flink_agents/integrations/chat_models/anthropic/tests/test_anthropic_response_parsing.py` — 77 passed.
- `ruff check` / `ruff format --check` on the touched Python files — clean.
- `spotless:apply` run on the touched Java test file to match AOSP formatting.

### DIFF
--- a/integrations/chat-models/anthropic/src/main/java/org/apache/flink/agents/integrations/chatmodels/anthropic/AnthropicChatModelConnection.java
+++ b/integrations/chat-models/anthropic/src/main/java/org/apache/flink/agents/integrations/chatmodels/anthropic/AnthropicChatModelConnection.java
@@ -223,6 +223,45 @@ public class AnthropicChatModelConnection extends BaseChatModelConnection {
         return !PREFILL_UNSUPPORTED_MODELS.contains(effectiveModel);
     }
 
+    // Models Anthropic documents as rejecting a non-default sampling parameter such as
+    // temperature. Source of truth:
+    // https://platform.claude.com/docs/en/build-with-claude/working-with-messages
+    //
+    // Sampling rejection starts at the 4.7 generation and also covers Claude Mythos 5, Claude
+    // Fable 5 and Claude Sonnet 5; a request that carries temperature for one of them is answered
+    // with a 400 rather than a completion. Claude 4.6-generation models still accept it, which is
+    // why this boundary sits one generation later than PREFILL_UNSUPPORTED_MODELS above and cannot
+    // reuse it: Claude 4.6 rejects a prefill while still accepting a temperature. Names carry no
+    // date and are pinned, so the name is itself the snapshot and is matched exactly, and a name
+    // outside the list is treated as accepting the parameter.
+    private static final Set<String> SAMPLING_UNSUPPORTED_MODELS =
+            Set.of(
+                    "claude-opus-4-7",
+                    "claude-opus-4-8",
+                    "claude-opus-5",
+                    "claude-sonnet-5",
+                    "claude-fable-5",
+                    "claude-mythos-5",
+                    "claude-mythos-preview");
+
+    /**
+     * Whether {@code effectiveModel} accepts a non-default sampling parameter such as {@code
+     * temperature}.
+     *
+     * <p>See the list above for the source of truth and for why it is one generation narrower than
+     * {@link #PREFILL_UNSUPPORTED_MODELS}. An unrecognized name reports {@code true}, matching the
+     * documented rule that sampling parameters are the long-standing behaviour and only the listed
+     * names withdraw support.
+     */
+    static boolean supportsSampling(String effectiveModel) {
+        // Load-bearing: the list is an immutable Set, whose contains(null) throws rather than
+        // reporting absence.
+        if (effectiveModel == null) {
+            return true;
+        }
+        return !SAMPLING_UNSUPPORTED_MODELS.contains(effectiveModel);
+    }
+
     /**
      * Derives the native {@code output_config} for a POJO class through the SDK's typed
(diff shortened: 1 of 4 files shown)


wenjin272:
Thanks for working on this. #1037 already addresses #1036 and is currently under review. Since the two PRs substantially overlap and conflict in several files, I think we should continue with #1037 and close this PR as a duplicate. 

Thanks again for the contribution.

Issue #1059, opened 2026-08-28
[Feature] Multimodal content blocks in ChatMessage


Zhuoxi2000:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Tracking issue for the "Support invoking multimodal models" 0.4 Must item from #862. Design discussed and agreed in #1031.

Agreed direction:

* Replace `ChatMessage.content: String` with `blocks: List<ContentBlock>` in both Java and Python. No dual representation. Existing text factories continue to work by creating a single `TextBlock`, and `getText()` / `.text` provides the ordered text projection.
* `ContentBlock` includes `TextBlock` plus `ImageBlock`, `AudioBlock`, `VideoBlock`, and `DocumentBlock`. Media blocks share `mime_type`, exactly one of base64 `data` or external `url`, and optional metadata such as `name`, `size_bytes`, and `sha256`.
* Java and Python share the same serialized shape using a `type` discriminator (`text|image|audio|video|document`), covered by the cross-language snapshots and pemja bridge.
* Block ordering is preserved end-to-end. TOOL messages use the same block model, which also provides the path to fix MCP image tool results being stringified today.
* Each `ChatModelConnection` maps blocks to its provider-native representation. Unsupported block types fail explicitly rather than being silently dropped or converted.
* Event Log records media metadata only at every log level, including VERBOSE. Inline payloads are omitted and URL query parameters are stripped.
* `tool_calls` and `extra_args` stay unchanged here. Their cleanup remains tracked in #1056.

Subtasks:

**Phase 1: framework**

* [ ] Add the `ContentBlock` model and migrate `ChatMessage` in Java and Python
* [ ] Update factories, text projection, prompt/message merging, and repo call sites
* [ ] Update chat-event serialization, cross-language snapshots, and pemja bridge
* [ ] Add Event Log media redaction and URL sanitization

**Phase 2: providers, 0.4 target**

* [ ] OpenAI Chat Completions support in Java and Python, covering OpenAI, Azure OpenAI, and vLLM
* [ ] Ollama support in Java and Python
* [ ] Explicit unsupported-block errors for providers not yet migrated

**Follow-ups**

* [ ] OpenAI Responses API
* [ ] Anthropic, replacing the existing `anthropic_content_blocks` workaround
* [ ] Gemini, Bedrock, and Tongyi
* [ ] MCP image tool results as TOOL message blocks
* [ ] Multimodal YAML prompts

[...]

PR #1060, opened 2026-08-28
[api] Introduce multimodal content blocks

Zhuoxi2000:
Linked issue: #1059 (Phase 1 — the tracking issue stays open for the provider phase)

### Purpose of change

Implements the framework part of the multimodal design agreed in #1031.

`ChatMessage.content: String` is replaced by `blocks: List<ContentBlock>` in both Java and Python. The initial block types are `TextBlock`, `ImageBlock`, `AudioBlock`, `VideoBlock`, and `DocumentBlock`. Media blocks share a MIME-typed shape with `mime_type`, exactly one of base64 `data` or external `url`, and optional metadata such as `name`, `size_bytes`, and `sha256`.

Existing text constructors/factories keep their signatures and create a single `TextBlock`. `getText()` / `.text` returns the ordered text projection.

Java and Python use the same `type`-discriminated JSON format, covered by the regenerated cross-language snapshots. The pemja bridge is updated to carry blocks across runtimes as well.

This PR also updates prompt handling so image-only messages are preserved, and `Prompt.formatMessages` applies substitutions only to text blocks while passing media blocks through unchanged.

Provider-specific multimodal conversion is intentionally left to the next #1059 phase. Existing providers continue using the text projection in this PR.

Behavior changes:

* CEL expressions using `response.content` need to migrate to `response.blocks`.
* Python `ChatMessage` now rejects unknown fields, so the removed `content=` argument fails explicitly instead of silently creating an empty message.
* `toString()` / `__str__`, equality, and hashing now use the block representation.
* `data` and `url` are mutually exclusive for media blocks.

### Tests

* Added Java and Python `ChatMessage` serialization tests covering the wire format, mixed block ordering, media fields, round trips, source validation, and rejection of the removed `content` argument.
* Regenerated the Java/Python cross-language snapshots using the existing snapshot flows; cross-deserialization passes in both directions.
* Added durable-state serde coverage for a mixed text + image `ChatMessage`.
* Java reactor tests and Python tests pass locally, aside from environment-specific dependency failures that are covered by CI.

### API

Yes. This is a Beta breaking change discussed in #1031.

`ChatMessage.content` is replaced by `blocks` in both Java and Python. Java removes `getContent()/setContent`; Python removes `.content`. Text-only access remains available through `getText()` / `.text` and `setText()` / `set_text()`.

New public types in both languages are `ContentBlock`, `TextBlock`, `MediaBlock`, `ImageBlock`, `AudioBlock`, `VideoBlock`, and `DocumentBlock`.

`ChatMessage` also adds `getBlocksAsMaps()` / `setBlocksFromMaps()` for the pemja bridge, where blocks cross as plain maps.

CEL expressions referencing `response.content` must migrate to `response.blocks`.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included` 

- [x] Yes
- [ ] No

Generated by: Claude Code  Fable 5


wenjin272:
Thanks for taking this on. Apologies for the delayed review due to our recent team outing. I will complete the review by the end of this week.

api/src/main/java/org/apache/flink/agents/api/chat/messages/MediaBlock.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.Objects;
+
+/**
+ * Shared shape for binary media blocks: modality is the concrete type, encoding is the MIME type.
+ *
+ * <p>The payload is carried by exactly one of base64 {@code data} or an externally managed {@code
+ * url} (enforced by the argument constructor and the per-type factories; the no-arg bean path is
+ * lenient for deserialization). URL-backed content is externally managed: URLs may expire, may not
+ * be reachable by the model provider, and may be invalid after recovery from a checkpoint. The
+ * optional {@code name}/{@code sizeBytes}/{@code sha256} metadata also serves the Event Log, which
```
wenjin272, line 35: We need an explicit serialization contract for Event Log output here. The regular `ChatMessage` JSON representation is also used by the Java/Python bridge, event serialization, and state recovery, where `data` and the original `url` must be preserved. Therefore, changing the global `ChatMessage` serialization to remove these fields would break the normal wire format. Reusing that same representation for Event Log output, however, cannot guarantee metadata-only and sanitized logging.

Could each `ContentBlock` define its log-safe representation—for example through `sanitize()`—and could the Event Log serialization module register a dedicated `ChatMessage` serializer that uses it? A media block should explicitly whitelist its loggable metadata, omit inline `data`, and sanitize credentials from URLs. This keeps `EventLogRecordJsonSerializer` generic, avoids hard-coding every block subtype into it, and lets future block types define their own logging policy. Tests should verify both that normal JSON serialization still preserves the complete payload and that Event Log serialization never exposes it at either STANDARD or VERBOSE level.

plan/src/test/java/org/apache/flink/agents/plan/resource/python/PythonPromptTest.java
```diff
@@ -75,9 +75,9 @@ public void testFromSerializedMapWithMessageListTemplate() {
                 prompt.formatMessages(MessageRole.SYSTEM, new HashMap<>());
         assertThat(formattedMessages).hasSize(2);
         assertThat(formattedMessages.get(0).getRole()).isEqualTo(MessageRole.SYSTEM);
-        assertThat(formattedMessages.get(0).getContent()).isEqualTo("You are a helpful assistant.");
+        assertThat(formattedMessages.get(0).getText()).isEqualTo("You are a helpful assistant.");
```
wenjin272, line 78: This test still constructs the serialized Python messages with the old `content` field, while the production Python resource now serializes them as `blocks`. `PythonPrompt.parseChatMessage()` still reads `content` and defaults it to an empty string, so a real message-based Python prompt loses all of its blocks when restored in Java. Could we parse `blocks` through the existing map/Jackson conversion and update this fixture to use the actual Python `model_dump()` shape, including text and media blocks?

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
```diff
@@ -97,7 +97,9 @@ private static List<ChatMessage> deepCopy(List<ChatMessage> messages) {
                     toolCalls.add(call == null ? null : new HashMap<>(call));
                 }
             }
-            copy.add(new ChatMessage(m.getRole(), m.getContent(), toolCalls, m.getExtraArgs()));
+            // The full constructor copies the block list; blocks themselves are shared, matching
```
wenjin272, line 100: This copies the block list but still shares every mutable `ContentBlock` instance with the original request. For example, a strategy can obtain the first `TextBlock` from `context.getMessages()`, call `setText("mutated")`, and the original message later sent to the model will also return `"mutated"`. The same aliasing exists for `MediaBlock.setData()` and `setUrl()`. This does not require a hostile strategy; it is exactly the kind of accidental mutation that the defensive copy is intended to isolate, and it is newly exposed because the previous text content was an immutable `String`.

Rather than adding a type-specific deep-copy branch whenever a new block type is introduced, could we make the concrete `ContentBlock` implementations immutable—`final` fields, no setters, and validated `@JsonCreator` constructors—and snapshot the block list in `ChatMessage`? `RoutingContext` could then safely share block instances while continuing to copy the mutable message/list structure. The routing isolation tests should also cover block-level aliasing rather than only replacing the copied message's text/list.

api/src/main/java/org/apache/flink/agents/api/chat/messages/MediaBlock.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Nullable private String name;
+
+    @JsonProperty("size_bytes")
+    @Nullable
+    private Long sizeBytes;
+
+    @Nullable private String sha256;
+
+    protected MediaBlock() {}
```
wenjin272, line 56: The constructor enforces exactly one of `data` and `url`, but the no-arg Jackson path and the independent setters bypass that check. Java can therefore accept a wire payload containing both or neither source while Python rejects the same payload. Could we use validated immutable `@JsonCreator` construction—or represent the source as `Base64Source | URLSource`—and freeze or validate assignment on the Python side as well?

docs/content/docs/development/chat_models.md
```diff
@@ -97,7 +97,7 @@ class MyAgent(Agent):
     @staticmethod
     def process_response(event: Event, ctx: RunnerContext) -> None:
         chat_response = ChatResponseEvent.from_event(event)
-        response_content = chat_response.response.content
+        response_content = chat_response.response.text
```
wenjin272, line 100: This response access was migrated, but the same documentation still contains `ChatMessage(..., content=...)` examples and Java calls to the removed `getContent()` method. The Python examples now fail because unknown fields are forbidden, while the Java examples no longer compile. Similar remnants remain in `prompts.md`, `react_agent.md`, `workflow_agent.md`, and the quickstarts. Could we complete a repository-wide documentation migration to `blocks`/the message factories and `text`/`getText()`?

wenjin272 (commented):
Thanks for the work on this comprehensive multimodal message refactor! I left a few comments on the Event Log serialization contract, cross-language migration and invariants, routing isolation, and documentation updates.

api/src/main/java/org/apache/flink/agents/api/chat/messages/MediaBlock.java
```diff
@@ -0,0 +1,155 @@
...
+ * lenient for deserialization). URL-backed content is externally managed: URLs may expire, may not
+ * be reachable by the model provider, and may be invalid after recovery from a checkpoint. The
+ * optional {@code name}/{@code sizeBytes}/{@code sha256} metadata also serves the Event Log, which
+ * records media metadata instead of payload bytes.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public abstract class MediaBlock extends ContentBlock {
+
+    @JsonProperty("mime_type")
+    private String mimeType;
```
wenjin272, line 42: Would mediaType be a better name than mimeType here? Media Type is the terminology used by the current standards, while MIME type is more of a commonly used historical name. Since this field is part of the public multimodal ContentBlock API rather than a MIME-specific protocol implementation, mediaType may better describe its semantics and also reads naturally alongside MediaBlock.

This is mostly an API naming suggestion, but since these block types are being introduced in this PR, it may be worth settling the terminology before the API becomes public.

api/src/main/java/org/apache/flink/agents/api/chat/messages/MediaBlock.java
```diff
@@ -0,0 +1,207 @@
...
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public abstract class MediaBlock extends ContentBlock {
+
+    @JsonProperty("media_type")
+    private final String mediaType;
+
+    @Nullable private final String data;
+
+    @Nullable private final String url;
```
wenjin272, line 53: Could we model the payload location as a typed `MediaSource` instead of keeping mutually exclusive nullable `data` and `url` fields directly on `MediaBlock`?

The current XOR validation works for the two initial sources, but it makes invalid states representable and does not scale well. Discussion #1031 explicitly leaves room for a future managed blob/reference source. With the current flat shape, adding `blob_id`, provider file IDs, or another source would require expanding the N-way exclusivity validation in Java and Python, as well as updating serializers and bridge conversions.

A discriminated source shape would make the invariant structural, for example:

```json
{
  "type": "image",
  "media_type": "image/png",
  "source": {
    "type": "base64",
    "data": "..."
  }
}
```

This PR only needs `Base64Source` and `UrlSource`; a managed `BlobSource` can remain future work. This is similar to AgentScope's `Source` model and gives us an explicit extension point without implementing blob storage now. Providers would still explicitly convert or reject supported source types.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/ChatMessageEventLogSerializer.java
```diff
@@ -0,0 +1,67 @@
...
+ * ChatMessage} and must never be reconstructed into one: an inline-backed media block drops its
+ * {@code data} (so reconstruction fails loudly), and a URL-backed one carries only the stripped
+ * URL.
+ */
+public class ChatMessageEventLogSerializer extends JsonSerializer<ChatMessage> {
+
+    /** The module Event Log mappers register to apply the sanitized {@link ChatMessage} shape. */
+    public static Module module() {
+        return new SimpleModule("flink-agents-event-log-chat-messages")
+                .addSerializer(ChatMessage.class, new ChatMessageEventLogSerializer());
```
wenjin272, line 50: Python-originated built-in events are currently deserialized as the base `Event` at the cross-language boundary. Therefore, infrastructure that runs before the action, including Event Log serialization, cannot see the concrete event type, and the `ChatMessage`-specific serializer may be bypassed.

This is a broader cross-language event restoration issue rather than something specific to this PR. I opened #1125 to track the framework-level fix for 0.4, so I don't think it needs to block this PR.

python/flink_agents/api/chat_message.py
```diff
@@ -43,6 +44,88 @@ class MessageRole(str, Enum):
...
+    provider, and may be invalid after recovery from a checkpoint. The optional
+    ``name``/``size_bytes``/``sha256`` metadata also serves the Event Log,
+    which records media metadata instead of payload bytes.
+    """
+
+    # Frozen keeps sharing a block (e.g. across a routing context copy) safe,
+    # and matches the validated immutable construction on the Java side.
+    model_config = ConfigDict(frozen=True)
+
+    media_type: str
```
wenjin272, line 75: Java rejects both `null` and an empty `media_type`, but the Python model currently uses an unconstrained `str`, so this is accepted:

```python
ImageBlock(media_type="", data="aGk=")
```

The object is therefore valid in Python but fails later when it crosses the Java bridge and is reconstructed as a Java `MediaBlock`. Could we enforce the same non-empty constraint in Python, for example with `Field(min_length=1)`, and add a negative cross-language test? The public Java and Python APIs should agree on which wire values are valid.

python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
```diff
@@ -426,18 +426,14 @@ def chat(
             ]
 
             extra_args["anthropic_content_blocks"] = message.content
-            return ChatMessage(
-                role=MessageRole(message.role),
-                content=text,
+            return ChatMessage.of(MessageRole(message.role), text,
```
wenjin272, line 429: Please rebase this PR onto the latest `main`, resolve the current merge conflict, and rerun CI. The current failures appear to come from call sites and tests on `main` that have not been migrated to the new `ChatMessage` API.

wenjin272 (commented):
Thanks for the substantial update and for addressing the earlier feedback. The immutable block model, construction-time validation, RoutingContext isolation, PythonPrompt restoration, and Event Log projection are solid improvements. I left four comments on the media-source abstraction, the separately tracked cross-language event restoration (#1125, non-blocking for this PR), Java/Python `media_type` parity, and rebasing onto the latest `main`.

python/flink_agents/integrations/chat_models/anthropic/anthropic_chat_model.py
Zhuoxi2000: Done. Rebased onto the latest main and fixed the conflicts.

python/flink_agents/api/chat_message.py
Zhuoxi2000: Fixed. Python now rejects empty media_type values up front. Same for empty Base64 data and empty URLs.

runtime/src/main/java/org/apache/flink/agents/runtime/eventlog/ChatMessageEventLogSerializer.java
Zhuoxi2000: Yep, #1125 is exactly the gap I hit here.

Right now Event.fromJson() turns Python-originated built-in events into a base Event with map attributes, so by the time the logger sees them we've already lost the typed event.

api/src/main/java/org/apache/flink/agents/api/chat/messages/MediaBlock.java
Zhuoxi2000: Done ! This shape is cleaner.

api/src/main/java/org/apache/flink/agents/api/chat/messages/MediaBlock.java
Zhuoxi2000: Yep, agreed.

docs/content/docs/development/chat_models.md
Zhuoxi2000: Done!

api/src/main/java/org/apache/flink/agents/api/chat/messages/MediaBlock.java
Zhuoxi2000: Fixed.

api/src/main/java/org/apache/flink/agents/api/chat/model/routing/RoutingContext.java
Zhuoxi2000: Yep, I went with the immutable approach you suggested.

plan/src/test/java/org/apache/flink/agents/plan/resource/python/PythonPromptTest.java
Zhuoxi2000: Good catch  this was an actual bug.Fixed it

api/src/main/java/org/apache/flink/agents/api/chat/messages/MediaBlock.java
Zhuoxi2000: Done. ContentBlock and MediaSource now handle their own log-safe form through sanitize().

For media blocks, we only keep the metadata and sanitized source. Base64 data is dropped, and URLs are stripped of credentials, query params, and fragments.

I also added a dedicated ChatMessageEventLogSerializer for the Event Log side rather than putting this into the generic serializer.

There are tests for both STANDARD and VERBOSE to make sure payloads and sensitive URL parts don't end up in the logs.

The Python-originated event case is still separate and tracked in #1125.

python/flink_agents/api/chat_message.py
```diff
@@ -43,6 +44,136 @@ class MessageRole(str, Enum):
...
+    text: str = ""
+
+    def __str__(self) -> str:
+        return self.text
+
+
+class Base64Source(BaseModel):
+    """An inline media payload, carried as base64 text."""
+
+    model_config = ConfigDict(frozen=True)
```
wenjin272, line 62: Could we configure the nested block and source models with `extra="forbid"` as well?

`ChatMessage` rejects unknown fields, but Pydantic configuration does not propagate into nested models. For example, Python currently accepts a `Base64Source` containing both `data` and an unexpected `url` field, then silently drops `url`, while Java rejects the same wire payload. The same issue applies to unknown fields on `TextBlock`, `UrlSource`, and `MediaBlock`.

Besides creating a Java/Python contract mismatch, this weakens the structural source invariant because malformed source objects can be accepted and normalized silently. Could we set `extra="forbid"` on these nested models and add mirrored negative wire-format tests?

api/src/main/java/org/apache/flink/agents/api/chat/messages/TextBlock.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Objects;
+
+/** A plain-text, immutable part of a {@link ChatMessage}. */
+public final class TextBlock extends ContentBlock {
+
+    private final String text;
+
+    @JsonCreator
+    public TextBlock(@JsonProperty("text") String text) {
+        this.text = text != null ? text : "";
```
wenjin272, line 35: Could we align the handling of an explicit `null` text value between Java and Python?

The Java constructor currently normalizes `"text": null` to an empty string, while the Python `TextBlock` rejects `None`. As a result, the same public wire payload is accepted and rewritten by Java but rejected by Python.

Since `TextBlock` is being introduced in this PR, I would prefer rejecting explicit `null` in Java as well, while keeping the omitted-field/default behavior consistent if needed. Either contract is possible, but both languages should make the same decision and cover it with mirrored tests.

wenjin272 (commented):
Thanks for the thorough update and for addressing the previous feedback. The typed MediaSource design, immutable blocks, Event Log sanitization, Python prompt restoration, media_type validation, and rebase all look much better. I found two remaining Java/Python wire-contract inconsistencies around nested unknown fields and explicit null text values; please take a look at the inline comments.

PR #1063, opened 2026-08-29
[integrations][java][python] Update stale Anthropic default model to claude-sonnet-4-5

Ashfaqbs:
Linked issue: #1038

### Purpose of change

The Anthropic chat model integration defaulted to the deprecated `claude-sonnet-4-20250514` snapshot in both language runtimes, restated across Javadoc, a Python docstring, and the docs parameter tables/examples. A caller who took the default silently got a deprecated model and would hit a hard failure once it retires.

- Java: `AnthropicChatModelSetup.DEFAULT_MODEL`, plus its Javadoc parameter entry and usage example.
- Python: `DEFAULT_ANTHROPIC_MODEL`, and its docstring now names the constant instead of restating the literal (matching the OpenAI and Tongyi integrations, which is why those two never drifted).
- Docs: `AnthropicChatModelSetup` parameter tables and usage examples for both languages, and pruned the "Available Models" list down to the current default — it previously recommended `claude-3-7-sonnet` and `claude-opus-4-1`, both already past retirement.

Anthropic was also the only chat model integration using a dated snapshot as its default; every other one (OpenAI, Gemini, Tongyi) defaults to an undated alias, which is why only this one went stale unnoticed.

Left untouched, deliberately:
- The Bedrock integration's own model-ID default — `us.anthropic.claude-sonnet-4-20250514-v1:0` is Bedrock's own model catalog naming, not the Anthropic API snapshot this issue is about.
- The `AnthropicChatModelConnectionTest` fixtures that use the old snapshot name as an example of a model *not* capable of native structured output — those intentionally exercise the "not currently capable" recognition path rather than pin a default.

### Tests

- Updated `AnthropicChatModelSetupTest.testGetParametersDefaults` to assert the new default (was pinning the old one) — 2/2 pass.
- Full `AnthropicChatModelConnectionTest` suite (58 tests) still passes unchanged, confirming the deliberately-untouched fixtures weren't affected.
- Python: full `anthropic` integration test suite (64 passed, 2 skipped) still passes, including `test_anthropic_chat_model.py`'s check against the `DEFAULT_ANTHROPIC_MODEL` constant.
- `spotless:check` (Java) and `ruff check` + `ruff format --check` (Python) both clean.

### API

No public API shape change — only the default value of an existing optional parameter.

### Documentation

- [x] `doc-included`

- [x] Yes

This fix follows from my own diagnosis of the bug. I used AI tooling for exploration and testing, directing it toward the change I had in mind.

### DIFF
--- a/docs/content/docs/development/chat_models.md
+++ b/docs/content/docs/development/chat_models.md
@@ -287,7 +287,7 @@ Anthropic provides cloud-based chat models featuring the Claude family, known fo
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
 | `connection` | str | Required | Reference to connection method name |
-| `model` | str | `"claude-sonnet-4-20250514"` | Name of the chat model to use |
+| `model` | str | `"claude-sonnet-4-5"` | Name of the chat model to use |
 | `prompt` | Prompt \| str | None | Prompt template or reference to prompt resource |
 | `tools` | List[str] | None | List of tool names available to the model |
 | `max_tokens` | int | `1024` | Maximum number of tokens to generate |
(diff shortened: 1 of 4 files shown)


wenjin272:
Thanks for working on this. #1039 was opened earlier for #1038 and already implements the same changes across the Java and Python integrations, tests, and documentation. The two PRs substantially overlap and also conflict in `chat_models.md`.

Let’s continue the review and remaining follow-ups in #1039 and close this PR as a duplicate. Thanks again!

PR #1071, opened 2026-08-30
[plan][runtime] Do not retry or persist interrupted chat calls on cancellation

### LINKED ISSUE #1070: [Bug][runtime] Do not retry or persist interrupted chat calls during job cancellation
### Search before asking

### Description

When a Flink job cancellation interrupts an in-flight chat-model call, the resulting `InterruptedException` is handled as an ordinary model failure on the current `main` branch.

There are two related problems:

1. `ChatModelInvoker.chatWithRetries()` catches every `Exception` and retries it when `ERROR_HANDLING_STRATEGY=RETRY`. Because throwing `InterruptedException` may clear the thread's interrupt status, retry backoff and additional model calls can continue after cancellation was requested, delaying task shutdown and issuing unnecessary external requests.
2. `RunnerContextImpl.durableExecuteCompletionOnly()` records every exception as a completed durable failure before rethrowing it. With an `ActionStateStore` configured, an interruption from the old execution attempt can therefore be replayed after recovery instead of re-executing the unfinished model call. Under `FAIL`, this can make the recovered action fail again; under `IGNORE`, the input can be skipped without producing a response; under `RETRY`, the replayed cancellation consumes retry budget.

Relevant paths on `main`:

- [`ChatModelInvoker.chatWithRetries()`](https://github.com/apache/flink-agents/blob/0417dad3177258e34af5d40507bae4d3f1e6b4bb/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelInvoker.java#L155-L203)
- [`RunnerContextImpl.durableExecuteCompletionOnly()`](https://github.com/apache/flink-agents/blob/0417dad3177258e34af5d40507bae4d3f1e6b4bb/runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java#L558-L588)
- [`ChatModelAction` IGNORE handling](https://github.com/apache/flink-agents/blob/0417dad3177258e34af5d40507bae4d3f1e6b4bb/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelAction.java#L482-L516)

Expected behavior:

- Cancellation signals should propagate immediately regardless of `FAIL`, `RETRY`, or `IGNORE` and should never trigger another model attempt.
- Cancellation should not be finalized as a durable success or failure. The call should remain unfinished, or its durable slot should be cleared, so recovery can re-execute or reconcile it.
- Ordinary provider and network failures should continue to follow the configured error-handling strategy.

#### Retry behavior

1. Register a test chat model whose `chat()` method blocks and throws `InterruptedException` when its execution thread is interrupted.
2. Configure `ERROR_HANDLING_STRATEGY=RETRY` with at least one retry.
3. Send a `ChatRequestEvent`, wait until the model call starts, and cancel/interrupt the executing task.
4. Observe that `chatWithRetries()` treats the interruption as a retryable model failure and may enter backoff or invoke the model again instead of propagating cancellation immediately.

#### Durable recovery behavior

1. Enable an `ActionStateStore` and execute the same blocking chat model through the built-in `ChatModelAction`.
2. Interrupt the model call and verify that the durable `chat` call is recorded as `FAILED` with `InterruptedException`.
3. Recover from the preceding checkpoint and reprocess the same event/action.
4. Observe that the stored interruption is replayed without invoking the model again. With `FAIL` the recovered action fails again; with `IGNORE` the request is skipped.

### Version and environment

- Apache Flink Agents `main` at `0417dad3177258e34af5d40507bae4d3f1e6b4bb`
- Java execution path
- The durable recovery case requires an `ActionStateStore` backend to be configured
- Provider-independent when the chat-model call surfaces cancellation as `InterruptedException`

Ashfaqbs:
Fixes #1070.

## Problem

When a job cancellation interrupts an in-flight chat-model call, the resulting `InterruptedException` was handled the same as any other model failure:

- `ChatModelInvoker.chatWithRetries()` caught it with the generic `catch (Exception e)` block and, under `ERROR_HANDLING_STRATEGY=RETRY`, could apply retry backoff and issue another model call after cancellation had already been requested — delaying task shutdown and firing unnecessary external requests.
- `RunnerContextImpl` (`durableExecuteCompletionOnly` and `executeAndFinalizeCurrentCall`) recorded the interruption as a completed durable success/failure before rethrowing it. With an `ActionStateStore` configured, that meant a stale interruption from a cancelled attempt could be replayed as a terminal result after recovery instead of the call being re-executed.

## Fix

Added an `InterruptedException`-specific catch ahead of the generic failure handling in both places:

- `ChatModelInvoker.chatWithRetries()` now restores the thread's interrupt status and rethrows immediately, skipping the retry/backoff branch entirely, regardless of `FAIL`, `RETRY`, or `IGNORE`.
- `RunnerContextImpl.durableExecuteCompletionOnly()` and `executeAndFinalizeCurrentCall()` now restore the interrupt status and rethrow without calling `recordDurableCompletion`/`finalizeCurrentCall`, so the durable slot is left unfinished (or pending) for recovery to re-execute or reconcile instead of replaying a stale interruption.

Ordinary provider/network failures are unaffected — they still go through the existing retry and durable-completion paths untouched.

## Testing

- Added `ChatModelInvokerTest` (new file) covering: an interrupted call is not retried (`durableExecute` invoked exactly once) and the interrupt status is restored; an ordinary `RuntimeException` still consumes the full retry budget, to confirm normal retry behavior is unchanged.
- Added two cases to `RunnerContextImplDurableExecuteTest` covering the legacy completion-only path and the pending-slot resume path: an interruption is not persisted as a durable result and the interrupt status is restored.
- Verified RED before the fix (`git stash` on the two source files, keeping the new tests) — all four new/updated cases failed as expected, then GREEN after restoring the fix.
- Full `plan` module suite: 288/290 passing; the 2 failures are pre-existing `BashToolTest` cases unrelated to this change (they require a working WSL bash on this Windows environment and fail the same way on `main` before this diff).
- Full `runtime` module suite: passing.
- `spotless:check` clean on both modules.

---

This fix follows from my own diagnosis of the bug. I used AI tooling for exploration and testing, directing it toward the change I had in mind.

### DIFF
--- a/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelInvoker.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelInvoker.java
@@ -182,6 +182,12 @@ final class ChatModelInvoker {
                 }
                 return new ChatAttemptResult(
                         model, chatModel, response, actualRetryCount, totalWaitTimeSec);
+            } catch (InterruptedException e) {
+                // A cancellation signal, not a model failure: restore the interrupt status and
+                // propagate immediately so task shutdown isn't delayed by retry backoff or an
+                // extra model call, regardless of the configured error-handling strategy.
+                Thread.currentThread().interrupt();
+                throw e;
             } catch (Exception e) {
                 if (strategy == Agent.ErrorHandlingStrategy.RETRY && attempt < numRetries) {
                     actualRetryCount = attempt + 1;
--- a/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelInvoker.java
+++ b/plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelInvoker.java
@@ -200,7 +200,12 @@ final class ChatModelInvoker {
                             numRetries,
                             currentWaitSec);
                     if (currentWaitSec > 0) {
-                        Thread.sleep(currentWaitSec * 1000L);
+                        try {
+                            Thread.sleep(currentWaitSec * 1000L);
+                        } catch (InterruptedException ie) {
+                            Thread.currentThread().interrupt();
+                            throw ie;
+                        }
                         totalWaitTimeSec += currentWaitSec;
                     }
                     continue;
--- a/plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelInvokerTest.java
+++ b/plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelInvokerTest.java
@@ -86,6 +86,49 @@ class ChatModelInvokerTest {
     }
 
     @Test
+    void testChatWithRetriesRestoresInterruptFlagFromRetryBackoffSleep() throws Exception {
+        RunnerContext ctx = mock(RunnerContext.class);
+        BaseChatModelSetup chatModel = mock(BaseChatModelSetup.class);
+        ReadableConfiguration config = mock(ReadableConfiguration.class);
+        when(ctx.getConfig()).thenReturn(config);
+        when(config.get(AgentExecutionOptions.CHAT_ASYNC)).thenReturn(false);
+        when(ctx.getResource("test-model", ResourceType.CHAT_MODEL)).thenReturn(chatModel);
+        when(ctx.getActionMetricGroup()).thenReturn(mock(FlinkAgentsMetricGroup.class));
+        // The interrupt fires from within the retry backoff's Thread.sleep, not from the call
+        // itself: set the flag first so Thread.sleep throws immediately, at no wall-clock cost.
+        when(ctx.durableExecute(any()))
+                .thenAnswer(
+                        invocation -> {
+                            Thread.currentThread().interrupt();
+                            throw new RuntimeException("transient failure");
+                        });
+
+        assertThrows(
+                InterruptedException.class,
+                () ->
+                        ChatModelInvoker.chatWithRetries(
+                                UUID.randomUUID(),
+                                "test-model",
+                                "durable-call-id",
+                                List.of(),
+                                Map.of(),
+                                null,
+                                ctx,
+                                Agent.ErrorHandlingStrategy.RETRY,
+                                1,
+                                1));
+
+        // Only the first attempt should have run: the sleep before the retry throws before a
+        // second call is made.
+        verify(ctx, times(1)).durableExecute(any());
+        // This is the only assertion that distinguishes the fix from the pre-fix code: the
+        // ChatAttemptFailed/times(1) shape above passes either way, since Thread.sleep still
+        // aborts the retry loop on both. Only the restored flag proves the backoff sleep's catch
+        // block restores it instead of leaving it cleared.
+        assertTrue(Thread.interrupted(), "interrupt status should be restored on the thread");
+    }
+
+    @Test
     void testChatWithRetriesRetriesOnOrdinaryFailure() throws Exception {
         RunnerContext ctx = mock(RunnerContext.class);
         BaseChatModelSetup chatModel = mock(BaseChatModelSetup.class);
--- a/plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelInvokerTest.java
+++ b/plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelInvokerTest.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.agents.plan.actions;
+
+import org.apache.flink.agents.api.agents.Agent;
+import org.apache.flink.agents.api.agents.AgentExecutionOptions;
+import org.apache.flink.agents.api.chat.model.BaseChatModelSetup;
+import org.apache.flink.agents.api.configuration.ReadableConfiguration;
+import org.apache.flink.agents.api.context.RunnerContext;
+import org.apache.flink.agents.api.metrics.FlinkAgentsMetricGroup;
+import org.apache.flink.agents.api.resource.ResourceType;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/** Tests for {@link ChatModelInvoker}. */
+class ChatModelInvokerTest {
+
+    @Test
+    void testChatWithRetriesDoesNotRetryOnInterruption() throws Exception {
+        RunnerContext ctx = mock(RunnerContext.class);
+        BaseChatModelSetup chatModel = mock(BaseChatModelSetup.class);
+        ReadableConfiguration config = mock(ReadableConfiguration.class);
+        when(ctx.getConfig()).thenReturn(config);
+        when(config.get(AgentExecutionOptions.CHAT_ASYNC)).thenReturn(false);
+        when(ctx.getResource("test-model", ResourceType.CHAT_MODEL)).thenReturn(chatModel);
+        when(ctx.getActionMetricGroup()).thenReturn(mock(FlinkAgentsMetricGroup.class));
+        when(ctx.durableExecute(any())).thenThrow(new InterruptedException("cancelled"));
+
+        // Clear any interrupt status left over from a previous test before asserting on it below.
+        Thread.interrupted();
+
+        assertThrows(
+                InterruptedException.class,
+                () ->
+                        ChatModelInvoker.chatWithRetries(
+                                UUID.randomUUID(),
+                                "test-model",
+                                "durable-call-id",
+                                List.of(),
+                                Map.of(),
+                                null,
+                                ctx,
+                                Agent.ErrorHandlingStrategy.RETRY,
+                                3,
+                                0));
+
+        // Only the first attempt should have run: retry backoff must not consume more attempts
+        // after a cancellation interrupts the call.
+        verify(ctx, times(1)).durableExecute(any());
+        assertTrue(Thread.interrupted(), "interrupt status should be restored on the thread");
+    }
+
+    @Test
+    void testChatWithRetriesRetriesOnOrdinaryFailure() throws Exception {
+        RunnerContext ctx = mock(RunnerContext.class);
+        BaseChatModelSetup chatModel = mock(BaseChatModelSetup.class);
+        ReadableConfiguration config = mock(ReadableConfiguration.class);
+        when(ctx.getConfig()).thenReturn(config);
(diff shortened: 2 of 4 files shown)


plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelInvoker.java
```diff
@@ -182,6 +182,12 @@ public ChatMessage call() throws Exception {
                 }
                 return new ChatAttemptResult(
                         model, chatModel, response, actualRetryCount, totalWaitTimeSec);
+            } catch (InterruptedException e) {
+                // A cancellation signal, not a model failure: restore the interrupt status and
+                // propagate immediately so task shutdown isn't delayed by retry backoff or an
+                // extra model call, regardless of the configured error-handling strategy.
+                Thread.currentThread().interrupt();
+                throw e;
             } catch (Exception e) {
```
weiqingy, line 191: `Thread.sleep` on line 203 sits inside this `catch (Exception e)` block, so a cancel during the backoff wait throws from in here rather than from the call above. A catch block is not covered by its own sibling catch, and `Thread.sleep` clears the interrupt status when it throws. The call still stops, so this is only about the flag, but on this one path it ends up cleared rather than restored. `RETRY_WAIT_INTERVAL` defaults to 1, so under RETRY there is a one second window on every retry.

Something like this, if useful:

```java
try {
    Thread.sleep(currentWaitSec * 1000L);
} catch (InterruptedException ie) {
    Thread.currentThread().interrupt();
    throw ie;
}
```

Worth restoring the flag there too?

plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelInvokerTest.java
```diff
@@ -0,0 +1,109 @@
...
+                                null,
+                                ctx,
+                                Agent.ErrorHandlingStrategy.RETRY,
+                                3,
+                                0));
+
+        // Only the first attempt should have run: retry backoff must not consume more attempts
+        // after a cancellation interrupts the call.
+        verify(ctx, times(1)).durableExecute(any());
+        assertTrue(Thread.interrupted(), "interrupt status should be restored on the thread");
```
weiqingy, line 85: nit: `Thread.interrupted()` on this line clears the flag, but it only runs if the `verify` above it passes. If that `verify` ever fails, the flag stays set on the JUnit thread. `ChatModelActionRetryTest.chatRetriesWithExponentialBackoff` is in this same package and drives a real one second backoff, so it can then fail with an unrelated `InterruptedException` and send someone chasing the wrong test. `RunnerContextImplDurableExecuteTest:91` and `:115` have the same shape. Would an `@AfterEach` calling `Thread.interrupted()` be worth adding, so the cleanup runs either way?

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -576,6 +576,12 @@ protected <T> T durableExecuteCompletionOnly(
         Exception exception = null;
         try {
             result = executionCallable.call();
+        } catch (InterruptedException e) {
+            // A cancellation signal, not a genuine call failure: leave the durable slot
+            // unfinished so recovery re-executes or reconciles the call instead of replaying a
+            // stale interruption as a completed success or failure.
+            Thread.currentThread().interrupt();
+            throw e;
```
weiqingy, line 584: `ModelRoutingResolver.java:97` runs the routing strategy through `ctx.durableExecute`, so an interrupt there reaches this rethrow. `ChatModelAction.java:609` then catches it, and under IGNORE line 621 returns normally. I could not trigger this with in-tree code. It needs a `MODEL_ROUTER` resource, and the only routing strategy in the repo today does no I/O. But `RoutingStrategy` is a user extension point, and the comment on line 612 already expects strategies that do I/O. The issue's first bullet asks for cancellation to propagate under IGNORE too. Should that catch let `InterruptedException` through before the IGNORE check?

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
```diff
@@ -939,6 +945,12 @@ protected <T> T executeAndFinalizeCurrentCall(
         Exception exception = null;
         try {
             result = callSupplier.call();
+        } catch (InterruptedException e) {
+            // A cancellation signal, not a genuine call failure: leave the pending call
+            // unfinalized so recovery re-executes or reconciles it instead of replaying a stale
+            // interruption as a completed success or failure.
+            Thread.currentThread().interrupt();
+            throw e;
```
weiqingy, line 953: Tool calls run through the two methods you patched, so this rethrow reaches them too, but the tool path still finishes normally after a cancel. `ToolCallAction.java:256` catches the `InterruptedException`, line 257 records it as a tool error, and the loop moves on to the next tool. Line 85 then sends the `ToolResponseEvent` anyway, which drives another chat call, and the action is persisted as finished. `executeParallel` has the same shape at line 209. That catch predates this PR. Is the tool path meant to be in scope here, or is it worth a separate issue?

weiqingy (commented):

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelInvoker.java
Ashfaqbs: Confirmed, good catch — the sleep sits in the sibling `catch (Exception e)` block, not the `InterruptedException` handler above it, so it wasn't covered by that fix. Wrapped it the same way (restore flag, rethrow) in c30ba894.

plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelInvokerTest.java
Ashfaqbs: Confirmed — assertTrue only consumes the flag when it passes, so a failing assertion here would leave it set on the JUnit thread. Added an @AfterEach in both this test class and RunnerContextImplDurableExecuteTest to clear it unconditionally, in c30ba894.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
Ashfaqbs: Traced it through — matches what you described. It's real but needs a RoutingStrategy that does I/O, which nothing in-tree does today, so it's about the extension point rather than this PR's own code. Given the issue is scoped to the chat/tool call paths, I'd rather keep this PR to what it already touches and file a follow-up for the routing-resolver + IGNORE interaction rather than pull it in here — let me know if you'd rather it go in this PR instead.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
Ashfaqbs: Confirmed, same shape in both executeSequentially and executeParallel, and it predates this PR. Tool-call cancellation handling looks like its own piece of work (the response still gets sent and the action still finishes), not a natural extension of the chat-retry fix here. I'd lean toward a separate issue/PR for it rather than scope-creeping this one — open to doing it here instead if you'd prefer to keep it together.

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelInvoker.java
weiqingy: The new block has no test. Both tests pass `retryWaitIntervalSec = 0` (lines 80 and 112), so the guard on line 202 skips lines 203 to 208.

A cheap deterministic test, in case it helps. If the stub sets the flag before it throws, `Thread.sleep` throws at once and costs no wall time:

```java
when(ctx.durableExecute(any()))
        .thenAnswer(
                inv -> {
                    Thread.currentThread().interrupt();
                    throw new RuntimeException("transient failure");
                });
```

with `RETRY`, `numRetries = 1`, `retryWaitIntervalSec = 1`.

One catch. `assertThrows` and the `times(1)` verify pass on the old code too, so only `assertTrue(Thread.interrupted())` proves the fix. Is a test here worth adding?

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
weiqingy: Agreed, leave it out of this PR. I would skip the follow-up issue too. #1042 already has the same fix. It adds `ModelRoutingResolver.isCancellation(...)` and calls it in `ChatModelAction.processChatRequest` just before the IGNORE check.

#1042 is also what makes this path reachable. `Strategies.llm(...)` runs a judge chat call through `chatWithRetries`, so the strategy that does I/O arrives with its own guard.

It is still open though, so this only holds if it lands as it stands. Does that look right to you?

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
weiqingy: Agreed, a separate issue is right. Nothing open covers tool-call cancellation, so it will not duplicate anything.

One thing that might be worth adding to it. Because the action returns normally, `ActionExecutionOperator` persists it as completed (line 491), and on recovery line 437 replays the output events without running the call again. That is problem 2 in #1070, carried by the action state rather than the durable slot. The chat path escapes it because the raw `InterruptedException` is rethrown on line 482, before the persist. Worth folding in?

plan/src/main/java/org/apache/flink/agents/plan/actions/ChatModelInvoker.java
Ashfaqbs: Added, in 5803bb1d. Set the interrupt flag from within the mocked `durableExecute` call itself so `Thread.sleep` throws immediately at no wall-clock cost, with RETRY / numRetries=1 / retryWaitIntervalSec=1 as you suggested. Verified it's the flag assertion alone that distinguishes the fix — temporarily restored the pre-c30ba894 sleep block and confirmed this new test fails there (times(1)/InterruptedException-thrown still pass either way) while the two existing tests stay green.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
Ashfaqbs: That reasoning matches what I see in #1042: `ModelRoutingResolver.isCancellation(...)` gets added and called from `ChatModelAction.processChatRequest` right before the IGNORE check, and `Strategies.llm(...)`'s judge chat call through `chatWithRetries` is what would make this path reachable in the first place. Agreed on skipping a separate follow-up issue, conditional on #1042 landing as-is — I'll keep an eye on it and flag here (or file the issue myself) if it merges in a materially different shape.

runtime/src/main/java/org/apache/flink/agents/runtime/context/RunnerContextImpl.java
Ashfaqbs: Good catch, worth folding in. Traced it: ActionExecutionOperator persists the action as completed at line ~491 (durableExecManager.maybePersistTaskResult) right after the tool-call catch swallows the interruption, and on recovery line ~437 (actionState.isCompleted()) skips re-execution and replays the stale output. Confirms it's the same class of problem as #1070's problem 2, just reached through the action-state path instead of the durable-slot path since the chat path's raw rethrow at line 482 escapes before the persist call. I'll fold this into the follow-up issue's description when I file it.

Ashfaqbs:
Filed the tool-call cancellation follow-up: #1088, including the persistence angle from the review thread.

plan/src/test/java/org/apache/flink/agents/plan/actions/ChatModelInvokerTest.java
```diff
@@ -0,0 +1,161 @@
...
+                                ctx,
+                                Agent.ErrorHandlingStrategy.RETRY,
+                                1,
+                                1));
+
+        // Only the first attempt should have run: the sleep before the retry throws before a
+        // second call is made.
+        verify(ctx, times(1)).durableExecute(any());
+        // This is the only assertion that distinguishes the fix from the pre-fix code: the
+        // ChatAttemptFailed/times(1) shape above passes either way, since Thread.sleep still
```
weiqingy: nit: the comment says `ChatAttemptFailed`, but this test asserts `InterruptedException` (line 106). The only `ChatAttemptFailed` assertion is line 143, in `testChatWithRetriesRetriesOnOrdinaryFailure`, so the sentence points a reader at the wrong test. `InterruptedException/times(1)` if that reads better.

weiqingy (approved):
Thanks for addressing the comments. One nit inline. LGTM.

Ashfaqbs:
Thanks for merging!

Issue #956, opened 2026-08-03
[Tech Debt][API][Tool] Align Java and Python Tool outcome contracts


joeyutong:
### Search before asking

- [x] I searched the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

## Background

Java and Python currently expose different Tool result contracts:

- Java `Tool.call` returns `ToolResponse`, which can explicitly represent success or failure without throwing an exception.
- Python `Tool.call` returns `Any`. An exception represents failure, while every normal return is treated as success because Python has no equivalent explicit Tool result type.

[PR #926](https://github.com/apache/flink-agents/pull/926) has introduced parallel Tool calls and the cross-language `Outcome` abstraction for durable batch execution. The resulting model has two distinct outcome layers:

```text
Outcome<T>   = whether the durable invocation returned or raised
ToolResponse = whether the Tool operation itself succeeded or failed
```

Java handles both layers for sequential and parallel calls: a successful `Outcome<ToolResponse>` may still contain an unsuccessful `ToolResponse`. Python can currently represent only the invocation layer, so every `Outcome.success(value)` is treated as a successful Tool operation.

This difference is now visible through several framework surfaces:

- `ToolResponseEvent.success` and `ToolResponseEvent.error`;
- Tool Execution Events and Agent Trace;
- Tool and MCP operational metrics proposed in [PR #955](https://github.com/apache/flink-agents/pull/955);
- Java/Python Tool bridges;
- MCP protocol-level error results.

There are also concrete built-in cases that cannot be represented consistently today. For example, Java and Python `load_skill` return normal success values when the Skill manager is unavailable, a Skill does not exist, or a requested Skill resource does not exist. These responses contain useful guidance for the model, but Trace and metrics classify them as successful Tool calls.

## Proposed direction

Define one Tool outcome contract across Java and Python while preserving compatibility for existing Python Tools:

- A normal raw Python return remains a successful Tool result.
- Both languages can explicitly return a successful or failed Tool result; Python exposes `ToolResponse.success(...)` and `ToolResponse.error(...)` for this purpose.
- An invocation exception remains a failed Tool result.
- Generic durable `Outcome` keeps its invocation-level meaning and does not absorb Tool-specific semantics.
- Sequential and parallel Tool-call paths normalize and record results through the same logic.
- Java/Python bridges preserve explicit Tool success, failure, and error details.
- `ToolResponseEvent`, Execution Events, and operational metrics consume the same normalized Tool outcome.
[...]

PR #1074, opened 2026-08-31
[API][Tool] Align Java and Python Tool outcome contracts

joeyutong:
Linked issue: #956

### Purpose of change

Java can distinguish whether a Tool invocation returned from whether the returned `ToolResponse` represents a successful Tool operation. Python previously exposed only raw return values, so every normal return was recorded as a successful Tool operation.

This change aligns the Java and Python Tool outcome contracts:

- adds Python `ToolResponse.success(...)` and `ToolResponse.error(...)` while preserving raw Python returns as successful results;
- normalizes sequential and parallel Python Tool calls through the same result-recording path;
- preserves explicit Tool outcomes across both Java-to-Python and Python-to-Java bridges without inspecting arbitrary user payloads;
- records missing Skill resources and MCP protocol error results as failed Tool operations while retaining their existing model-facing messages;
- keeps `Outcome<T>` focused on whether the durable invocation returned or raised.

`ToolResponseEvent`, Execution Events, and downstream metrics can therefore consume the same normalized Tool outcome in both languages. The existing `ToolResponseEvent.responses` wire representation is unchanged.

### Tests

- `mvn -pl api,plan,runtime,integrations/mcp -am -DskipITs test`
- `pytest` for the affected Python Tool API, Tool call, MCP, Skill, and Java bridge suites (66 passed)
- Ruff format and lint checks for all changed Python files

### API

Adds the public Python `flink_agents.api.tools.ToolResponse` type. Existing Python Tools that return raw values remain source- and behavior-compatible.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Codex 0.151.0-alpha.7.2 (GPT-5.6 Sol)


python/flink_agents/integrations/mcp/mcp.py
```diff
@@ -65,12 +65,20 @@ def tool_type(cls) -> ToolType:
...
+            )
 
-        return asyncio.run(
-            self.mcp_server.call_tool_async(self.metadata.name, *args, **kwargs)
-        )
+        try:
+            return asyncio.run(
+                self.mcp_server.call_tool_async(self.metadata.name, *args, **kwargs)
+            )
+        except Exception as e:
```
wenjin272, line 77: Could we also normalize explicit outcomes in `PythonMCPTool.call()`? That Java wrapper still wraps every normal Python return in `ToolResponse.success(...)`. Returning `ToolResponse.error(...)` here therefore causes Java agents using Python MCP tools to record failures as successes. This reproduces with real Pemja for both transport exceptions and MCP protocol errors. Please cover this path with a cross-language regression test.

python/flink_agents/integrations/mcp/mcp.py
```diff
@@ -237,6 +249,10 @@ async def call_tool_async(self, tool_name: str, *args: Any, **kwargs: Any) -> An
 
             content = [extract_mcp_content_item(item) for item in result.content]
 
+            if result.isError:
```
wenjin272: Could we move the protocol-error check outside `async with self._get_session()`? With a real `ClientSession`, raising here gets wrapped in an AnyIO `ExceptionGroup`. `MCPTool.call()` then retains only the outer message, “unhandled errors in a TaskGroup”, losing the original MCP error content and recovery hints. A regression test retaining the real session lifecycle would catch this; the current fake session does not.

wenjin272 (commented):
Thanks for working on aligning the Java and Python tool outcome contracts!

python/flink_agents/integrations/mcp/mcp.py
joeyutong: Fixed. `PythonMCPTool` now uses the same bridge encoder/decoder as `FunctionTool`, so an explicit Python `ToolResponse.error(...)` remains a failed Java `ToolResponse`. I also added a unit regression and extended `MCPCrossLanguageTest` with a real FastMCP protocol error over Pemja.

python/flink_agents/integrations/mcp/mcp.py
joeyutong: Fixed by moving MCP content/error interpretation outside the session context. The regression now uses a real `ClientSession` lifecycle and verifies that the original protocol error content is retained instead of only the outer `ExceptionGroup` message.

wenjin272 (approved):

PR #1090, opened 2026-09-02
[integrations][python] Apply DashScope native structured output

weiqingy:
Linked issue: #280

### Purpose of change

Passing an `output_schema` to the Tongyi connection raised `NotImplementedError`, because the connection had no native translation. DashScope does support schema-enforced structured output, so this adds it.

The catch is that it applies to fewer models than you might expect. DashScope documents `json_schema` for five model families, but four of them are served on its multimodal interface, while this connection calls `Generation.call` on the text-generation interface. Calling one through the other fails with `url error`. The Qwen3.7-Max family is served on the text interface, except its `qwen3.7-max-2026-06-08` snapshot, which is multimodal-only as well. That leaves four models where native enforcement is reachable here.

So the connection reports the capability for an exact list of those four models (`qwen3.7-max`, `qwen3.7-max-preview`, `qwen3.7-max-2026-05-17`, `qwen3.7-max-2026-05-20`), and for everything else, including the `qwen-plus` default, it sends nothing new and keeps the existing prompt-based fallback. No behavior changes for anyone not on one of those four.

Sources for the two halves of that intersection, both worth checking against the allowlist in the diff:
- [structured output](https://help.aliyun.com/zh/model-studio/json-mode) for which models support `json_schema`
- [text generation](https://help.aliyun.com/zh/model-studio/text-generation) for which models are text-interface

Heads-up on the docs: the DashScope API reference pages still list only `text` and `json_object` for `response_format` and never mention `json_schema`. The feature guide above is the current source, and the error reference corroborates it by naming `json_schema` as a recognised key.

### Tests

New file `test_tongyi_native_structured_output.py`, 18 cases, no API key or network needed. It asserts the request the SDK would actually receive: the parameter is present and correctly shaped on a capable model, absent on every other model, absent when no schema is passed, and a non-Pydantic schema falls back instead of failing. The capability check is tested against all four capable names, incapable names including the multimodal-only `qwen3.7-max-2026-06-08` snapshot, empty string and `None`.

Each assertion was verified by mutating the source and confirming the intended test fails, so the tests are not passing by accident.

**One limitation, stated plainly.** No vendor source shows `json_schema` on `Generation.call`. It is documented as a parameter of the native API, with the placement rule for HTTP callers, but the only native sample uses `MultiModalConversation.call` with a model this connection excludes. Every test here is a mock and I had no DashScope key, so the suite cannot settle it. What is established: the parameter provably reaches the request body, and if the provider rejects it the call fails loudly with a `RuntimeError` rather than silently returning an unconstrained answer. **If a reviewer has a key, one live call against `qwen3.7-max` would close this.**

### API

No public API change. `supports_native_structured_output` is an existing hook this connection now overrides, and no dependency changes. Python only, since there is no Java Tongyi connection.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

- [x] Yes
- [ ] No

### DIFF
--- a/python/flink_agents/integrations/chat_models/tongyi_chat_model.py
+++ b/python/flink_agents/integrations/chat_models/tongyi_chat_model.py
@@ -22,9 +22,10 @@ import uuid
 from typing import Any, Dict, List, Sequence, cast
 
 from dashscope import Generation
-from pydantic import Field
+from pydantic import BaseModel, Field
+from typing_extensions import override
 
-from flink_agents.api.agents.types import OutputSchema
+from flink_agents.api.agents.types import OutputSchema, render_output_schema
 from flink_agents.api.chat_message import ChatMessage, MessageRole
 from flink_agents.api.chat_models.chat_model import (
     BaseChatModelConnection,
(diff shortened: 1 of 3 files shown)


weiqingy:
The one failing check is unrelated to this PR:

- **it-java [ubuntu-latest] [java-21] [flink-2.2]**: `429 Too Many Requests` from Maven Central while resolving the Apache parent POM `org.apache:apache:pom:34`. It failed in the "Install flink-agents Java" step after 42 seconds, so nothing was compiled and no test ran. [Job log](https://github.com/apache/flink-agents/actions/runs/33671037333/job/100384423887).

This PR only changes Python and docs. It touches no `.java` and no `pom.xml`, and the other four `it-java` jobs in the same run passed on the same commit.

I have re-triggered the failed job.

python/flink_agents/integrations/chat_models/tongyi_chat_model.py
```diff
@@ -35,6 +36,73 @@
...
+# snapshots behind it, so a name matches the prefix itself or a name continuing it
+# after a "-" separator. That expresses the documented unit instead of a snapshot
+# census that goes stale, and it keeps out a different family that merely extends
+# the prefix, such as qwen3.7-maximum. The one snapshot the rule admits without
+# json_schema reaching it, qwen3.7-max-2026-06-08, is multimodal-routed and answers
+# this connection with "url error" whether or not a response_format rides along.
+#
+# A name outside the rule reports not-capable and degrades to the prompt-engineering
+# fallback rather than failing at the provider.
+_NATIVE_STRUCTURED_OUTPUT_ALIAS_PREFIXES = ("qwen3.7-max",)
```
wenjin272, line 59: `qwen3.7-max-2026-06-08` is incorrectly classified as supported by this prefix match. `TongyiChatModel` invokes `dashscope.Generation.call`, but the [Alibaba Cloud documentation](https://www.alibabacloud.com/help/en/model-studio/text-generation#dashscope-h4) states that this model uses the multimodal API and that directly substituting it into a text-generation request causes an error. Please replace the prefix match with an exact allowlist containing `qwen3.7-max`, `qwen3.7-max-preview`, `qwen3.7-max-2026-05-17`, and `qwen3.7-max-2026-05-20`.

wenjin272 (commented):
Thanks for taking this on. Only one comment.

wenjin272 (approved):

Issue #1072, opened 2026-08-30
[Bug][skills] Bound download and extraction size for URL skill sources


weiqingy:
### Description

`Skills.from_url(...)` / `Skills.fromUrl(...)` download a remote zip and extract it without bounding how many bytes get written, in either runtime. A URL skill source can therefore fill the TaskManager's temp filesystem. That is a resource exhaustion property rather than a transport or integrity one, so it is tracked apart from #1003.

The download streams the response straight to a temp file with no cap, and without consulting `Content-Length`: `shutil.copyfileobj(resp, out)` at `python/flink_agents/runtime/skill/repository/_materialize.py:162`, and `Files.copy(in, tmpZip, StandardCopyOption.REPLACE_EXISTING)` at `runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java:263`. A source that streams without ending, or simply serves a very large body, runs until the disk is full.

Extraction is unbounded too. `extract_zip_safely` validates every entry against zip-slip and checks nothing else, then extracts: `zf.extractall(extract_dir)` at `_materialize.py:133`, and the per-entry `Files.copy(in, target)` at `SkillMaterializer.java:142`. There is no per-entry size cap, no cumulative uncompressed cap and no entry count cap, so a small archive can expand well past its transfer size.

Digest pinning does not close this. In the change proposed on #1005 the SHA-256 is computed after the download helper returns, so the file is already fully on disk before any operator expectation is checked. The Java side has the same ordering.

This came out of reviewing #1005 and was raised on #1003 first: https://github.com/apache/flink-agents/issues/1003#issuecomment-5383724087

### Expected behavior

- Bound the bytes written during download, failing closed with an error that names the limit.
- Bound extraction on cumulative uncompressed size, per-entry size and entry count, checked before writing past the limit rather than after.
- Where the server declares a `Content-Length` above the cap, reject before streaming. That is a cheap early exit, not a substitute for counting bytes as they arrive, since the header can be absent or wrong.
- Keep the existing partial-output cleanup on the failure path, so a rejected download or extraction leaves nothing behind.
- Keep the limits and any configuration surface aligned across Java and Python.
- Add focused Java and Python tests for a download past the cap, an archive that expands past the cap, and unchanged behavior for an archive inside the limits.

### How to reproduce

1. Serve a response at an HTTPS URL that streams without ending, or whose body is far larger than any real skill archive.
[...]

PR #1091, opened 2026-09-02
[runtime][java][skills] Bound download and extraction size

pranavshuklaa:
Linked issue: #1072

### Note on branch base

This branch is built on top of #1005 (commit 5eb91e73), which is still open.
Until #1005 merges, this PR's diff will include #1005's commits in addition
to mine. My changes are isolated to the most recent commit on this branch 
SkillMaterializer.java and its tests. I'll rebase onto main once #1005 lands,
at which point the diff will show only these changes. 

### Purpose of change

This change adds resource bounds for skill archives materialized from URL
sources.

A skill archive downloaded from a remote URL can otherwise consume an
unbounded amount of disk space during download and ZIP extraction, a
malicious or compromised source could exhaust disk space via an oversized
download or a zip bomb. This PR adds explicit limits around both stages and
ensures the limits are enforced using the actual data being processed, not
just metadata the source controls.

The proposed limits are:
* **512 MiB** maximum downloaded archive size
* **200 MiB** maximum size of a single extracted entry
* **1 GiB** maximum total extracted size
* **10,000** maximum number of extracted entries

These values are proposed for discussion, we can adjust based on sense of realistic skill archive sizes.

#### Download protection

The download path now:

* Checks the declared `Content-Length` when available and rejects archives
  already declared larger than the download limit, before reading any body
  bytes.
* Independently tracks the number of bytes actually read from the response
  stream and rejects the download before writing bytes that would cross the
  limit.
* Treats the streamed byte count as authoritative, so the protection holds
  even when `Content-Length` is missing, zero, or understated by the server.
* Deletes the temporary file immediately if the download fails or is
  rejected, rather than relying solely on the JVM shutdown hook.

This closes the gap where a server could bypass a `Content-Length`-only
check simply by omitting or lying about the header, declared size is only
ever used as a cheap early exit, never as the actual enforcement.

#### ZIP extraction protection

The extraction path applies layered validation:

* Rejects archives with more than the configured maximum number of entries,
  before any bytes are extracted.
* Checks declared per-entry and cumulative uncompressed sizes from the ZIP
  central directory as an early, cheap rejection where that metadata is
  present.
* Independently tracks actual decompressed bytes written per entry and
  cumulatively across the archive, rejecting extraction before writing bytes
  that would cross either limit.
* Preserves the existing zip-slip (path traversal) validation, run before
  any extraction begins.
* Deletes the extraction directory immediately on any failure, in addition
  to the existing JVM shutdown hook fallback.

As with the download path, declared ZIP metadata is attacker-controlled and
is only used as an early exit ,the actual decompressed byte count during
extraction is the real enforcement. This is deliberate: a crafted archive
can declare an entry as 1 byte while its actual DEFLATE stream expands to
gigabytes (a zip bomb), so trusting `ZipEntry.getSize()` alone would not be
sufficient.

#### Smoke test

`SkillMaterializerSmokeTest` is included alongside the unit test coverage.
It exercises the same core scenarios as the unit tests but with verbose
console output, useful for manually verifying the enforcement and cleanup
behavior end-to-end against the real HTTP and ZIP code paths. Happy to
remove it before merge as you'd rather keep only the unit tests 
flagging it now since it's additive rather than required.

### Tests

Added unit tests in `SkillMaterializerTest` covering:

* Download size limit via declared `Content-Length`
* Download size limit via actual streamed byte count, including understated
  and missing `Content-Length`
* Download failure cleanup (temp file removed on rejection)
* ZIP entry count limit
* Per-entry declared size limit (metadata pre-check)
* Per-entry actual decompressed size limit, using a fixture with forged ZIP
  metadata so the declared size passes but the actual DEFLATE stream exceeds
  the limit, proving the byte counter, not the metadata check, is doing the
  enforcement
* Cumulative extraction size limit
* Extraction failure cleanup (temp directory removed on rejection)
* Zip-slip rejection cleanup (confirms cleanup covers the pre-existing path
  traversal check too, not just the new size checks)
* Happy-path extraction unaffected by the new checks

`SkillMaterializerSmokeTest` covers the same core scenarios with readable
console output for manual verification.

Verified locally:

* `SkillMaterializerTest` — all tests passing
* `SkillMaterializerSmokeTest` — all tests passing

### API

No

### Documentation

- [ ] `doc-needed` 
- [x] `doc-not-needed` 
- [ ] `doc-included` 

- [x] Yes
- [ ] No

`Generated-by: Generated-by: ChatGPT (GPT-5.6 Luna)`


runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
```diff
@@ -58,6 +61,27 @@ public final class SkillMaterializer {
 
     private static final int JAR_URL_PREFIX_LEN = "jar:".length();
 
+    // --- Size caps for download and extraction (issue #1072) ---
+
+    /** Maximum number of bytes accepted from a single HTTP download. */
+    public static final long MAX_DOWNLOAD_BYTES = 512L * 1024 * 1024; // 512 MiB
```
rob-9, line 67: these limits currently only exist on Java side. #1072 requires aligned Java and Python enforcement, but Python still uses unbounded `shutil.copyfileobj(...)` and `ZipFile.extractall(...)`. could you add the corresponding Python limits and regression tests so `Skills.from_url(...)` does not retain the original vulnerability?

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
```diff
@@ -119,36 +143,154 @@ private static Thread registerCleanup(Path path) {
...
-        // Register cleanup before validation so the empty tempdir is always reclaimed,
-        // even if validation raises.
+
+        // Register the fallback cleanup hook before any work so the empty dir is always reclaimed,
+        // even if validation or extraction raises.
         Thread hook = registerCleanup(extractDir);
+
+        try {
+            extractZipSafelyInto(zipPath, extractDir);
+        } catch (IOException e) {
```
rob-9, line 176: the path here deletes the directory but leaves `hook` registered, so every rejected archive retains a shutdown-hook thread until JVM exit.

it also only cataches `IOException`: for example, an entry containing NUL makes `Path.resolve(...)` throw `InvalidPathException`, bypassing this cleanup and leaving the temp directory until shutdown.

could we create the `Materialized` handle before extraction and call `close()` for both `IOException` and `RuntimeException`?

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
```diff
@@ -119,36 +143,154 @@ private static Thread registerCleanup(Path path) {
...
+    /**
+     * Core extraction logic: validates, checks bounds, then extracts. Separated from {@link
+     * #extractZipSafely} so the caller can handle cleanup on failure.
+     */
+    private static void extractZipSafelyInto(Path zipPath, Path extractDir) throws IOException {
         try (ZipFile zf = new ZipFile(zipPath.toFile())) {
-            Enumeration<? extends ZipEntry> entries = zf.entries();
-            while (entries.hasMoreElements()) {
-                ZipEntry entry = entries.nextElement();
+            List<? extends ZipEntry> entries = Collections.list(zf.entries());
```
rob-9, line 202: the entry-count check at line 202 happens only after `Collections.list(...)` has enumerated and retained every entry. 

memory consumption before rejection therefore remains proportional to the attacker-controlled entry count rather than the 10,000-entry limit. could we check `zf.size()` before constructing the list, or stop during bounded enumeration, so no more than the permitted entries are materialized?

runtime/src/test/java/org/apache/flink/agents/runtime/skill/SkillMaterializerTest.java
```diff
@@ -173,18 +177,6 @@ void malformedUrlDoesNotLeakRawInput() {
         assertTrue(ex.getCause() == null);
     }
 
-    @Test
-    void rejectsScopedIpv6BeforeConnection() {
```
rob-9: was this deletion intentional? this direct `downloadToTempFile(...)` regression test was added in `5eb91e73` following the scoped-IPv6 downloader issue raised in #1005. 

the surviving tests cover validation through the API and `URLSkillRepository`, but none directly preserve this downloader contract. could we retain the test since its removal is unrelated to the resource-bound change?

runtime/src/test/java/org/apache/flink/agents/runtime/skill/SkillMaterializerTest.java
```diff
@@ -592,4 +584,612 @@ private List<String> getMessages() {
...
+                                                    && Files.isDirectory(p))
+                            .count();
+        }
+
+        assertEquals(
+                before, after, "failed extraction must not leave a temporary directory behind");
+    }
+
+    @Test
+    void rejectsCumulativeBytesOverTotalCap(@TempDir Path tempDir) throws IOException {
```
rob-9, line 1137: this fixture's declared uncompressed total already exceeds the limit, so it is rejected by the metadata precheck before extraction begins.

the assertion confirms that path by expecting `total uncompressed size`; the authoritative `totalWritten` guard instead reports `total extracted size`. could the fixture understate its declared sizes, or use injectable test limits, so the test actually exercises cumulative decompressed-byte enforcement?

rob-9 (changes requested):

rob-9:
with code style check, I believe there's an issue with `SkillMaterializerSmokeTest.java` missing a header.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
pranavshuklaa: yes thanks for pointing that out. materialized is now constructed before extraction runs, using the already-registered shutdown hook. The extraction block now calls materialized.close() for both IOException and RuntimeException before re-throwing. Since close() removes the shutdown hook and deletes the extraction directory, malformed entries such as one containing a NUL character can no longer leave either the hook or temporary directory behind.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
pranavshuklaa: changed this. now checking zf.size() before calling Collections.list The entry count is read directly from the archive's central directory, so an archive exceeding the configured entry-count limit is rejected before constructing the ZipEntry list. This avoids making memory usage before rejection proportional to an attacker controlled number of entries.

runtime/src/test/java/org/apache/flink/agents/runtime/skill/SkillMaterializerTest.java
pranavshuklaa: I deleted that accidently, The rebase ate this one. Restored it. Thanks for pointing it out

runtime/src/test/java/org/apache/flink/agents/runtime/skill/SkillMaterializerTest.java
pranavshuklaa: Fixed this. the test now forges each entry's declared uncompressed size down to 1 byte via forgeDeclaredSizesForAllEntries() before extraction, so the metadata precheck passes and the real cumulative decompressed-byte counter (totalWritten) is what rejects the archive. I also updated the assertion to check for "total extracted size", matching the message from that guard.

pranavshuklaa:
> with code style check, I believe there's an issue with `SkillMaterializerSmokeTest.java` missing a header.

Yes, regarding this file, I have removed it completely. I had done the smoketest to get human-readable manual verification during development, but its scenarios are now fully covered by the regression tests in SkillMaterializerTest, so it was redundant.

runtime/src/test/java/org/apache/flink/agents/runtime/skill/SkillMaterializerTest.java
```diff
@@ -592,4 +598,670 @@ private List<String> getMessages() {
...
+                            "http://127.0.0.1:" + port + "/skill.zip", 5_000, true);
+            try {
+                assertEquals(1024, Files.size(file));
+            } finally {
+                Files.deleteIfExists(file);
+            }
+        } finally {
+            server.stop(0);
+        }
+    }
```
rob-9, line 777: Could this test use a smaller limit, say 1 KiB, and check that 1,024 bytes succeeds but 1,025 fails? Right now it downloads 1 KiB against the 512 MiB limit, so it doesn't reach the boundary.

python/flink_agents/runtime/skill/tests/test_materialize.py
```diff
@@ -186,6 +191,273 @@ def test_raises_on_http_error(
...
+        # zipfile, unlike declaring an honest (accurate) total that simply
+        # exceeds the cap, which is covered separately by
+        # test_rejects_declared_entry_size_over_cap.
+        actual_size = 200
+        zip_path = tmp_path / "cumulative.zip"
+        entries = {f"entry-{i}.bin": b"B" * actual_size for i in range(6)}
+        _make_zip_with_patched_declared_sizes(zip_path, entries, 1)
+
+        with pytest.raises(zipfile.BadZipFile, match="Bad CRC-32"):
+            extract_zip_safely(zip_path)
```
rob-9, line 439: This test fails on the first entry's CRC, so it never checks the total-size limit. Could you use a valid ZIP with, say, two 600- byte files and set the total limit to 1 KiB for the test? Each file fits, but together they should exceed the limit.

rob-9 (commented):
thanks; just a couple fixes in the tests.

rob-9 (approved):

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
```diff
@@ -58,6 +61,27 @@ public final class SkillMaterializer {
...
+    public static final long MAX_EXTRACT_ENTRY_BYTES = 200L * 1024 * 1024; // 200 MiB
+
+    /**
+     * Maximum cumulative uncompressed bytes written across all entries during a single zip
+     * extraction. Enforced against actual bytes written.
+     */
+    public static final long MAX_EXTRACT_TOTAL_BYTES = 1024L * 1024 * 1024; // 1 GiB
+
+    /** Maximum number of entries permitted in a single zip archive. */
+    public static final int MAX_EXTRACT_ENTRIES = 10_000;
```
wenjin272, line 83: These limits are currently hard-coded, but realistic skill archive sizes and TaskManager temporary-disk budgets can vary significantly between deployments. Could we expose them through a single configuration object, with aligned Java, Python, and YAML semantics? The values should remain positive and should not support an unlimited setting.

The proposed defaults also seem quite permissive for a skill archive: a 512 MiB download, 200 MiB single entry, 1 GiB extraction, or 10,000 entries can still consume substantial disk space and inodes, especially when multiple skills are materialized concurrently. Unless we have usage data supporting these values, could we start with more conservative defaults—for example, 64 MiB download, 64 MiB per entry, 256 MiB total extraction, and 1,000 entries—while allowing deployments to raise them explicitly?

runtime/src/test/java/org/apache/flink/agents/runtime/skill/SkillMaterializerTest.java
wenjin272: Agreed. More broadly, could we pass the configured limits into the internal materializer methods so the tests can use small thresholds? `acceptsBodyExactlyAtDownloadCap` currently sends only 1 KiB against the 512 MiB production limit, while the other Java and Python tests either perform multi-GiB I/O or fail at ZIP CRC validation before reaching the cumulative byte counter. Small test limits would let us verify `limit` versus `limit + 1`, distinguish understated and missing `Content-Length`, and exercise the actual per-entry and cumulative byte counters directly without the resource cost.

wenjin272 (commented):
Thanks @pranavshuklaa for working on this resource-exhaustion hardening and for aligning the Java and Python paths. The core bounded-streaming and extraction approach looks sound. I left two comments about making the limits configurable with more conservative defaults, and using injectable small limits so the boundary and byte-counter tests are exact and lightweight.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
```diff
@@ -58,6 +61,104 @@ public final class SkillMaterializer {
...
+            this.maxExtractEntries = maxExtractEntries;
+        }
+    }
+
+    /**
+     * Backward-compatible alias for {@link Limits#DEFAULT#maxDownloadBytes}.
+     *
+     * @deprecated Use {@link Limits#DEFAULT} or inject a {@link Limits} instance.
+     */
+    public static final long MAX_DOWNLOAD_BYTES = Limits.DEFAULT.maxDownloadBytes;
```
wenjin272, line 139: Thanks for lowering the defaults and grouping the limits into one object. However, `Limits` / `MaterializerLimits` currently only provides an injection point for tests: all production repository paths still use the defaults, while the actual configuration wiring is deferred by the TODO. I think this configuration surface needs to be completed in this PR, with the four positive limits exposed consistently through the Java, Python, and YAML APIs and propagated through the real materialization path. Also, the `MAX_*` aliases were introduced in this still-unmerged PR, so there is no released API compatibility to preserve. Could we remove these already-deprecated public constants and use the configured limits object directly? This would also eliminate the current compiler warnings caused by Javadoc `@deprecated` tags without corresponding `@Deprecated` annotations.

python/flink_agents/runtime/skill/repository/_materialize.py
```diff
@@ -160,37 +218,179 @@ def copy_dir_to_temp(src_dir: Path) -> Materialized:
...
+        f.read(4)  # uncompressed size
+        fn_len = struct.unpack("<H", f.read(2))[0]
+        ex_len = struct.unpack("<H", f.read(2))[0]
+        f.read(fn_len + ex_len)  # filename + extra field
+        compressed = f.read(member.compress_size)
+
+        if member.compress_type == zipfile.ZIP_STORED:
+            return io.BytesIO(compressed)
+        if member.compress_type == zipfile.ZIP_DEFLATED:
+            return io.BytesIO(zlib.decompress(compressed, -15))
```
wenjin272, line 309: `zlib.decompress()` expands the entire entry in memory before the size checks run, so a zip bomb can cause OOM before being rejected. It also bypasses CRC validation. Could we keep `zf.open(member)` for streaming extraction and test the byte counter through a separate bounded-copy helper?

wenjin272 (commented):
Could you please rebase this branch onto the latest `main` instead of merging `main` into the feature branch? Keeping the PR history linear avoids an unnecessary merge commit and makes the final change set easier to review and maintain. Please force-push the rebased branch afterward.

wenjin272 (changes requested):

runtime/src/test/java/org/apache/flink/agents/runtime/skill/SkillMaterializerTest.java
pranavshuklaa: I had pushed the changes and missed on replying to these, sorry for that. To update the Limits object is now injectable into all materializer methods, so tests use a 1 KiB cap instead of the 512 MiB production default.
acceptsBodyExactlyAtDownloadCap now sends exactly 1,024 bytes against a 1,024-byte cap and asserts success. rejectsBodyOneByteOverDownloadCap sends 1,025 bytes against the same cap with no Content-Length header, so only the streaming counter can reject it. The same injectable limits are used across all download and extraction cap tests,no test allocates more than a few KiB of data.

python/flink_agents/runtime/skill/tests/test_materialize.py
pranavshuklaa: Thanks for the check i have added a fix for this. Now the test is renamed `test_rejects_cumulative_bytes_over_total_cap`, which uses a plain (unpatched) ZIP with two 200-byte entries and a total cap of 300 bytes. Each entry fits individually under the per-entry cap (250 bytes), so the per-entry check passes. The declared total (400bytes) exceeds the cap, so the declared-total pre-check fires and rejects the archive. The assertion matches on "total", which is present in both the pre-check message ("declared total uncompressed size ... limit") and the streaming counter message ("total extracted size ... limit"), so the test is valid regardless of which path fires first.

runtime/src/main/java/org/apache/flink/agents/runtime/skill/repository/SkillMaterializer.java
pranavshuklaa: That makes sense, thanks for the detailed suggestion! the four limits are now exposed in AgentConfigOptions (Java) and AgentConfigOptions in core_options.py, with YAML keys skill.source.url.max-download-bytes, skill.source.url.max-extract-entry-bytes, skill.source.url.max-extract-total-bytes, and skill.source.url.max-extract-entries. All four values must be strictly positive there is no unlimited setting. The new options are documented in docs/content/docs/operations/configuration.md under a new "URL Skill Source Options" section. Removed the MAX_* constants entirely. 
Limits.fromConfig(ReadableConfiguration) / limits_from_config(config) build a Limits / MaterializerLimits from the fourConfigOption values, falling back to the conservative defaults when a key is absent or config is null. The limits are now threaded through the full production path in both runtimes: `ActionExecutionOperator `→ `ResourceCache `→ `ResourceContextImpl `→ `SkillManager `→ `SkillSourceRegistry `→ `URLSkillRepository `/ `FileSystemSkillRepository `→ Limits.fromConfig(config). Tests for fromConfig are added in both `SkillMaterializerTest.java` (limitsFromConfigReadsAllFourKeys, limitsFromConfigNullReturnsDefault) and `test_materialize.py` (TestLimitsFromConfig). Pls suggest if there is an alternative which would be better.

python/flink_agents/runtime/skill/repository/_materialize.py
pranavshuklaa: Fixed this. _open_entry_stream is deleted.  _extract_zip_to_dir now uses zf.open(member) for streaming decompression, reading in 64 KiB chunks. Size checks run per chunk before any bytes are written to disk. BadZipFile
(bad CRC-32 or truncated data) is now being caught at the outer try level and re-raised as ValueError("... failed integrity check ..."), so the archive is still rejected via CRC validation rather than the byte counter. When a size limit fires mid-read, src.close() is called inside a try/except BadZipFile block before re-raising the original limit error, suppressing the spurious CRC error that would otherwise fire on an incomplete read.

PR #1092, opened 2026-09-03
[runtime][java] Make embedded Python interpreters thread-confined

### LINKED ISSUE #1087: [Bug] Concurrent access to a shared PythonInterpreter can crash the JVM in async cross-language calls
### Search before asking

### Description

The embedded `pemja.core.PythonInterpreter` owned by an operator is shared by cross-language action and resource paths. When multiple threads concurrently enter the same interpreter through Pemja APIs such as `invoke`, the JVM can terminate with a native `SIGSEGV` instead of returning a managed Java or Python exception.

This is broader than the Mem0 path where the problem was first observed. Any asynchronous cross-language path that can concurrently access the same interpreter is potentially affected. The participating threads may originate from:

- Flink's Java async executor;
- the mailbox or lifecycle thread;
- Python-created workers that call into Java and then re-enter the same interpreter through a Java adapter; or
- another JVM thread carrying a cross-language callback.

This does not mean that every asynchronous operation is unsafe. A Python worker that only runs Python code, a callback that only executes Java code without touching the interpreter, sequential cross-language calls, and calls using different interpreter instances are outside the demonstrated failure condition.

The failure was first observed in the cross-language Mem0 long-term-memory E2E test in this CI job:

https://github.com/apache/flink-agents/actions/runs/33615851952/job/100201365522

The original failing path was:

```text
Java durable async worker
  -> Java MemorySet.add
  -> Java Mem0LongTermMemory wrapper
  -> shared Pemja PythonInterpreter.invoke
  -> Python Mem0LongTermMemory.add
  -> Mem0 chat / embedding / vector-store processing
```

A sanitized excerpt from a local reproduction is:

```text
# A fatal error has been detected by the Java Runtime Environment:
# SIGSEGV (0xb)
# JRE version: Java(TM) SE Runtime Environment (21.0.8+12)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (bsd-aarch64)

Current thread: JavaThread "flink-agents-java-async-pool-5-thread-2" [_thread_in_native]

C  [Python]  _PyFrame_Clear
C  [Python]  _PyEval_EvalFrameDefault
C  [pemja_core.cpython-311-darwin.so]  JcpPyObject_Call
C  [pemja_core.cpython-311-darwin.so]  Java_pemja_core_PythonInterpreter_invoke
j  pemja.core.PythonInterpreter.invoke(...)
j  org.apache.flink.agents.runtime.python.utils.PythonResourceAdapterImpl.callMethod(...)
```

### Root-cause isolation

The failure was isolated from Mem0 with a ChatModel cross-language E2E test:

1. Configure `AgentExecutionOptions.NUM_ASYNC_THREADS` to `2`.
2. Send two events with different keys so that two slow chat requests overlap.
3. Use a Java agent calling a Python ChatModel setup through the shared interpreter.

Both connection variants are affected:

- Java -> Python setup -> Java connection; and
- Java -> Python setup -> Python connection.

In the Python-connection variant, the JVM crashed before the Python connection wrapper's `chat()` entry log was emitted. The fatal stack was already inside `PythonResourceAdapterImpl.callMethod -> PythonInterpreter.invoke`, which excludes the connection implementation, Mem0's internal executor, and Elasticsearch as required causes.

Expected behavior: asynchronous cross-language execution completes safely, or an unsupported concurrent access pattern fails with a managed exception without terminating the JVM.

Actual behavior: concurrent access to the shared interpreter can terminate the entire JVM in CPython evaluation through Pemja.

#### Isolated ChatModel reproduction

1. Build Flink Agents and install its Python dependencies.
2. Start Ollama with a chat model such as `qwen3:1.7b`.
3. Configure the cross-language ChatModel test with two async threads and two different input keys.
4. Use a Python ChatModel setup backed by either a Java or Python connection.
5. Make the two requests slow enough to overlap and run `ChatModelCrossLanguageTest`.

When the requests overlap, the Surefire JVM may abort before both requests complete and produce an `hs_err_pid*.log`.

#### Original Mem0 reproduction

1. Start Ollama with the `nomic-embed-text` model.
2. Start Elasticsearch and export `ES_HOST=http://localhost:9200`.
3. Export a valid `ACTION_API_KEY` and, if needed, `ACTION_BASE_URL` for the OpenAI-compatible chat model.
4. Run:

```bash
mvn -pl e2e-test/flink-agents-end-to-end-tests-resource-cross-language \
  -Dtest=Mem0LongTermMemoryTest test
```

The test sends interleaved facts for two keys. `Mem0LongTermMemoryAgent.addItems()` invokes `memorySet.add()` using `durableExecuteAsync()`. One of the overlapping additions may abort the Maven Surefire JVM and produce an `hs_err_pid*.log`.

### Version and environment

- Flink Agents: `0.4-SNAPSHOT`
- Flink: `2.3.0`
- Pemja: `0.5.7`
- CI: `ubuntu-latest`, Temurin JDK 21, Python 3.12
- Local reproduction: macOS aarch64, Oracle JDK 21.0.8, Python 3.11

wenjin272:
Linked issue: Closes #1087

### Purpose of change

Concurrent cross-language calls no longer enter one embedded `PythonInterpreter` from multiple threads, avoiding the native JVM crashes observed in async ChatModel and Mem0 paths without globally serializing calls.

#### Runtime flow

1. Each Python-enabled operator subtask creates an owner interpreter and a `PythonInterpreterManager`.
2. Every Java-to-Python action, resource, adapter, and conversion call goes through that manager.
3. The mailbox owner and managed Java async workers execute inline on their own thread-local interpreters. Other callers use a stable lane in a bounded callback-worker pool.
4. Multi-step operations retaining Python objects complete within one interpreter operation.
5. Operator shutdown stops the managed async workers first. Each managed or callback worker closes its interpreter on the owning thread, then the owner interpreter and Python environment are closed.

#### Key decisions

- Use hybrid routing instead of a global lock because a global lock would serialize all bridge calls and negate `num-async-threads`. Managed Java workers can safely own an interpreter, while creating one on a CPython-created thread can attach a second Python thread state to that native thread; only the latter calls are forwarded.
- Give each unmanaged source thread a stable callback lane so related opaque Python handles stay on one interpreter. Lane count reuses `num-async-threads`; threads and interpreters remain lazy.
- Reuse Python resource instances across thread-confined interpreters rather than initializing one resource graph per worker.
- Keep the fix inside Flink Agents; no Pemja or Flink release change is required.

#### Interaction decisions

| Python bridge present | Calling context | Result |
|---|---|---|
| No | Any | No interpreter manager or callback worker is created. |
| Yes | Mailbox/owner thread | Execute inline with the owner interpreter. |
| Yes | Managed Java async worker | Lazily create and reuse that worker's interpreter; execute inline. |
| Yes | Python-created or other unmanaged thread | Submit synchronously to that caller's stable callback lane. |
| Yes | Callback worker re-entering Java-to-Python | Reuse its bound interpreter inline; do not submit recursively. |

#### Behavioral contracts

1. Bridge calls on different execution threads do not concurrently use one `PythonInterpreter`.
2. Independent managed-worker and callback-lane calls can overlap; the fix does not globally serialize Python bridge traffic.
3. Python-to-Java-to-Python callbacks use bounded interpreter-owning workers and preserve caller-to-lane affinity.
4. Python object conversion, invocation, result retention, and nested Java-to-Python callbacks stay on the selected interpreter.
5. Interpreters are closed on their owning worker threads, accepted work completes before teardown, and calls after manager closure are rejected.
6. Plans without Python actions, Python resources, or Mem0 retain the existing no-Python-runtime path.

#### Failure behavior

- A callback operation or interpreter initialization failure is propagated; this layer does not retry or fall back to sharing another interpreter. A partially initialized interpreter is closed before the failure escapes.
- An interrupted caller finishes an accepted callback before its interrupt status is restored, preventing premature native-state teardown.
- Closing the manager from a non-owner thread, configuring a non-positive callback-worker count, or invoking after close raises a managed exception.
- Shutdown attempts all interpreter cleanup and reports later cleanup failures as suppressed exceptions. External chat, embedding, and vector-store failures keep their existing behavior.

### Tests

| Contract | Tests |
|---|---|
| 1. Per-thread isolation | Manager isolation test; Java concurrent ChatModel E2E |
| 2. Calls remain concurrent | Manager overlap test; Java and Python concurrent ChatModel E2Es |
| 3. Bounded, affine callbacks | Unmanaged routing, bounded-reuse, and callback-reentry tests; Python E2E |
| 4. Same-interpreter conversion | Python action and Java chat-message adapter tests |
| 5. Owning-thread lifecycle | Managed/callback close, post-close rejection, and thread-factory tests |
| 6. No-Python path stays lazy | Python bridge no-op test |

Coverage targets the native-crash risks: concurrent interpreter entry, Python-created-thread re-entry, opaque-object affinity, and shutdown. The prerequisite-gated Mem0 E2E was also verified manually with its external dependencies available.

Not verified: none identified for the changed behavior.

<details>

- The owner thread is captured when the manager is constructed; only that thread may close the manager.
- Managed-worker identity is scoped to `AsyncExecutorThreadFactory` execution and removed after its exit cleanup.
- Callback lanes are single-thread executors. Executor termination is followed by an actual worker-thread join so interpreter cleanup finishes before Python runtime teardown.
- The lifecycle read lock covers accepted operations; close first excludes new work, then drains callback executors and aggregates cleanup failures.
- Each newly created interpreter imports the same bridge modules before use.

</details>

### API

No public API or configuration key is added. Existing callers require no source changes. `num-async-threads` now also bounds the per-subtask callback lanes; unused lanes create neither threads nor interpreters. Java-only and no-Python plans retain their existing execution behavior, apart from stricter async-worker shutdown waiting.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Codex 0.144.5 (GPT-5)


joeyutong (commented):
[`deleteCollection()`](https://github.com/apache/flink-agents/blob/2db48c356bcdf08d35617ab3bde4d35451c8f812/api/src/main/java/org/apache/flink/agents/api/vectorstores/python/PythonCollectionManageableVectorStore.java#L73) still calls `PyObject.invokeMethod()` directly, bypassing the interpreter manager. Could we use `adapter.callMethod()` here, as `createCollectionIfNotExists()` does, to preserve thread confinement?

joeyutong (approved):
Thanks for the updates! All my concerns have been addressed. Overall LGTM.

PR #1094, opened 2026-09-03
[runtime] Cover the cross-loader action key failure and document action renaming

purushah:
Linked issue: N/A (follow-up to the action-key fix that landed through #1100)

### Purpose of change

Durable action-state keys used to derive the action segment from `Action.hashCode()`. For Java actions that hash folds in `JavaFunction`'s `Class<?>[] parameterTypes`, and `Class.hashCode()` is a per-JVM identity hash, so every key changed on restart and recovery never replayed completed actions. The fix, deriving the segment from the plan-unique `action.getName()`, is already on `main` through #1100, which was stacked on this branch.

This PR carries what remains: a regression test that reproduces the restart failure, removal of a test that could not detect it, and two clarifications to the upgrade guidance #1100 added.

#### Behavioral contracts

1. Two `Action`s with the same name whose `JavaFunction` parameter types are distinct `Class` objects, loaded by different class loaders, produce the same state key.
2. Recovery reuses completed results for an action only while its name is unchanged; renaming an action makes its recorded results unreachable and it executes again on the next recovery.

#### Failure behavior

Nothing in this change raises or changes a runtime path. The test skips, rather than fails, in the rare case that the two `Class` objects happen to receive equal identity hashes, because there is then nothing to discriminate.

### Tests

| Contract | Test |
|---|---|
| 1, key stable across class loaders | `ActionStateUtilCrossClassLoaderTest`, new |
| 2, renaming semantics | documented; pinned indirectly by `ActionStateUtilTest.testActionUUIDSegmentDerivesFromActionName`, already on `main` |

`ActionStateUtilCrossClassLoaderTest` loads an `Event` subclass through two isolating class loaders, which yields two distinct `Class` objects with different identity hashes, the same effect a JVM restart has on `JavaFunction.hashCode()`. The loader redefines only the subclass and delegates everything else to the parent, so `Action`'s signature check against `Event` still passes. Verified against `main` with the action segment temporarily switched back to `Action.hashCode()`: the test fails on the action-UUID segment; with `main` as is, it passes.

`ActionStateUtilTest.testSameActionNameYieldsSameKeyAcrossInstances` is removed. Both instances share the same `Class` objects inside one JVM, so it passed before the fix too.

The earlier cross-JVM test that launched two `java` processes is not carried over. As noted in review, separate JVMs do not guarantee different identity hashes and the test passes on JDK 17 with the old derivation restored, so it added CI time without discriminating the bug.

`JAVA_HOME=<jdk17> mvn test -pl runtime -am -Dtest='ActionStateUtil*'`: 29 tests, 0 failures. `mvn spotless:check` under JDK 17 passes.

### API

No public API or runtime change.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

Two edits to the upgrade hint under Exactly-Once Action Consistency in `deployment.md`: the incompatible key-format change is dated to Flink Agents 0.4, and a sentence states that an action's recorded results survive a code change only while its name is unchanged.

- [x] Yes
- [ ] No

Generated-by: Claude Code (Claude Fable 5.1)


runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
```diff
@@ -222,8 +222,11 @@ private static String generateUUIDForEvent(Event event) throws IOException {
...
 
     private static String generateUUIDForAction(Action action) throws IOException {
+        // Action.hashCode() folds in JavaFunction's Class[] parameterTypes, and Class.hashCode()
+        // is the per-JVM identity hash — so the hash-derived UUID changes on every process
+        // restart and recovery lookups can never hit. Derive from the plan-unique action name,
+        // which is stable across restarts.
         return String.valueOf(
-                UUID.nameUUIDFromBytes(
-                        String.valueOf(action.hashCode()).getBytes(StandardCharsets.UTF_8)));
+                UUID.nameUUIDFromBytes(action.getName().getBytes(StandardCharsets.UTF_8)));
```
rob-9, line 230: Question about compatibility: Python actions previously used a hash of stable strings, so their identifiers should've already been stable across restarts. 

If a Python action saved a completed result before this change, I believe an upgraded job would end up looking for it under the new identifier and run the action again. 

The name-based identifier makes sense, but I think we should clarify this upgrade case instead of saying no working deployment regresses, and explain how users should handle existing action state.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtilTest.java
```diff
@@ -116,6 +118,43 @@ public void testGenerateKeyRejectsNonPositiveMaxParallelism() throws Exception {
...
+    @Test
+    public void testSameActionNameYieldsSameKeyAcrossInstances() throws Exception {
+        InputEvent event = new InputEvent("test-input");
+        String first =
+                ActionStateUtil.generateKey(
+                        "test-key", 7, new NoOpAction("stable-name"), event, MAX_PARALLELISM);
+        String second =
+                ActionStateUtil.generateKey(
+                        "test-key", 7, new NoOpAction("stable-name"), event, MAX_PARALLELISM);
+        assertEquals(first, second);
```
rob-9: Both actions here are created in the same JVM and share the same classes, so this test would also pass before the fix. 

Could we add a test using separate JVMs or classloaders to exercise the original failure? The preceding test already catches reverting the identifier calculation; this would add coverage of the restart behavior described.

rob-9 (changes requested):

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
purushah: Good catch, thanks. `PythonFunction` hashes `module` and `qualName`, so Python action keys were already stable and this change does alter them. I'll fix the compatibility note: a Python action in flight at the restoring checkpoint re-executes once after upgrading. To avoid that, drain in-flight work before the stop-with-savepoint, or start the upgraded job with a fresh action-state topic/table. No fallback to the old key, per the beta breaking-change policy. I'll add a note to `deployment.md`.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtilTest.java
purushah: Agreed, that test passes before the fix too. I'll replace it with a test that loads an `Event` subclass through two isolating class loaders, so the parameter `Class` objects get different identity hashes, which is the same effect a JVM restart has on `JavaFunction.hashCode()`. Verified it fails on the pre-fix code and passes with the fix.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
rob-9: Ok! A fresh topic/table would still allow completed actions to repeat if the job restores from an older checkpoint, since their saved results are unavailable.

purushah: Right, a fresh topic/table doesn't help; it just makes the old results unreachable the other way. I'll drop that from the note. The guidance will be: upgrade from a stop-with-savepoint taken after the job has gone idle with no pending actions, and don't restore the upgraded job from an older checkpoint. Restoring from an older checkpoint re-executes, once, the Python actions completed since that checkpoint.

purushah: Great catch — you're right, and my compatibility note overclaimed. PythonFunction hashes stable strings (module, qualName), so Python-action keys were already stable across restarts and this change does invalidate their existing state. Will correct the note: Java action state was unrecoverable across restarts (the bug — no regression possible); Python action state was recoverable, and the first recovery after upgrading re-executes those actions once and re-persists under the new format. I'll spell that out as the upgrade guidance. If you'd prefer a legacy-key fallback read for Python actions during a deprecation window I'm happy to add one, though given the project's pre-1.0 status a release note may be the better trade.

purushah: Agreed — a fresh topic only avoids stale entries; restore from an older checkpoint re-executes either way since the saved results are unreadable under the new format. Will fold that into the note.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtilTest.java
purushah: Right — one JVM shares the same Class objects, so that test can't discriminate. Will add a cross-JVM test that launches two separate java processes, prints the generated key for the same action in each, and asserts they match — it fails on the pre-fix code and keeps the in-JVM tests as fast pins.

docs/content/docs/operations/deployment.md
```diff
@@ -104,6 +104,10 @@ The same persisted action state is also used by fine-grained durable execution.
 
 See [Action State Store Configuration]({{< ref "docs/operations/configuration#action-state-store" >}}) for configuration options.
 
+{{< hint warning >}}
+**Upgrading**: Action state keys are derived from the action name since Flink Agents 0.4. Action state recorded by earlier versions is not consulted after the upgrade. Upgrade from a stop-with-savepoint taken after the job is idle with no pending actions, and do not restore the upgraded job from an older checkpoint; otherwise actions completed since that checkpoint are executed once more.
```
rob-9: could we also explain what this means for future action changes? keeping the name means recovery can reuse results from the prev implementation, while renaming it makes those results unreachable and can cause re-execution. 

the release upgrade guidance looks good; these rules would help users update their own agents too.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/CrossJvmKeyStabilityTest.java
```diff
@@ -0,0 +1,94 @@
...
+                    "STATE_KEY="
+                            + ActionStateUtil.generateKey("cross-jvm-key", 5, action, event, 8));
+        }
+    }
+
+    @Test
+    void sameActionYieldsSameKeyAcrossSeparateJvms() throws Exception {
+        String first = keyFromFreshJvm();
+        String second = keyFromFreshJvm();
+        assertEquals(
```
rob-9: nit : with the old hash calculation restored, this test still passes on a JDK 17 setup. separate JVMs don't guarantee different identity hashes. 

the class-loader test catches the old bug, so my earlier request is covered. but could we soften the PR description's claim that this test fails before the fix?

rob-9 (commented):
thanks! these address my earlier comments.

everything looks good. I've added two non-blocking suggestions below about documenting future action changes and qualifying the separate-JVM test claim.

wenjin272:
Thanks for working on this, and apologies—I mistakenly merged #1100 before reviewing and merging this PR, even though it was based on your changes. I preserved your `Co-authored-by` attribution in the squash commit.

The core action-name-based identifier change is already on `main` through #1100. Could you rebase this PR onto the latest `main` and keep the remaining test updates, adapting them to the new key encoder API? We can retain the upgrade guidance already added by #1100 and clarify that the incompatible state-format change starts with Flink Agents 0.4, rather than adding a separate upgrade note.

docs/content/docs/operations/deployment.md
purushah: Added to the hint: keeping an action's name across a code change lets recovery reuse its completed results; renaming it makes those results unreachable, so it runs again on the next recovery.

runtime/src/test/java/org/apache/flink/agents/runtime/actionstate/CrossJvmKeyStabilityTest.java
purushah: Thanks for checking that. You're right that separate JVMs don't guarantee different identity hashes, so rather than soften the claim I dropped the cross-JVM test in the rebase. The class-loader test covers the restart failure and does fail with the old derivation restored, so it stays.

purushah:
Rebased onto `main`. The action-name derivation is dropped from this PR since #1100 carries it, thanks for preserving the attribution. What remains: the class-loader regression test adapted to `ActionStateKeyEncoder` via `ActionStateTestUtils`, removal of the same-JVM test that could not detect the bug, and two edits to the upgrade hint from #1100: the format change is dated to 0.4, and a sentence covers action renaming. The cross-JVM test is not carried over, see the thread on it.

wenjin272 (approved):
Thanks for the rebasing. LGTM

PR #1096, opened 2026-09-04
[integrations][gemini] Cover buildConfig with unit tests

weiqingy:
Linked issue: #280

### Purpose of change

`GeminiChatModelConnection.buildConfig` assembles the request configuration for every Gemini call: the system instruction, sampling parameters, forwarded extra arguments, and tool declarations. It had no test coverage at all, because it was private and nothing could reach it.

This makes it package-private, following the convention five other members of the same class already use, and covers it.

The gap that mattered most: nothing verified that `buildConfig` actually forwards extra arguments to the code that applies them. The existing tests for that code call it directly, so the connection between the two was never exercised.

While adding the tests, three things turned out to be executed by the new tests but checked by none of them. They are now checked: the description and parameter schema on a tool declaration, and the guard that keeps a user's message from being sent as part of the system instruction.

Groundwork for native structured output on Gemini, which follows separately.

### Tests

Four unit tests added to `GeminiChatModelConnectionTest`. Module suite goes from 25 to 29, no failures. They run offline and need no API key.

Each test was checked by breaking the production code it covers and confirming the test fails. Each break is caught by exactly one test, so a failure points at one cause.

### API

No public API change. `buildConfig` moves from private to package-private, which widens it only within its own package. No signature, behaviour, or dependency change.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

### DIFF
--- a/integrations/chat-models/gemini/src/main/java/org/apache/flink/agents/integrations/chatmodels/gemini/GeminiChatModelConnection.java
+++ b/integrations/chat-models/gemini/src/main/java/org/apache/flink/agents/integrations/chatmodels/gemini/GeminiChatModelConnection.java
@@ -251,7 +251,8 @@ public class GeminiChatModelConnection extends BaseChatModelConnection {
         return map;
     }
 
-    private GenerateContentConfig buildConfig(
+    // Package-visible for unit testing of the request-config assembly.
+    GenerateContentConfig buildConfig(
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):
Thanks for catching this. LGTM

PR #1097, opened 2026-09-04
[integrations][bedrock] Apply Bedrock native structured output

weiqingy:
Linked issue: #280

### Purpose of change

Bedrock was the last Java chat model with no native structured-output path. Asking it for a schema did not fall back quietly, it threw, because the connection never overrode the foundation's hook. This wires Converse's `outputConfig` so the provider constrains the response, with the existing prompt-engineered path still handling everything else.

Bedrock has no Python connection in this repo, so this is Java-only.

The feature needed an SDK bump. `outputConfig` appears on `ConverseRequest` at `bedrockruntime` 2.41.22 and does not exist at 2.41.21, so the pin moves to the first version that has it rather than to the latest. That property is shared with three other modules, so the bump moves them too. They compile and test unchanged, and `NOTICE` follows the resulting dependency set.

It is three commits: a refactor that adds a seam for testing the request, the bump on its own, then the feature. Each is meant to be readable without the other two.

Two decisions are worth a reviewer's attention.

**Which models get the native path.** Support on Bedrock is documented per model, not per family, so the check is an exact match against the ids AWS lists, retried once after stripping a leading inference-profile segment such as `us.` or `eu.`. A family-prefix match would be wrong in a way that matters: AWS documents one Qwen model as unsupported while six of its siblings are supported. Anything unrecognised takes the prompt fallback, and that includes ARNs, which carry no model information at all, and prompt routers, which do not pick a model until the request runs. Guessing wrong in that direction costs a fallback; guessing wrong in the other costs a runtime error.

**How the schema is built.** The AWS SDK has no schema generator, unlike the OpenAI and Anthropic ones, so the schema is derived locally the way the Ollama connection does it. A couple of constraints there are easy to get wrong and are covered by tests: without Jackson awareness the generator ignores `@JsonProperty` names, emits `@JsonIgnore` fields and lists enum constants by their Java names rather than the values Jackson reads back, and without a required check it marks nothing required, which would let an empty document satisfy any schema.

Two limitations to be aware of rather than surprised by. Recursive types cannot be used as output schemas, since the generated self-reference is something Bedrock rejects up front. And Bedrock warns that the first request against a new schema can spend up to a few minutes compiling it, which here happens inside an operator, so the first record through a new schema may be far slower than the rest.

### Tests

The Bedrock chat-model module goes from 12 tests to 84, all offline.

Most of that is coverage this module simply never had. Nothing previously asserted what the connection actually sends, so tools, system messages, inference config and message merging are all tested for the first time. The rest covers the new behaviour: which model ids are accepted and rejected, that the derived schema keeps Jackson property names and enum values and marks fields required, and that the native path engages and disengages when it should. Tools and a schema on the same request are also covered, since Bedrock allows both.

The three other modules affected by the SDK bump run unchanged, and the full build was run so the packaging step is exercised, since a dependency move is exactly what can break it.

What is not covered: none of this calls Bedrock. Those tests need credentials CI does not have, so the request shape is verified and the service's acceptance of it is not. I am verifying that separately against a real endpoint and will report the results here before this is merged. The open questions are whether `outputConfig` is accepted as built, which schema details Bedrock tolerates, and whether an unsupported model errors or quietly ignores the request.

### API

No public API is added or changed, and no new configuration. The visible difference is that a Bedrock connection given an output schema on a supported model now returns a schema-constrained response instead of throwing.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

Nothing user-facing changes in configuration, and the repo has no structured-output docs page for any provider today.

- [x] Yes
- [ ] No


integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
```diff
@@ -173,19 +326,73 @@ public ChatMessage chat(
...
+    //
+    // A self-referencing class derives its own field as a reference back to the schema root,
+    // whatever the required check says. Bedrock does not accept a recursive schema and rejects the
+    // request before the model runs, so declaring the field Optional does not rescue it; only
+    // flattening the recursion does.
+    private static JsonNode toNativeSchema(Class<?> schemaClass) {
+        SchemaGeneratorConfigBuilder configBuilder =
+                new SchemaGeneratorConfigBuilder(
+                                SchemaVersion.DRAFT_2020_12, OptionPreset.PLAIN_JSON)
+                        .with(new JacksonModule());
```
wenjin272, line 391: Could we align the generated schema with Jackson's enum wire values? The bare `JacksonModule` emits Java enum names instead of their `@JsonProperty` or `@JsonValue` values—for example, `IN_PROGRESS` instead of `in-progress`—so a schema-valid response can fail during subsequent `ObjectMapper` deserialization. Ollama and #1098 use similar schema-generation logic, so this may also be a good opportunity to extract the shared configuration into a `chat-models/common` component. Would you prefer to introduce that abstraction here or handle the cross-provider migration separately?

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
```diff
@@ -88,6 +103,51 @@
...
+    // and the version in one string, so there is no alias for a prefix to cover, and a prefix would
+    // over-capture: "qwen.qwen3" admits qwen.qwen3-vl-235b-a22b, which AWS documents as not
+    // supported, and "anthropic.claude-sonnet-4" admits anthropic.claude-sonnet-4-20250514-v1:0,
+    // whose card carries no answer at all. Exact matching also keeps irregular id shapes correct
+    // with no normalisation rule: mistral.mistral-large-3-675b-instruct carries no version suffix,
+    // openai.gpt-oss-120b-1:0 carries "-1:0" rather than "-v1:0".
+    //
+    // A card whose capability table carries the bullet in neither column is undocumented rather
+    // than negative, and is absent from this set for that reason.
+    private static final Set<String> NATIVE_STRUCTURED_OUTPUT_MODELS =
```
wenjin272, line 127: AWS now documents Structured outputs support for Claude Opus 4.6 and lists `anthropic.claude-opus-4-6-v1` as its Bedrock model ID, but it is absent from this allowlist. Consequently, both the base ID and its `us.`, `eu.`, `au.`, and `global.` inference-profile forms are classified as incapable and fall back to prompting. Could we add the base ID and cover the direct and prefixed forms in the capability tests? Reference: [AWS Claude Opus 4.6 model card](https://docs.aws.amazon.com/en_en/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-6.html).

wenjin272 (commented):
Thanks for adding native structured-output support for Bedrock and for documenting the provider-specific constraints. I left two inline comments on schema generation and the current model capability list.

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
weiqingy: Good catch, thanks. Fixed in 8d763889 by enabling `FLATTENED_ENUMS_FROM_JSONPROPERTY` and `FLATTENED_ENUMS_FROM_JSONVALUE`. The new `testDerivedSchemaFollowsJacksonEnumValues` reads every listed value back with a plain `ObjectMapper`, for one `@JsonProperty` enum and one `@JsonValue` enum. #1098 has the same fix in d233bead, and I'll add it to #1117 too.

For the shared config, I'd lean toward a separate PR once these land. There is no shared chat-model module yet, and the recipes differ a bit per provider (map values, closed objects, `$ref` handling). Adding it here would also tie #1098 and #1117 to this PR. The follow-up would move Ollama, Gemini, Bedrock and watsonx onto one config, and fix the same enum gap in Ollama on main. Does that plan work for you?

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
weiqingy: Thanks for the pointer. Added in 3f82ff90. While checking, I went through all the Bedrock model cards and found 28 more models that document structured outputs on `bedrock-runtime`, including Claude Sonnet 4.6. I added those too, so the list has 41 ids now.

`capableModels()` covers the direct ids, and the prefix test now runs `us.`, `eu.`, `apac.`, `au.`, `jp.` and `global.` against `anthropic.claude-opus-4-6-v1`.

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
```diff
@@ -173,19 +356,78 @@ public ChatMessage chat(
...
+    //   - The required check marks every field required except an Optional one. The default marks
+    //     nothing required, which lets a model omit fields at will, while marking everything
+    //     required would force the fields a caller declared omissible.
+    //   - The Jackson module makes the schema name properties the way Jackson names them, and
+    //     list enum constants mapped by @JsonProperty or by a @JsonValue method under the values
+    //     Jackson reads. The response is read back into the same class with an ObjectMapper, so a
+    //     property that @JsonProperty renames or @JsonIgnore drops, and a mapped enum constant,
+    //     have to appear in the schema as the mapper reads them, or a response that satisfies the
+    //     schema still fails to deserialize. An enum annotating only some constants falls back to
+    //     Java names for all of them, so its annotated constants do not read back. The two enum
```
wenjin272, line 407: One user-visible limitation remains: for a partially annotated enum, victools falls back to Java names for the whole enum, so the generated schema may differ from the values Jackson accepts. Since the existing structured-output integrations do not yet have user-facing documentation either, this does not need to block this PR. Could we document it consistently once the structured-output work across providers is complete, noting that `@JsonProperty` should be applied to every enum constant?

wenjin272 (commented):
Thanks for addressing the previous feedback and expanding the model and schema coverage. 

There is only one question regarding the document, but I believe we can address it as a group at the end.

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
weiqingy: Yes, that works. Once the remaining provider PRs land, I'll add a structured output section to the chat model docs that covers every provider, including that an enum used in an output schema needs `@JsonProperty` on every constant.

integrations/chat-models/bedrock/src/main/java/org/apache/flink/agents/integrations/chatmodels/bedrock/BedrockChatModelConnection.java
weiqingy: #1117 was merged before the enum fix reached it, so the watsonx part is in #1120 now.

Issue #1034, opened 2026-08-21
[Feature] Add checkpoint-aligned cleanup for Kafka action state


rob-9:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

The Kafka action-state store records action progress so that a recovering job can avoid running completed actions again. Each Flink checkpoint saves a recovery marker with the next Kafka offset to read for every action-state partition. During recovery, `rebuildState()` finds the earliest required offset for each partition and replays the log from there.

PR #885 adds opt-in tombstones, which are null-valued Kafka records written for state pruned after a checkpoint completes. Replay treats a tombstone as deletion of that key. On a compacted action-state topic, Kafka may later remove the tombstone and the values it replaced, reclaiming storage and reducing long-term topic growth. Compaction happens asynchronously, so tombstones do not impose a fixed size limit.

The recovery trade-off is that an older checkpoint may replay both a completed action-state value and a later tombstone. The tombstone removes the recovered completion state, so the action may run again. The checkpoint that triggered a prune remains usable with respect to that prune, but later pruning may invalidate it once it becomes an older recovery point.

Safe physical cleanup therefore needs a clearly defined oldest supported recovery point. The operator and action-state store know the marker saved in an individual checkpoint, but they do not have a complete view of every retained checkpoint, externalized checkpoint, or savepoint that users may still want to restore.

For example, suppose C0 needs partition 0 from offset 100 and C1 needs it from offset 150. If the user chooses C1 as the oldest supported recovery point, Kafka may discard records before offset 150. C1 and newer recovery points must still work. An attempt to restore C0 should fail with a clear error instead of silently rebuilding incomplete action state.

### User interaction model

- The first version could let users explicitly choose and advance the oldest supported recovery point. The boundary can only move forward.
- A later policy could move it automatically, for example by retaining the latest N checkpoints. Users would need to register or pin external savepoints they want the policy to protect.
- Once a boundary or policy is configured, cleanup and retries can run without asking for confirmation each time. Without one, coordinated prefix cleanup stays off. The existing opt-in tombstones keep their current recovery trade-off.

### Possible design

1. Keep a durable mapping from each globally completed checkpoint to its effective Kafka recovery marker. For each partition, use the earliest offset required by any subtask so recovery remains correct after rescaling.
[...]

PR #1100, opened 2026-09-04
[runtime] Preserve typed Flink key identity in action state

### LINKED ISSUE #1099: [Bug] Preserve Flink key identity in ActionState
### Search before asking

### Description

`ActionStateUtil.generateKey()` persists the key-group computed from the original typed Flink key, but represents the logical business key using `key.toString()`. The matching helpers compare only that textual value.

Distinct Flink keys can therefore be treated as the same business key during lookup and pruning. Examples include `Long(1)` and `String("1")`, or distinct custom keys with the same `toString()` result. When the key-group also matches, their complete state keys can be identical.

### Impact

- When distinct keys with the same string form are held by one store/subtask, Kafka and Fluss lookup cleanup can evict the other key's cached action state.
- Kafka pruning can emit tombstones for another key's exact records when both keys are held by the same store/subtask.
- Evicting recovered completed state can make the operator miss it on the next lookup, re-execute the action, and repeat external side effects.
- With `maxParallelism = 1`, distinct keys with the same string form can produce identical complete state keys, allowing one key's state to overwrite or replay another key's state.

PR #1024 fixed recovery ownership by persisting the key-group computed from the typed key, but the business-key segment still stores only `key.toString()`, which does not preserve the key's type or uniquely identify distinct keys. When its option is enabled, PR #885 exposes the remaining problem more severely because tombstones make accidental deletion durable.

PR #1094 fixes a different part of the state key: the action UUID was derived from an `Action.hashCode()` that can change after a JVM restart. It does not change the `key.toString()` business-key identity or the `matchesBusinessKey*` helpers described here.

### Proposed direction

Persist the key's serialized bytes, or a collision-resistant digest of them, and use that value consistently for key generation, lookup, pruning, and Kafka partitioning. Using the operator's Flink key serializer would make the durable identity follow the same typed representation used by keyed state. Comparing the key-group in addition to `toString()` would fix only cases where the keys belong to different groups; distinct keys can still share a key-group.

Define one explicit versioned format and do not interpret old records by matching `key.toString()`. If records written in the old format cannot be mapped to a Flink key unambiguously, recovery should fail with a clear compatibility error rather than guessing.

### Acceptance criteria

- `Long(1)` and `String("1")` remain isolated with `maxParallelism = 128` and `maxParallelism = 1`.
- Distinct custom keys with the same `toString()` remain isolated.
- Key identity remains stable across serialization and task recovery.
- Kafka and Fluss lookup, divergence cleanup, and pruning use the same serialized key identity.
- Pruning one key never emits a tombstone for another key.
- Regression tests cover Kafka, Fluss, and the operator recovery path.

### How to reproduce

Using the existing `NoOpAction` and `InputEvent` test fixtures:

```java
Action action = new NoOpAction("test-action");
InputEvent event = new InputEvent("test-input");

String numericKey = ActionStateUtil.generateKey(1L, 1L, action, event, 128);
String stringKey = ActionStateUtil.generateKey("1", 1L, action, event, 128);

assertNotEquals(numericKey, stringKey);
assertTrue(ActionStateUtil.matchesBusinessKey(stringKey, 1L));
```

The keys had different key-groups (`Long(1)` → 86, `String("1")` → 54), but the cross-key match passed.

With both records cached, `KafkaActionStateStore.pruneState(1L, 1L)` evicted both and emitted two tombstones. A `Long(1)` cache-miss lookup also evicted a newer `String("1")` record in both Kafka and Fluss.

With `maxParallelism = 1`, the two generated state keys were identical.

Expected: distinct Flink keys remain isolated regardless of their string representation.

Actual: keys with equal `toString()` values cross-match; when their key-group also matches, their complete state keys collide.

These helper and store paths were executed directly. The resulting action re-execution follows the operator's cache-miss path but was not reproduced end to end through a Flink restart.

### Version and environment

- Flink Agents: `0.4-SNAPSHOT`
- Reproduced against `main` commit `19e22488` and PR #885 worktree commit `1f371bde`; both include the key format from PR #1024
- Apache Flink: `2.3.0`
- Java: JDK 17
- OS: macOS 15.4.1, arm64
- No live Kafka or Fluss service is required; the Kafka pruning reproduction uses `MockProducer`, and the lookup reproduction exercises the stores' in-memory cache paths

rob-9:
Closes #1099.

This branch is stacked on PR #1094, which makes the action identifier stable across JVM restarts. The two changes fix separate parts of the action-state key.

### Purpose of change

Flink Agents stores action progress under the current key so recovery can find completed work and avoid repeating its side effects. The existing format turns that key into text. Different typed keys with the same text, such as `Long(1)` and `String("1")`, can therefore point to the same action state. A lookup or prune for one key can remove the other key's completed state, allowing recovery to run that action again.

This change gives each typed key a durable identity derived from its serialized bytes. Kafka and Fluss use that identity consistently for writes, lookups, cleanup, partitioning, and recovery.

#### Runtime flow

1. During operator initialization, the store receives the keyed backend's serializer and maximum parallelism.
2. `ActionStateKeyEncoder` serializes the typed key and hashes the bytes with SHA-256. The state key contains that digest, the key-group, sequence number, event ID, and action ID.
3. Kafka partitions by the business-key digest. Fluss uses it as the table's distribution identity. Both stores use it for lookup, divergence cleanup, and pruning.
4. Recovery validates the fields and key-group range before applying the subtask ownership filter and caching each owned record.

#### Key decisions

- Serialized bytes preserve type information and align the identity with keyed state. A fixed-length digest keeps raw key data out of backend keys and gives every record a bounded key size.
- Serializer snapshots can change during normal operation as POJO subclass caches grow. The key therefore uses serialized key bytes without a snapshot fingerprint.
- The store supports one record format. Existing records and changes to key types or serializer configuration during recovery are unsupported; field validation cannot reliably detect every incompatible record.

#### Related work

Issue #1034 manages the Kafka offset boundary for deleting an older log prefix, while this PR defines the identity of each record. PR #885 manages Kafka tombstones. When those changes are combined, replay must validate the record key before applying a value or tombstone, and pruning must derive the same typed identity.

#### Interaction decisions

| Record during recovery | Ownership | Result |
|---|---|---|
| Valid current-format fields | Owned | Cache the record |
| Valid current-format fields | Foreign | Skip the record |
| Malformed field or out-of-range key-group | Any | Stop and identify the invalid field and key |

#### Behavioral contracts

- `Long(1)`, `String("1")`, and custom keys sharing a string representation receive separate identities, including at maximum parallelism `1`.
- Equivalent serializers produce the same key identity.
- Unchanged POJO jobs recover completed action state even when runtime subclass caching changes their serializer snapshots.
- Kafka and Fluss share one typed identity across storage, lookup, cleanup, distribution, and recovery.
- Recovery validates each record first, then caches records owned by the current subtask.
- Generated keys use canonical fields and a nonnegative sequence number.

#### Failure behavior

- Key serialization failures raise immediately and retain their original causes.
- Malformed fields, noncanonical values, and out-of-range key-groups stop recovery with a specific error. Recovery errors bound every echoed key and field.
- Pruning preserves records whose format or sequence number prevents safe attribution.

### API

The supported user-facing API and configuration remain unchanged. The stores, key utilities, and Kafka partitioner now carry explicit `@Internal` annotations. Their construction path receives an `ActionStateKeyEncoder`, so direct callers of these implementation classes must update their constructor calls.

The durable record format changes incompatibly. Existing action-state records are unsupported; use a fresh Kafka topic or Fluss table and start without an older checkpoint or savepoint when upgrading. For subsequent recovery, preserve the key type and serializer configuration. Flink may accept a serializer change that produces different key bytes; the action-state store does not detect that change and may miss completed work.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

The deployment and configuration documentation explains the format transition, serializer compatibility, backend isolation, and digest security boundary.

- [x] Yes
- [ ] No

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
@@ -222,8 +222,11 @@ public class ActionStateUtil {
     }
 
     private static String generateUUIDForAction(Action action) throws IOException {
+        // Action.hashCode() folds in JavaFunction's Class[] parameterTypes, and Class.hashCode()
+        // is the per-JVM identity hash — so the hash-derived UUID changes on every process
+        // restart and recovery lookups can never hit. Derive from the plan-unique action name,
+        // which is stable across restarts.
         return String.valueOf(
-                UUID.nameUUIDFromBytes(
-                        String.valueOf(action.hashCode()).getBytes(StandardCharsets.UTF_8)));
+                UUID.nameUUIDFromBytes(action.getName().getBytes(StandardCharsets.UTF_8)));
     }
 }
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
@@ -22,25 +22,29 @@ import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.json.JsonMapper;
 import org.apache.flink.agents.api.Event;
 import org.apache.flink.agents.plan.actions.Action;
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.api.common.typeutils.TypeSerializerSnapshotSerializationUtil;
+import org.apache.flink.core.memory.DataOutputSerializer;
 import org.apache.flink.runtime.state.KeyGroupRangeAssignment;
 import org.apache.flink.util.Preconditions;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
 import java.util.List;
 import java.util.UUID;
 import java.util.function.IntPredicate;
 import java.util.function.LongPredicate;
 
 /** Utility class for action state related operations. */
-public class ActionStateUtil {
-
-    private static final Logger LOG = LoggerFactory.getLogger(ActionStateUtil.class);
+@Internal
+public final class ActionStateUtil {
 
     private static final JsonMapper MAPPER =
             JsonMapper.builder()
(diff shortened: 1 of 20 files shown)


runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
```diff
@@ -81,16 +99,49 @@ public static String generateKey(
...
+
+    /**
+     * Fingerprints the serializer's snapshot, including its version and configuration, once per
+     * store. Custom serializers must describe all encoding changes in their snapshots.
+     */
+    static String generateSerializerFingerprint(TypeSerializer<?> keySerializer) {
+        DataOutputSerializer output = new DataOutputSerializer(128);
+        try {
+            TypeSerializerSnapshotSerializationUtil.writeSerializerSnapshot(
+                    output, keySerializer.snapshotConfiguration());
```
wenjin272: Could we remove the serializer snapshot fingerprint from both the state key and recovery validation? Flink’s `PojoSerializer` snapshots include a runtime subclass cache, so an unchanged job can produce a different fingerprint after checkpoint recovery even though the key bytes remain identical. I reproduced this with a POJO subclass: Flink keyed state restored successfully, but Kafka action-state recovery failed. For this PR, documenting that changes to key types or serializer configurations are unsupported during recovery seems sufficient.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
```diff
@@ -22,57 +22,75 @@
...
-    private static final Logger LOG = LoggerFactory.getLogger(ActionStateUtil.class);
+@Internal
+public final class ActionStateUtil {
 
     private static final JsonMapper MAPPER =
             JsonMapper.builder()
                     .configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true)
                     .configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true)
                     .build();
     private static final String KEY_SEPARATOR = "_";
```
wenjin272, line 53: Could we also remove the `v2` prefix and keep a single state-key format? Since we’re still in 0.x and don’t need backward compatibility for existing action-state records, the version marker adds unnecessary complexity and may suggest a compatibility mechanism we don’t intend to provide.

wenjin272 (commented):
Thanks for working on this and for adding the key-isolation coverage! I left two suggestions to simplify the state-key format and avoid rejecting valid recovery.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
rob-9: removed. I added a Pojo recovery test that fails before this fix, and documented that key type and serializer config changes are unsupported during recovery.

runtime/src/main/java/org/apache/flink/agents/runtime/actionstate/ActionStateUtil.java
rob-9: yep, makes sense!

wenjin272 (approved):

Issue #939, opened 2026-07-30
[Bug] Bounded workflows silently drop records in batch mode with multiple keys


Sxnan:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

When a Flink Agents workflow runs with `RuntimeExecutionMode.BATCH`, records may be silently dropped if one operator subtask processes multiple keys. The job finishes successfully without errors, but usually only the last key processed by each subtask produces output.

`ActionExecutionOperator` stores pending action tasks in keyed state while scheduling workflow continuations through the mailbox. Flink's batch keyed-state backend assumes that a key is fully processed before moving to the next key and clears keyed state whenever the current key changes. Therefore, when the operator processes key B before the mailbox continuation for key A runs, A's pending tasks are cleared. Switching back to A can also clear B's tasks.

Expected: every input record completes its workflow and produces output.

Actual: pending workflows are silently discarded while the Flink job reports `FINISHED`.

Setting `execution.batch-state-backend.enabled=false` avoids the issue.

### Possible fix

In batch mode, the operator must finish the complete action chain for the current key before allowing input processing to move to another key. Alternatively, Flink Agents jobs should disable the batch keyed-state backend. A regression test should use a real BATCH execution environment with at least two keys assigned to the same subtask.

### How to reproduce

1. Create a Python workflow with multiple synchronous stages:

   `InputEvent -> Stage1Event -> Stage2Event -> Stage3Event -> OutputEvent`

2. Run it with:

   - `RuntimeExecutionMode.BATCH`
   - parallelism `1`
   - five input records
   - a different key for every record

3. Observe that only one record reaches the output:

   `5 inputs -> 1 output`

Control results:

- Using one fixed key: `5 inputs -> 5 outputs`
- Setting `execution.batch-state-backend.enabled=false`: `5 inputs -> 5 outputs`
- The issue remains reproducible when all action delays are removed.

### Version and environment

- Flink Agents: 0.2.1
- Apache Flink: 1.20.3
- Python: 3.11
- Java: OpenJDK 11
- Runtime mode: BATCH
- OS: macOS ARM64

The issue was reproduced locally with the official `flink-agents==0.2.1` package and does not require a vendor-specific Flink distribution.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

weiqingy:
[...]

PR #1104, opened 2026-09-05
[runtime] Reject RuntimeExecutionMode.BATCH with the default batch state backend

Ashfaqbs:
Linked issue: #939

### Purpose of change

A job configured with `RuntimeExecutionMode.BATCH` and Flink's batch keyed-state backend (`execution.batch-state-backend.enabled`, enabled by default in `BATCH` mode) now fails immediately at job-graph construction with a clear error, instead of running to completion and silently dropping records.

`ActionExecutionOperator` keeps pending action tasks in keyed state across mailbox continuations that can span multiple keys of the same subtask. The batch keyed-state backend assumes a key is fully processed before the operator moves to the next one and clears keyed state on every key switch, so a continuation for one key can be discarded when the operator processes another key first — the job still reports `FINISHED`; only the affected records go missing.

**Runtime flow:** `CompileUtils.connectToAgent` (both the Python and Java entry points funnel into the same private overload) now calls `checkBatchStateBackendCompatibility` before constructing the `ActionExecutionOperator`. That reads `ExecutionOptions.RUNTIME_MODE` and `ExecutionOptions.USE_BATCH_STATE_BACKEND` off `keyedInputStream.getExecutionEnvironment().getConfiguration()` and throws before the operator is ever wired into the graph.

**Key decisions:**
- Fail fast at graph construction rather than attempt the deeper fix (making the operator safe under batch key-switch semantics), which is a separate, larger change out of scope here.
- Only an explicitly configured `BATCH` mode is detected. `RuntimeExecutionMode.AUTOMATIC` (the default) that later resolves to batch execution because every source happens to be bounded is not caught, since that resolution hasn't happened yet at graph-construction time — inspecting source boundedness to predict `AUTOMATIC`'s eventual resolution was rejected as disproportionate for a case the explicit, documented `BATCH` setting already covers.

#### Interaction decisions

| `RUNTIME_MODE` | `USE_BATCH_STATE_BACKEND` | Behavior |
|---|---|---|
| `STREAMING` | `true` / `false` | Unaffected — check only fires for `BATCH` |
| `BATCH` | `true` (default) | Rejected: `IllegalStateException` before the operator is constructed |
| `BATCH` | `false` | Allowed — the documented workaround |
| `AUTOMATIC` | `true` / `false` | Unaffected — not detected (see Key decisions) |

#### Behavioral contracts
1. `connectToAgent` throws `IllegalStateException` before constructing the operator when `RUNTIME_MODE=BATCH` and `USE_BATCH_STATE_BACKEND=true`.
2. `connectToAgent` behaves exactly as before this change for every other `RUNTIME_MODE`/`USE_BATCH_STATE_BACKEND` combination.
3. The exception message names both workarounds (`RuntimeExecutionMode.STREAMING`, or setting `execution.batch-state-backend.enabled=false`) and links to #939.

#### Failure behavior
The only new failure path: an unsupported `BATCH` + batch-state-backend-enabled configuration raises `IllegalStateException` synchronously, before any operator is created — not a retry, not silently absorbed. For the Python entry point, this check runs after the existing pickled-input-type and agent-plan-JSON validations, so it doesn't change their ordering or messages. No other failure path changes.

### Tests

| Contract | Test |
|---|---|
| 1 | `rejectsExplicitBatchModeWithDefaultBatchStateBackend` |
| 2 (BATCH + disabled) | `allowsExplicitBatchModeWithBatchStateBackendDisabled` |
| 2 (STREAMING, unaffected) | pre-existing `testJavaNoKeyedStreamConnectToAgent` / `testJavaKeyedStreamConnectToAgent` |
| 3 | `rejectsExplicitBatchModeWithDefaultBatchStateBackend` asserts the message contains both `RuntimeExecutionMode.BATCH` and `execution.batch-state-backend.enabled` |

Not verified: `AUTOMATIC` mode resolving to batch execution via all-bounded sources; an actual executed `BATCH` job reproducing the original multi-key data-loss scenario (would need a heavier MiniCluster run) — the causal chain for *why* the collision happens relies on the issue's own trace of Flink's batch state-backend semantics, not a reproduction inside this PR.

<details>
<summary>Verification evidence</summary>

Ran `CompileUtilsTest` (9/9 pass) plus the full `operator`, `context`, and `skill` packages after a clean rebuild (`mvn -pl runtime clean`, then `mvn -pl api,plan -am install -DskipTests` to pick up unrelated upstream API changes already on `main`). 249 tests, 7 pre-existing failures unrelated to this change: 4 are a Windows-only temp-directory-deletion quirk in `ClasspathSkillRepositoryTest` (locked JAR files), and 3 (`FileSystemSkillRepositoryTest`, `SkillManagerTest` x2) pick up this machine's own local Claude Code skills directory contents (`github`, `nano-banana-pro`) rather than the repo's test fixtures — both reproduce identically on unmodified `main` in this environment. `mvn -pl runtime spotless:check` clean.

</details>

### API

No public API changes. `connectToAgent`'s signature and behavior for every previously-supported configuration are unchanged; this adds a new, additive rejection for a configuration that was previously accepted but silently broken.

### Documentation

- [x] `doc-not-needed`

- [x] Yes

### DIFF
--- a/runtime/src/main/java/org/apache/flink/agents/runtime/CompileUtils.java
+++ b/runtime/src/main/java/org/apache/flink/agents/runtime/CompileUtils.java
@@ -22,15 +22,19 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.flink.agents.plan.AgentPlan;
 import org.apache.flink.agents.runtime.operator.ActionExecutionOperatorFactory;
+import org.apache.flink.api.common.RuntimeExecutionMode;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.java.functions.KeySelector;
 import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.configuration.ExecutionOptions;
+import org.apache.flink.configuration.ReadableConfig;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.datastream.KeyedStream;
 import org.apache.flink.streaming.api.typeinfo.python.PickledByteArrayTypeInfo;
 import org.apache.flink.types.Row;
 
 import static org.apache.flink.util.Preconditions.checkArgument;
+import static org.apache.flink.util.Preconditions.checkState;
 
 /** A utility class that bridges Flink DataStream/SQL with the Flink Agents agent. */
 public class CompileUtils {
@@ -97,6 +101,7 @@ public class CompileUtils {
             TypeInformation<OUT> outTypeInformation,
             boolean inputIsJava,
             boolean pythonKeyIsPickled) {
+        checkBatchStateBackendCompatibility(keyedInputStream);
         return (DataStream<OUT>)
                 keyedInputStream
                         .transform(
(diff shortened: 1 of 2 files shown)


wenjin272 (approved):
Thanks for adding this fail-fast check. I’m happy to merge it as an interim safeguard against silent data loss. Our goal is still to support batch execution correctly with the default batch state backend, so #939 should remain open to track the underlying runtime fix.

Issue #1106, opened 2026-09-06
[integrations] Preserve finish_reason for Anthropic and Watsonx responses


emecii:
### Motivation

Follow-up to #936 and merged #1040. The shared response path now checks `finish_reason`, but Anthropic still has the capture TODO identified in #936; Watsonx parses a reason without preserving it in message metadata. Those providers therefore do not consistently participate in the shared handling.

#1040 explicitly checks the common path before text, structured output, or tool dispatch, and its approval acknowledges that design. This proposal follows that accepted behavior rather than moving the gate.

### Proposed scope

Capture `finish_reason` in the Java and Python Anthropic and Watsonx connection responses, preserving content and the existing provider-specific content blocks.

For Anthropic, propose `max_tokens` → `length`, `end_turn`/`stop_sequence` → `stop`, and `tool_use` → `tool_calls`. Preserve unrecognized reasons, including `model_context_window_exceeded` and `refusal`, verbatim for now. Mapping context exhaustion to `length` would currently tell users to increase output tokens, which may be the wrong remedy; treating a refusal as `content_filter` would add a new rejection decision.

For Watsonx, retain its reported reason verbatim. Missing reasons should not add a metadata key. Keep finish-reason metadata out of outbound provider requests, while preserving Anthropic content-block round trips.

### Scope status

The narrow Anthropic portion is now submitted in PR #1107. It implements only the directly
equivalent `max_tokens` → `length` mapping in Java and Python, and preserves every other supplied
Anthropic reason unchanged. Its scope follows the explicit Anthropic follow-up identified by
@weiqingy on #936 and the shared-path behavior accepted in #1040.

Watsonx remains an independent, unstarted follow-up. Because it already emits the shared
vocabulary, retaining the reason is mechanically separate from the Anthropic mapping; it should
be reviewed independently rather than enlarging #1107.

### Validation and scope decision

Use SDK response fixtures and Java/Python tests for capture with and without usage, missing and unknown reasons, outbound serialization, and the existing shared rejection/error-strategy path. No live-provider performance or model-quality claims are proposed.

The direct committer guidance in #936 supports metadata capture and the shared action's handling of
canonical reasons. The two Anthropic reasons above remain native until shared diagnostics and
rejection semantics are decided.

Prepared with OpenAI Codex assistance.

CodeTrainerMan:
Thanks for the detailed breakdown. I would like to take the **Watsonx** portion.

[...]

PR #1107, opened 2026-09-07
[integrations][anthropic][java][python] Preserve Anthropic finish reason

emecii:
Linked issue: #1106 

### Purpose of change

Anthropic responses that stop at the configured output-token limit now reach the shared chat action
as `extra_args["finish_reason"] == "length"`. The action already rejects that canonical value before
it dispatches incomplete content, so callers receive its clear truncation error instead of parsing
or consuming a partial response.

#### Runtime flow

Both Anthropic connections read the SDK response stop reason after converting the content. They map
Anthropic's `max_tokens` value to the existing cross-provider `length` value and retain every other
reported value unchanged. The shared chat action reads that metadata before it dispatches the
response, while direct connection callers can inspect the metadata themselves.

#### Key decisions

Only `max_tokens` is normalized because it is Anthropic's name for the shared action's established
token-limit condition. Other reasons remain provider values so that an unknown future reason is
observable without changing control flow.

#### Interaction decisions

| Anthropic stop reason | Stored `finish_reason` | Shared action behavior |
| --- | --- | --- |
| absent | absent | Existing behavior |
| `max_tokens` | `length` | Rejects incomplete response before dispatch |
| any other supplied value | same value | Existing accepted-reason behavior |

#### Behavioral contracts

- A Java or Python Anthropic response stopped by `max_tokens` exposes the canonical `length`
  metadata that the shared action recognizes.
- A supplied non-token-limit stop reason remains available without being reclassified.
- A response without a stop reason gains no metadata entry.

#### Failure behavior

This converter change does not catch, retry, or transform provider failures. When the model returns
`max_tokens`, the existing shared action raises its established truncation error; direct connection
callers can inspect the response metadata before invoking that action.

### Tests

| Contract | Tests |
| --- | --- |
| `max_tokens` becomes `length`; ordinary reasons remain observable | Java `AnthropicChatModelConnectionTest` (94 passed); Python `test_anthropic_response_parsing.py` |
| Shared action rejects `length` before dispatch | Java `ChatModelActionTest` and `ChatModelActionRetryTest` (36 passed); Python `test_chat_model_action.py` |

Focused Python run: 107 passed across the Anthropic response parser and shared-action tests.
Spotless, Ruff format/check, and Apache RAT license checks pass.

#### Coverage and gaps

Focused Java and Python tests cover the provider-to-shared-reason mapping and the shared action's
rejection path. Live Anthropic API calls were not run; the tests use SDK response objects and local
action fixtures, so they require no credentials or network access.

<details>
<summary>Implementation evidence</summary>

The Python converter stores the key independently of token-usage metrics. The Java converter reads
the optional SDK stop reason after constructing the response and does the same. This keeps the
mapping available for responses with or without recorded usage and for both plain-text and tool-call
conversion paths.
</details>

#### Compatibility impact

No public API signature changes. Existing consumers of `extra_args` gain `finish_reason` on
Anthropic responses; the shared action now classifies Anthropic token-limit responses consistently
with OpenAI-family responses.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Codex CLI 0.153.4 (GPT-6)


wenjin272 (approved):

Issue #1088, opened 2026-09-02
[Bug][runtime] Interrupted tool calls are recorded as tool errors, letting the action loop continue and persist as completed


Ashfaqbs:
## Description

`RunnerContextImpl.executeSequentially`/`executeParallel`'s tool-call path (`ToolCallAction.java:256`) catches `InterruptedException` the same as any ordinary tool failure, records it as a tool error, and lets the action loop continue — unlike the chat-call path (fixed in #1071), which now rethrows a raw `InterruptedException` immediately.

Concretely, on a cancellation during a tool call:
1. `ToolCallAction.java:256` catches the `InterruptedException`, `:257` records it as a tool error.
2. The loop moves on and (`:85`) sends the `ToolResponseEvent` anyway, which drives another chat call.
3. `ActionExecutionOperator` then persists the action as completed (~line 491, `durableExecManager.maybePersistTaskResult`), since nothing tells it the action was actually interrupted.
4. On recovery, `actionState.isCompleted()` (~line 437) skips re-execution and replays that stale output instead of re-running the call.

Step 4 is the same failure mode as #1070's "problem 2" (a stale interruption replayed as terminal after recovery), just reached through the action-state persistence path instead of the durable-call-slot path that #1071 closed for the chat side — because here the raw `InterruptedException` never escapes far enough to skip the persist call.

`executeParallel` has the equivalent shape at `RunnerContextImpl.java:209` (parallel tool call execution added in #926).

This predates #1071 and is out of that PR's scope (chat-call retry/persistence only); filing separately per discussion on https://github.com/apache/flink-agents/pull/1071#discussion_r3892137335 and follow-up at https://github.com/apache/flink-agents/pull/1071#discussion_r3915950776.

## Expected behavior

A cancellation during a tool call should propagate the same way the chat-call fix in #1071 does: restore the interrupt flag and rethrow immediately, rather than being recorded as a tool error and let the action loop continue — so the action is never persisted as completed off the back of a cancelled tool call, and recovery re-executes rather than replaying stale output.

## Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #1111, opened 2026-09-09
[runtime] Do not retry or persist interrupted tool calls on cancellation

Ashfaqbs:
Linked issue: #1088

### Purpose of change

`ToolCallAction.executeSequentially`/`executeParallel` caught `InterruptedException` the same as any ordinary tool failure, recorded it as a tool error, and let the loop continue — `processToolRequest` still sent a `ToolResponseEvent` afterward, driving another chat call and letting `ActionExecutionOperator` persist the action as completed. On recovery, `isCompleted()` then skipped re-execution and replayed the stale output instead of re-running the interrupted call — the same failure mode #1071 closed for the chat-call path, reached here through the tool-call path #1071 didn't cover (out of that PR's scope, filed separately per the discussion on #1071).

`executeParallel`'s batch path had the same gap one layer down too: `RunnerContextImpl.durableExecuteAllAsync`'s per-callable `catch (Exception e)` folded an interruption into `Outcome.failure` and kept scheduling the rest of the batch, even though `durableExecute` itself (fixed in #1071) already rethrows `InterruptedException` correctly — this outer catch was just re-swallowing it before it could propagate.

### Behavioral Semantics

Fixed both the same way #1071 fixed the chat-call path: an explicit `catch (InterruptedException e)` ahead of the generic failure handling restores the interrupt flag and rethrows immediately, in three places:

- `RunnerContextImpl.durableExecuteAllAsync` — stops scheduling the remaining callables in the batch and propagates, instead of recording a failed `Outcome` and continuing.
- `ToolCallAction.executeParallel` — propagates instead of recording every execution in the batch as a tool error.
- `ToolCallAction.executeSequentially` — propagates instead of recording the current execution as a tool error and moving on to the next one in the loop.

`processToolRequest` now declares `throws InterruptedException` and lets it propagate, so `sendEvent` for the `ToolResponseEvent` is skipped entirely on a cancellation — no further chat call gets driven off a cancelled tool call, and the action is never persisted as completed on the back of it. Ordinary tool failures (missing tool, execution exceptions, infrastructure errors) are unaffected — they still go through the existing recording/reporting paths untouched.

### Tests

- Added 3 cases to `ToolCallActionTest`: an interrupted single (sequential) call propagates and sends no `ToolResponseEvent`; a second sequential tool call does not run once the first is interrupted; an interrupted parallel batch propagates and sends no `ToolResponseEvent`. All three assert the interrupt status is restored on the thread.
- Added a case to `RunnerContextImplDurableExecuteTest`: in a 3-callable batch where the 2nd is interrupted, the 1st (already-succeeded) callable's result is correctly persisted, the 2nd's interruption is not recorded as a failed outcome, and the 3rd never runs.
- `ToolCallActionTest`: 19/19 passing (16 pre-existing + 3 new). `RunnerContextImplDurableExecuteTest`: 14/14 passing (13 pre-existing + 1 new).
- Full `plan` module suite: 313/316 passing; the 3 failures are the pre-existing `BashToolTest` cases (require a working WSL bash on this Windows environment, fail identically on `main` before this diff — confirmed via `git stash`).
- Full `runtime` module suite: 847/854 passing; all 14 failures are pre-existing, environment-specific (Fluss coordinator server not starting, Windows path/file-locking issues in unrelated skill-loading and Python-environment tests) — none in `ToolCallActionTest` or `RunnerContextImplDurableExecuteTest`.
- `spotless:check` clean on both modules.

### API

No public API changes. `ToolCallAction.processToolRequest`/`executeSequentially`/`executeParallel` now declare `throws InterruptedException`, but these are internal, reflectively-invoked action handlers (`JavaFunction`), not part of the public surface.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No


plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
```diff
@@ -206,6 +212,12 @@ private static void executeParallel(
             for (int i = 0; i < outcomes.size(); i++) {
                 recordOutcome(executions.get(i), outcomes.get(i), ctx, success, error, responses);
             }
+        } catch (InterruptedException e) {
```
wenjin272, line 215: The new catch only handles cases where `durableExecuteAllAsync()` throws directly, but the production Java path uses `JavaRunnerContextImpl`, whose JDK 11 and JDK 21 batch executors convert a tool's `InterruptedException` into `Outcome.failure`. `finalizeExecutedOutcomes()` then persists it as a durable failure, and `recordOutcome()` still turns it into a tool error, so the action can send a `ToolResponseEvent` and complete after cancellation. Could we propagate the interruption from the batch execution boundary before finalizing it, leaving the interrupted/unstarted slots pending while preserving already completed outcomes? A regression test should exercise `JavaRunnerContextImpl` rather than mocking `durableExecuteAllAsync()` to throw directly.

wenjin272 (commented):
Thanks for following up on the tool-call cancellation path and for adding focused regression coverage. I found one remaining issue in the production parallel Java path.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
Ashfaqbs: Good catch, thanks. Pushed c27c9647: `finalizeExecutedOutcomes()` in `JavaRunnerContextImpl` now checks whether an executed outcome's error is an `InterruptedException` (which is exactly what both the JDK 11 fallback and JDK 21 continuation batch executors fold a cancelled supplier into), restores the interrupt flag, and rethrows before calling `finalizeCallAt` for that slot. Slots finalized earlier in the loop keep their persisted outcome; that slot and everything after it in iteration order stay in their already-reserved pending state, so recovery re-executes them instead of replaying a false failure. The rethrow propagates up through `durableExecuteAllAsync()` into `executeParallel()`'s existing `catch (InterruptedException e)`, which was already correct.

Added `testDurableExecuteAllAsyncInterruptionLeavesRemainingSlotsPendingAndPropagates` using the existing `InspectingContinuationActionExecutor` double (mirrors the real executors' catch-and-wrap behavior) rather than mocking `durableExecuteAllAsync()` directly, so it exercises the actual `JavaRunnerContextImpl` path this bug lived in.

runtime/src/main/java/org/apache/flink/agents/runtime/context/JavaRunnerContextImpl.java
```diff
@@ -192,6 +192,15 @@ private <T> void finalizeExecutedOutcomes(
                 plan.outcomes.set(callIndex, outcome);
                 continue;
             }
+            if (outcome.isFailure() && outcome.getError() instanceof InterruptedException) {
```
wenjin272, line 195: One path is still missing: without a durable store, `durableExecuteAllAsync()` returns the wrapped `InterruptedException` as a failure outcome, so `ToolCallAction` continues processing it as a tool error. Could we rethrow the interruption in this path as well and add a regression test?

wenjin272 (commented):
Thanks for the update and for addressing the durable-state cancellation path.

runtime/src/main/java/org/apache/flink/agents/runtime/context/JavaRunnerContextImpl.java
Ashfaqbs: Good catch, fixed in b0b672d1. `executeAllWithoutDurableState()` now checks each returned `Outcome` for a wrapped `InterruptedException` and rethrows it (restoring the interrupt flag) instead of letting it flow back to `ToolCallAction` as a normal failed outcome. Since there's no durable store on this path, there's nothing to leave pending -- the fix is purely about not misclassifying the cancellation as a tool failure. Added `testDurableExecuteAllAsyncWithoutDurableStateInterruptionPropagates`, which builds a `JavaRunnerContextImpl` without calling `setDurableExecutionContext` (so `durableExecuteAllAsync` takes the no-durable-store branch) and asserts the interruption propagates with the flag restored.

wenjin272 (approved):

PR #1113, opened 2026-09-10
[integrations][java][python] Preserve Watsonx finish_reason in extra_args

Ashfaqbs:
Linked issue: #1106 (Watsonx portion; the Anthropic portion is covered separately by #1107)

### Purpose of change

Watsonx's chat response already carries a `finish_reason` in the shared OpenAI-compatible vocabulary (`stop`, `tool_calls`, etc.), but the connection only used it to emit a log warning and never surfaced it to callers. This preserves it verbatim in `extra_args["finish_reason"]`, mirroring the existing OpenAI/Azure OpenAI connection behavior, for both the Java and Python connections.

- A present, non-null reason is copied verbatim into `extra_args["finish_reason"]`, including values outside `stop`/`tool_calls`.
- A missing or JSON-null reason adds no key (matches the OpenAI connection's documented contract).
- The existing "response may be truncated" log warning is unchanged.
- Capture is independent of the token-usage metadata branch (verified with a case that has a finish reason but no usage block).

### Tests

- Java: `WatsonxChatModelConnectionTest` — added 4 tests covering a documented reason, an unrecognized/vendor-specific reason stored verbatim, absent `finish_reason` member, JSON-null `finish_reason`, and a reason captured without usage metadata present. `mvn test -Dtest=WatsonxChatModelConnectionTest` → 24 passed.
- Python: `test_watsonx_chat_model.py` — added 2 tests (unknown reason verbatim, no key when `None`) and asserted `finish_reason` in the existing mocked-chat test. `pytest -m "not integration"` → 14 passed.
- `mvn spotless:check` and `ruff check`/`ruff format --diff` both clean on the touched files.

### API

No public API signature changes; adds a new, optional key to the existing `extra_args` metadata map returned by both connections.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes


wenjin272 (approved):

PR #1114, opened 2026-09-10
[api][plan][python] Expose sub-agents to chat models as callable tools

### LINKED ISSUE #1082: [Feature] Introduce subagent framework and AGENT resource type
### Search before asking

### Description

As discussed in #660, there are scenarios where a Flink Agent needs to call external agents — for example, delegating a subtask to a remote LLM agent service or a third-party agent platform. Currently, users can only do this by manually writing remote calls inside actions, which means they miss out on durable execution, checkpoint recovery, and tool call integration. 

This issue proposes sub-agent support to close that gap: agents can declare sub-agents as an AGENT resource and invoke them from actions (and later from chat-model tool calls) through a future-style API, with durable execution semantics so in-flight calls survive failover, and full Java/Python parity including cross-language setups. This makes agent composition declarative and recovery-safe, and lays the groundwork for concurrency optimizations such as eager submit and batch execution. 

A solution is already in mind and under development — design discussion and implementation details are tracked in #909 and #938.

### LINKED ISSUE #1112: [Feature] Expose sub-agents to chat models as callable tools
### Search before asking

### Description

This is a sub-increment of the sub-agent work already designed in #909 and scoped in #1082 — specifically the "chat-model tool calls" step that #1082 listed as a later deliverable. The framework and `AGENT` resource type landed in #938; the remaining work is being split into several sub-issues/PRs, and this one covers exposing declared sub-agents to the chat model as callable tools, so the model can delegate a sub-task to a sub-agent the way it calls a function.

As agreed in #909, this reuses the existing tool-call dispatch path (`ToolCallAction`): internally the same durable `submit` / `await` sub-agent invocation, wrapped by the chat model's ReAct-style loop. Full Java/Python parity.

### Scope

- Sub-agent callables are advertised under a reserved `subagent_` prefix, keeping the `AGENT` and `TOOL` namespaces distinct; a chat-model setup declares its delegatable sub-agents via a `subagents` argument.
- A sub-agent is exposed only if it can produce a JSON input schema (explicit or derived from its input type); otherwise it is skipped. Cross-language bridge handles are rejected.
- Dispatch stays durable; results are normalized to JSON-compatible content, and a failed sub-agent returns its reason to the model instead of failing the job.

This is one of several sub-issues/PRs splitting the sub-agent roadmap; it does not add a built-in internal sub-agent (a separate track). End-to-end out-of-the-box UX lands when that track completes.

### References

- Design discussion: #909
- Parent feature issue: #1082
- Framework PR: #938
- Background: #660

yunfengzhou-hub:
Linked issue: #1112

### Purpose of change

A chat model can now delegate to a sub-agent by issuing a tool call. The setup declares each sub-agent it names as one callable under the reserved `_subagent_` prefix; at execution a prefixed call is routed to that sub-agent and its result is handed back to the model. Which sub-agent to call, and with what arguments, becomes the model's decision — previously a sub-agent could only be invoked from action code.

A sub-agent the caller names but gives no usable input shape is rejected when the model sets up, and a delegation that fails tells the model why, so it can correct the call rather than repeat it blindly.

#### Runtime flow

`open()` walks the `subagents` argument after the tools and, for each name resolving to a `SubagentSetup` with a usable input schema, adds a metadata-only `SubagentTool` named `_subagent_<name>`. The model builds a function call against that schema. `ToolCallAction` reads the prefix off the callable name, resolves the AGENT resource once, and passes the model's arguments to `submit(...).await()` unchanged; the result is normalized to JSON-generic form and rendered as the tool-message content. In the batched tool-call path sub-agent calls are dispatched concurrently — every call is submitted before any is awaited, so the runs overlap like the batched tools instead of proceeding one by one. Python mirrors this in its setup and tool-call action.

#### Key decisions

The reserved prefix, not a distinct tool type, separates the namespaces: tool registration rejects `_subagent_` at plan-construction, so a prefixed name can only address a sub-agent, and a tool and a sub-agent may share a resource name. No listing message is injected — every sub-agent description ends with a marker, so the model learns a callable is a delegation from its description alone.

`SubagentTool` is metadata-only (`call()` throws); dispatch resolves the AGENT resource at execution, keeping the sub-agent's own durable execution as the single source of the result and avoiding nested durable cursors. An input schema is derived from the declared input type through the same Jackson generator the ReAct output schema uses, then aligned to the cross-language form (object-level `required`, `byte[]` as string/binary), adding no dependency.

#### Interaction decisions

| subagents entry resolves to | Input schema | Offered to the model |
| --- | --- | --- |
| SubagentSetup, explicit schema | that schema | yes |
| SubagentSetup, input type rendering as an object | derived schema | yes |
| SubagentSetup, type rendering as no object / none | none | no — open() fails (checkState); job fails at setup |
| bridge handle (owned by the other language) | carries none | no — open() fails |
| a tool registered under the prefix | — | impossible — rejected at plan-construction |

At dispatch a `_subagent_` name resolves in the AGENT namespace and a plain name in TOOL; a success returns normalized JSON content, and any failure returns an error carrying the reason.

#### Behavioral contracts

1. Each named sub-agent with a usable input schema is declared as exactly one callable `_subagent_<name>`, after the tool callables, carrying its description and schema.
2. A sub-agent with neither a schema nor a shape-bearing input type is rejected when the chat model sets up — `open()` fails through `checkState` — rather than dropped, so a declaration with no arguments for the model to build a call from surfaces immediately.
3. Every sub-agent description ends with the marker; an undescribed one gets a generic delegation description, and no separate listing message is injected.
4. The schema is the explicit one if declared, else derived from the input type; explicit wins.
5. A tool and a sub-agent may share a resource name; a callable name resolves to exactly one namespace.
6. A tool name must not carry the prefix — rejected at plan-construction, including a name an MCP server advertises.
7. A prefixed call hands the model's arguments over unchanged; injection stays tool-only.
8. A success is reported as JSON-generic content; a declared result type narrows it (extras ignored), an undeclared one takes it as it arrived.
9. A result JSON cannot express — including one that reaches back into itself — is refused with the path where it was found, and reported as a failed delegation, not dropped, stringified, or made a job failure. No fixed nesting cap is imposed: a deep-but-finite result still has a JSON form, so only a cycle, which has none, is refused; a result too deep to walk overflows and is folded into a failed delegation.
10. A failed delegation reaches the model with the reason.
11. Re-opening rebuilds the callables from scratch and does not duplicate them.
12. Java and Python decide alike on declaration, dispatch, and result handling, and derive the same input schema for the same declaration.

#### Failure behavior

Construction and plan-building raise; call-time failures are absorbed into the tool response.

- Blank explicit input schema → IllegalArgumentException at construction.
- An input type that cannot be rendered, including a self-referential one → IllegalArgumentException naming the remedy.
- A subagents entry resolving to a bridge handle, a sub-agent with no schema and no shape-bearing input type, or a repeated callable name → open() fails (checkState), not silently dropped.
- A tool under the reserved prefix, including one an MCP server advertises → raises at plan-construction.
- At call time, an absent/non-SubagentSetup resource, an exception from submit/await, a failed result, a cyclic result, or a result too deep to normalize are each caught and returned as ToolResponse.error with the reason; the job continues. A StackOverflowError from a result nested deeper than the stack can walk is folded into the same failed delegation on both the serial and the batched path, rather than escaping as an Error that fails the job.
- A sub-agent call whose durable execution is cancelled surfaces an InterruptedException, propagated the way the tool paths do, so no further chat call is driven off a cancelled delegation; the concurrent path also cancels every handle submitted but no longer awaited.

### Tests

The new suites run offline, with no live model or external agent. Java: BaseChatModelSetupSubagentTest, SubagentSetupTest, AgentPlanSubagentResourceTest, ToolResultUtilsTest, ToolCallActionSubagentTest; Python: the mirrored test_chat_model_subagents, test_subagent, test_agent_plan, test_tool_result_utils, test_tool_call_action_subagent. Every contract is pinned on both sides:

| Contract | Java | Python |
| --- | --- | --- |
| 1 declared as one prefixed callable, after tools | ✅ | ✅ |
| 2 no schema and no shape → rejected at open() | ✅ | ✅ |
| 3 description marker; no listing message | ✅ | ✅ |
| 4 explicit schema wins, else derived | ✅ | ✅ |
| 5 tool and sub-agent share a name | ✅ | ✅ |
| 6 tool under the prefix rejected (incl. MCP-advertised) | ✅ | ✅ |
| 7 arguments handed over unchanged | ✅ | ✅ |
| 8 result type narrows; undeclared as-is | ✅ | ✅ |
| 9 non-JSON / cyclic result refused with its path; no depth cap | ✅ | ✅ |
| 10 failed delegation carries the reason | ✅ | ✅ |
| 11 re-open does not duplicate | ✅ | ✅ |
| 12 Java and Python decide alike, same derived schema | ✅ | ✅ |

Highest risk is result normalization (9): non-string map keys, non-finite numbers, a POJO inside a JSON tree, arrays walked by index, a cyclic result refused at its path, and a 200-deep acyclic result still normalized, each asserted on the path in the message where one applies. The derived input schema (12) is pinned whole on both sides — object-level `required` and each property's type/format, including a `byte[]`/`bytes` field rendered as string/binary — so a divergence fails a build rather than passing contract 12 while the schemas differ.

Not verified: no test drives a live model or a real external sub-agent, so these pin what the framework declares, dispatches, and reports, never that a provider accepts the derived schema. A sub-agent owned by the other language is checked only up to the rejection at open(); cross-language delegation end-to-end is out of scope. Parallel-mode routing of a sub-agent beside tools, and each call id landing on its own result under the batch, are covered; concurrency timing is not.

Implementation invariants (not caller-observable)
- SubagentTool.getToolType() is FUNCTION and call() throws UnsupportedOperationException; the callable is never invoked through Tool.call().
- getInputType()/getResultType()/getResourceType() are @JsonIgnore (behavior, not state); Python pins the same via test_the_declared_types_stay_out_of_the_plan_json and test_the_metadata_serializes_under_the_cross_language_keys.
- A declared result type is read with FAIL_ON_UNKNOWN_PROPERTIES off, mirroring pydantic's ignore-extra, then re-checked for JSON compatibility because a declared type can still render an inexpressible field.
- Result normalization tracks the containers on the current path in an identity-based set, so a cycle is refused while the walk still can and a diamond (one node reached by two distinct paths) stays legal; there is no depth counter.
- resolveSubagent resolves and type-checks the AGENT resource once and carries the setup down, so a plain tool call never attempts an AGENT resolution.
- A derived input schema is carried by the plan JSON; an explicit one is kept verbatim.

### API

Additive, aligned across Java / Python / YAML, building on the AGENT resource type and SubagentSetup:

- SubagentSetup gains getDescription(), getInputSchema(), getInputType(), getResultType() and the CALLABLE_NAME_PREFIX constant (`_subagent_`).
- The chat-model setup takes a new subagents argument (Java/Python constructor + YAML subagents:): the AGENT resources it may delegate to.
- ToolResultUtils (plan) normalizes a result; SubagentTool and InputSchemas are package-private, not public API.

One thing changes for a caller who does nothing differently: a tool named `_subagent_*` is now rejected at plan-construction, including a name an MCP server advertises. Otherwise nothing changes unless the new subagents argument is set, and a plain tool call routes as before. A sub-agent owned by the other language cannot yet be exposed as a callable — declaring one fails open().

### Documentation

- doc-needed
- doc-not-needed
- doc-included

doc-needed but deferred, matching the AGENT-resource change: document the subagents chat-model argument and the sub-agent-as-tool flow once the internal sub-agent lands and the API stabilizes.

- [x] Yes
- [ ] No

Generated-by: Qoder 1.29.0 (Qwen3.8-Max)


pltbkd:
This is a follow-up PR of the subagent framework, allowing the LLM to offload some work to the subagents. Could you take a look when you have time? @weiqingy @alnzng

python/flink_agents/api/subagent.py
```diff
@@ -15,18 +15,68 @@
...
 if TYPE_CHECKING:
     from flink_agents.api.runner_context import RunnerContext
 
 _LOG = logging.getLogger(__name__)
 
+# Prefix of the callable name a sub-agent is exposed to a chat model under.
+# Tools are forbidden to register under this prefix, so a prefixed callable
+# name unambiguously addresses a sub-agent and the executing side routes it to
+# the AGENT namespace.
+CALLABLE_NAME_PREFIX = "subagent_"
```
alnzng: if this is intended to be internal only, I would prefer using "_subagent_" as prefix to keep consistent with existing patterns, e.g. `_TOOL_CALL_CONTEXT` in chat_model_action.py

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
```diff
@@ -104,11 +116,54 @@ public void open() throws Exception {
...
+            if (inputSchema == null) {
+                // Unlike a bridge handle this is a sub-agent the caller could have described, so
+                // it is dropped with a warning rather than failing the job: the rest of the
+                // callables stay usable.
+                LOG.warn(
+                        "Sub-agent {} declares neither an input schema nor an input type, so there"
+                                + " are no arguments for the model to build a call from and it is"
+                                + " not offered as a callable.",
+                        name);
+                continue;
```
alnzng: If empty input schema is not expected, should we fail it directly? It seems to be this drop the sub-agent silently(with a warning). also is this behavior consistent with Python API?

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
```diff
@@ -104,11 +116,54 @@ public void open() throws Exception {
...
+            Preconditions.checkState(
+                    resource instanceof SubagentSetup,
+                    "Sub-agent %s must resolve to a SubagentSetup, but was %s",
+                    name,
+                    resource.getClass().getName());
+            SubagentSetup setup = (SubagentSetup) resource;
+            String inputSchema = setup.getInputSchema();
+            if (inputSchema == null) {
+                // Unlike a bridge handle this is a sub-agent the caller could have described, so
+                // it is dropped with a warning rather than failing the job: the rest of the
```
alnzng: I wonder why we decided to drop sub-agent silently, not fail the job directly.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
```diff
@@ -197,21 +218,32 @@ private static void executeParallel(
             Map<String, Boolean> success,
             Map<String, String> error,
             Map<String, ToolResponse> responses) {
-        List<DurableCallable<ToolResponse>> callables = new ArrayList<>(executions.size());
+        // Sub-agent calls already run through durable execution inside the setup, so they stay
+        // synchronous here and only the tool calls enter the durable batch.
+        List<ToolCallExecution> toolExecutions = new ArrayList<>();
         for (ToolCallExecution execution : executions) {
+            if (execution.agent != null) {
+                dispatchAgentExecution(execution, ctx, success, error, responses);
```
alnzng: It looks like all the sub-agent tool calls not run parallel, this is different with the regular tool call execution, right? any reason behind this?

alnzng (commented):
Thanks for making the changes.

python/flink_agents/api/subagent.py
yunfengzhou-hub: Done — renamed to `_subagent_` in both languages.

api/src/main/java/org/apache/flink/agents/api/chat/model/BaseChatModelSetup.java
yunfengzhou-hub: It was supposed that we may have some built-in sub-agents that may not be exposed to LLM, which means they do not need schemas. But this assumption is not that solid so I agree to reject this corner case for now.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
yunfengzhou-hub: Thanks for the reminder. This code was developed before the batch tool-call feature, along with #938, so we missed this feature. I've added support for parallel tool calls.

python/flink_agents/plan/agent_plan.py
```diff
@@ -293,6 +293,22 @@ def _to_plan_function(func: ApiFunction) -> PythonFunction | JavaFunction:
     raise TypeError(msg)
 
 
+def _check_tool_name_not_reserved(name: str) -> None:
```
weiqingy, line 296: This guard runs at `:341` for the decorator and `:386` for `add_resource`, but I don't see it on the MCP discovery loop at `:486-492`, which registers each tool under the name the remote server chose (`name=tool.name`).

Java looks covered. MCP discovery there goes through `addResourceProvider`, which calls `checkToolNameNotReserved` (`AgentPlan.java:694`). I walked the other Python TOOL paths, and YAML funnels into `add_resource`, so this looked like the only gap. I may have missed an entry point though.

So a server advertising `_subagent_lookup` registers a TOOL, and dispatch then routes it to the AGENT namespace where it can never resolve. Contract 6 says a tool under the prefix is rejected at plan-construction, which holds when the name is ours to choose. Here it isn't.

Adding `_check_tool_name_not_reserved(tool.name)` to that loop would close it, but then one badly named remote tool fails the whole plan. Which would you prefer, failing at discovery, or skipping just that tool with a warning so the rest of the server still works?

plan/src/test/java/org/apache/flink/agents/plan/actions/ToolCallActionSubagentTest.java
```diff
@@ -0,0 +1,673 @@
...
+                                "function",
+                                Map.of(
+                                        "name",
+                                        callableName,
+                                        "arguments",
+                                        Map.of("prompt", "review the diff")))));
+    }
+
+    /** One request carrying two sub-agent calls, so the batched path has more than one to run. */
+    private static ToolRequestEvent twoSubagentRequest(String first, String second) {
```
weiqingy, line 401: Both tests that enable the batched path use this builder, so every call in them is a sub-agent and `toolExecutions` ends up empty. `recordOutcome(toolExecutions.get(i), outcomes.get(i), ...)` at `ToolCallAction.java:251` therefore never runs against a non-empty list.

That pairing is what keeps each tool's result on its own call id once sub-agents are filtered out of the batch. If the two lists drift, one call's result lands on another call's id and both tests still pass.

A mixed call also seems like the common case, since a model with both tools and sub-agents available can call one of each in a turn.

Would a third parallel case, one `_subagent_` plus one tool, checking each id gets its own result, be worth adding? `withParallelToolCalls()` and `durableExecuteAllAsync` are already in the fixture, so it looks mostly like a new request builder.

api/src/main/java/org/apache/flink/agents/api/subagent/InputSchemas.java
```diff
@@ -0,0 +1,89 @@
...
+        if (type == null || type == Object.class) {
+            return null;
+        }
+        JsonNode schema = render(type);
+        return "object".equals(schema.path("type").asText()) ? schema.toString() : null;
+    }
+
+    private static JsonNode render(Class<?> type) {
+        try {
+            return MAPPER.generateJsonSchema(type).getSchemaNode();
```
weiqingy: I ran the pinned jackson-databind 2.18.2 (`pom.xml:48`) and pydantic 2.11.4 over the `Review` type this PR mirrors on both sides.

Java:

```
{"type":"object","properties":{"path":{"type":"string"},"lines":{"type":"integer"}}}
```

Python:

```
{"properties":{"path":{"title":"Path","type":"string"},"lines":{"default":0,"title":"Lines","type":"integer"}},"required":["path"],"title":"Review","type":"object"}
```

The two sides express `required` differently. Jackson's legacy generator writes it per property, and only under `@JsonProperty(required = true)`. Nothing on this path sets that, so Java marks no field required at all. Pydantic writes the object-level `"required":["path"]`. Same declaration, two different messages to the model about which arguments it must send.

The mirror tests check `type` plus the two property types (`SubagentSetupTest.java:156-166`, `test_subagent.py:98-104`), which is exactly where the two outputs agree. So contract 12 passes while the schemas differ on the field that changes what the model sends.

Is that deliberate? If not, would you rather fill `required` from the non-defaulted properties on the Java side, or have both mirror tests assert the whole schema so the next drift fails a build?

nit: from the same run, a `byte[]` field renders in Java as `{"type":"array","items":{"type":"byte"}}`, and `byte` is not a JSON Schema type. Python gives `{"type":"string","format":"binary"}`.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
```diff
@@ -344,6 +387,151 @@ private static void recordInlineResponse(
...
+            // wrapping the call again here would nest durable cursors.
+            SubagentResult result = execution.agent.submit(ctx, execution.agentArguments).await();
+            recordAgentResult(execution, result, ctx, success, error, responses);
+        } catch (InterruptedException e) {
+            // A cancellation, not a sub-agent failure: propagate it exactly like the tool paths do
+            // (#1111) so the caller skips sendEvent instead of folding the cancellation into a
+            // tool-error response and driving a further chat call off it.
+            Thread.currentThread().interrupt();
+            throw e;
+        } catch (Exception e) {
```
weiqingy, line 447: Both sub-agent dispatch paths catch `InterruptedException` then `Exception`, but neither has a `catch (Error e)`. The tool paths do, at `:264` and `:327`.

`ToolResultUtils.requireJsonCompatible` walks `Map`, `List` and arrays with no cycle guard and no depth limit (`ToolResultUtils.java:120-143`), so a result that contains itself recurses until the stack runs out. That `StackOverflowError` is an `Error`, so it passes these catches and fails the job. Python's same recursion raises `RecursionError`, which is an `Exception`, so dispatch absorbs it into a `ToolResponse` error. Contract 9 says an inexpressible result is reported as a failed delegation, not made a job failure.

`InputSchemas.render` already carries a `catch (StackOverflowError)` for this same shape (`InputSchemas.java:74-87`).

I can't say how likely a cyclic result is in practice. Neither suite has a cycle or a deep-nesting case though, and the body calls result normalization the highest-risk area. Would the same clause here work, or would you rather thread an identity set through `requireJsonCompatible`?

api/src/main/java/org/apache/flink/agents/api/subagent/SubagentSetup.java
```diff
@@ -19,22 +19,105 @@
...
  * resource.
  */
 public abstract class SubagentSetup extends SerializableResource {
 
+    /**
+     * Prefix of the callable name a sub-agent is exposed to a chat model under. Tools are forbidden
+     * to register under this prefix, so a prefixed callable name unambiguously addresses a
+     * sub-agent and the executing side routes it to the {@code AGENT} namespace.
+     */
+    public static final String CALLABLE_NAME_PREFIX = "_subagent_";
```
weiqingy, line 40: nit: the prefix is `_subagent_` here, but the description still says `subagent_` throughout, including the runtime-flow paragraph and contract 1.

Two behavior claims also read the other way now. Contract 2 says a sub-agent with no usable schema "is not offered, and does not stop the others", and the table row says "dropped with a warning; job continues". `BaseChatModelSetup.java:149` fails `open()` through `checkState` instead, and `chat_model.py:451-457` raises to match.

Worth a pass over the description before merge? It is the part people read later, and the contract list is detailed enough that the stale rows stand out.

weiqingy (commented):

python/flink_agents/plan/agent_plan.py
yunfengzhou-hub: Fixed. the MCP discovery loop now runs `_check_tool_name_not_reserved(tool.name)` and raises, so a remote `advertising _subagent_*` fails at plan-construction. I went with failing at discovery rather than skipping with a warning.

plan/src/test/java/org/apache/flink/agents/plan/actions/ToolCallActionSubagentTest.java
yunfengzhou-hub: Added — `parallelDispatchKeepsToolAndSubagentResultsOnTheirOwnIds` runs one `_subagent_` call alongside two tools under withParallelToolCalls(), and asserts call-1/2/3 each land on their own result ("agent-result", "alpha called", "beta called").

api/src/main/java/org/apache/flink/agents/api/subagent/InputSchemas.java
yunfengzhou-hub: Updated. I fixed the behavior in Java so that it now returns a json schema with the same `required` property like Python. It also renders `byte[]` as `{"type":"string","format":"binary"}` instead of an array of the non-standard `byte` type. Both mirror tests now assert `required` too, so a future drift fails the build.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
yunfengzhou-hub: I've added both guarantees. One the one hand, `ToolResultUtils.requireJsonCompatible` checks for pontential loops and fail early in that case. On the other hand, `catch StackOverflowError` is also added to invoker's try-catch blocked.

api/src/main/java/org/apache/flink/agents/api/subagent/SubagentSetup.java
yunfengzhou-hub: Done. The description has been updated to match the latest code behavior.

alnzng (approved):
LGTM. 

Let's please add end-to-end example and docs to explain the usage of sub-agents.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
```diff
@@ -384,6 +426,180 @@ private static void recordInlineResponse(
...
+            throws Exception {
+        if (result.isSuccess()) {
+            success.put(execution.id, true);
+            responses.put(
+                    execution.id,
+                    ToolResponse.success(
+                            ToolResultUtils.toChatMessageContent(
+                                    ToolResultUtils.normalizeAgentResult(
+                                            result.getResult(), execution.agent.getResultType()))));
+            ExecutionReporters.succeeded(
```
weiqingy, line 577: With the new operational metrics on main, a sub-agent call is reported as a tool named `_subagent_<name>`. The tool metrics only keep names of registered tools (`ToolExecutionMetricRecorder.java:64-67`), and a `_subagent_` name can never be one. So every sub-agent call is counted as `tool=unknown`, together with tool names the model made up. These calls also never report a start time, so they get no latency. Python looks the same.

For someone watching a dashboard, sub-agent calls and bad tool calls would look alike, and what are likely the slowest calls would be missing from latency. How would you like sub-agent calls to show up? For example, they could report a start time and get their own group. I found this by reading the code and haven't run it, so please correct me if I missed a path.

weiqingy (commented):
Thanks for working through these. The MCP check, the mixed parallel test and the cycle handling look good to me. One new question inline about metrics, plus a follow-up on the schema thread.

api/src/main/java/org/apache/flink/agents/api/subagent/InputSchemas.java
weiqingy: Thanks, the top-level fields match Python now.

I tried a few more input types and found two cases where Java and Python still give different schemas:

- A nested object gets no `required` list in Java (`InputSchemas.java:130-133`). The model might then skip a field Java needs, and the sub-agent would get a null. A `byte[]` inside a nested object also still shows up as an array of `byte`.
- Some fields are marked required when they shouldn't be. Take `boolean isActive`: the schema calls it `active`, but the check at `:141` looks for `isActive`, doesn't find it, and marks it required. The same happens to a field renamed with `@JsonProperty` and to a getter with no field.

What do you think about matching on the schema name and running the same step on nested objects? If nested types feel out of scope for this PR, narrowing contract 12 to top-level fields would work too.

api/src/main/java/org/apache/flink/agents/api/subagent/SubagentSetup.java
weiqingy: Thanks, the description matches the code now.

nit: the API section says "One thing changes" for existing callers. A repeated tool name now also fails `open()` (`BaseChatModelSetup.java:121`), where before it was accepted. Would it be worth listing that too, so someone upgrading isn't surprised?

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
yunfengzhou-hub: Took your suggestion. A resolved sub-agent call now reports a start time and lands in its own `subagent` group keyed by the registered name, with success and failure counts and latency. A `_subagent_` name that resolves to no registered sub-agent stays under `tool=unknown` with the made-up names. Both the sequential and the batch path emit the start event, and Python mirrors it.

api/src/main/java/org/apache/flink/agents/api/subagent/SubagentSetup.java
yunfengzhou-hub: Listed. The API section now says two things change: the `_subagent_*` rejection, and a repeated tool name failing `open()` where it was previously accepted.

api/src/main/java/org/apache/flink/agents/api/subagent/InputSchemas.java
yunfengzhou-hub: Took the first option: match on the schema name, and run the same step on nested objects and array items. Both cases now agree with Python, and the mirror tests on both sides cover them.

yunfengzhou-hub:
> LGTM.
> 
> Let's please add end-to-end example and docs to explain the usage of sub-agents.

@alnzng Thanks for the comment. I tested the end-to-end workflow of using sub-agents in an online Flink cluster, and found a bug related to the serialization process of the subagent framework. I plan to add the end-to-end example and the usage docs in a follow-up PR, once the bugfix #1132 stabilizes the sub-agent setup API that a Java example compiles against.

plan/src/main/java/org/apache/flink/agents/plan/actions/ToolCallAction.java
weiqingy: Thanks for resolving the comments. Nothing further from me.

Issue #1122, opened 2026-09-15
[Tech Debt][API] Make asynchronous durable calls composable


wenjin272:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

This is a child issue of #1055. It tracks a review of the public asynchronous durable execution APIs before the 0.4 compatibility baseline.

#### Background

Java and Python currently expose substantially different models for asynchronous durable execution.

Java exposes a synchronous-looking API even for asynchronous execution:

```java
<T> T durableExecuteAsync(DurableCallable<T> callable) throws Exception;
```

The runtime submits the callable to a worker pool and uses a Continuation to yield the Action, but the caller receives the final value rather than a future. As a result, individual calls cannot be created first and then composed into a concurrent batch.

Python returns an awaitable:

```python
result = await ctx.durable_execute_async(
    func, *args, durable_id="logical-call-id"
)
```

The current `AsyncExecutionResult` is deferred: execution is submitted when it is awaited. It does not support standard `asyncio.gather()`. Parallel execution therefore uses the separate `durable_execute_all_async()` API introduced in #926, which requires users to construct `DurableCall` descriptors.

The APIs differ not only syntactically, but also in what "async" means:

- Java hides suspension internally and returns `T`.
- Python exposes a deferred awaitable.
- Batch execution uses a second invocation form rather than composing single-call handles.

This also complicates the `durable_id` discussion in #1016: the single-call Python API already accepts `durable_id`, while the public `DurableCall` batch descriptor needs a separate way to carry or wrap the same identity.

#### Proposed direction

Make an asynchronous durable invocation a first-class, composable handle in both languages while keeping the surface idiomatic for each language.

Illustrative Python API:

```python
first = ctx.durable_execute_async(foo, x, durable_id="foo-x")
second = ctx.durable_execute_async(bar, y, durable_id="bar-y")

outcomes = await ctx.gather(first, second)
```

`DurableCall` may remain an internal batch-planning representation, but users should not need to construct it. Each single-call handle already carries its recovery identity and invocation arguments.

Illustrative Java API:

```java
DurableFuture<T> first = ctx.durableExecuteAsync(call1);
DurableFuture<T> second = ctx.durableExecuteAsync(call2);

List<Outcome<T>> outcomes =
        ctx.await(ctx.gather(List.of(first, second)));
```

For example, the corresponding contracts could be:

```java
[...]

Issue #1016, opened 2026-08-16
[Tech Debt] Align Java and Python recovery model


da-daken:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Description

Following the issues discovered in [#926](https://github.com/apache/flink-agents/pull/926), the tool recovery models for Java and Python are currently inconsistent:

- Java: ("tool-call", "")
- Python: (func_id, digest(args))

According to AGENT.md, the two languages need to maintain consistency. After unifying to a single recovery model, there is an upgrade cost to consider. During failure recovery, mismatches may occur, which could lead to re‑execution. This point may need to be documented or clarified somewhere.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

da-daken:
@weiqingy @wenjin272 Before starting this work, I think we can first align on the recovery model. I believe that for recovery, the function only needs to maintain a unique `ID`. Currently, matching is done using both the `ID `and the `argsDigest`. The `argsDigest `is essentially an enhancement added by the system to account for the possibility of user operation errors. However, in practice, even if the parameters and the method remain the same, if the data source of that method has changed, it should logically be re‑run. But currently, both the Python side and the Java side can still incorrectly recover in such cases. Fundamentally, we need to maintain a unique ID for all persistent methods within a single action for matching purposes. The multiple tool IDs returned by the LLM are naturally unique; other built‑in actions, such as the chat method of chatAction, are called only once and are thus also unique. For user‑invoked calls, we will hand over control to the user. WDYT?

wenjin272:
I think using a unique ID as the recovery identifier makes sense. That said, I’m not very familiar with the Durable Execution mechanism. @joeyutong  and @pltbkd, could you also take a look and share your thoughts?

pltbkd:
Yes — this problem also surfaced in the sub-agent framework work, and it was discussed in #926. The durable id mechanisms of Java and Python differ, and Python's original mechanism cannot distinguish multiple lambdas captured by the same method. Given the different iteration needs on each side, and to keep coordination cost low, we did not previously force the two onto the same scheme. But now both PRs are merged, and the pre-0.4 API cleanup is coming, so I think this is a good moment to address it.

[...]

PR #1127, opened 2026-09-16
[api][runtime][java][python] Add composable durable futures

wenjin272:
Linked issue: #1122

### Purpose of change

This makes asynchronous durable calls composable in Java and Python without requiring users to construct a separate batch-call descriptor.

#### Runtime flow

1. `durableExecuteAsync` / `durable_execute_async` creates a cold, runtime-owned `DurableFuture`; creation neither reserves a durable slot nor starts the callable.
2. Resolving a single handle through Java `ctx.await(...)` or Python `await` delegates to the existing recovery state machine, which replays a terminal result, reconciles or re-executes a pending call, or reserves and executes a new call.
3. Resolving `ctx.gather(...)` plans the complete ordered batch, reserves all required pending slots, and only then submits uncached calls.
4. Gathered outcomes complete the child handles in input order, so a later await replays the locally cached value or exception without consuming another durable slot.

#### Key decisions

- `DurableFuture` is an opaque, context-owned handle rather than a `java.util.concurrent.Future`. Java waits go through `ctx.await()` so JDK 21 execution can yield through the runtime continuation instead of blocking with `Future.get()`.
- Both language APIs remain deferred. This preserves the reserve-before-execute invariant; eager submission would require a separate reservation protocol.
- `gather` accepts only unique, unresolved single-call handles created by the same context. This prevents ambiguous slot ownership and reuse.
- The old public batch descriptor and batch execution APIs are removed; their recovery and parallel execution machinery remains internal.
- Durable identity matching is unchanged. The broader identity model remains tracked by #1016.

### Behavioral Semantics

| Interaction | State | Behavior |
| --- | --- | --- |
| Await one handle | New call | Reserve, execute through the async runtime path, persist, and cache locally. |
| Await one handle | Terminal or pending recovery state | Replay the terminal outcome, or reconcile/re-execute the pending call. |
| Await a gathered handle | Mixed cached, pending, and new calls | Preserve input order; reserve every required new slot before starting any uncached call. |
| Await a resolved child | Locally completed by `await` or `gather` | Return or rethrow the cached outcome without advancing durable state. |

Behavioral contracts:

- Creating or abandoning a handle has no execution side effect; this change does not add action-end validation for unresolved handles.
- Single-call failure is rethrown. Gathered callable failures are returned as per-call `Outcome.failure` values.
- `InterruptedException` remains a cancellation signal in Java: it is propagated, the interrupt flag is restored, and interrupted/unstarted slots remain pending.
- On batch timeout, started calls are finalized according to their outcomes while unstarted calls remain pending for recovery.
- A serialization/finalization failure is returned as a failure outcome and leaves that slot pending.
- JDK 21 uses the continuation-backed async path; older JDKs retain the synchronous fallback. Python handles remain directly awaitable but do not become general-purpose `asyncio` futures.

### Tests

| Contract | Coverage |
| --- | --- |
| Cold and reusable handles | `testDurableExecuteAsyncCreatesDeferredReusableHandle`; `test_flink_runner_context_async_future_is_deferred_and_reusable` |
| Reserve before execute, ordering, and child completion | `testGatherReservesBatchBeforeExecutionAndCompletesChildren`; `test_flink_runner_context_gather_reserves_before_execution` |
| Replay, reconciliation, and partial recovery | Java and Python gather recovery/reconciler suites in the runtime context tests |
| Failure, timeout, interruption, and pending-slot behavior | Java gather timeout/interruption/finalization tests and Python gather failure/timeout/finalization tests |
| Production tool-call migration | Java and Python `ToolCallAction` tests, including parallel business and infrastructure failures |

Verified:

- Java plan tests: 52 passed.
- Java runtime/context compatibility tests: 44 passed.
- Python non-integration suite: 1347 passed, 13 skipped, 172 deselected.
- `./tools/lint.sh --check`.
- `./tools/build.sh --java` on JDK 17: all 35 modules built successfully with tests skipped.
- JDK 21 runtime multi-release and test-source compilation.

Not verified: external-service E2E execution, live Flink checkpoint recovery, and execution of the JDK 21 continuation path (compile-only in this environment).

<details>
<summary>Implementation invariants and evidence</summary>

- Runtime-only future implementations keep callables and internal batch plans out of the public API.
- Java and Python validate owner, handle kind, duplicate inputs, and already-resolved inputs before constructing a gathered handle.
- Existing tool, chat-model, context-retrieval, subagent, memory, E2E, and documentation call sites were migrated to direct await or gather composition.

</details>

### API

This is a breaking pre-0.4 Java API change: `durableExecuteAsync` now returns `DurableFuture<T>`, and immediate callers migrate to `ctx.await(ctx.durableExecuteAsync(...))`. Java adds `await` and `gather`, and removes `durableExecuteAllAsync`.

Python direct-await usage remains source-compatible. Python adds public `DurableFuture` and `ctx.gather(...)`, removes public `DurableCall` / `AsyncExecutionResult`, and removes `durable_execute_all_async` in favor of composing single-call handles.

### Documentation

- [ ] `doc-needed`
- [ ] `doc-not-needed`
- [x] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Codex (GPT-5)


wenjin272:
@pltbkd @da-daken @joeyutong Would you mind taking a look at this PR at your convenience?

api/src/main/java/org/apache/flink/agents/api/context/RunnerContext.java
```diff
@@ -146,11 +146,11 @@ default boolean hasResource(String name, ResourceType type) {
...
 
     /**
-     * Asynchronously executes the provided callable with durable execution support.
+     * Creates a deferred asynchronous durable call.
      *
-     * <p>On JDK 21+, this method uses Continuation to yield the current action execution, submits
-     * the callable to a thread pool, and resumes when complete. On JDK &lt; 21, this falls back to
-     * synchronous execution.
+     * <p>Creating the returned handle does not execute the callable or reserve durable state. Use
+     * {@link #await(DurableFuture)} to execute a single call, or {@link #gather(List)} to compose
```
da-daken, line 152: Would it be better to add a description about parallelism?

da-daken (approved):
LGTM! Just a non-blocking comment—take a look when you get a chance.

pltbkd:
Hi @wenjin272, thanks for the PR — overall LGTM.

I have a suggestion about the API. Do you think it's better to put `await` in `DurableFuture`, like `SubagentFuture`? It's close to the conventions of both Java and Python, and using `await` rather than `get` avoids misleading users into thinking this is a standard future.

`gather` can stay on the context, consistent with Python. Also, with this change `requireOwnedFuture` can live inside `GatherDurableFuture`, so we don't have to introduce that method and its check on the context. WDYT?

btw, `java.util.Set` shows up inline in the code — maybe we should add something to the AGENTS guide, or a format enforcer?

PR #1138, opened 2026-09-18
[api][plan][runtime][python] Support registering an Agent as an in-process internal sub-agent

### LINKED ISSUE #1137: [Feature] Support in-process internal sub-agent
### Search before asking

#### Background

Since #938 / #1082, an in-process ("deferred") sub-agent is authored by hand-writing a `DeferredSubagentSetup` subclass that implements `prepare()`. That setup is a single deferred callable: it has no action loop, no events, and no resources of its own. Reusing an existing, fully-formed `Agent` as a sub-agent therefore means re-expressing its logic inside a deferred setup, and a sub-agent cannot itself own sub-agents.

Sub-agent registration is deliberately uniform: an `AGENT` resource is declared via `addResource(name, ResourceType.AGENT, instance)` or the YAML `subagents:` block, with no annotation (see #909). An internal sub-agent should not introduce a second registration surface.

#### Goal

Allow a plain `Agent` instance registered as an `AGENT` resource to be run by the framework as an in-process **internal sub-agent** that executes its own action loop, so an existing agent can be reused as a child without rewriting it as a deferred setup. The caller invokes it like any other sub-agent and receives its accumulated output.

Illustrative caller API (shape open for discussion):

```python
class RootAgent(Agent):
    @action(InputEvent.EVENT_TYPE)
    @staticmethod
    async def orchestrate(event: Event, ctx: RunnerContext) -> None:
        prompt = InputEvent.from_event(event).input
        child = ctx.get_resource("reviewer", ResourceType.AGENT)
        future = await child.submit(ctx, prompt)   # no explicit ids
        result = await future
        ctx.send_event(OutputEvent(output=result.result[0]))

root = RootAgent()
root.add_resource("reviewer", ResourceType.AGENT, EchoChildAgent())
```

The behavior the feature must provide:

- **Isolation** between parent and child: the child's events, resource resolution, and memory stay within its own scope and do not leak into the parent.
- **Durability**: a call survives failover — its identity and result are recovered deterministically rather than re-executed to a different outcome. The no-id `submit(ctx, prompt)` form is the intended public API; explicit ids stay an internal concern.
- **Composition**: a child may itself register and await its own sub-agents (nesting).
- **Failure containment**: a failing child surfaces to the caller as a failed result rather than failing the whole job.
- **Cross-language**: a child agent written in one language can be driven by a caller in the other (relates to umbrella step 6).

#### Semantics to define

The design should explicitly decide and document:

- The isolation boundary: which of memory, action state, resource scope, and emitted events are isolated, and how the child's output is returned to the caller.
- How a call is identified, and how that identity is derived and replayed on recovery; how it relates to the durable-identity model in #1016.
- Whether the child runs on the parent's mailbox/operator or is dispatched separately, and how the caller suspends and resumes while awaiting.
- Nesting depth / recursion limits and cycle prevention (e.g. a child registering its own parent).
- Failure, timeout, and cancellation propagation between parent and child.
- Interaction with #1131 (plan rebuild across the JobManager→TaskManager boundary) for both Java- and Python-authored children.
- JDK constraints for a Java caller awaiting in-process, and the fallback.

#### Compatibility

Additive: internal sub-agents are a new way to use an existing `AGENT` resource and should not change the deferred/async setup APIs or the registration surface. Whether this lands in 0.4 or is deferred is a maintainer scoping decision under #1134.

#### Acceptance criteria

- A plain `Agent` registered as an `AGENT` resource can be invoked as an in-process sub-agent and returns its accumulated output to the caller.
- Parent/child isolation is demonstrated for emitted events, resource scope, and memory.
- A sub-agent call recovers deterministically after a failover, with no duplicate or divergent execution.
- Nesting works, and a failing child is reported through the caller's result without failing the job.
- Cross-language invocation (Java↔Python) is covered.
- Tests cover the happy path, isolation, failure containment, nesting, and recovery.
- User documentation and examples are added for both Java and Python.

yunfengzhou-hub:
Linked issue: #1137

### Purpose of change

Registering an `Agent` instance as an `AGENT` resource now runs it as an in-process **internal sub-agent**: the caller resolves it via `ctx.get_resource(scope, AGENT)`, calls `submit(ctx, prompt)`, and awaits a `Result` holding the child's accumulated output — reusing an existing agent as a child instead of hand-writing a `DeferredSubagentSetup`. The child runs its own action loop, isolated and durable, and may nest.

#### Runtime flow

1. **Compile.** `AgentPlan` splits an `AGENT` resource into three exclusive cases — `SubagentSetup`, YAML descriptor, `Agent` instance. Only the `Agent` case compiles, via `InternalSubagentCompilationHelper`: the child plan is compiled once, shared across names registering the same child, with cycles and self-references rejected, and served by `InternalSubagentProvider`. A Python-authored child travels as a `PythonSerializableResourceProvider`.
2. **Materialize.** The provider builds `InternalSubagentSetup` (Java `extends BaseDeferredSubagentSetup`, Python subclasses `DeferredSubagentSetup`) holding child plan and scope. Per-call statuses, per-scope child caches, and the per-key session index live in the setup, not the operator.
3. **Submit + bootstrap.** `submit(ctx, prompt)` assigns `(session_id, call_id)` from the executing task's namespace at await time, so a replay awaiting in the same order re-derives the same ids, and returns a deferred `SubagentFuture`. On first resolve `prepare` registers the call status and sends one `InternalSubagentCallEvent` on the mailbox thread, outside the durable boundary so a replayed send carries identical attributes.
4. **Dispatch + suspend.** The callable waits off the mailbox thread — a Java caller on a JDK 21 continuation, a Python coroutine at `await`. The operator matches the child's actions to the envelope and runs them scope-aware (`ActionTaskContextManager` binds the child's `SubagentScope`; child resources resolve against the child plan, including a nested sub-agent) with isolated memory and a child `ResourceCache` inheriting the root.
5. **Accumulate + resume + recover.** Child `OutputEvent`s accumulate into the call status; on quiesce the operator completes its future and the caller resumes with `SubagentResult.ok`. `findCallStatus` recurses through child caches so nested envelopes resolve depth-agnostically. On failover the replayed bootstrap reproduces the identity, so the child replays its recorded output; `onRecordFinished(key)` drops that key's sessions.

#### Key decisions

- **Compile the child `Agent`** rather than require a `DeferredSubagentSetup` subclass: that boilerplate is what this removes, and a real plan and action loop compose (nesting) where a single deferred callable cannot.
- **Opaque deferred handle, not a raw future.** Awaiting must release the mailbox so the operator can dispatch the child; a `CompletableFuture.get()` would block the mailbox and deadlock.
- **Identity assigned at await, not minted at send**, so failover re-derives it and the replayed envelope resolves the child's persisted state instead of re-running it.
- **Orchestration state in the setup, not the operator**, keeping per-scope and per-key state off the operator's shared dispatch path.

#### Interaction decisions

| Caller context | On resolve |
| --- | --- |
| Java action, JDK 21 continuation | Caller suspends on the continuation; the mailbox dispatches the child; caller resumes with the `Result`. |
| Java action, no stackful suspension (≤ JDK 17) | `prepare` fails fast with `IllegalStateException` rather than blocking the mailbox and deadlocking. |
| Python `async` action | `await` releases the mailbox; a Python-compiled child's lazily-built handle adopts the prepared namespace, so no-id `submit` mints deterministic ids. |
| Python synchronous action | Holds the mailbox for its whole pemja call and cannot yield, so awaiting an internal sub-agent would deadlock; the caller action must be `async`. |
| Any caller, after failover | Replayed bootstrap carries identical attributes; the child replays its recorded output and ids re-derive identically. |

#### Behavioral contracts

- An `Agent` registered as an `AGENT` resource is invocable in-process; awaiting `submit(ctx, prompt)` returns a `Result` whose `result` holds the child's accumulated `OutputEvent` payloads.
- A child action that raises surfaces as a failed `Result` (`success=false`, `error_message`); the job continues.
- A child may register and await its own sub-agents; nested envelopes resolve to the right depth.
- The child is isolated: its own resource cache (inheriting the root), scope-aware context, and isolated memory; its events do not land on the parent context.
- Successive calls under one record run independently under distinct identities.
- A completed call is durable: after failover it resolves to the same identity and replays the child's recorded output.

#### Failure behavior

- Resolve without stackful suspension (Java caller): raises `IllegalStateException` (fail-fast, no deadlock).
- Python `prepare` with a context that is not an `InternalSubagentCallFactory`: raises `NotImplementedError`.
- Bootstrap send fails, or await for an unregistered `(sessionId, callId)`: raises `IllegalStateException`.
- No-id `submit` on a handle never wired to the task lifecycle: raises `RuntimeError` ("No prepared action task to assign sub-agent ids from"); the operator's Python-lifecycle gate keeps a Python caller driving a Python-compiled child from reaching it.
- Child action raises: captured into `SubagentResult.error(e)`; the caller's `Result` reports failure and the job proceeds.
- Unresolved handles at action finish: the deferred base enforces resolution; a failed invocation skips cleanup because the run fails and replays.

### Tests

| Contract | Coverage |
| --- | --- |
| `Agent` compiles to an internal provider; shared plan; cycles/self-reference rejected; other two `AGENT` cases unaffected | `childAgentCompilesIntoInternalProvider`, `sharedChildAgentCompilesToSinglePlan`, `cycleThroughRootIsRejected`, `selfReferenceIsRejected`, `nonSubagentAgentResourceIsRejected`; Python `test_agent_plan_subagent_resources.py` mirrors all |
| Caller receives accumulated output | `callerReceivesTheChildsAccumulatedOutput`; e2e `test_python_internal_subagent_call_end_to_end` |
| Child failure surfaces via `Result` | `childFailureSurfacesThroughTheResult`; e2e `test_python_internal_subagent_failure_surfaces_via_result` |
| Nesting; scope matching | `nestedCallReachesTheGrandchild`, `scopeMatchingAppliesTypeAndConditionFilters`; e2e `test_python_internal_subagent_nested_call` |
| Independent identities per call | `successiveCallsRunIndependentlyUnderDistinctIdentities` |
| Fail-fast without continuations | `resolvingWithoutContinuationsFailsFast`, `continuationsAreAvailableOnTheTestClasspath` |
| Durability / failover replay | `InternalSubagentRecoveryTest`; `replayOfCompletedActionEmitsPreparedAndReusedPair` |
| Cross-language lifecycle and no-id namespace replay | `TaskLifecycleListenerNotificationTest`; Python `test_task_lifecycle_bridge.py` |

Highest-risk paths — cross-language identity replay, mailbox suspension/deadlock, and cycle rejection — are pinned by dedicated Java and Python tests plus a real-Flink Python e2e (call, failure, nested); plan-compilation parity is asserted in both languages.

Verified: Java `ut-java` green (internal sub-agent, plan, lifecycle, continuation suites); Python non-integration suite green; internal sub-agent Python e2e green on a real Flink job; `./tools/build.sh --java` on JDK 17.

Not verified: live-cluster checkpoint/failover of an in-flight call (recovery is harness-level, not a real cluster restart); a Java caller awaiting end-to-end through the JDK 21 continuation on a live cluster; external-service e2e. Residual fork CI failures are pre-existing LLM-auth (HTTP 401) integration tests, unrelated to this change.

<details>
<summary>Implementation invariants, full test roster, and change size</summary>

- `prepare` splits a mailbox-thread `bootstrap` (registers `InternalSubagentCallStatus`, indexes the record key, registers the session owner on the shared context, sends the envelope) from an off-mailbox `call`/`_call` that waits on the status's response future and converts a child exception into `SubagentResult.error`.
- `awaitSubagentCall` unwraps `ExecutionException` so the caller's `Result` carries the failing child action's exception, not future plumbing.
- Cross-language no-id: `ActionExecutionOperator` registers the `PythonTaskLifecycleListener` when the root plan declares a Python-compiled sub-agent (`hasPythonCompiledSubagent()`), so `notify_action_prepared` reaches `FlinkRunnerContext`, which captures the prepared namespace and replays it onto the lazily-materialized handle via `adopt_prepared_namespace`; `SubagentIdAllocator` is scoped to one action execution so ordinals restart per action and replays re-derive the same ids.
- Full roster: Java `InternalSubagentCallTest`/`RecoveryTest`/`GuardTest`/`MatchTest`, `AgentPlanSubagentResourceTest`, `ActionTaskContextManagerTest`, `ContinuationSupportTest`, `TaskLifecycleListenerNotificationTest`; Python `test_agent_plan_subagent_resources.py`, `test_task_lifecycle_bridge.py`; e2e `internal_subagent_test.py`.
- Change size: 39 files, +2963/−66 across `api`, `plan`, `runtime` (Java), and `python`.

</details>

### API

No public signature changes. `Agent.addResource(String, ResourceType, Object)` is unchanged; for an `AGENT` resource it now also accepts an `Agent` instance (javadoc-documented widening, source-compatible) alongside the existing `SubagentSetup` and YAML descriptor forms, which are unaffected. Python `submit(ctx, prompt)` gains working no-id support for internal sub-agents; `submit_with_identity` remains the explicit-id hook. New `plan` and `runtime` classes are additive. A Java caller awaiting in-process requires JDK 21 (continuation); the Python caller path does not.

### Documentation

- [x] `doc-needed`
- [ ] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

Generated-by: Qoder 1.29.0 (Qwen3.8-Max)


purushah:
Thanks for picking this up. Before we get into line-by-line review, could we open a discussion for the runtime model? This PR settles a few things #909 left open, and they affect most of the operator changes.

The three I'd like agreement on:

1. Where in-flight call state lives across a checkpoint. Right now it's only on the heap and rebuilt when the parent replays, but child tasks are checkpointed before the parent is re-enqueued. On restore the child can run first, find no status, and the job loops.
2. Whether the parent should wait on a `ContinuationActionExecutor` thread. Children use the same pool, so enough concurrent calls on one subtask can starve it.
3. What "done" means for a child with more than one action. Replay only re-sends the child's output events, so a chained child comes back with partial output after failover.

I have more detailed notes and am happy to share them once the direction is set.

yunfengzhou-hub:
> Thanks for picking this up. Before we get into line-by-line review, could we open a discussion for the runtime model? This PR settles a few things #909 left open, and they affect most of the operator changes.
> 
> The three I'd like agreement on:
> 
> 1. Where in-flight call state lives across a checkpoint. Right now it's only on the heap and rebuilt when the parent replays, but child tasks are checkpointed before the parent is re-enqueued. On restore the child can run first, find no status, and the job loops.
> 2. Whether the parent should wait on a `ContinuationActionExecutor` thread. Children use the same pool, so enough concurrent calls on one subtask can starve it.
> 3. What "done" means for a child with more than one action. Replay only re-sends the child's output events, so a chained child comes back with partial output after failover.
> 
> I have more detailed notes and am happy to share them once the direction is set.

@purushah Thanks for the comment. Worth aligning on these before this moves further. The preview does have some known persistence gaps that we're addressing.

**1. In-flight state and recovery.** We'd like to settle the principle first and get your read. A sub-agent call inherits durable execution's consistency model:
- Without ActionState, an in-flight call tolerates recomputation and takes the final result.
- With ActionState, the call is consistent but requires the request to be replayable; there is no reconcile mechanism, and anything needing reconcile belongs in durable execution.

Mechanically, the parent's initiation and the child action's start each create a sub-agent call record, the later one linking to the earlier. On failover: without ActionState, in-flight calls are cancelled and re-run; with ActionState, the replayed action records serve as a reconciler for the call — when the parent has initiated but not completed, its request is matched against the persisted action records, completed actions are skipped and their results reused (including an already-produced OutputEvent, the sub-agent's response), and only the unfinished child actions re-run. Consistency therefore still rests on durable execution. I can write the full plan up separately if that's easier to react to.

**2. Shared continuation executor.** The call awaits on the async pool, so it does hold async threads; under heavy parallel fan-out where the children also need async capacity, this can block. We expect large fan-out to be rare, so the near-term guidance is to raise the async thread count, with auto-scaling as a later option.

**3. Multi-action child "done".** A call completes when all actions in its sub-agent scope quiesce — the same condition the main agent uses to finish a record: every output action is an OutputEvent and none is pending. Consistency follows point 1.

---

We'd like to continue discussions in this PR or #1137, given that after #909 we have created an umbrella issue for sub-agent mechanism to better track discussions and development in each child issue. What do you think?

PR #1144, opened 2026-09-18
[api][java][python] Reject null and non-positive limit in VectorStoreQuery

CodeTrainerMan:
Linked issue: #1145

#### User-visible outcome

Building a `VectorStoreQuery` with a `null` or negative `limit` is now rejected when the query is constructed, with an `IllegalArgumentException` (Java) or a pydantic `ValidationError` (Python) that names the parameter. Previously the Java path failed later with a bare `NullPointerException` at the query call site for `null`, and a negative limit was forwarded to the backend as an invalid request in both languages.

#### Runtime flow

Callers build a Java `VectorStoreQuery` through any of the four constructors; all four delegate to the fully specified one, which validates `limit` before assigning any field, so no partially built object escapes. `BaseVectorStore#query(VectorStoreQuery)` then unboxes `getLimit()` into the `int limit` parameter of `queryEmbedding` exactly as before; after this change the unboxing can no longer see `null`, and the value reaching the backend can no longer be negative. On the Python side `VectorStoreQuery` is a pydantic model and `limit` now carries `ge=0`, so pydantic rejects a negative value during model construction; `null` was already rejected there by the `int` annotation.

#### Key decisions

- Validate in the constructor rather than in `BaseVectorStore#query`: query objects are also built and passed across the Java/Python boundary, and rejecting at construction keeps an invalid value from being carried anywhere.
- Reject `null` instead of treating it as "unbounded": `BaseVectorStore#get`'s Javadoc advertises `limit=null` as unbounded, but `query(VectorStoreQuery)` has always unboxed it unconditionally, so accepting null and reading it as unbounded would be a new behavior, not a fix. Alternative rejected: defaulting a null limit to 10.
- Apply the same bound to the Python field so both languages agree, rather than fixing Java only.

#### Interaction decisions

| `limit` | Java | Python |
|---|---|---|
| `null` | `IllegalArgumentException`: `` `limit` must not be null. `` | `ValidationError` (unchanged: `int` annotation) |
| `< 0` | `IllegalArgumentException`: `` `limit` must not be negative, but was N. `` | `ValidationError` from `ge=0` |
| `0` | accepted | accepted |
| `> 0` | accepted, unchanged | accepted, unchanged |
| omitted | n/a | accepted, default `10` unchanged |

#### Behavioral contracts

1. A Java `VectorStoreQuery` cannot be constructed with a `null` limit, through any constructor.
2. A Java `VectorStoreQuery` cannot be constructed with a negative limit, through any constructor.
3. Both rejections raise `IllegalArgumentException` with a message naming `limit`.
4. Zero and positive limits construct successfully, and `getLimit()` returns the value passed.
5. A Python `VectorStoreQuery` cannot be constructed with a negative limit, accepts `0`, and still defaults to `10`.

#### Failure behavior

An invalid `limit` is the only new failure path, and it raises: the constructor throws and no object is created, so nothing falls back, retries, or is silently absorbed. No external-service error is involved - the check runs before any store is contacted. A `null` or negative limit arriving from configuration or from another component now surfaces at query construction instead of at query execution.

#### Contracts to tests

| Contract | Tests |
|---|---|
| 1 | `VectorStoreQueryTest.rejectsNullLimit` (covers all four constructors) |
| 2 | `VectorStoreQueryTest.rejectsNegativeLimit` |
| 3 | `VectorStoreQueryTest.rejectsNullLimit`, `rejectsNegativeLimit` (assert type and message text) |
| 4 | `VectorStoreQueryTest.acceptsZeroLimit`, `acceptsPositiveLimit` |
| 5 | `test_vector_store_query.py::test_rejects_negative_limit`, `::test_accepts_zero_and_positive_limit`, `::test_default_limit_unchanged` |

#### Coverage and what was not verified

The check is a pure constructor contract, covered from both languages at the boundary values (null, -1, 0, 1, omitted). Not verified: no test drives a real store backend through `BaseVectorStore#query`, so the claim that a negative limit used to reach the backend as an invalid request is read from the code path rather than observed against a live Chroma or OpenSearch instance; no test covers a `limit` deserialized from YAML or JSON configuration.

<details>

- All four Java constructors delegate to the six-argument one, and none assigns fields itself, so one check covers every entry point.
- The message text is asserted verbatim, so changing it breaks a test instead of quietly changing the contract.
- Before/after evidence: with the change stashed, 3 of the 4 Java cases fail (the null case with `NullPointerException`); with it applied, all pass. Full `api` module suite: 447 tests, 0 failures. Python: 3 passed.
</details>

#### Compatibility impact

`VectorStoreQuery` is public API in both languages. A caller that currently passes `null` or a negative limit and gets a late `NullPointerException` or a backend error now gets an `IllegalArgumentException` or `ValidationError` at construction. Everything else is unchanged: the signatures, the `Integer` type of `limit`, the Python default of `10`, zero and positive limits, and every other field.

### Documentation

- [ ] `doc-needed`
- [x] `doc-not-needed`
- [ ] `doc-included`

- [x] Yes
- [ ] No

Generated-by: CodeBuddy (Auto model)


python/flink_agents/api/vector_stores/vector_store.py
```diff
@@ -69,7 +69,9 @@ class VectorStoreQuery(BaseModel):
     query_text: str = Field(
         description="Text query to be converted to embedding for semantic search."
     )
-    limit: int = Field(default=10, description="Maximum number of results to return.")
+    limit: int = Field(
+        default=10, ge=0, description="Maximum number of results to return."
```
wenjin272: Should `0` really be accepted here? The production query path cannot honor it consistently: Elasticsearch, OpenSearch, and S3 Vectors coerce `0` to `1`, while Chroma rejects `n_results=0`. Please either require `limit > 0` in both Java and Python, or handle `0` at the common query entry point by returning an empty result, with a production-path regression test.

wenjin272 (changes requested):
Thanks for the contribution! The validation direction looks good overall. I left one concern about allowing `limit=0`, because the current backends do not honor that value consistently.

CodeTrainerMan:
Thanks for the review @wenjin272. I went with the first option: require `limit > 0` in both Java and Python (commit 66577096).

Reason: handling `0` at the common query entry point would only cover `BaseVectorStore#query(VectorStoreQuery)`. `queryEmbedding(...)` is a public abstract method that can still be called directly with `limit=0`, so on that path the backend inconsistency would remain (Elasticsearch / OpenSearch / S3 Vectors coerce `0` to `1`, Chroma rejects `n_results=0`). Rejecting `0` when the query is constructed also fails fast with a message naming the parameter, rather than silently returning an empty list when `limit` comes from a computed value.

Changes:
- Java: `VectorStoreQuery` rejects `limit <= 0` (previously `< 0`); the message is now "`limit` must be positive, but was N."
- Python: `VectorStoreQuery.limit` uses `gt=0` (previously `ge=0`).
- Tests in both languages now expect `limit=0` to be rejected alongside negative values; positive limits keep their current behaviour and the Python default of 10 is unchanged.

Verified locally: Java `api` module full test run (449 tests, 0 failures, spotless check included) and the Python limit-validation tests (4 passed).

Happy to switch to option B if the community prefers to keep `limit=0` as a legal "return nothing" semantic - in that case I will add the production-path regression test you mentioned.

wenjin272 (approved):

Issue #1119, opened 2026-09-15
[Feature][runtime] Support parallel action execution on JDK < 21


eemario:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/flink-agents/issues) and found nothing similar.

### Problem
On JDK < 21, `durableExecuteAsync` runs inline on the mailbox thread — a single 500ms LLM call blocks the entire operator instance. Other keys' input records, watermarks, and checkpoints are stalled. JDK 21+ avoids this via Continuation.yield (Loom), but most production Flink deployments are still on JDK 8/11/17.
### Proposed Solution
A coroutine-free parallel execution engine that runs action tasks on an elastic worker pool:
1. Worker pulls an action task from Flink keyed state under a global lock.
2. Sync phase (under lock): context setup → user action code runs until `durableExecuteAsync`.
3. Async phase (lock released): the worker thread blocks on the async call (LLM, I/O) without holding the lock — other workers and the mailbox can proceed.
4. Commit (mailbox thread): worker marks done → submits commit mail → mailbox drains completed tasks in FIFO order per key.
The engine is automatically gated: activates only when `!ContinuationActionExecutor.isContinuationSupported()` (JDK < 21) and all actions are pure Java. JDK 21+ and Python agents keep the existing engine. The `durableExecuteAsync` / `RunnerContext` API is unchanged.
Key guarantees preserved: per-key ordering (staged commits drain the queue in FIFO order), exactly-once (`maybePersistTaskResult` during mailbox commit), checkpoint correctness (quiesce in-flight tasks before barrier), and backpressure (via configured max-in-flight-input-records).
### Relationship to [#1109](https://github.com/apache/flink-agents/issues/1109)
[#1109](https://github.com/apache/flink-agents/issues/1109) fixes JDK 21+ busy-wait polling with completion callbacks; this proposal provides the JDK < 21 solution. The two are complementary — one addresses the Loom path, the other the no-Loom path.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wenjin272:
Thanks for taking this on @eemario. Assigned.

PR #1147, opened 2026-09-20
[api][runtime] Support parallel action execution on JDK < 21

eemario:
Linked issue: #1119

### Purpose of change

Execute plans whose actions are implemented in Java on coordinator workers when continuation-based execution is unavailable. Shared operator state remains protected by the parallel execution lock, while blocking asynchronous calls release the lock and restore task context after worker re-acquisition.

#### Runtime flow:

The mailbox thread admits each input record under the in-flight budget and appends its action tasks to the per-key queue. The coordinator grants workers the parallel execution lock in ascending (recordIndex, taskIndex) order, with absolute priority for the mailbox thread. A worker pulls its task from keyed state, prepares and runs the action; a blocking asynchronous call releases the lock before awaiting and restores the task context after re-acquisition. Commits happen on the mailbox thread, so keyed state is only ever mutated under the lock.

#### Key decisions:

- Worker-pool engine on JDK < 21, where continuation-based execution is unavailable; engine eligibility is validated at setup and ineligible plans fall back to the previous engine.
- Drain in-flight action tasks at prepareSnapshotPreBarrier instead of snapshotting them: pulled-but-uncommitted results live only in memory and would be lost on restore, so the barrier waits until the engine is quiesced and the snapshot captures no half-pulled intermediate state. Dispatch resumes only after the synchronous snapshot capture, because a worker woken any earlier could pull a queued task out of keyed state before it is captured. Restore semantics are therefore identical to the serial engine.
- Mailbox-priority lock, so checkpoint and commit work on the mailbox thread is never starved by workers.
- Bound concurrent input records with in-flight backpressure instead of unbounded admission.
- Align lifecycle notifications with the serial path: prepared/started/reused fire in execute(), finishing/finished/ failed fire in commit(), with noop-silent guards.
- Register the durable-execution context in the lookup map on setup so workers resolve it during replay.

#### Interaction decisions (engine selection):

| Plan | `parallel-execution.enabled` | Engine |
|---|---|---|
| pure-Java, JDK < 21 | true (default) | parallel |
| pure-Java, JDK < 21 | false | serial |
| coroutine engine available | any (ignored) | coroutine |
| contains Python actions | any (ignored) | serial |

#### Behavioral contracts:

- Same-key input records commit in input order.
- Fan-out siblings, broadcast listeners and multi-key inputs run concurrently; each action observes its own action-scoped context.
- At most max-in-flight-input-records (default 100) records are in flight; admission then blocks until an in-flight record retires.
- A checkpoint contains no in-flight execution state; after restore the engine behaves exactly as the serial path and exactly-once output is preserved.
- A failing action task fails the job on the mailbox thread without hanging.
- Task lifecycle notifications are indistinguishable from the serial path.

#### Failure behavior:

- Pull/prepare/run failures on a worker are captured and rethrown on the mailbox thread; the lock is always returned and no pool thread dies silently.
- A failed prepareSnapshotPreBarrier rolls the drain back via stopDraining and resumes dispatch, because such a failure never reaches snapshotState's stopDraining.
- Aborting an old checkpoint does not resume a drain belonging to the current checkpoint.
- Unsupported configurations (Python actions, coroutine engine) are not errors: the parallel options are ignored and execution falls back to the previous engine.

### Tests

| Behavioral contract | Tests |
|---|---|
| Same-key input records commit in input order | `ParallelExecutionE2ETest.sameKeyInputsCommitInOrder`; `ParallelExecutionTaskQueueTest` |
| Fan-out / broadcast / multi-key inputs run concurrently, each action keeping its action-scoped context | `ParallelExecutionE2ETest.fanoutRunsSiblingActionsInParallel`, `broadcastRunsAllListenersInParallel`, `multiKeyInputsRunInParallel`, `parallelSiblingsKeepActionScopedContext`; `ActionTaskContextManagerTest` |
| At most `max-in-flight-input-records` (default 100) records in flight; admission then blocks | `ParallelExecutionE2ETest.maxInFlightInputRecordsThrottlesAdmission`; `ParallelExecutionCoordinatorTest` |
| A checkpoint contains no in-flight state; after restore behavior equals the serial path with exactly-once output | `ActionExecutionOperatorDrainDeadlockIntegrationTest`; `ActionExecutionOperatorCapacityIntegrationTest.actionsBeyondWorkerCountRecoverFromCheckpointWithExactlyOnceOutput`; `ParallelExecutionE2ETest.tasksBeyondWorkerCountCompleteExactlyOnce`; `ParallelExecutionLockTest` |
| A failing action task fails the job on the mailbox thread without hanging | `ParallelExecutionE2ETest.asyncFailureFailsTheJob`, `failingSiblingFailsTheJobWithoutHanging` |
| Task lifecycle notifications are indistinguishable from the serial path | `TaskLifecycleListenerNotificationTest`; `ActionExecutionOperatorTest` |

Coverage by risk: checkpoint consistency and exactly-once recovery are pinned
by the drain-deadlock and capacity integration tests; concurrency and
action-scoped context by the E2E suite; ordering, backpressure and failure
propagation each by dedicated tests.

Not verified: plans containing Python actions (the engine never applies by design); JobManager failover while a drain is in progress.

### API

- Adds AgentExecutionOptions.PARALLEL_EXECUTION_ENABLED (default true) and MAX_IN_FLIGHT_INPUT_RECORDS (default 100) to the Java and Python APIs; no existing option changes meaning.
- Pure-Java agents on JDK < 21 switch from serial to parallel execution by default; per-key ordering and exactly-once semantics are unchanged, and parallel-execution.enabled=false restores the previous behavior.
- JDK 21 coroutine execution and plans containing Python actions are unaffected.

### Documentation

- [ ] `doc-needed` 
- [ ] `doc-not-needed` 
- [x] `doc-included` 

- [x] Yes
- [ ] No

Generated-by: Qoder


eemario:
Thanks @pltbkd and @Ryan-Nightwish - this builds on your earlier work. PTAL when you get a chance!

runtime/src/main/java/org/apache/flink/agents/runtime/operator/parallel/ParallelExecutionCoordinator.java
```diff
@@ -0,0 +1,299 @@
...
+            Supplier<ParallelExecutionTask> workFactory,
+            int maxWorkers,
+            long workerIdleTimeoutMillis) {
+        this.parallelExecutionLock = checkNotNull(parallelExecutionLock);
+        this.schedule = new ParallelExecutionTaskQueue(parallelExecutionLock);
+        this.mailboxDispatcher = checkNotNull(mailboxDispatcher);
+        this.workFactory = checkNotNull(workFactory);
+        DispatchQueue dispatchQueue = new DispatchQueue();
+        this.workerExecutor =
+                new ThreadPoolExecutor(
```
wenjin272, line 93: These workers need to participate in the existing managed-worker interpreter lifecycle from #1092. With the default thread factory, `PythonInterpreterManager` treats them as unmanaged callers and forwards Python invocations to a separate callback worker. This introduces two regressions for Java actions using Python resources:

- **Cross-thread Python object access:** `PythonVectorStore` receives `PyObject`s created on the callback worker, but reads their attributes and closes them on the original action worker.
- **Resource-initialization deadlock:** the action worker holds the `ResourceCache.getResource()` monitor while initializing a Python chat model setup. It waits for the callback worker to run Python `open()`. That `open()` resolves the model connection through the same resource cache, so the callback worker waits for the monitor held by the action worker. Neither can proceed. Previously, this lookup ran on the same thread and could re-enter the monitor.

I reproduced both with the production Java components and mocked native interpreter calls. Could we integrate the new pool with managed-worker interpreter ownership and cleanup, and add regression coverage for these paths?

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -1049,17 +1398,256 @@ private void tryResumeProcessActionTasks() throws Exception {
...
+        }
+
+        @Override
+        public void execute() {
+            try {
+                // Pull + prepare under the lock hold: pairing the FIFO poll with the under-lock
+                // taskIndex keeps the (taskIndex, task) binding correct across concurrent workers.
+                // createAndSetRunnerContext already binds the context on this thread.
+                setCurrentKey(key);
+                ActionTask task = stateManager.pollNextActionTask();
```
wenjin272, line 1465: This direct call to `stateManager.pollNextActionTask()` bypasses the operator's existing dequeue helper, which calls `builtInMetrics.markActionTaskDequeued()`.

Enqueueing an action still stores its execution ID and timestamp in `initialTaskEnqueueNanos`. The skipped dequeue notification is the only place that removes this entry; action completion does not clean it up. Consequently, the parallel path retains an entry for every completed action indefinitely, causing unbounded memory growth as well as an incorrect pending-action gauge.

Using the actual operator harness, 10 fully completed input records producing 20 actions left **20 retained entries and a pending count of 20** in parallel mode; serial mode left **zero**.

Could we preserve the existing dequeue bookkeeping here, accounting for internal noop tasks? The direct `pollNextPendingInputEvent()` call in `maybeFinishCurrentInput()` also skips the corresponding pending-input metric decrement.

wenjin272 (commented):
Thanks for implementing parallel action execution on JDK < 21 and for the detailed explanation and tests! I found two runtime issues that should be addressed before merging: the new workers do not participate in the existing interpreter thread-ownership lifecycle, and the parallel dequeue path skips metrics cleanup. I have left the call paths and reproduction results inline.

runtime/src/main/java/org/apache/flink/agents/runtime/operator/ActionExecutionOperator.java
```diff
@@ -236,10 +347,23 @@ public void open() throws Exception {
...
                 jobIdentifier,
                 getRuntimeContext().getUserCodeClassLoader());
 
+        if (parallelExecutionWithoutCoroutineEnabled) {
+            executionCoordinator =
+                    new ParallelExecutionCoordinator(
+                            parallelExecutionLock,
+                            mailboxExecutor::execute,
+                            Work::new,
+                            pythonBridge::releaseCurrentThreadInterpreter,
```
wenjin272: Thanks for addressing the earlier findings! The managed-worker cleanup introduces a resource-lifetime issue.

On the parallel path, `ctx.getResource()` checks lock ownership rather than mailbox-thread identity, so a Python resource can be lazily created on an action worker. Its cached `PyObject` retains that worker interpreter's native state.

During operator shutdown, the coordinator closes first, and worker cleanup closes the interpreter. `ResourceCache.close()` then closes the cached resource, whose `PyObject.close()` accesses the already-freed native state.

I reproduced this with the actual coordinator, resource cache, adapters, and native Pemja: the current shutdown order crashes in `PyObject.decRef → PyEval_AcquireThread`. Closing the cache while the worker interpreter is still alive succeeds, even from the mailbox thread, so this is a lifetime issue rather than cross-thread access alone.

Could we ensure cached resource handles cannot outlive their interpreter state? Reordering shutdown alone would not cover surplus workers retiring after the idle timeout. The current mocked tests do not exercise cached resource cleanup after interpreter teardown.

